PR #2, opened 2024-01-20
specs: 4844 fault-proof preimage key type specs

protolambda:
Migrated specs PR https://github.com/ethereum-optimism/optimism/pull/9065

This adds two new preimage oracle key types:
- global SHA2-256: full preimage by hash
- global EIP-4844 point-evaluation: a blob point by its combination of blob-commitment and point index.


ajsutton (commented):
Looks good but files all got rearranged in the meantime. Sorry.

ajsutton (approved):
Merged with latest main so this is good to go.

PR #6, opened 2024-01-25
feat: interop specs

tynes:
**Description**

WIP interop specs


specs/interop.md
```diff
@@ -0,0 +1,188 @@
...
+| baseFeeScalar     | uint32                  | 4-7                     | 1       |
+| blobBaseFeeScalar | uint32                  | 8-11                    |         |
+| sequenceNumber    | uint64                  | 12-19                   |         |
+| l1BlockTimestamp  | uint64                  | 20-27                   |         |
+| l1BlockNumber     | uint64                  | 28-35                   |         |
+| basefee           | uint256                 | 36-67                   | 2       |
+| blobBaseFee       | uint256                 | 68-99                   | 3       |
+| l1BlockHash       | bytes32                 | 100-131                 | 4       |
+| batcherHash       | bytes32                 | 132-163                 | 5       |
+| interopSetSize    | uint8                   | 164-165                 | 6       |
```
mslipper, line 168: This would limit the number of chains in the interop set to 255. Is this enough?

specs/interop.md
```diff
@@ -0,0 +1,188 @@
...
+
+The `setL1BlockValuesInterop()` function MUST be called on every block after the interop upgrade block.
+The interop upgrade block itself MUST include a call to `setL1BlockValuesEcotone`.
+
+## SystemConfig
+
+The `SystemConfig` is updated to manage the interop set. The chain operator can add or remove chains from the interop set
+through the `SystemConfig`. A new `ConfigUpdate` event `UpdateType` enum is added that corresponds to a change in the interop set.
+
+The `SystemConfig` MUST enforce that the maximum size of the interop set is `type(uint8).max` or 255.
```
mslipper: Is this enough?

specs/interop.md
```diff
@@ -0,0 +1,188 @@
...
+
+### CrossL2Outbox
+
+Address: `0x4200000000000000000000000000000000000023`
+
+The `CrossL2Outbox` is responsible for initiating a cross chain message on the local chain.
+For every cross chain message that is sent, it MUST emit an event that indicates the existence
+of the cross chain message.
+
+An initiating message
```
mslipper: Missing a section?

specs/interop.md
```diff
@@ -0,0 +1,188 @@
...
+network upgrade will be included in this document in the future.
+
+## Overview
+
+| Term          | Definition                                               |
+|---------------|----------------------------------------------------------|
+| Local Chain   | A blockchain that includes an initiating message |
+| Remote Chain  | A blockchain that includes a relaying message |
+| Initiating Message | A transaction submitted to a local chain that authorizes execution on a remote chain |
+| Relaying Message | A transaction submitted to a remote chain that corresponds to an initiating message |
```
protolambda: Do we allow relaying messages that relay generic "block header" information (state-root, maybe receipts root, etc.), rather than an actual initiated message? I think the fault-proof can support that easily, since outgoing initiating messages can be accumulated into the block-header, and that's all accessible in the current proposed consolidation step.
This would allow state-reads and such. The relaying message pays for relay anyway.

tynes: This is a very interesting idea and am very curious about what can be built with it. Do you have any specific examples of things that can be built with it? Just want to make sure we are going in the straightest line towards so we need less testing and edge cases.

I could certainly imagine a type of withdrawal that isn't a call but instead "teleports" the block header into a remote chain. For what its worth, its possible to do this not natively at the application layer with this spec. You create a contract on both sides, the local contract uses the blockhash opcode and sends it across, the remote is `onlyOtherContract` so nobody else can store hashes in it and it stores the blockhash, then the user unwraps the blockhash with the header.

I think there is something interesting here with regards to credible commitments, ie sequencer of chain A includes a payment to bob and then a proof of that is sent to another chain where the other chain can statically analyze the tx (unwrapped from blockhash) but its still a bit researchy. Our sequencer MEV ecosystem isnt mature enough for this.

I propose that we don't bake this feature into the protocol directly now because it can be accomplished at the app layer and then if people do begin to use this functionality at the app layer then we can "precompile" or "nationalize" the functionality by pulling it into the protocol

hamdiallam: @tynes this was the original reason for collecting the `roots` in the the v2 implementation. Not interop critical code but unlocks immediate cross-chain state reads.

However this does introduce the system transaction it looks like v3 is trying to avoid

specs/interop.md
```diff
@@ -0,0 +1,188 @@
...
+entity in practice.
+
+## The Interop Set
+
+The interop set defines the set of chains that a remote chains allows as local chains. Another way of
+saying it is that the interop set defines the set of initiating messages that are valid for a relaying
+message to be included. A relaying message MUST have an initiating message that is included in a chain
+in the interop set.
+
+The interop set is defined by chain id. Since it is impossible to enforce uniqueness of chain ids,
```
hamdiallam: Does it make sense to hold reference of the system_config_address which maps to a chain id in the interop state?

That way even though chain ids are not unique, it can provable

specs/interop.md
```diff
@@ -0,0 +1,188 @@
...
+For every cross chain message that is sent, it MUST emit an event that indicates the existence
+of the cross chain message.
+
+An initiating message 
+
+### CrossL2Inbox
+
+Address: `0x4200000000000000000000000000000000000022`
+
+The `CrossL2Inbox` is responsible for relaying a cross chain message on the remote chain.
```
hamdiallam: @protolambda @tynes

Carrying over the principle that we should be creating a design that works for us first (Messaging & StandardBridge).

In slack, @karlfloersch mentioned leaving the state bloat problem as a future improvement is fine. Since the primary use case of interop is the bridge and this use case requires replay protection, why not simply have the `CrossL2Inbox` record these message hashes such that we can make the `CrossDomainMessenger` stateless. My reasoning for this

1. We have single expected op-chain behavior for deposits, withdrawals, interop transactions. These transactions are processed only once. It feels like strange to have a fork in protocol UX for interop transcations
2. It becomes a nice improvement to the CrossDomainMessenger and simplifies it's implementation
3. Every app that uses the inbox must re-implement replay-protection

specs/interop.md
hamdiallam: until we have zk, this 255 feels like a nice upper bound the set given the scaling constraints

specs/interop.md
hamdiallam: how does social consensus work with fault proofs? The superchain registry is off-chain as of now right?

specs/interop.md
```diff
@@ -0,0 +1,188 @@
...
+The derivation pipeline enforces invariants on safe blocks that include relaying transactions.
+
+- The relaying message MUST have a corresponding initiating message
+- The initiating message that corresponds to a relaying message MUST come from an allowed chain
+
+Blocks that contain transactions that relay cross domain messages to the remote chain where the
+initiating transaction does not exist MUST be considered invalid and MUST not be allowed by the
+derivation pipeline to be considered safe.
+
+There is no concept of replay protection within the protocol since there is no replay protection
```
hamdiallam: is there evidence of this
> no replay protection mechanism that fits well for all applications

app-level msg hash based replay protection seems sufficient. I can think of any significant use case that would entail multiple replay messages for a single initiated message

hamdiallam: Yes it simplifies the implementation (by not much - 1 mapping). But allows for strange outcomes

specs/interop.md
hamdiallam: hmm oh @protolambda, are you suggesting relaying generic header information external to the protocol? I wonder what party would be incentivized to this for free

specs/interop.md
```diff
@@ -0,0 +1,188 @@
...
+  - L1Block deployment
+  - Update L1Block Proxy ERC-1967 Implementation Slot
+
+The execution payload MUST set `noTxPool` to `false` for this block.
+
+The exact definitions for these upgrade transactions are still to be defined.
+
+## Derivation Pipeline
+
+The derivation pipeline enforces invariants on safe blocks that include relaying transactions.
```
hamdiallam: @protolambda @tynes 

General thought that popped up with safe head progression now become async for chain dependencies.

When message inclusion was a apart of the derivation pipeline, a great way to scale the interop set to many chains was through light clients. The verifier would only need to retain a depth of the interop peers relative to the max time an which a remote message would be included. This reduces resource constraints substantially

However now that message inclusion is up the user. I believe verifiers must now run archival nodes of all peers. Unless i'm missing something in my reasoning, are we okay with this tradeoff? If we want to scale interop to many chains which was a goal previously, I think this does pose a problem

specs/interop.md
tynes: I cleaned this up, there may still be some dangling lines

specs/interop.md
tynes: I added in a section under security concerns that talks about why this value was chosen. There are 2 reasons that are both security related

specs/interop.md
tynes: > I believe verifiers must now run archival nodes of all peers.

Can you explain why this is the case? I do not believe it is the case. Verifiers need logs, not historical state

specs/interop.md
```diff
@@ -0,0 +1,324 @@
...
+
+After the full execution of a block, the set of `MessagePassed` events emitted from the `CrossL2Outbox` are collected and merklized.
+This commitment MUST be included as the block header's [extra data field](https://github.com/ethereum/execution-specs/blob/1fed0c0074f9d6aab3861057e1924411948dc50b/src/ethereum/frontier/fork_types.py#L115).
+The events are serialized with using [Simple Serialize](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md) aka SSZ.
+
+| Name          | Value |
+|---------------|-------------------------------------|
+| `MAX_LOG_DATA_SIZE` | `uint64(2**24)` or 16,777,216 |
+| `MAX_MESSAGES_PER_BLOCK` | `uint64(2**20)` or 1,048,576 |
+
```
tynes: I decided for ssz here because I am under the impression that it is better for the fault proof. I don't love how there are 2 consensus serializations of the cross chain message. One is ABI for the smart contracts (see the `messageHash` definition) and then ssz for in the extradata. Thoughts @protolambda?

specs/interop.md
tynes: All applications that want the same sort of replay protection that we want as the developers of the bridge can use the messenger. The messenger is responsible for replay protection and can give it to any application that needs replay protection. For people that want to use the lower level interface, they will need to implement their own type of replay protection. The idea is that the second order effect of forcing everybody to use our replay protection scheme is that it will end up bloating the state more

specs/interop.md
```diff
@@ -0,0 +1,324 @@
...
+The `messageHash` and chain id uniquely identify the existence of an initiating message and SHOULD
+be used by a block builder to ensure the existence of the initiating message when verifying the relaying
+transaction. `messageHash` SHOULD be `indexed` to allow for easy filtering of logs.
+
+A cross chain message is defined in the following format:
+
+```solidity
+struct CrossChainMessage {
+    uint256 nonce;
```
tynes: There are 3 ways to do upgradability with this. One is to introduce a `CrossChainMessageV2` and another is to put the version in the nonce. Another possibility is to do something like:

```solidity
struct CrossChainmessage {
  uint256 version;
  bytes data;
}
```

Where the version defines the encoding of the data

specs/interop.md
```diff
@@ -0,0 +1,324 @@
...
+The `CrossL2Inbox` MUST only transfer out ether if the caller is the `CrossL2Outbox` from a remote
+chain.
+
+#### Relaying Messages
+
+All of the information required to satisfy the invariants MUST be included to or committed to in the calldata
+of the function that is used to execute relaying messages. Both the block builder and the smart contract use
+this information to ensure that all system invariants are held.
+
+```solidity
```
tynes: I don't love how much wasted calldata there is here but it is simple. I recommend we optimize in the future

specs/interop.md
```diff
@@ -0,0 +1,324 @@
...
+
+The `CrossL2Inbox` is responsible for relaying a cross chain message on the remote chain.
+It is permissionless to finalize a cross chain message on behalf of any user. Certain protocol
+enforced invariants must be preserved to ensure safety of the protocol.
+
+#### Invariants
+
+- The timestamp of relaying message MUST be greater than or equal to the timestamp of the initiating message
+- The chain id of the initiating message MUST be in the interop set
+- The relaying message MUST be initiated by an externally owned account such that the top level EVM call frame enters the `CrossL2Inbox`
```
Spacesai1or: This would mean EIP-4337 accounts/paymasters are barred from relaying messages

specs/interop.md
hamdiallam: a message can be replayed at any time. So even if we prune the storage data, we still need to permanently retain all emitted outbox logs from all remotes

specs/interop.md
hamdiallam: > The idea is that the second order effect of forcing everybody to use our replay protection scheme is that it will end up bloating the state more

This is only true if there are even any apps that would use a lower-level interface without replay protection. Is there any supporting evidence of apps like this that has any substantial usage?

This also leads into another point that the state bloat still exists as now all message receipts from all remotes must be permanently retained since they can be relayed as many times as they want (even for msgs that revert) as the revert is at the evm level

specs/interop.md
tynes: Logs are part of history, not state. You do not need an archive node to access them

hamdiallam: yea that's right we dont an archive node but we have growing storage requirements with this design that scales poorly. the state bloat is pushed into permanent receipts that need to be collected and stored

specs/interop.md
tynes: > is there evidence of this

I can think of a few examples off of the top of my head:
- messages must go exactly in order, only once
- messages must be monotonic but can jump ahead by more than 1, only once
- messages can go in any order, only once
- messages can go in any order, as many times as possible

>- messages must be monotonic but can jump ahead by more than 1, only once

This is particularly useful for invalidating signed offchain messages, similar to the way permit2 does nonce invalidation

No replay protection mechanism exists that fits well for all applications similarly to how no cache invalidation mechanism works well for all applications. Application specific solutions should be adopted based on the application

What strange outcomes do you mean? Anybody that wants replay protection can use the messenger

specs/interop.md
tynes: Hmm I don't understand what you mean by this. No matter what, there is no canonical mapping between a chain id and rpc url. Social consensus is used to determine the "type" of rpc url aka what chain it points to. Social consensus tracks the canonical chain that uses a chain id, just like how we use https://chainlist.org/ today

specs/interop.md
tynes: > but unlocks immediate cross-chain state reads

This can be built at the application layer, we should be careful about adding scope to the protocol unless we know for a fact it solves a problem

tynes: > hmm oh @protolambda, are you suggesting relaying generic header information external to the protocol? I wonder what party would be incentivized to this for free

I interpreted this as internal to the protocol. I wouldn't ask the question "what party would be incentivized to do this for free", I would ask the question "what is a concrete usecase for this that is unlocked by baking it into the protocol instead of having it at the application layer?"

specs/interop.md
tynes: > This is only true if there are even any apps that would use a lower-level interface without replay protection. Is there any supporting evidence of apps like this that has any substantial usage?

This is also true if the particular type of replay protection that we use is useful for all applications. Which is fundamentally not true. See [this](https://github.com/ethereum-optimism/specs/pull/6#discussion_r1468216950) comment

Message receipts do not lead to state bloat as they are not part of state. They are part of history. It is a well known issue that history grows forever, which is why eip4444 exists. We will need to contribute to building eip4444 eventually

specs/interop.md
tynes: This is a known limitation of the design. There is no way to have static analysis on a transaction without an `onlyEOA` check unfortunately. Without this, it will be very expensive for the block builder as it will need to do a full simulation to get the information required to verify the existence of the initiating transaction. This introduces denial of service attacks for the sequencer. App level builders will figure out other types of paymaster systems for this

specs/interop.md
```diff
@@ -0,0 +1,396 @@
...
+derivation pipeline to be considered safe.
+
+There is no concept of replay protection within the protocol since there is no replay protection
+mechanism that fits well for all applications. Users MAY submit an arbitrary number of relaying
+messages per initiating message. This greatly simplifies the protocol and allows for
+applications to build application specific replay protection.
+
+### Depositing a Relaying Message
+
+Deposit transactions (force inclusion transactions) give censorship resistance to layer two networks.
```
tynes: Need to add explicit call out the the deposit is dropped by the derivation pipeline

specs/interop.md
```diff
@@ -0,0 +1,396 @@
...
+The derivation pipeline must gracefully handle the case in which a user uses a deposit transaction to
+relay a cross chain message. To not couple preconfirmation security to consensus, deposit transactions
+that relay cross chain messages MUST have an initiating message that is considered safe. This relaxes
```
ajsutton: What does considered safe actually mean here? Are we trying to say that a deposit tx that relays cross chain messages, is only valid if it occurs in an L1 block after the channel including the initiating transaction is included on L1? Can it be in the same block but after the batcher transaction?

Are our derivation pipeline specs actually that strict on the order of processing batches from L1 vs using an L1 block as the origin for new blocks?

specs/interop.md
ajsutton: Yeah SSZ is definitely better here to avoid the large preimage problem. The SSZ Merkle Tree breaks everything down into a maximum leaf size of 32 bytes whereas ABI format would represent the entire event data as a single preimage.  If the event data has a size limit that would be fine, but if it has arbitrary size (or could be quite large) then we should use SSZ.

specs/interop.md
```diff
@@ -0,0 +1,396 @@
...
+##### Only EOA Invariant
+
+The `onlyEOA` invariant on relaying a cross chain message enables static analysis on relaying messages.
+This allows for the derivation pipeline and block builders to reject relaying messages that do not
+have a corresponding initiating message.
+
+##### Transferring Ether in a Cross Chain Message
+
+The `CrossL2Inbox` MUST be initially set in state with an ether balance of `type(uint256).max`.
+There is no reason for a user to send [value](https://github.com/ethereum/execution-specs/blob/1fed0c0074f9d6aab3861057e1924411948dc50b/src/ethereum/frontier/fork_types.py#L53)
```
ajsutton: This isn't a particularly reassuring way to phrase this.  Might be worth noting that transactions that cause a balance overflow revert. Reverting those transaction is ok because there's no reason for a user to send value...

specs/interop.md
```diff
@@ -0,0 +1,396 @@
...
+```
+
+By including the hash in the calldata, it makes static analysis much easier for block builders.
+Without the check onchain, the block builder would need to hash the `CrossChainMessage` to build the
+`eth_getLogs` query. With the hash onchain, the block builder can directly trust the hash.
+
+### L1Block
+
+The `L1Block` contract is updated to include the set of allowed chains. The L1 Atrributes transaction
+sets the set of allowed chains. The `L1Block` contract MUST provide a public getter to check is a particular
```
ajsutton:
```suggestion
sets the set of allowed chains. The `L1Block` contract MUST provide a public getter to check if a particular
```

specs/interop.md
```diff
@@ -0,0 +1,396 @@
...
+| basefee           | uint256                 | 36-67                   | 2       |
+| blobBaseFee       | uint256                 | 68-99                   | 3       |
+| l1BlockHash       | bytes32                 | 100-131                 | 4       |
+| batcherHash       | bytes32                 | 132-163                 | 5       |
+| interopSetSize    | uint8                   | 164-165                 | 6       |
+| chainIds          | [interopSetSize]uint256 | 165-(32*interopSetSize) | 6+      |
+
+## Fault Proof
+
+Pseudocode for the `op-program`
```
ajsutton: I probably just need the `TODO: more description` part to give me the right context here, but this code doesn't map to my mental model of op-program at all so I can't make any sense of it. :)

specs/interop.md
```diff
@@ -0,0 +1,396 @@
...
+class CrossChainMessage(Container):
+    nonce: uint256
+    sender: ExecutionAddress
+    target: ExecutionAddress
+    value: uint256
+    timestamp: uint256
+    chainid: uint256
+    data: ByteList[MAX_LOG_DATA_SIZE]
+
+messages = List[CrossChainMessage, MAX_MESSAGES_PER_BLOCK]
```
tynes: This is the wrong syntax, will update

specs/interop.md
```diff
@@ -0,0 +1,396 @@
...
+The events are serialized with using [Simple Serialize](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md) aka SSZ.
+
+| Name          | Value |
+|---------------|-------------------------------------|
+| `MAX_LOG_DATA_SIZE` | `uint64(2**24)` or 16,777,216 |
+| `MAX_MESSAGES_PER_BLOCK` | `uint64(2**20)` or 1,048,576 |
+
+```python
+class CrossChainMessage(Container):
+    nonce: uint256
```
tynes: nonce exists to ensure that only 1 log will ever exist per message. If it did not exist, then the same cross chain message hash could exist at multiple timestamps, which then the honest block builder would need to look at the latest timestamp. this adds edge cases so i recommend we keep nonce for simplicity here

specs/interop.md
```diff
@@ -0,0 +1,502 @@
...
+    - [Cross Chain Message Latency](#cross-chain-message-latency)
+    - [Dynamic Size of L1 Attributes Transaction](#dynamic-size-of-l1-attributes-transaction)
+    - [Maximum Size of the Dependency Set](#maximum-size-of-the-dependency-set)
+    - [Mempool](#mempool)
+    - [Reliance on History](#reliance-on-history)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Interoperability
+
```
tynes: Should add in the exact mapping to the `eth_getLogs` query from the event in the outbox

specs/interop.md
hamdiallam: After reviewing the latest changes, I see that we're choosing to to separate away from a universal CDM base contract as designed between L1<>L2.

With a different messenger, diverging in behavior is fine. My main concern was around getting to an end-state where the CDM interface remains universal but it looks like we are explicitly choosing to handle L2<>L2 separately.

specs/interop.md
hamdiallam: Since the L1 address space is shared between chains, my question was if there's any value in using it so that we canonical label for a chain that's verifiable.

I'm not super familiar with fault proofs and was wondering how the right preimage inputs would be provided to a proof on L1 when a chain-dependency input is invalid if the chain id plays into that (if it doesn't then that answers my q)

specs/interop.md
hamdiallam: forsure, even with eip4444 since a message can be replayed >1. Verifiers must have _some_ remote source of pruned history if someone choose to relay a really old message. Just wanted to note this concern

thanks for the distinction between state/history

specs/interop.md
tynes: This is being done explicitly to handle ether

specs/interop.md
tynes: Thank you for the feedback!

specs/interop.md
```diff
@@ -0,0 +1,540 @@
...
+#### Invariants
+
+- The timestamp of executing message MUST be less than or equal to the timestamp of the initiating message
+- The chain id of the initiating message MUST be in the dependency set
+- The executing message MUST be initiated by an externally owned account such that the top level EVM call frame enters the `CrossL2Inbox`
+- The chain id of the initiating message MUST not be the chain id of the executing message
+
+##### Only EOA Invariant
+
+The `onlyEOA` invariant on executing a cross chain message enables static analysis on executing messages.
```
annieke: Just checking on this constraint: it's only on the relayer, which means we cannot use a 4337-like paymaster system to sponsor relays.

However, the initiating transaction _can_ be from a contract, right? so we're not barred for use cases where an app automatically pushes a cross-chain tx when certain app state updates?

specs/interop.md
```diff
@@ -0,0 +1,540 @@
...
+| Destination Chain  | A blockchain that includes an executing message |
+| Initiating Message | A transaction submitted to a source chain that authorizes execution on a destination chain |
+| Executing Message | A transaction submitted to a destination chain that corresponds to an initiating message |
+| Cross Chain Message | The cumulative execution and side effects of the initiating message and executing message |
+| Dependency Set | The set of chains that originate initiating transactions where the executing transactions are valid |
+
+A total of two transactions are required to complete a cross chain message. The first transaction is submitted
+to the source chain which authorizes execution on a destination chain. The second transaction is submitted to a
+destination chain, where the block builder SHOULD only include it if they are certain that the first transaction was
+included in the source chain.
```
annieke: Note: this current design doesn't guarantee atomicity (eg the first transaction doesn't _guarantee_ the second transaction's success, in the same way that an L2 withdrawal doesn't guarantee L1 finalization). 

Atomicity is a potential project on top of this minimal protocol.

specs/interop.md
```diff
@@ -0,0 +1,540 @@
...
+
+The term "block builder" is used interchangeably with the term "sequencer" for the purposes of this document but they need not be the same
+entity in practice.
+
+## The Dependency Set
+
+The dependency set defines the set of chains that a destination chains allows as source chains. Another way of
+saying it is that the dependency set defines the set of initiating messages that are valid for an executing
+message to be included. An executing message MUST have an initiating message that is included in a chain
+in the dependency set.
```
annieke: Note: the dependency is currently configured on a per-chain basis (not bundled to a specific chain group such as the superchain-registry), on its L1 contract.

specs/interop.md
```diff
@@ -0,0 +1,540 @@
...
+  {
+    "address": "0x4200000000000000000000000000000000000023",
+    "topics":["0x52f9c5a044b4fc98e5bcc3f7790fc36930ad9e20cbee36bfaa0232b016c925b2", "$HASH"]
+  }
+]
+```
+
+## Sponsorship
+
+If a user does not have ether to pay for the gas of an executing message, application layer sponsorship solutions can be created.
```
annieke: Note: this applies our learning from the L1 to L2 deposit gas market, which is one of the most complex portions of our code. 

A user of this protocol currently has to pay for gas on both chains, which can be abstracted away in the app layer.

specs/interop.md
```diff
@@ -0,0 +1,540 @@
...
+The `L2ToL2CrossDomainMessenger` is a higher level abstraction on top of the `CrossL2Outbox` that
+provides features necessary for secure transfers of `ether` and ERC20 tokens between L2 chains.
+Messages sent through the `L2ToL2CrossDomainMessenger` on the source chain receive both replay protection
+as well as domain binding, ie the executing transaction can only be valid on a single chain.
+
+TODO: versioned messages in nonce like L1 and L2 path?
+
+#### Transferring Ether in a Cross Chain Message
+
+The `L2ToL2CrossDomainMessenger` MUST be initially set in state with an ether balance of `type(uint248).max`.
```
protolambda: `uint248`? As long as it's not `uint256` to not overflow it makes sense. Worth a note about the choice of value here.

specs/interop.md
```diff
@@ -0,0 +1,540 @@
...
+By including the hash in the calldata, it makes static analysis much easier for block builders.
+Without the check onchain, the block builder would need to hash the `CrossChainMessage` to build the
+`eth_getLogs` query. With the hash onchain, the block builder can directly trust the hash.
+
+### L2ToL2CrossDomainMessenger
+
+Address: `0x4200000000000000000000000000000000000024`
+
+The `L2ToL2CrossDomainMessenger` is a higher level abstraction on top of the `CrossL2Outbox` that
+provides features necessary for secure transfers of `ether` and ERC20 tokens between L2 chains.
```
annieke: As of now, the `L2ToL2CrossDomainMessenger` handles ETH natively, but ERC20s will be built on top of another layer of abstraction above that (like the StandardBridge)

annieke (commented):
Notes from sync spec review

@karlfloersch 's userland features:
1) Shared liquidity / Super-ERC20 tokens
2) Transaction sponsorship for x-chain messages
3) Low latency or same-block atomic transactions
4) cross-L2 oracles, and more, with the WIP cross-L2 outbox variant with events

specs/interop.md
annieke: Confirmed during spec review

specs/interop.md
```diff
@@ -0,0 +1,540 @@
...
+
+A total of two transactions are required to complete a cross chain message. The first transaction is submitted
+to the source chain which authorizes execution on a destination chain. The second transaction is submitted to a
+destination chain, where the block builder SHOULD only include it if they are certain that the first transaction was
+included in the source chain.
+
+The term "block builder" is used interchangeably with the term "sequencer" for the purposes of this document but they need not be the same
+entity in practice.
+
+## The Dependency Set
```
Spacesai1or: If part of the reason for the dependency set is to reduce the chance of a dos, there's still a possibility that it _could_ happen even with a limited/trusted dependency set. What happens in this scenario? How can you recover from it? Would the chain operator need to update the L1 config to remove the spamming chain from the dependency set? Would it be possible (although maybe not ideal) for the sequencer just to ignore the spam?

I don't think interop messages have a higher priority of being included in a block over non-interop txs, right? So a dos would mean a degraded performance for inclusion of every tx?

specs/interop.md
tynes: The rationale is in the spec, saying its for outbound liquidity w/o risk of overflow

specs/interop.md
tynes: The spec has changed and this no longer applies

specs/interop.md
tynes: This no longer applies to the latest spec

specs/interop.md
tynes: pseudocode has been added

specs/interop.md
tynes: > there's still a possibility that it could happen even with a limited/trusted dependency set

Can you explain how? The executing message isn't force included and pays for gas.

The denial of service comes from looking for data that doesn't exist

The interop executing messages do not have a higher priority for being included over regular messages.

specs/interop.md
Chomtana: What if different chains have the same chain ID? For example, chain ID 42069 which is a well-known chain ID for getting started OP stack deployment.

specs/interop.md
tynes: I have included much more detail in the new `safety` section on this topic

specs/interop.md
tynes: The spec states the following:

```
The dependency set is defined by a set of chain ids. Since it is impossible to enforce uniqueness of chain ids,
social consensus MUST be used to determine the chain that represents the canonical chain id. This
particularly impacts the block builder as they SHOULD use the chain id to assist in validation
of executing messages.
```

specs/interop.md
```diff
@@ -0,0 +1,708 @@
...
+#### `cross-unsafe` Inputs
+
+- MUST have valid cross chain messages
+
+`cross-unsafe` represents the `unsafe` blocks that had their cross chain messages fully verified.
+The network can be represented as a graph where each block across all chains are represented as
+a node and then a directed edge between two blocks represents the source block of the initiating
+message and the block that included the executing message.
+
+An input can be promoted from `safe` to `cross-unsafe` when the full dependency graph is resolved
```
snario, line 217:
```suggestion
An input can be promoted from `cross-unsafe` to `safe` when the full dependency graph is resolved
```

I think this is a typo, right?

specs/interop.md
```diff
@@ -0,0 +1,708 @@
...
+such that all cross chain messages are verified to be valid and at least one message in the dependency
+graph is still `unsafe`.
```
snario: This was / is kind of confusing to me when reading it. Why would an input be `safe` is at least one message in the dependency graph is still `unsafe`, since the definition of `safe` requires all messages in the graph to be resolved? Also, wouldn't it be "demoted", not "promoted" from `safe` to `cross-unsafe`?

specs/interop.md
```diff
@@ -0,0 +1,708 @@
...
+    uint256 chainid;
+}
+```
+
+| Name          | Type | Description |
+|---------------|----------------------------------|--|
+| `origin` | `address` | Account that emits the log |
+| `blocknumber` | `uint256` | Block number in which the log was emitted |
+| `logIndex` | `uint256` | The index of the log in the array of all logs emitted in the block |
+| `timestamp` | `uint256` | The timestamp that the log was emitted. Used to enforce the timestamt invariant |
```
snario:
```suggestion
| `timestamp` | `uint256` | The timestamp that the log was emitted. Used to enforce the timestamp invariant |
```

specs/interop.md
```diff
@@ -0,0 +1,708 @@
...
+It is impossible to check that the `Identifier` matches the cross chain message on chain. If the block
+builder includes a message that does not correspond to the `Identifier`, their block will be reorganized
+by the derivation pipeline.
+
+TODO upgrade path: think of `Identifier` as an enum that determines the mapping function for a log to a message
+a change in the id will result in a change in the 4byte selector
+
+#### `Identifier` Getters
+
+The `Indentifier` MUST be exposed via `public` getters so that contracts can call back to authenticate
```
snario:
```suggestion
The `Identifier` MUST be exposed via `public` getters so that contracts can call back to authenticate
```

specs/interop.md
```diff
@@ -0,0 +1,708 @@
...
+}
+```
+
+### L1Block
+
+| Constant         | Value                   |
+|----------------- | ----------------------- |
+| Address | `0x4200000000000000000000000000000000000015` |
+
+The `L1Block` contract is updated to include the set of allowed chains. The L1 Atrributes transaction
```
snario:
```suggestion
The `L1Block` contract is updated to include the set of allowed chains. The L1 Attributes transaction
```

specs/interop.md
```diff
@@ -0,0 +1,708 @@
...
+The `setL1BlockValuesInterop()` function MUST be called on every block after the interop upgrade block.
+The interop upgrade block itself MUST include a call to `setL1BlockValuesEcotone`.
+
+## SystemConfig
+
+The `SystemConfig` is updated to manage the dependency set. The chain operator can add or remove
+chains from the dependency set through the `SystemConfig`. A new `ConfigUpdate` event `UpdateType`
+enum is added that corresponds to a change in the dependency set.
+
+The `SystemConfig` MUST enforce that the maximum size of the dependency set is `type(uint8).max` or 255.
```
snario: Where does this number come from, and will be upgradeable in the future?

snario: I'd be curious to know if there has been consideration as to how this evolves over time, also since the size of the data has to be small enough. 

Related questions:
- The values will change over time, right, so if we're at 255 and we want to add a new one, it will be possible to remove one and add a new one? If so, this is kind of like the "current" dependency set.
- In a future where we want, say, 1000 dependant chains for whatever reason, is the approach going to involve merklizing the list? Does that break anything?

specs/interop.md
```diff
@@ -0,0 +1,708 @@
...
+|---------------|----------------------------------------------------------|
+| Source Chain   | A blockchain that includes an initiating message |
+| Destination Chain  | A blockchain that includes an executing message |
+| Initiating Message | An event emitted from a source chain |
+| Executing Message | A transaction submitted to a destination chain that corresponds to an initiating message |
+| Cross Chain Message | The cumulative execution and side effects of the initiating message and executing message |
+| Dependency Set | The set of chains that originate initiating transactions where the executing transactions are valid |
+
+A total of two transactions are required to complete a cross chain message. The first transaction is submitted
+to the source chain emits an event that can be consumed on a destination chain. The second transaction is submitted to a
```
mds1:
```suggestion
A total of two transactions are required to complete a cross chain message. The first transaction is submitted
to the source chain and emits an event that can be consumed on a destination chain. The second transaction is submitted to a
```

specs/interop.md
```diff
@@ -0,0 +1,708 @@
...
+- L1 Attributes Transaction calling `setL1BlockValuesEcotone`
+- User deposits from L1
+- Network Upgrade Transactions
+  - L1Block deployment
+  - CrossL2Inbox deployment
+  - L2ToL2CrossDomainMessenger deployment
+  - Update L1Block Proxy ERC-1967 Implementation Slot
+  - Update CrossL2Inbox Proxy ERC-1967 Implementation Slot
+  - Update L2ToL2CrossDomainMessenger Proxy ERC-1967 Implementation Slot
+  - Mint Cross Chain Ether Liquidity in L2ToL2CrossDomainMessenger
```
mds1: Is this a new deposit transaction type, similar to the [L1 Attributes](https://specs.optimism.io/protocol/deposits.html#l1-attributes-deposited-transaction) deposit transaction, that sets the `type(uint248).max` initial balance?

specs/interop.md
```diff
@@ -0,0 +1,708 @@
...
+arbitrary number of executing messages per initiating message. Applications MUST build their own replay
+protection mechanisms if they are interacting with the lowest level abstractions.
+
+### Depositing an Executing Message
+
+Deposit transactions (force inclusion transactions) give censorship resistance to layer two networks.
+The derivation pipeline must gracefully handle the case in which a user uses a deposit transaction to
+relay a cross chain message. To not couple preconfirmation security to consensus, deposit transactions
+that execute cross chain messages MUST have an initiating message that is considered [safe](#safety) by the remote
+chain's derivation pipeline. This means that the remote chain's data including the initiating message
```
mds1: Should this say "source chain" instead of "remote chain"? I see remote chain used in a few other places too, if those are all intended we should add remote chain to the definitions table above

specs/interop.md
```diff
@@ -0,0 +1,708 @@
...
+If a deposit transaction references an initiating transaction that is not yet safe or does not exist,
+it MUST be dropped by the derivation pipeline.
+
+This inclusion property prevents a class of attacks where the user can trick the derivation pipeline
+into reorganizing the sequencer.
+
+### Safety
+
+Safety is an abstraction that is useful for reasoning about security. It should be thought about
+as a spectrum from `unsafe` to `final`. Users can choose to operate on information based on its
```
mds1: nit, `finalized` is used everywhere else
```suggestion
as a spectrum from `unsafe` to `finalized`. Users can choose to operate on information based on its
```

specs/interop.md
```diff
@@ -0,0 +1,708 @@
...
+
+Anything before `safe` technically uses a "preconfirmation" based security model which is not part
+of consensus. While useful to have definitions of the default meanings of these terms, they are
+technically policy and may be changed in the future.
+
+The `unsafe` label has the lowest latency while the `finalized` label has the highest latency.
+A set of properties must be matched before an input or an output can be promoted to the next
+label.
+
+The initiating messages for all dependent executing messages MUST be resolved as safe before an L2 block can transition
```
mds1: typo?
```suggestion
The initiating messages for all destination executing messages MUST be resolved as safe before an L2 block can transition
```

specs/interop.md
```diff
@@ -0,0 +1,708 @@
...
+- MUST NOT be reorganized based on Ethereum economic security
+
+`finalized` represents full Proof of Stake economic security on top of the data. This means that
+if the data is reorganized, then validators will be slashed.
+
+## State Transition Function
+
+After the full execution of a block, the set of [logs][log] that are emitted MUST be merklized and
+included in the block header. This commitment MUST be included as the block header's
+[extra data field][block-extra-data]. The events are serialized with using [Simple Serialize][ssz] aka SSZ.
```
mds1: The yellow paper says extra data is limited to 32 bytes but from the linked execution spec it seems to be arbitrary length. Did it change, or is the execution spec incorrect? If arbitrary length, is the requirement that the commitment is the only data in the extra data field, or can other data be prepended/appended?

specs/interop.md
protolambda: Dependent is correct, but it can be worded better. Executing messages depend on specific initiating messages. I'll revisit this after organizing the specs. Merging other typo fixes first now.

specs/interop.md
tynes: This number is chosen because it limits the size of the L1 attributes tx + higher numbers make the node prohibitively expensive. The security considerations section talks about this a bit, but the tldr is that we cannot consume more than the block gas limit with force include txs and the cost of promoting txs from unsafe to safe  uses a lot of bandwidth

This number feels quite sane for the first version, especially since the process of verifying the initiating messages in the derivation pipeline will use a lot of bandwidth. We will figure out how to make it cheaper over time but the first implementation will not be super scalable. Anything over 10 seems like it would be kind of crazy but will need to benchmark to know for sure.

If you want many dependent chains, you can have a multi hub and spoke kind of model. All transitive deps technically count, so just because you cannot directly accept incoming messages from a chain doesn't mean that you cannot, you will just need a hop in between. In the future we can consider relaxing this restriction depending on how cheap we can make the verification in the derivation pipeline. We have some fun ideas to use kzg since the witness size is constant, perhaps worldcoin would like to help with the cryptography?

specs/interop.md
tynes: The existing deposit tx type can mint ether, so its not a new type, its the same deposit tx type, it just mints ether out of thin air to that particular predeploy

specs/root.md
```diff
@@ -42,6 +42,7 @@ Specifications of new features in active development.
   - [Fault Dispute Game](./experimental/fault-proof/fault-dispute-game.md)
     - [Honest Challenger Behavior](./experimental/fault-proof/honest-challenger-fdg.md)
   - [Cannon VM](./experimental/fault-proof/cannon-fault-proof-vm.md)
+- [Interop](./interop.md)
```
tynes: This will need an update based on the latest commits that move the single file into a directory

specs/interop/predeploys.md
```diff
@@ -0,0 +1,243 @@
...
+**Table of Contents**
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Predeploys
+
+Two new system level predeploys are introduced for managing cross chain messaging along with
+an update to the `L1Block` contract with additional functionality.
+
+## CrossL2Inbox
```
tynes, line 30: Ok so if we end up shipping with the feature where you can pull in events from L1 then having this be named the `CrossL2Inbox` really only displays a subset of its functionality. It may be worth just renaming it `Inbox` for simplicity (or `Portal`) but also don't want to bikeshed on names too much.

h/t @annieke

annieke: Either's fine with me! +1 on not bikeshedding, we can just make diagrams to show how contracts relate to one another (instead of matching names)

specs/interop/fault_proof.md
```diff
@@ -0,0 +1,24 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+TODO
+
+No changes to the dispute game bisection are required. The only changes required are to the fault proof program itself.
+The insight is that the fault proof program can be a superset of the state transition function.
+
+## Security Considerations
+
+TODO
```
annieke, line 19: I'm not sure if this is the best section to add it, but would love a section somewhere in the specs that answers "how does enabling interop on a chain change what an user is trusting"? 
- is it still a single honest node, or now n honest nodes for n interop-ing chains? 
- any difference for being only the source chain or only destination chain?
- what are the implications of trusting an unsafe head of another sequencer, worst case scenario? (just a reorg?)

annieke: Separate but somewhat relevant Q - would also love an explicit write out on what must be true about participating chains to uphold these trust assumptions
- share the same fault proof contracts? 
- share L1 implementation contracts?
- have the same block time? 


i believe @tynes you phrased this + the other ask as "sketching out more directly what is needed for consensus" :)

specs/interop.md
tynes: This is now covered in https://github.com/ethereum-optimism/specs/pull/31

specs/interop.md
tynes: It must be at most 32 bytes - https://github.com/ethereum/go-ethereum/blob/8a76a814a2b9e5b4c1a4c6de44cd702536104507/params/protocol_params.go#L31

specs/interop/predeploys.md
```diff
@@ -0,0 +1,260 @@
...
+#### Sending Messages
+
+The initiating message is represented by the following event:
+
+```solidity
+event SentMessage(bytes message) anonymous;
+```
+
+The `bytes` are an ABI encoded call to `relayMessage`. The event is defined as `anonymous` so that no topics
+are prefixed to the abi encoded call.
```
mds1, line 174: The reason we want no topics is to save gas or to avoid having to slice off the first 32 bytes when parsing? With only a single event name emitted by this contract an anonymous one seems ok, but we probably want an event in `relayMessage` and then it becomes harder to distinguish the two (e.g. now you rely on number of topics to distinguish them which doesn't sound ideal)

specs/interop.md
tynes: This was a typo, should be updated

specs/interop/predeploys.md
tynes: The identifier points to a log where the log must be encoded like so:

```
msg := make([]byte, 0)
for _, topic := range log.Topics {
    msg = append(msg, topic.Bytes()...)
}
msg = append(msg, log.Data...)
```

This is because sometimes there is important information in the topics, but then you need an adapter contract to map a topic to a 4byte selector. It is because this `msg` is the calldata passed into a contract

specs/interop/predeploys.md
```diff
@@ -0,0 +1,260 @@
...
+#### `_id`
+
+A pointer to the `_msg` in a remote (or local) chain.
+
+The message [`Identifier`] of the executing message.
+This is required to enforce the message executes an existing and valid initiating message.
+
+By including the [`Identifier`] in the calldata, it makes static analysis much easier for block builders.
+It is impossible to check that the [`Identifier`] matches the cross chain message on chain. If the block
+builder includes a message that does not correspond to the [`Identifier`]`, their block will be reorganized
```
Spacesai1or:
```suggestion
builder includes a message that does not correspond to the [`Identifier`], their block will be reorganized
```

specs/interop/predeploys.md
```diff
@@ -0,0 +1,260 @@
...
+    require(_id.timestamp <= block.timestamp);
+    require(L1Block.isInDependencySet(_id.chainid));
+
+    assembly {
+      tstore(ORIGIN_SLOT, _id.origin)
+      tstore(BLOCKNUMBER_SLOT, _id.blocknumber)
+      tstore(LOG_INDEX_SLOT, _id.logIndex)
+      tstore(TIMESTAMP_SLOT, _id.timestamp)
+      tstore(CHAINID_SLOT, _id.chainid)
+    }
```
Spacesai1or, line 101: What ends up reading these slots? Is this info made available for `_target` to query if needed since `_id` is `calldata` and wouldn't otherwise be available?

Spacesai1or: > The Identifier MUST be exposed via public getters so that contracts can call back to authenticate properties about the _msg.

Yep that's it

specs/interop.md
tynes: Yes good catch, will need to do a pass and replace the old language everywhere

specs/interop/predeploys.md
tynes: > The reason we want no topics is to save gas or to avoid having to slice off the first 32 bytes when parsing? With only a single event name emitted by this contract an anonymous one seems ok, but we probably want an event in `relayMessage` and then it becomes harder to distinguish the two (e.g. now you rely on number of topics to distinguish them which doesn't sound ideal)

The `Identifier` points exactly to a single log and if the identifier doesn't exactly match the serialized log and the tx is included in a block, its considered an invalid block

tynes: > The reason we want no topics is to save gas or to avoid having to slice off the first 32 bytes when parsing? With only a single event name emitted by this contract an anonymous one seems ok, but we probably want an event in `relayMessage` and then it becomes harder to distinguish the two (e.g. now you rely on number of topics to distinguish them which doesn't sound ideal)

The `Identifier` points exactly to a single log and if the identifier doesn't exactly match the serialized log and the tx is included in a block, its considered an invalid block

specs/interop.md
tynes: This has been completed

specs/interop/fault_proof.md
snario: Discussed a similar thing recently. It seems like the trust requirements are now that there is a single honest validator for each chain and all of its dependent chains, and potentially any of the chains that those dependent chains have as dependencies?

specs/interop.md
tynes: We don't necessarily want to put labels on L1 and it also doesn't solve the problem because a label is just an alias for a chainid, who decides what the canonical labels are for chainids? No matter what there is a subjective nature to resolving a chainid to a particular chain where people agree on what chain id maps to what chain via social consensus

specs/interop/fault_proof.md
tynes: This is true, there needs to be an honest validator for each chain in the full transitive dependency set. Will be sure to add this explicitly

PR #13, opened 2024-01-30
feat: diff based L1 attributes transaction

tynes:
**Description**

Adds an experimental spec for a diff based L1 Attributes transaction.
This helps to reduce the historical growth of the chain by making
the L1 attributes transaction only push values that have changed.

This spec is placed in an `experimental` directory temporarily
until we figure out better ways to manage possible changes to the
protocol. By creating the spec now, we can more easily schedule
it for a future hardfork.

### DIFF
--- a/specs/experimental/diff-based-attributes-tx.md
+++ b/specs/experimental/diff-based-attributes-tx.md
@@ -1,0 +1,52 @@
+# Diff Based L1Attributes Transaction
+
+The `L1Attributes` transaction is the first transaction included in every L2 block.
+It can be thought of as an extension to the L2 block header that is 100% backwards
+compatible with L1 Ethereum by not changing the structure of the block header itself.
+A major problem with the `L1Attributes` transaction is that it constantly submits the
+same information over and over again. This adds chain history that must be preserved
+forever. Ideally, the `L1Attributes` transaction only includes diffs to reduce the
+growth of the chain history.
+
+## Implementation
+
+The `L1Attributes` transaction already commits to a schema that is ABI encoded.
+There is no concept of an `Optional` type in ABI encoding, so instead we can
+utilize a custom data structure to indicate the presence of a field in the schema.
+A leading bitfield can be used to use presence. Each bit indicates a specific item
+in the schema where `1` means it is present and `0` means it is not present. Each
+item MUST have at least a fixed size portion so that the decoder knows how far
+forward to seek after a read when the item is present.
+
+### Ecotone L1Attributes Example
+
+The following fields are required as part of the `L1Attributes` transaction after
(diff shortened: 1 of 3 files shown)


specs/experimental/diff-based-attributes-tx.md
```diff
@@ -0,0 +1,52 @@
...
+| baseFeeScalar     | 4            | 0b1000\_0000\_0000\_0000 |
+| blobBaseFeeScalar | 4            | 0b0100\_0000\_0000\_0000 |
+| sequenceNumber    | 8            | 0b0010\_0000\_0000\_0000 |
+| l1BlockTimestamp  | 8            | 0b0001\_0000\_0000\_0000 |
+| l1BlockNumber     | 8            | 0b0000\_1000\_0000\_0000 |
+| basefee           | 32           | 0b0000\_0100\_0000\_0000 |
+| blobBaseFee       | 32           | 0b0000\_0010\_0000\_0000 |
+| l1BlockHash       | 32           | 0b0000\_0001\_0000\_0000 |
+| batcherHash       | 32           | 0b0000\_0000\_1000\_0000 |
+
```
tynes: We can also optimize this so only 1 byte is necessary by enshrining the sequenceNumber as always being present since it always changes

tynes: We cannot enshrine the sequence number because we need different tx hashes each time

specs/fjord/diff-based-attributes-tx.md
```diff
@@ -0,0 +1,62 @@
...
+The following fields are required as part of the `L1Attributes` transaction after
+the Ecotone network upgrade.
+
+| Field             | Size (bytes) | Bitfield Representation  |
+| ----------------- | ------------ | ------------------------ |
+| baseFeeScalar     | 4            | 0b1000\_0000\_0000\_0000 |
+| blobBaseFeeScalar | 4            | 0b0100\_0000\_0000\_0000 |
+| sequenceNumber    | 8            | 0b0010\_0000\_0000\_0000 |
+| l1BlockTimestamp  | 8            | 0b0001\_0000\_0000\_0000 |
+| l1BlockNumber     | 8            | 0b0000\_1000\_0000\_0000 |
```
mds1, line 43: Can these be merged into a single bit, because on L1 if one of these changes the other must also change? This also gets you down to 8 bits, so you can use 1 byte. `l1BlockHash` can be merged into this too, bringing it down to 7 total bits

tynes: Good looks, you are right

tynes: I added in a note for this, thank you

PR #14, opened 2024-01-30
feat: fjord

tynes:
**Description**

Add placeholder for coordination of the fjord network upgrade.
Experimental and meant to centralize discussion of what goes into
the fjord network upgrade in a single place and ahead of time.

Also using an experiment directory structure where the network
upgrade name is a directory and includes all of the changes for
that particular upgrade. It would be good to refactor the rest of
the repo to follow this standard so it is easy to know exactly what
is changing between network upgrades. At the moment, the information
is spread out across many files.

### DIFF
--- a/specs/fjord/overview.md
+++ b/specs/fjord/overview.md
@@ -1,0 +1,7 @@
+# Fjord Network Upgrade
+
+This document is not finalized and should be considered experimental.
+
+## Execution Layer
+
+- [RIP-7212: Precompile for secp256r1 Curve Support](https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7212.md)


K-Ho:
I would strongly advocate for https://github.com/ethereum-optimism/op-geth/pull/202 FastLZ compression

As of Jan 2024, OP Mainnet has been negative on L1 Fees (due to worldcoin pre-compressing), Base has been fine, but lower volume chains have also been negative historically. Because we don't take compressability into account when we charge fees, we have teams like Worldcoin taking advantage of this and pre-compressing. We currently lose money on pre-compressed transactions, but make money on uncompressed transactions, but ideally our fees take into account compression, and we make a bit of money on *all* transactions whether or not they are pre-compressed. If we don't do this, there is real risk of all teams taking advantage of precompression, and causing OP Chains like OP Mainnet to unintentionally lose money. This uncertainty in whether chains will be making or losing on their L1 Fees makes chain operators' financial modelling difficult

PR #21, opened 2024-02-01
fault-proof: Specify bond incentives

Inphi:
Update the Fault Proof specification to include bonds.

The current FDG spec is beefy enough to understand, so I've opted to specify bonds as a "mod" or addon to the game. I think this makes it easier for readers to grok.

As promised, this also includes changes to the honest challenger behavior to deal with freeloader claims. The revised honest challenger spec is straight forward. But we may want to sketch out its proof of correctness somewhere, though not sure where exactly.

### DIFF
--- a/specs/experimental/fault-proof/bond-incentives.md
+++ b/specs/experimental/fault-proof/bond-incentives.md
@@ -1,0 +1,30 @@
+# Bond Incentives
+
+Bonds is an add-on to the core [Fault Dispute Game](./fault-dispute-game.md). The core game mechanics are
+designed to ensure honesty as the best response to winning subgames. By introducing financial incentives,
+Bonds makes it worthwhile for honest challengers to participate.
+
+- Moves
+- Subgame Resolution
+
+## Moves
+
+Moves must be adequately bonded to be added to the FDG. This document does not specify a
+scheme for determining the minimum bond requirement. FDG implementations should define a function
+computing the minimum bond requirement with the following:
+
+```solidity
+function getRequiredBond(Position _movePosition) public pure returns (uint256 requiredBond_)
+```
+
+As such, attacking or defending requires a check for the `getRequiredBond()` amount against the bond
+attached to the move. To incentivize participation, the minimum bond should cover the cost of a possible
+counter to the move being added. Thus, the minimum bond depends only on the position of the move that's added.
+
+## Subgame Resolution
+
+If a subgame root resolves incorrectly, then its bond is distributed to the **leftmost claimant** that countered
+it. This creates an incentive to identify the earliest point of disagreement in an execution trace.
+
+At maximum game depths, where a claimant counters a bonded claim via `step`, the bond is instead distributed
+to the account that successfully called `step`.
--- a/specs/experimental/fault-proof/bond-incentives.md
+++ b/specs/experimental/fault-proof/bond-incentives.md
@@ -11,7 +11,7 @@ Bonds makes it worthwhile for honest challengers to participate.
 
 Moves must be adequately bonded to be added to the FDG. This document does not specify a
 scheme for determining the minimum bond requirement. FDG implementations should define a function
-computing the minimum bond requirement with the following:
+computing the minimum bond requirement with the following signature:
 
 ```solidity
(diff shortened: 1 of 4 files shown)


specs/experimental/fault-proof/bond-incentives.md
```diff
@@ -0,0 +1,30 @@
...
+Bonds makes it worthwhile for honest challengers to participate.
+
+- Moves
+- Subgame Resolution
+
+## Moves
+
+Moves must be adequately bonded to be added to the FDG. This document does not specify a
+scheme for determining the minimum bond requirement. FDG implementations should define a function
+computing the minimum bond requirement with the following:
```
refcell: > with the following

With the following `signature`?

specs/experimental/fault-proof/bond-incentives.md
```diff
@@ -0,0 +1,31 @@
+# Bond Incentives
+
+Bonds is an add-on to the core [Fault Dispute Game](./fault-dispute-game.md). The core game mechanics are
+designed to ensure honesty as the best response to winning subgames. By introducing financial incentives,
+Bonds makes it worthwhile for honest challengers to participate.
+
+- Moves
+- Subgame Resolution
+
+## Moves
```
tynes, line 17: How is a bond sent? Which contract is it sent to?

specs/experimental/fault-proof/bond-incentives.md
```diff
@@ -0,0 +1,31 @@
+# Bond Incentives
+
+Bonds is an add-on to the core [Fault Dispute Game](./fault-dispute-game.md). The core game mechanics are
+designed to ensure honesty as the best response to winning subgames. By introducing financial incentives,
+Bonds makes it worthwhile for honest challengers to participate.
```
tynes, line 5: What happens without bonds?

specs/experimental/fault-proof/bond-incentives.md
```diff
@@ -0,0 +1,31 @@
...
+
+```solidity
+function getRequiredBond(Position _movePosition) public pure returns (uint256 requiredBond_)
+```
+
+As such, attacking or defending requires a check for the `getRequiredBond()` amount against the bond
+attached to the move. To incentivize participation, the minimum bond should cover the cost of a possible
+counter to the move being added. Thus, the minimum bond depends only on the position of the move that's added.
+
+## Subgame Resolution
```
refcell, line 31: I think this is missing discussion on the freeloader vs opposing teams as mentioned in `specs/experimental/fault-proof/honest-challenger-fdg.md`

specs/experimental/fault-proof/bond-incentives.md
Inphi: Intentionally unspecified. But the answer is the FDG contract receives the bond.

These specs describe the FDG at a high level. There are a couple cases where it aids user understanding to get into the details of contract implementation. But I don't think this is one of them. A different implementation of the FDG may decide to use a separate contract to store bonds.

specs/experimental/fault-proof/bond-incentives.md
Inphi: Without bonds, honest actors will be guaranteed to win the game. However, it'll be very expensive to do so, given the costs of making transactions and running cannon, as they don't get compensated for participating.

specs/experimental/fault-proof/bond-incentives.md
Inphi: Yeah, I'll add a note on this. I think a section that explicitly notes what happens when there are multiple successful counters should cover this.

Inphi: Added a section on the incentives implied by the leftmost payoff rule. It covers what claims are defensible and why an honest challenger will want to ensure their claim is the unique leftmost claim.

tynes:
Would super appreciate if you added the responses to the doc as it will help to scale context :D

clabby (approved):

ajsutton (approved):
LGTM.

specs/experimental/fault-proof/bond-incentives.md
Inphi: Added a note on this to the specs Thanks for the review.

PR #25, opened 2024-02-02
fjord: Add FastLZ L1-Cost function specs

mdehoog:
**Description**

Specs for the FastLZ L1 cost function changes, targeting Fjord. See https://hackmd.io/@mdehoog/compression-analysis for a detailed write up.

Related: 
* https://github.com/ethereum-optimism/op-geth/pull/249
* https://github.com/ethereum-optimism/optimism/pull/10534
* https://github.com/ethereum-optimism/optimism/pull/9618


specs/fjord/exec-engine.md
```diff
@@ -0,0 +1,97 @@
...
+The L1 cost is computed as:
+
+```
+l1FeeScaled = l1BaseFeeScalar*l1BaseFee*16 + l1BlobFeeScalar*l1BlobBaseFee
+l1CostSigned = (intercept + fastlzCoef*fastlzSize + uncompressedTxCoef*uncompressedTxSize) * l1FeeScaled / 1e12
+l1Cost = uint256(max(0, l1CostSigned))
+```
+
+Where:
+
```
tynes, line 33: This could be good to have as a markdown table that defines variables up front. It could include their type and a description

tynes: For values that are constants, you could define them as constants up front similar to this: https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#constants

Then a rationale section would be great to explain them

tynes:
Are we going to want to adjust `intercept`, `fastlzCoef`,  and `uncompressedTxCoef` dynamically? Or do you think that hardcoding these values for all networks will be sufficient? If we want to make them dynamic, then we will need a file `system-config.md` that describes the setters and a `l1-attributes.md` that describes the updates to the L1 Attributes transaction. We just optimized the L1 Attributes tx as part of ecotone to make it smaller so I am fine with adding new fields to it, I also have a diff based proposal here that could be helpful in allowing us to add many possible fields to the L1 Attributes tx without it adding bloat to the history: https://github.com/ethereum-optimism/specs/pull/13

specs/protocol/deposits.md
```diff
@@ -287,6 +286,50 @@ and also set the following new attributes:
...
+
+In the first L2 block after the Fjord activation block, the Fjord L1 attributes are first used.
+
+The pre-Fjord values are migrated over 1:1.
+Blocks after the Fjord activation block contain all pre-Fjord values 1:1,
+and also set the following new attributes:
+
+- The `l1CostIntercept` is set to `0`.
+- The `l1CostFastlzCoef` is set to `1e6`.
+- The `l1CostTxSizeCoef` is set to `0`.
```
mdehoog: should we default these values to the model from the linear regression?

specs/protocol/exec-engine.md
```diff
@@ -156,6 +156,98 @@ The two base fee values and their respective scalars can be accessed in two inte
...
+```
+
+As an example, we generated a dataset from all transactions on Optimism Mainnet in October 2023,
+and the resulting linear regression was:
+
+`-27.321890037208703 + 1.03146206*fastlzSize - 0.08866427*txSize`
+
+Scaling these values by `1e6` gives the constants used in the L1-Cost fee estimator:
+
+- `intercept = -27_321_890`
```
trianglesphere: How much precision do we want / need here?

mdehoog: we are currently proposing signed 32-bit integers, as we don't see a need to go above 2^31 for any of these values

mdehoog:
> Are we going to want to adjust `intercept`, `fastlzCoef`, and `uncompressedTxCoef` dynamically? Or do you think that hardcoding these values for all networks will be sufficient? If we want to make them dynamic, then we will need a file `system-config.md` that describes the setters and a `l1-attributes.md` that describes the updates to the L1 Attributes transaction. We just optimized the L1 Attributes tx as part of ecotone to make it smaller so I am fine with adding new fields to it, I also have a diff based proposal here that could be helpful in allowing us to add many possible fields to the L1 Attributes tx without it adding bloat to the history: #13

~We plan to make then configurable. I've updated the relevant parts of the spec.~

EDIT: we met about this and decided to derisk the upgrade by hardcoding the constants for now. We may want to introduce ability for this to be configured, either by L1 attributes or by putting the logic in an L2 contract, but for now hardcoding is fine.

specs/protocol/deposits.md
```diff
@@ -287,6 +286,50 @@ and also set the following new attributes:
...
+| l1BlockNumber     | uint64  | 28-35          |         |
+| basefee           | uint256 | 36-67          | 2       |
+| blobBaseFee       | uint256 | 68-99          | 3       |
+| l1BlockHash       | bytes32 | 100-131        | 4       |
+| batcherHash       | bytes32 | 132-163        | 5       |
+| l1CostIntercept   | int32   | 164-167        | 6       |
+| l1CostFastlzCoef  | int32   | 168-171        |         |
+| l1CostTxSizeCoef  | int32   | 172-175        |         |
+
+Total calldata length MUST be exactly 176 bytes. This helps to slow database growth as every
```
tynes: The second sentence here doesn't feel useful. `This helps to slow database growth as...` not super clear what the alternative is? It also *adds* history size based on ecotone

tynes:
Generally looks good

specs/protocol/deposits.md
mdehoog: agreed (this is a copy/paste from above), i will remove it

tynes:
This looks great! Thank you. We would like to modularize the specs a bit so the diff that comes with a particular network upgrade is very obvious. If you could rebase and move the new specs into the `fjord` directory, that would be great. We will be going through and making directories for each network upgrade in the future

specs/fjord/deposits.md
```diff
@@ -0,0 +1,93 @@
...
+meaning that `setL1BlockValuesFjord` is not guaranteed to exist yet.
+
+Every subsequent L1 Attributes transaction should include a call to the `setL1BlockValuesFjord()` function.
+The input args are no longer ABI encoded function parameters,
+but are instead packed into 5 32-byte aligned segments and one 12-byte segment (starting after the function selector).
+Each signed and unsigned integer argument is encoded as big-endian using a number of bytes corresponding to the
+underlying type. The overall calldata layout is as follows:
+
+[fjord-upgrade-txs]: derivation.md#fjord
+
```
tynes: This is ignoring the diff based L1 attributes design. How do you feel about using the diff based design?

mdehoog: As discussed in the meeting, we simplified and removed the L1 attributes altogther

specs/fjord/predeploys.md
```diff
@@ -16,4 +16,23 @@ Following the Fjord upgrade, three additional values used for L1 fee computation
...
 - costTxSizeCoef
 
-These values are hard-coded constants in the `GasPriceOracle` contract.
+These values are hard-coded constants in the `GasPriceOracle` contract. The
+calculation follows the same formula outlined in the
+[Fjord L1-Cost fee changes (FastLZ estimator)](./exec-engine.md#fjord-l1-cost-fee-changes-fastlz-estimator)
+section.
+
+A new method is introduced: `getL1FeeUpperBound(uint256)`. This method returns an upper bound for the L1 fee
+for a given transaction size. It is provided for callers who wish to estimate L1 transaction costs in the
```
roberto-bayardo, line 26: I think you want to clarify this is for the *signed* transaction size?   Maybe also add that if it's unsigened, users should add 68 bytes to the length.

mdehoog: thanks, fixed 👍 (made it unsigned)

specs/fjord/predeploys.md
```diff
@@ -0,0 +1,39 @@
...
+write path, and is much more gas efficient than `getL1Fee`. It assumes the worst case of no compression,
+implemented as follows:
+
+```solidity
+function getL1FeeUpperBound(uint256 unsignedTxSize) external view returns (uint256) {
```
mdehoog, line 35: this may also be helpful for protocols that currently depend on constants directly and are doing their own L1 fee calculation (example: https://sips.synthetix.io/sips/sip-356/)

roberto-bayardo (approved):

specs/fjord/exec-engine.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+| Input arg         | Type      | Description                                                       |
+|-------------------|-----------|-------------------------------------------------------------------|
+| `l1BaseFee`       | `uint256` | L1 base fee of the latest L1 origin registered in the L2 chain    |
+| `l1BlobBaseFee`   | `uint256` | Blob gas price of the latest L1 origin registered in the L2 chain |
+| `fastlzSize`      | `uint256` | Length of the FastLZ-compressed RLP-encoded signed tx             |
+| `txSize`          | `uint256` | Length of uncompressed RLP-encoded signed tx                      |
+| `l1BaseFeeScalar` | `uint32`  | L1 base fee scalar, scaled by `1e6`                               |
+| `l1BlobFeeScalar` | `uint32`  | L1 blob fee scalar, scaled by `1e6`                               |
```
sebastianst: In this PR, you correct the fee scalar variable names in `specs/protocol/predeploys.md` to `baseFeeScalar` and `blobBaseFeeScalar`, because that's how they are called in `system_config.md`. Do you on purpose deviate from those names here?

specs/fjord/exec-engine.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### Fjord L1-Cost fee changes (FastLZ estimator)
+
+Fjord updates the L1 cost calculation function to use a FastLZ-based compression estimator.
+The L1 cost is computed as:
+
+```pseudocode
+l1FeeScaled = l1BaseFeeScalar*l1BaseFee*16 + l1BlobFeeScalar*l1BlobBaseFee
+l1CostSigned = (intercept + fastlzCoef*fastlzSize + txSizeCoef*txSize) * l1FeeScaled / 1e12
+l1Cost = uint256(max(0, l1CostSigned))
```
sebastianst: We're trying to prove that we always produce values >0. If we cannot do this, we need to reconsider which minimum we set here.

https://github.com/ethereum-optimism/protocol-quest/issues/211

specs/fjord/exec-engine.md
```diff
@@ -0,0 +1,114 @@
...
+| `fastlzCoef`      | `int32`   | FastLZ coefficient, scaled by `1e6` (can be negative)             |
+| `txSizeCoef`      | `int32`   | Transaction size coefficient, scaled by `1e6` (can be negative)   |
+
+Where:
+
+- the `intercept` is hard-coded as `-27_321_890`
+
+- the `fastlzCoef` is hard-coded as `1_031_462`
+
+- the `txCoef` is hard-coded as `-88_664`
```
sebastianst: We're reevaluating if these numbers are still good for the tx makeup we see now post-Ecotone.

https://github.com/ethereum-optimism/protocol-quest/issues/210

specs/fjord/derivation.md
```diff
@@ -0,0 +1,133 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Deriving the Transaction List
+
+### Network upgrade automation transactions
+
+#### Fjord
+
+The Fjord hardfork activation block, contains the following transactions in this order:
```
sebastianst:
```suggestion
The Fjord hardfork activation block contains the following transactions, in this order:
```
feel free to also fix it in the Ecotone text.

specs/fjord/exec-engine.md
```diff
@@ -0,0 +1,114 @@
...
+- `fastlzCoef = 1_031_462`
+- `txCoef = -88_664`
+
+Note that the linear regression takes into account the current compression ratio, so the
+scalars `l1BaseFeeScalar` and `l1BlobFeeScalar` should be adjusted to account for any previous
+compression ratio they encoded. For example, if the previous compression ratio was `0.684`, then
+they should be divided by `0.684` to get the new scalars, e.g.:
+
+- `l1BaseFeeScalar = 11_111` (`~= 7600/0.684`)
+- `l1BlobFeeScalar = 1_250_000` (`~= 862000/0.684`)
```
roberto-bayardo: Where do the 7600 & 862000 constants come from? Are those just hypothetical "previous" scalar values?

roberto-bayardo: Also I guess when Fjord goes live, we'll be overcharging l1 fees a bit until the scalars are adjusted in this manner?

specs/fjord/exec-engine.md
```diff
@@ -0,0 +1,114 @@
...
+
+1. Calculate the RLP-encoded transaction payload. Record the length of this payload as `txSize`.
+2. Compress the payload using FastLZ. Record the length of the compressed payload as `fastlzSize`.
+3. Compress the payload using a "best estimate" compression, defined as:
+
+- Create two zlib streams, both using the maximum compression level.
+- Bootstrap one of the streams with at least 64kb of transaction data (after compression).
+- Write each transaction to both streams (flushing after each write), and record the size increase
+  of the larger buffer as `bestEstimateSize`.
+- Each time the larger buffer grows larger than 128kb, empty the stream and swap the buffers.
```
roberto-bayardo: FWIW Now that we batch 6 blobs at a time we found that you get about 2% more compression, which is probably not enough difference to affect the results much, but wondering if we should confirm that by increasing this to 128kb*6.

specs/fjord/derivation.md
```diff
@@ -0,0 +1,133 @@
...
+
+```bash
+cast compute-address --nonce=0 0x4210000000000000000000000000000000000002
+Computed Address: 0xa919894851548179A0750865e7974DA599C0Fac7
+```
+
+Verify `sourceHash`:
+
+```bash
+❯ cast keccak $(cast concat-hex 0x0000000000000000000000000000000000000000000000000000000000000002 $(cast keccak "Fjord: Gas Price Oracle Deployment"))
```
sebastianst:
```suggestion
cast keccak $(cast concat-hex 0x0000000000000000000000000000000000000000000000000000000000000002 $(cast keccak "Fjord: Gas Price Oracle Deployment"))
```

specs/fjord/exec-engine.md
sebastianst: Would be great to add a short explanation why this is a good process to get to a "best estimate"

specs/fjord/exec-engine.md
sebastianst: We're reevaluating this, taking into account larger 768kB 6-blob channels with better compression ratios.

specs/static/bytecode/fjord-gas-price-oracle-deployment.txt
sebastianst, line 1: Confirmed that this matches the `Bin` field of the GPO bindings in the monorepo PR. We need to adjust this if we change the model params.

sebastianst (commented):
This looks great! Just need to confirm the model parameters and possibly adjust them here.

specs/fjord/derivation.md
```diff
@@ -0,0 +1,133 @@
...
+transactions by setting `noTxPool: true`.
+
+##### GasPriceOracle Deployment - Fjord
+
+The `GasPriceOracle` contract is upgraded to support the new Fjord L1 data fee computation. Post fork this contract
+will use FastLZ to compute the L1 data fee.
+
+To perform this upgrade, a deposit transaction is derived with the following attributes:
+
+- `from`: `0x4210000000000000000000000000000000000002`
```
sebastianst, line 169: [Implementation](https://github.com/ethereum-optimism/optimism/pull/9618/files#diff-af006f09ce5a43a4503f2e664dcf7536cd393a36ef99582c32adb20d65c4d45fR17) uses `...3`. Changing it here has some downstream effects on other parts of the spec.

specs/fjord/exec-engine.md
xiaoxiaff: I believe those value where the value when this PR is created. Given it is being fetched live through the contract, these value is expected to change.

specs/fjord/derivation.md
danyalprout: Great catch, the implementation is wrong -- I'll update that PR to match.

specs/fjord/exec-engine.md
```diff
@@ -22,7 +22,8 @@ The L1 cost is computed as:
 
 ```pseudocode
 l1FeeScaled = l1BaseFeeScalar*l1BaseFee*16 + l1BlobFeeScalar*l1BlobBaseFee
-l1CostSigned = (intercept + fastlzCoef*fastlzSize + txSizeCoef*txSize) * l1FeeScaled / 1e12
+estimatedLen = intercept + fastlzCoef*fastlzSize 
+l1CostSigned = max(71, estimatedLen) * l1FeeScaled / 1e12
```
sebastianst: I'd prefer to make the `71` a named parameter in the specs, like `fastlzMinCost`. It might also still change.

I think we can also now rename `l1CostSigned` to `l1Cost` because we now already guarantee in its definition that it will be > 0.

specs/fjord/exec-engine.md
```diff
@@ -0,0 +1,79 @@
...
+
+We generated a dataset from two weeks of post-Ecotone transactions on Optimism Mainnet, as we found that was
+the most representative of performance across multiple chains and time periods. More details on the linear regression
+and datasets used can be found in this [repository](https://github.com/roberto-bayardo/compression-analysis/tree/main).
+
+### L1 Gas Usage Estimation
+
+The `L1GasUsed` property on the transaction receipt is updated to take into account the improvement in
+[compression estimation](./exec-engine.md#fees) accuracy. The value will be calculated by
+multiplying the `estimatedSize` of the transaction from the above L1 cost formula by 16. The value of 16, assumes most
```
sebastianst:
```suggestion
multiplying the `estimatedSize` of the transaction from the above L1 cost formula by 16. The value of 16 assumes most
```

specs/fjord/exec-engine.md
```diff
@@ -0,0 +1,79 @@
...
+of the bytes in the compressed data are non-zero.
+
+The `L1GasUsed` property will be deprecated. Due to it not accurately calculating the L1 gas used
```
sebastianst:
```suggestion
The `L1GasUsed` property will be deprecated due to it not accurately calculating the L1 gas used
```

specs/fjord/exec-engine.md
```diff
@@ -0,0 +1,79 @@
...
+| `minTransactionSize` | `uint32`  | A lower bound on transaction size, in bytes                       | 100                      |
+
+Previously, `baseFeeScalar` and `blobFeeScalar` were used to encode the compression ratio, due to the inaccuracy of
+the L1 cost function. However, the new cost function takes into account the compression ratio, so these scalars should
+be adjusted to account for any previous compression ratio they encoded.
+
+##### FastLZ Implementation
+
+All compression algorithms must be implemented equivalently to this
+[commit](https://github.com/ariya/FastLZ/commit/344eb4025f9ae866ebf7a2ec48850f7113a97a42).
```
sebastianst: Maybe also mention explicitly that the implementation can be found in `fastlz.c` and also reference the function `fastlz_compress`.

specs/static/bytecode/fjord-gas-price-oracle-deployment.txt
sebastianst, line 1: This transaction and the tx in the monorepo PR at `op-node/rollup/derive/fjord_upgrade_transactions.go` match. But the hex code at `op-bindings/bindings/gaspriceoracle.go` (`Bin` field) differs.

specs/fjord/exec-engine.md
sebastianst: it's now described to vary

sebastianst (approved):
Looks great, thank you! Also confirmed that the new GPO deployment code matches the monorepo PR.

specs/fjord/predeploys.md
```diff
@@ -0,0 +1,77 @@
...
+      }
+
+      // Assume the compressed data is mostly non-zero, and would pay 16 gas per calldata byte
+      return estimatedSize * 16;
+    }
+    // ...
+}
+```
+
+The `getL1GasUsed` method will be deprecated. This is due to it not accurately estimating the
```
tynes, line 74: When you say deprecated, is it still expected to function? Or will it revert when called?

specs/static/bytecode/fjord-gas-price-oracle-deployment.txt
```diff
@@ -0,0 +1 @@
...006105c9565b8160018114611723576002811461172d57611749565b60019150506105c9565b60ff84111561173e5761173e6115a5565b50506001821b6105c9565b5060208310610133831016604e8410600b841016171561176c575081810a6105c9565b6117768383611690565b807fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff048211156117a8576117a86115a5565b029392505050565b6000610aba83836116f1565b6000808212827f7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff038413811516156117f6576117f66115a5565b827f800000000000000000000000000000000000000000000000000000000000000003841281161561182a5761182a6115a5565b50500190565b60007f7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff600084136000841385830485118282161615611871576118716115a5565b7f800000000000000000000000000000000000000000000000000000000000000060008712868205881281841616156118ac576118ac6115a5565b600087129250878205871284841616156118c8576118c86115a5565b878505871281841616156118de576118de6115a5565b505050929093029392505050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052603260045260246000fd5b60007fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff820361194c5761194c6115a5565b506001019056fea164736f6c634300080f000a
```
tynes: Where is the source code for this? I don't see it [here](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L2/GasPriceOracle.sol)

specs/fjord/predeploys.md
danyalprout: The intention is for it to continue to function, with the updated implementation above -- which takes the fastlz compression into account.

Our plan was to mark it as deprecated this network upgrade then update the implementation to revert in a future upgrade.

specs/static/bytecode/fjord-gas-price-oracle-deployment.txt
danyalprout: It's the bytecode of the updated GasPriceOracle from this [PR](https://github.com/ethereum-optimism/optimism/pull/9618/files#diff-7eeb2cdd3a766512a162b2921f45ae64a3ae91f99d7d408ca11defb6142f4ba5).

You can verify it by grabbing the latest commit/running forge build and grabbing the bytecode via:

```
cat forge-artifacts/GasPriceOracle.sol/GasPriceOracle.json | jq ".bytecode.object"
```

specs/fjord/predeploys.md
tynes: It could be noted that this function may revert in a future upgrade

tynes (approved):

specs/fjord/predeploys.md
danyalprout: Good call -- I'll add that 👍

trianglesphere (approved):

PR #27, opened 2024-02-04
experimental: transaction ordering policy

tynes:
**Description**

Explore what applying a predicate to the inputs as part of the
derivation pipeline would look like. The predicate is defined
using evm bytecode and can therefore be quite flexible based
on the chain operator's needs. The payload is dropped if the
predicate returns false.

The chain operator can configure the predicate's bytecode
on L1 using the system config contract.

In practice, this allows for static analysis of the `ExecutionPayload`
before execution. If some chain operator defined invariant
about the `ExecutionPayload` isn't held to be true, then
the payload is dropped. This can be used to enforce constraints
on the way that block builders can build blocks.

### DIFF
--- a/specs/experimental/tx-ordering-policy.md
+++ b/specs/experimental/tx-ordering-policy.md
@@ -1,0 +1,124 @@
+# Transaction Ordering Policy
+
+## Overview
+
+Enforcement of arbitrary transaction ordering rules can enable application specific rollups
+to operate more efficiently. The ability to further constrict the set of valid blocks
+based on the properties of the transactions included in the block can be called
+"enforcement of a transaction ordering policy". The enforcement can be done using a
+predicate that is opaque to the protocol itself, meaning that chain operators can
+design their own policies without needing to modify consensus.
+
+## Derivation Pipeline
+
+The derivation pipeline is modified to add a new stage in between the attributes queue and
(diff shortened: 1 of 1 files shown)


tynes:
It would be good to have a `view` function on the contract or in another predeploy that accepts a block body and returns a bool if the block is valid or not. This will help block builders to know they are creating valid blocks

tynes:
For this to work with deposits, dropped blocks would need to be mapped to deposits only blocks per https://github.com/ethereum-optimism/design-docs/pull/20

specs/experimental/tx-ordering-policy.md
```diff
@@ -0,0 +1,124 @@
...
+#### General Invariants
+
+General invariants do not observe anything particular about the applications that are called as part of the
+execution of a `PayloadAttributes`. An example invariant is that the transactions are exactly ordered by
+their tip. This sort of rule may eliminate some timing games, although searchers that learn information faster
+will always be able to use more information when determining how much to bid.
+
+### L1 Attributes Transaction
+
+The L1 Attributes transaction is assumed to be updated with a [diff based serialization](https://github.com/ethereum-optimism/specs/pulls/13).
```
emhane, line 107:
```suggestion
The L1 Attributes transaction is assumed to be updated with a [diff based serialization](https://github.com/ethereum-optimism/specs/pull/13).
```

PR #30, opened 2024-02-06
fix details of Ecotone scalar management in GasPriceOracle spec

roberto-bayardo:
**Description**

Fixes the details of how Ecotone scalars are managed by the GasPriceOracle predeploy.


tynes (approved):
Thank you!

tynes (approved):

PR #36, opened 2024-02-09
interop: describe messaging graph

protolambda:
**Description**

This expands the interop `messaging.md` with a description of how dependencies can be looked as as a graph, and how safety can be determined in terms of finding a graph component with specific properties.

This also defines how the graph may be pruned, how it is bounded (incl. message expiry, and definition of upper relevant timestamp as a horizon).

This PR is based on top of #6 

**Metadata**

Fix https://github.com/ethereum-optimism/protocol-quest/issues/138

### DIFF
--- a/specs/interop/messaging.md
+++ b/specs/interop/messaging.md
@@ -14,8 +14,16 @@
   - [Timestamp Invariant](#timestamp-invariant)
   - [ChainID Invariant](#chainid-invariant)
   - [Only EOA Invariant](#only-eoa-invariant)
-  - [Message Expiry](#message-expiry)
-- [Message dependency resolution](#message-dependency-resolution)
+  - [Message Expiry Invariant](#message-expiry-invariant)
+- [Message Graph](#message-graph)
+  - [Invalid messages](#invalid-messages)
+    - [Block reorgs](#block-reorgs)
+    - [Block Recommits](#block-recommits)
+  - [Intra-block messaging: cycles](#intra-block-messaging-cycles)
+  - [Resolving cross-chain safety](#resolving-cross-chain-safety)
+  - [Horizon timestamp](#horizon-timestamp)
+  - [Pruning the graph](#pruning-the-graph)
+  - [Bounding the graph](#bounding-the-graph)
 - [Security Considerations](#security-considerations)
   - [Cyclic dependencies](#cyclic-dependencies)
   - [Transitive dependencies](#transitive-dependencies)
@@ -149,6 +157,7 @@ This trade-off is in review. This invariant may be ignored in initial interop te
 
 The expiry invariant invalidates inclusion of any executing message with
 `id.timestamp + EXPIRY_TIME < executing_block.timestamp` where:
+
 - `id` is the [`Identifier`] encoded in the executing message, matching the block attributes of the initiating message.
 - `executing_block` is the block where the executing message was included in.
 - `EXPIRY_TIME = 180 * 24 * 60 * 60 = 15552000` seconds, i.e. 180 days.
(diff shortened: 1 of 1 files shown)


specs/interop/messaging.md
```diff
@@ -136,13 +149,121 @@ This feature could be added in a backwards compatible way by adding a new functi
 One possible way to handle explicit denial of service attacks is to utilize identity
 in iterated games such that the block builder can ban identities that submit malicious transactions.
 
-### Message Expiry
+### Message Expiry Invariant
+
+Note: Message Expiry as property of the protocol is in active discussion.
```
tynes, line 154: We cannot ignore this invariant but we can relax it and we do so via making `EXPIRY_TIME = 0`. In practice this applies in the case of a reorg, to ensure that the executing message always happens at the same time or in the future

specs/interop/messaging.md
```diff
@@ -136,13 +149,121 @@ This feature could be added in a backwards compatible way by adding a new functi
...
+When tracking message dependencies, edges are maintained for *all* identified source blocks.
+
+Reorgs are resolved by filtering the view of the solver to only canonical blocks.
+If the source block is not canonical, the dependency is invalid.
+
+The canonical L2 block at the identified block-height is the source of truth.
+
+#### Block Recommits
+
+Blocks may be partially built, or sealed but not published, and then rebuilt.
```
tynes, line 202: There is a lot of hidden context in this paragraph, it could be made more approachable

specs/interop/messaging.md
```diff
@@ -136,13 +149,121 @@ This feature could be added in a backwards compatible way by adding a new functi
...
+
+### Intra-block messaging: cycles
+
+While messages cannot be initiated by future blocks,
+they can be initiated by any transactions within the same timestamp,
+as per the [Timestamp Invariant](#timestamp-invariant).
+
+This property allows messages to form cycles in the graph:
+blocks with equal timestamp, of chains in the same dependency set,
+may have dependencies on one another.
```
tynes, line 215: The total ordering doesn't matter, an initiating message may have a larger transaction index than the executing message that references it given they have the same timestamp

specs/interop/messaging.md
```diff
@@ -136,13 +149,121 @@ This feature could be added in a backwards compatible way by adding a new functi
...
+I.e. the graph must not have any inward edges towards invalid blocks within the safety-view.
+
+A safety-view is the subset of canonical blocks of all chains with the specified safety label or a higher safety label.
+Dependencies on blocks outside of the safety-view are invalid,
+but may turn valid once the safety-view changes (e.g. a reorg of unsafe blocks).
+
+By resolving in terms of graph-components, cyclic dependencies and transitive dependencies are addressed.
+
+### Horizon timestamp
+
```
tynes, line 231: I am confused how this works in practice, like basically queue any blocks past a timestamp and then only begin to add them to the graph once the full graph with lower timestamps is solved as much as possible, only leaving a graph that has dependencies with a larger timestamp?

specs/interop/messaging.md
tynes: Ah i see this is in addition to the timestamp invariant

tynes (approved):

specs/interop/messaging.md
protolambda: Basically yes. Thanks to the timestamp invariant we can determine what is relevant to solving existing dependencies, and what can wait till later solving passes.

specs/interop/messaging.md
protolambda: Can clarify, but that property is already mentioned in the timestamp-invariant section, and I didn't want to duplicate it.

tynes:
I am going to merge this so that I can squash the feature branch, generally think this is good to go

PR #43, opened 2024-02-13
plasma: versioned commitments and max input size

tchardin:
In this PR:
- op-plasma commitment versioning using a byte type as prefix.
- enforcing a max input commitment size based on l1 calldata max size.

### DIFF
--- a/specs/experimental/plasma.md
+++ b/specs/experimental/plasma.md
@@ -68,13 +68,13 @@ to be submitted to the inbox address, the data is uploaded to the DA storage lay
 commitment (keccak256 hash) is submitted as the bacher inbox transaction call data.
 
 Commitments are encoded as `commitment_type_byte ++ commitment_bytes`, where `commitment_bytes` depends
-on the `commitment_type_byte` where [0, 128] are reserved for official implementations:
+on the `commitment_type_byte` where [0, 128) are reserved for official implementations:
 
 | `commitment_type` | `commitment`                    |
 | ----------------- | ------------------------------- |
 | 0                 | `keccak256(tx_payload)`         |
 
-The batcher SHOULD cap input payloads to the maximum L1 calldata size or risk the input to be skipped
+The batcher SHOULD cap input payloads to the maximum L1 tx size or risk the input to be skipped
 during derivation.
 
 The batcher SHOULD NOT submit a commitment onchain unless input data was successfully stored on the service.
@@ -190,7 +190,7 @@ In addition, an expired challenge will reorg out `[r_start, r_end]` L2 blocks so
 block derived from the expired challenge's input and `r_end` the last L2 block derived before the pipeline
 was reset.
 
-Derivation MUST skip input data such as `input_data_size > max_l1_calldata_size` to ensure data can be submitted
+Derivation MUST skip input data such as `input_data_size > max_l1_tx_size` to ensure data can be submitted
 onchain if challenged.
 
 [pipeline]: ../protocol/derivation.md#resetting-the-pipeline
--- a/specs/experimental/plasma.md
+++ b/specs/experimental/plasma.md
@@ -105,7 +105,7 @@ function challenge(uint256 challengedBlockNumber, bytes commitment) external pay
 ```
 
 - The L1 block number in which it was included.
-- Versionned commitment bytes (i.e. `0 ++ keccak256(frame.. )`)
+- Versioned commitment bytes (i.e. `0 ++ keccak256(frame.. )`)
 
 Users with access to the input data then have another window of L1 blocks (`resolveWindow`)
 during which they can submit it as calldata to the chain by calling the `resolve` method of the contract.
--- a/specs/experimental/plasma.md
+++ b/specs/experimental/plasma.md
@@ -74,8 +74,8 @@ on the `commitment_type_byte` where [0, 128) are reserved for official implement
 | ----------------- | ------------------------------- |
 | 0                 | `keccak256(tx_payload)`         |
 
-The batcher SHOULD cap input payloads to the maximum L1 tx size or risk the input to be skipped
-during derivation.
+The batcher SHOULD cap input payloads to the maximum L1 tx size or the input will be skipped
+during derivation. See [derivation section](#derivation) for more details.
 
 The batcher SHOULD NOT submit a commitment onchain unless input data was successfully stored on the service.
 In addition, a DA provider storage service SHOULD return an error response if it is unable to properly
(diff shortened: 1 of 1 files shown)


specs/experimental/plasma.md
```diff
@@ -67,6 +67,16 @@ The [batching][batcher] and compression of input data remain unchanged. When a b
 to be submitted to the inbox address, the data is uploaded to the DA storage layer instead, and a
 commitment (keccak256 hash) is submitted as the bacher inbox transaction call data.
 
+Commitments are encoded as `commitment_type_byte ++ commitment_bytes`, where `commitment_bytes` depends
+on the `commitment_type_byte` where [0, 128] are reserved for official implementations:
```
tynes: This should be `[0, 128)` to signify non inclusion of 128

specs/experimental/plasma.md
```diff
@@ -67,6 +67,16 @@ The [batching][batcher] and compression of input data remain unchanged. When a b
...
+
+| `commitment_type` | `commitment`                    |
+| ----------------- | ------------------------------- |
+| 0                 | `keccak256(tx_payload)`         |
+
+The batcher SHOULD cap input payloads to the maximum L1 calldata size or risk the input to be skipped
```
tynes: What is the maximum calldata size? There is a policy level for p2p gossip at max tx size but there is no max L1 calldata size besides buying 30 million gas with calldata. What do you mean?

tynes: I see that you added a line about the max size. It should be called something like max batch size since that is what it describes right? The commitments could be posted in blobs right? So we aren't necessarily tied to calldata

specs/experimental/plasma.md
```diff
@@ -180,6 +190,9 @@ In addition, an expired challenge will reorg out `[r_start, r_end]` L2 blocks so
 block derived from the expired challenge's input and `r_end` the last L2 block derived before the pipeline
 was reset.
 
+Derivation MUST skip input data such as `input_data_size > max_l1_calldata_size` to ensure data can be submitted
```
tynes: This should be included in a constants table that includes the value

specs/experimental/plasma.md
tchardin: @tynes this is the value I had in mind https://github.com/ethereum-optimism/optimism/blob/3a5fb5e854ad3c42080ec3fa0e5fc2c3585dffba/op-batcher/batcher/config.go#L55 it's set with the `max-l1-tx-size-bytes` flag so could just call it that but I believe the max batch size is different since it can be split across multiple txs.

specs/experimental/plasma.md
tchardin: Good catch, thanks!

specs/experimental/plasma.md
tchardin: Removed calldata naming, but I'm not sure what the value should be if constant since it is a parameter that can be tweaked for performance

specs/experimental/plasma.md
alvrs: In the case of `op-plasma` the limit would be the max input size to the `resolve` function that still fits in the 30 million gas limit (that includes the computation inside the function, so potentially a smaller limit than the non-`op-plasma` limit where we're just limited by `calldata`/`blob` cost)

specs/experimental/plasma.md
```diff
@@ -180,6 +190,9 @@ In addition, an expired challenge will reorg out `[r_start, r_end]` L2 blocks so
 block derived from the expired challenge's input and `r_end` the last L2 block derived before the pipeline
 was reset.
 
+Derivation MUST skip input data such as `input_data_size > max_l1_tx_size` to ensure data can be submitted
```
tynes: Can you define the exact value of `max_l1_tx_size`?

specs/experimental/plasma.md
```diff
@@ -67,6 +67,16 @@ The [batching][batcher] and compression of input data remain unchanged. When a b
...
 commitment (keccak256 hash) is submitted as the bacher inbox transaction call data.
 
+Commitments are encoded as `commitment_type_byte ++ commitment_bytes`, where `commitment_bytes` depends
+on the `commitment_type_byte` where [0, 128) are reserved for official implementations:
+
+| `commitment_type` | `commitment`                    |
+| ----------------- | ------------------------------- |
+| 0                 | `keccak256(tx_payload)`         |
+
+The batcher SHOULD cap input payloads to the maximum L1 tx size or risk the input to be skipped
```
tynes: "or risk" makes it sound like the skipping is optional but it is not optional according to later in the spec

specs/experimental/plasma.md
tynes: I think the default value for `max-l1-tx-size-bytes` makes sense altho it would make it into a consensus parameter rather than being a config option. Perhaps we want to add something to the derivation pipeline and the batcher that checks if its in plasma mode, this adds additional branching logic which isn't ideal but i don't know if we want to add this consensus restriction to rollup mode. The batcher should check the size before the call to `PlasmaDA.SetInput` and error if its larger than that and then maybe on start check the configured param isn't larger than the default  (now consensus) param. Would also need to add something to the derivation pipeline on the get call to check the size of the returned data and throw it out if its too large. This is a slightly annoying edge case to have to test for

specs/experimental/plasma.md
tynes: The max size has to be a consensus constant that is enforced in the derivation pipeline to guarantee that everything can be challenged. Otherwise you need to use a streaming keccak implementation like in the dispute game which is complex

specs/experimental/plasma.md
tchardin: Added.

specs/experimental/plasma.md
tchardin: Removed

specs/experimental/plasma.md
tchardin: Added a consensus constant based on Geth's maxTxSize so any batch can be easily challenged/resolved. It could be updated later in a hardfork if that max value is eventually raised.

specs/experimental/plasma.md
```diff
@@ -180,6 +190,11 @@ In addition, an expired challenge will reorg out `[r_start, r_end]` L2 blocks so
 block derived from the expired challenge's input and `r_end` the last L2 block derived before the pipeline
 was reset.
 
+Derivation MUST skip input data such as `input_data_size > MAX_L1_TX_SIZE` where `MAX_L1_TX_SIZE` is a consensus
+constant of 131072 bytes. In theory `MAX_L1_TX_SIZE` could be increased up to
+`(tx_gas_limit - fixed_resolution_cost) / dynamic_resolution_cost` based on the cost of resolving challenges in
+the contract implementation however to make challenging accessible it is capped based on geth's maxTxSize.
```
trianglesphere:
```suggestion
the contract implementation however to make challenging accessible it is capped based on geth's txMaxSize.
```
`txMaxSize` is the grep'able constant in geth.

trianglesphere (approved):
one tiny wording thing but LGTM

specs/experimental/plasma.md
tchardin: Updated, thanks!

PR #45, opened 2024-02-14
Add Buffer for Extra Data

MSilb7:
Add buffer in `compressed tx size` for extra data, per: https://github.com/ethereum-optimism/optimism/blob/dad21c03834a3164ae7fbba2850dfd29b5e4dea7/packages/contracts-bedrock/src/L2/GasPriceOracle.sol#L159

cc @K-Ho

### DIFF
--- a/specs/protocol/exec-engine.md
+++ b/specs/protocol/exec-engine.md
@@ -164,7 +164,7 @@ Where:
   header of the latest registered L1 origin block.
 
 Conceptually what the above function captures is the formula below, where `compressedTxSize =
-(zeroes*4 + ones*16) / 16` can be thought of as a rough approximation of how many bytes the
+(zeroes*4 + ones*16 + 68*16) / 16` can be thought of as a rough approximation of how many bytes the
 transaction occupies in a compressed batch.
 
 `(compressedTxSize) * (16*l1BaseFee*lBaseFeeScalar + l1BlobBaseFee*l1BlobBaseFeeScalar) / 1e6`


sebastianst:
@MSilb7 given our recent discussions about this, is this PR still relevant?

tynes:
cc @MSilb7 @sebastianst should we still consider this PR?

specs/protocol/exec-engine.md
```diff
@@ -164,7 +164,7 @@ Where:
   header of the latest registered L1 origin block.
 
 Conceptually what the above function captures is the formula below, where `compressedTxSize =
-(zeroes*4 + ones*16) / 16` can be thought of as a rough approximation of how many bytes the
+(zeroes*4 + ones*16 + 68*16) / 16` can be thought of as a rough approximation of how many bytes the
```
emhane, line 167: where did 68 come from? seems like magic number

emhane:
> cc @MSilb7 @sebastianst should we still consider this PR?

it has been merged to impl, so it seems we must https://github.com/ethereum-optimism/optimism/blob/6d9d43cb6f2721c9638be9fe11d261c0602beb54/packages/contracts-bedrock/src/L2/GasPriceOracle.sol#L236-L250

specs/protocol/exec-engine.md
sebastianst: I think that's an estimate of the overhead data of a tx, mostly coming from its signature?

protolambda: In some places the 68 is hardcoded, to estimate cost before signing. But after the tx has been signed and encoded there's no need to add the 68. So it completely depends on context whether or not it's a necessary addition.

PR #47, opened 2024-02-16
interop: fix  underspecification on predeploys

0xfuturistic:
nit: isInDependencySet(uint256) should return true IF AND ONLY IF the chain id is in the dependency set. currently the specs would mean that the function returns true if the chain id is in the set but possibly also if the chain id is not.

### DIFF
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -230,7 +230,8 @@ function relayMessage(uint256 _destination, uint256 _nonce, address _sender, add
 The `L1Block` contract is updated to include the set of allowed chains. The L1 Attributes transaction
 sets the set of allowed chains. The `L1Block` contract MUST provide a public getter to check if a particular
 chain is in the dependency set called `isInDependencySet(uint256)`. This function MUST return true when
-the chain's chain id is passed in as an argument.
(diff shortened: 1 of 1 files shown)


specs/interop/predeploys.md
```diff
@@ -230,7 +230,8 @@ function relayMessage(uint256 _destination, uint256 _nonce, address _sender, add
 The `L1Block` contract is updated to include the set of allowed chains. The L1 Attributes transaction
 sets the set of allowed chains. The `L1Block` contract MUST provide a public getter to check if a particular
 chain is in the dependency set called `isInDependencySet(uint256)`. This function MUST return true when
-the chain's chain id is passed in as an argument.
+the chain's chain id, passed as an argument, is in the dependency and FALSE when it isn't in there. 
+Additionally, this ensures that the function returns FALSE when no chain id is passed in.
```
tynes, line 234: When you say no chain id, do you mean when chainid=0? If no chainid is passed in, the bytecode will revert because of abi decoding issues

tynes:
@0xfuturistic thoughts on this PR? I am not sure if it is necessary. If you think it is then can you reword it a bit and address my comment?

PR #49, opened 2024-02-16
interop: keep predeploy addresses sequention

mds1:
There was no predeploy at `0x42...23`, which @protolambda said may have been previously used for an obsoleted predeploy. So this shifts L2ToL2CrossDomainMessenger down to `0x42...23` to keep addresses sequential


protolambda (approved):
Thanks!

tynes (approved):

PR #52, opened 2024-02-21
feat: Stage one bridge integration specs

clabby:
## Overview

Adds specifications for the modifications to the `OptimismPortal` performed in the stage one project.


Inphi (approved):
LGTM! Very nice to see this in the spec.

PR #53, opened 2024-02-21
feat: Add bond formula and fallback mechanism to bond specs

clabby:
## Overview

Expands the bond incentives spec to add information about the bond formula as well as the fallback mechanism to break incentive compatibility if a game resolves incorrectly during the stage one phase of the fault proof system.

### DIFF
--- a/specs/experimental/fault-proof/stage-one/bond-incentives.md
+++ b/specs/experimental/fault-proof/stage-one/bond-incentives.md
@@ -131,5 +131,3 @@ function claimCredit(address _recipient) external {
     if (!success) revert BondTransferFailed();
 }
 ```
-
-This fallback mechanism will be removed in the stage two project.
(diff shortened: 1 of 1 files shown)


Inphi (commented):
The bond formula presented here is only for the cost of countering a position.

But this doesn't include other bond incentives including the minimum deterrence deposit and over-collateralization factor. Do we want to specify those here as well?
See this [post](https://github.com/ethereum-optimism/client-pod/issues/551#issue-2121652375) for the details.

specs/experimental/fault-proof/stage-one/bond-incentives.md
```diff
@@ -58,3 +97,39 @@ are located on the same level as honest claims, but situated to its right (i.e.
...
+        credit[_recipient] = 0;
+    }
+
+    // Transfer the credit to the recipient.
+    (bool success,) = _recipient.call{ value: recipientCredit }(hex"");
+    if (!success) revert BondTransferFailed();
+}
+```
+
+This fallback mechanism will be removed in the stage two project.
```
ajsutton: We should not make promises about future changes. Just need to document how things work at this stage.

ajsutton (approved):
To Mofi's comment I think we do want to include things like overcollateralisation here as well.  I'm happy to merge this as is and then add those in separately if that's easier or if they're not set yet.

clabby:
Closing in favor of #60

PR #57, opened 2024-02-22
fault-proofs: Fix freeloader handling

Inphi:
Fixes two issues in the honest challenger spec:

1. Honest claims were misidentified as freeloaders. If the claim is the one an honest challenger would have made, then it should be ignored.
2. The recursion used to generate the correct response always fails to support an invalid path (by design). But in the freeloader case, we know it's safe to support such paths.

The algorithm also contains fixes to properly identify freeloaders at any depth. Freeloaders are identified by enumerating the set of correct responses, and checking whether the claim we're responding to is a leftmost sibling of a correct response. Once the freeloader is identified, we greedily respond, ignoring path validity.

### DIFF
--- a/specs/experimental/fault-proof/stage-one/honest-challenger-fdg.md
+++ b/specs/experimental/fault-proof/stage-one/honest-challenger-fdg.md
@@ -115,6 +115,8 @@ class Response(Enum):
 
 MAX_TRACE = 2**MAX_GAME_DEPTH
 
+ALL_CLAIMS: List[Claim] = []
+
 def agree_with(claim: Claim, chal_trace: List[ClaimHash, MAX_TRACE]) -> bool:
     if chal_trace[trace_index(claim.position)] != claim.claim_hash:
         return False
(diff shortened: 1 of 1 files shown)


specs/experimental/fault-proof/stage-one/honest-challenger-fdg.md
```diff
@@ -115,6 +115,8 @@ class Response(Enum):
 
 MAX_TRACE = 2**MAX_GAME_DEPTH
 
+ALL_CLAIMS: List[Claim] = []
```
Inphi, line 118: placeholder for all claims in a game.

specs/experimental/fault-proof/stage-one/honest-challenger-fdg.md
```diff
@@ -132,17 +134,32 @@ def respond_claim(claim: Claim, correct_trace: List[ClaimHash, MAX_TRACE]) -> Re
...
 
+def claim_response(claim: Claim, response: Response, trace: List[ClaimHash, MAX_TRACE]) -> Claim:
+    new_position = claim.position << 1 if Response.ATTACK else (claim.position+1) << 1
+    claim_hash = chal_trace{trace_index(new_position)]
+    return Claim(parent=claim, position=new_position, claim_hash=claim_hash)
+
+def is_freeloader(claim: Claim, team: Team, trace: List[ClaimHash, MAX_TRACE]) -> Claim:
+    if claim.parent is None:
+        return None
+    correct_claims = {claim_response(c, respond(c, team, trace), trace) for c in ALL_CLAIMS}
```
Inphi, line 145: this doesn't work.
There are two scenarios;
- regular dishonest claims is where we consider the path validity of dishonest claims
- freeloader mode where we ignore path validity.

When responding to a claim, we need to figure out which scenario we're dealing with and adapt the response logic appropriately. But first we need a more rigorous definition of a freeloader claim. One definition I'm working with is the following:

"Freeloaders are simply dishonest claims that are left-most positioned". This definition encompasses freeloaders (using the original definition) that switch teams as they counter other freeloaders and so on.
I'm only 60% confident that this is the correct definition. But we cannot proceed until this is proven.

specs/experimental/fault-proof/stage-one/honest-challenger-fdg.md
ajsutton: Maybe add a comment to explain that?

ajsutton:
@Inphi  I think we can close this one now right?

Inphi:
> @Inphi I think we can close this one now right?

Yeah. This is already fixed by https://github.com/ethereum-optimism/specs/pull/63

PR #58, opened 2024-02-22
interop: superchain backend service

hamdiallam:
Introduction of the new interfaces and abstractions introduced that encapsulates
the logic described in the interop spec. 

Spec for ethereum-optimism/protocol-quest#158


specs/interop/services.md
```diff
@@ -0,0 +1,83 @@
...
+
+- [Superchain Backend](#superchain-backend)
+  - [Message Safety](#message-safety)
+- [Optimism Transaction Pool Policies](#optimism-transaction-pool-policies)
+  - [SuperchainMessagingPolicy](#superchainmessagingpolicy)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+
+## Superchain Backend
```
tynes: Can you make this single `#` and at the top of the file?

specs/interop/services.md
```diff
@@ -0,0 +1,83 @@
...
+
+The superchain backend provides the methods available for both
+the block builder and and verifier for message safety checks
+```go
+type MessageSafetyLabel int
+
```
tynes: Spacing errors, maybe tabs vs spaces

specs/interop/services.md
```diff
@@ -0,0 +1,83 @@
...
+	Safe
+	Finalized
+)
+
+type SuperchainBackend interface {
+    MessageSafety(id Identifier, payload []byte) (MessageSafetyLabel, error)
+}
+```
+
+The backend is instantiated per chain executing interop messages, maintaning a view
```
tynes: Does this mean there is 1 service per chain in the transitive dep set?

specs/interop/services.md
```diff
@@ -0,0 +1,83 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
```
tynes: A big thing missing here is the RPC api that i saw was in the PR

specs/interop/services.md
```diff
@@ -0,0 +1,83 @@
...
+```go
+type OptimismTxPolicyStatus uint
+
+const (
+	OptimismTxPolicyInvalid OptimismTxPolicyStatus = iota
+	OptimismTxPolicyValid
+)
+
+type OptimismTxPoolPolicy interface {
```
tynes: Why are there 2 different interfaces instead of just using a single one? I suppose the tradeoff is deserializing the calldata vs sending the whole tx. I don't think we want to send all txs thru this interface

specs/interop/services.md
```diff
@@ -0,0 +1,83 @@
...
+const (
+	Invalid MessageSafetyLabel = iota
+    Unsafe
+    CrossUnsafe
+	Safe
+	Finalized
+)
+
+type SuperchainBackend interface {
+    MessageSafety(id Identifier, payload []byte) (MessageSafetyLabel, error)
```
tynes: `Identifier` technically isn't defined in this doc but you can link to its definition elsewhere

specs/interop/services.md
hamdiallam: I'll remove this section as it's specific to op-geth and the txpool/sequencer spec already clearly describes what the execution engine must do

specs/interop/services.md
hamdiallam: Ah no i'll reword this. Just a connection to the chains in the dep set

specs/interop/services.md
hamdiallam: Added a bit about consuming via a json-rpc server or go library

specs/interop/superchain_backend.md
```diff
@@ -0,0 +1,156 @@
...
+number_to_block_hash = dict() # (chain_id, int) -> hash
+latest_blocks = dict() # (chain_id) -> hash
+
+## Graph Edge
+##   - The initiating side is represented with the block number since the corresponding payload may
+##   not have been seen at the time the executing message has been processed.
+dependents = dict() # (initiating_chain_id, block_number) -> List[(executing_chain_id, block_hash)]
+dependencies = dict() # (executing_chain_id, block_hash) -> List[(initiating_chain_id, block_number)]
+
+class Block:
```
tynes, line 80: I recommend working with the actual block type as in the [execution specs](https://github.com/ethereum/execution-specs/blob/119208cf1a13d5002074bcee3b8ea4ef096eeb0d/src/ethereum/shanghai/fork_types.py#L217). This is helpful to ensure that you have the right data present during actual implementation because there is no concept of a chain id on a block. The chain id is part of transactions and of the chain config but not the block/header. There is a parent block hash but not an actual reference to the real block type

specs/interop/superchain_backend.md
```diff
@@ -0,0 +1,156 @@
...
+### Unsafe to Cross-Unsafe
+
+In order to discern between unsafe and cross-unsafe blocks, the backend maintains a graph of
+dependencies between blocks can include interop messages between them.
+
+We describe a psuedo-implementation of cross-unsafe message resolution using this graph. Every node in
+this graph represents a block with edges representing the dependent relationship between them. Upon
+receipt of unsafe payloads, the graph is extended with edges added on detection of interop transactions.
+
+```python
```
tynes, line 65: nit: there are sometimes `;` but not always, recommend getting rid of them everywhere for consistency

specs/interop/superchain_backend.md
```diff
@@ -0,0 +1,156 @@
...
+##   not have been seen at the time the executing message has been processed.
+dependents = dict() # (initiating_chain_id, block_number) -> List[(executing_chain_id, block_hash)]
+dependencies = dict() # (executing_chain_id, block_hash) -> List[(initiating_chain_id, block_number)]
+
+class Block:
+    chain_id: int
+    number: int
+    parent: Block
+
+    executing_messages: List[(MsgId, MsgPayload)]
```
tynes: You can define a function that takes a block and returns its executing messages, this will make it look much more similar to the actual implementation

specs/interop/superchain_backend.md
```diff
@@ -0,0 +1,156 @@
...
+    unverified_msg_ref_count: int
```
tynes: This will need to go in a global dict similar to the `dependents` i think

specs/interop/superchain_backend.md
```diff
@@ -0,0 +1,156 @@
...
+def block_safety(chain_id, block_hash):
+    block = blocks[block_hash]
+    if block.unverified_msg_ref_count > 0:
+        return UNSAFE
+
+    ## All messages were validated but we also need to ensure all
+    ## transitive block dependencies have had their messages also
+    ## validated
+    for initiating_chain_id, block_number in dependencies[(chain_id, block_hash)]:
+        if block_bumber > latest_heights[tx.msg_origin]:
```
tynes: What is `msg_origin`? I don't see it defined anywhere

specs/interop/superchain_backend.md
```diff
@@ -0,0 +1,156 @@
...
+
+class Block:
+    chain_id: int
+    number: int
+    parent: Block
+
+    executing_messages: List[(MsgId, MsgPayload)]
+    unverified_msg_ref_count: int
+
+def block_safety(chain_id, block_hash):
```
tynes: Is this different than `MessageSafety` from the Go interface defined above?

specs/interop/superchain_backend.md
```diff
@@ -0,0 +1,156 @@
...
+        if block_bumber > latest_heights[tx.msg_origin]:
+            return UNSAFE
+
+        dependency_block = blocks[number_to_block_hash[(initiating_chain_id, block_number)]]
+        if block_safety(initiating_chain_id, dependency_block.hash) == UNSAFE:
+            return UNSAFE
+
+    return CROSS_UNSAFE
+
+def add_unsafe_payload(chain_id, payload):
```
tynes: Can you add a comment on each of these functions that states how its called? ie what triggers it being called

specs/interop/superchain_backend.md
```diff
@@ -0,0 +1,156 @@
...
+    # try resolve any messages for this block
+    resolve_unverified_messages(chain_id, payload.hash);
+
+    # for existing dependents, trigger resolution
+    for executing_chain_id, block_hash in dependents[(chain_id, payload.number)]:
+        resolve_unverified_messages(executing_chain_id, payload.hash);
+
+def resolve_unverified_messages(chain_id, block_hash):
+    block = blocks[block_hash]
+    for tx in block.executing_txs:
```
tynes: There is no `executing_txs` property on the block type that you defined. Do you mean `executing_messages`?

specs/interop/superchain_backend.md
tynes: Then you can define a function that turns a standard tx into an Identifier + Payload and use that when you want to reference properties of the Identifier

specs/interop/superchain_backend.md
hamdiallam: Yeah I added a comment about this. This function would be internally used by the Go interface. After checking the validity rule, it would need to check the safety level of the entire remote block to discern between unsafe/cross-unsafe/safe.

specs/interop/superchain_backend.md
hamdiallam: Sg, i moved this stuff out globally. Originally was encapsulating this information within the `Block` class but not necessary

specs/interop/superchain_backend.md
hamdiallam: sounds good, created `parse_executing_messages` from the block

specs/interop/superchain_backend.md
```diff
@@ -0,0 +1,205 @@
...
+
+This method MUST enforce the attributes listed in the [Identifier](./messaging.md#message-identifier) matches the log
+and block attributes returned by peer's `eth_getLogs` request with the `blockNumber` and `logIndex` parameters.
+
+This method MUST enforce the provided payload matches the bytes computed when
+constructing the [MessagePayload](./messaging.md#message-payload) against the fetched log.
+
+This method MUST return the applicable [safety label](./verifier.md#safety) for the message, or `Invalid` upon failures.
+
+This method MUST NOT enforce the [timestamp invariant](./messaging.md#timestamp-invariant) on the provided messages. The
```
tynes: I am slightly confused by this - the timestamp invariant asserts that the message must not come from the future. We probably still want to assert this but we don't want to assert anything stricter than that (aka the policy)

specs/interop/superchain_backend.md
```diff
@@ -0,0 +1,205 @@
...
+    Extract all executing messages in this block
+    """
+    messages = []
+    for tx in block.transactions:
+        if is_executing_tx(tx):
+            msg_id, payload = parse_executing_tx_data(tx.data)
+            messages.append(msg_id, payload)
+    return messages
+
+def block_safety(chain_id: int, block_hash: Bytes32):
```
tynes, line 95: I notice that this never returns `SAFE`?

specs/interop/superchain_backend.md
```diff
@@ -0,0 +1,205 @@
...
+    blocks[block.header.hash] = block
+    chain_block_head[chain_id] = block.header.hash
+
+    executing_messages = parse_executing_messages(block)
+    block_unverified_executing_messages[block.header.hash] = executing_messages
+
+    # add edges based on the executing messages present
+    for msg_id, _ in executing_messages:
+        # initiating block linked with the executing block
+        dependents[(msg_id.origin, msg_id.block_number)].add((chain_id, block.header.hash))
```
tynes: I think we want to replace `msg_id.origin` with `msg_id.chain_id` - see https://specs.optimism.io/interop/messaging.html#message-identifier

specs/interop/superchain_backend.md
```diff
@@ -0,0 +1,205 @@
...
+    #  (child blocks should be pruned from the graph. left out for simplicity)
+    block = blocks[block_hash]
+    chain_block_head[chain_id] = block.header.parent_hash
+
+    # invalidate the dependent blocks
+    for executing_chain_id, dependent_block_hash in dependents[(chain_id, block.header.number)]:
+        handle_invalidation(executing_chain_id, dependent_block_hash)
+
+        # remove dependencies as the dependency block shouldn't exist
+        delelete(dependencies[(executing_chain_id, dependent_block_hash)])
```
tynes: This looks like a typo in `delete`

specs/interop/superchain_backend.md
hamdiallam: The reason we don't enforce it here is because we don't exactly know the exact context of the executing message.

For the block builder -> Timestamp is known when locally building or sealing the block
For the verifier -> Timestamp is known ahead of time

We could opt to pass the timestamp of the executing message as an argument in the `MessageSafety` but I was suggesting here that both the block builder / verifier locally enforce the invariant as the timestamp is apart of the message id and this api will ensure it's validity.

specs/interop/superchain_backend.md
hamdiallam: ah yea that's right

protolambda (commented):
Reorgs and intra-block cyclic dependencies are hard problems, and need more attention in the spec. Recursive functions without state don't handle cycles, and block-number based identifiers are prone to break in fun ways upon reorgs. Other than that this looks pretty good.

specs/interop/superchain_backend.md
tynes: If we have the blocknumber then we can fetch the timestamp by getting the block by number right? I also believe that both L1 and L2 timestamps are deterministic, so its possible we could have a cache of "starting timestamp" (which i believe exists already on the output oracle) and compute timestamps that way

hamdiallam: ah another example as well is the mempool validation. We wouldn't know the executing timestamp ahead of time

hamdiallam: changed this to "SHOULD NOT" instead and added a bit of comment amount why

specs/interop/superchain_backend.md
hamdiallam: added a comment about Safe here in the psuedocode. And the section below covers as well

tynes:
Closing this due to it being stale

PR #70, opened 2024-02-27
fault-proof: Global precompile oracle

Inphi:
To accommodate the acceleration of more precompiles, including `bn256Pairing` and `ecrecover`, we introduce a generic precompile oracle. This oracle replaces the KZG point evaluation oracle as it's agnostic to the precompile operation being accelerated.

We assume that future upgrades do not alter the precompile function. Which is a reasonable assumption given that precompiles can be modeled as pure functions. So far this has never happened, as only tweaks to the precompile gas schedule have been made.

### DIFF
--- a/specs/experimental/fault-proof/index.md
+++ b/specs/experimental/fault-proof/index.md
@@ -14,7 +14,7 @@
     - [Type `3`: Global generic key](#type-3-global-generic-key)
     - [Type `4`: Global SHA2-256 key](#type-4-global-sha2-256-key)
     - [Type `5`: Global EIP-4844 Point-evaluation key](#type-5-global-eip-4844-point-evaluation-key)
-    - [Type `6`: Global EIP-4844 Point-evaluation precompile key](#type-6-global-eip-4844-point-evaluation-precompile-key)
+    - [Type `6`: Global Precompile key](#type-6-global-precompile-key)
     - [Type `7-128`: reserved range](#type-7-128-reserved-range)
     - [Type `129-255`: application usage](#type-129-255-application-usage)
   - [Bootstrapping](#bootstrapping)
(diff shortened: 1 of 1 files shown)


clabby (approved):
This LGTM. Did we figure out how to handle EVM versioning? We'll always only be able to execute the most recent EVM version in the `PreimageOracle` contract for input verification.

We might be saved by the fact that we can replace preimages, if one already exists for the global precompile key that was executed on a prior hardfork w/ different precompile behavior. Though there's still an interesting case where the challenger can be playing a dispute game, L1 forks, and the precompile execution now differs.

Np if not- think this is reasonable:
> We assume that future upgrades do not alter the precompile function. Which is a reasonable assumption given that precompiles can be modeled as pure functions. So far this has never happened, as only tweaks to the precompile gas schedule have been made.

Inphi:
> This LGTM. Did we figure out how to handle EVM versioning? We'll always only be able to execute the most recent EVM version in the `PreimageOracle` contract for input verification.
> 
> We might be saved by the fact that we can replace preimages, if one already exists for the global precompile key that was executed on a prior hardfork w/ different precompile behavior. Though there's still an interesting case where the challenger can be playing a dispute game, L1 forks, and the precompile execution now differs.
> 
> Np if not- think this is reasonable:
> 
> > We assume that future upgrades do not alter the precompile function. Which is a reasonable assumption given that precompiles can be modeled as pure functions. So far this has never happened, as only tweaks to the precompile gas schedule have been made.

There's one scheme for EVM versioning that involves regular upgrades to the PreimageOracle. You'd basically have to know in advance if a future upgrade breaks precompile behavior and then update a hardcoded set of supported blocktimes in the PreimageOracle. It's not a good scheme as it breaks the immutable aspect of the PreimageOracle.
But yeah, I don't think we need to do this in any case.

ajsutton (approved):

PR #81, opened 2024-03-05
interop: address alias uncertainty

### LINKED ISSUE #76: Interop: Address Aliasing
# L2 to L2 Address Aliasing

> 💡 tldr; no address aliasing is required for security, just an exposed “cross domain sender” and “initiating chain id”. Its functionally the same as using aliasing but more simple

Address aliasing exists for L1 to L2 deposits because there is no guarantee that owner of the account on L1 is the owner of the account on L2. For EOAs, this is not an issue because the same keypair can generate signatures for both chains but for smart contracts, this is a valid concern. It prevents somebody from assuming the identity of a smart contract on L2 that they do not actually control.

Aliasing is built directly into the protocol via the `OptimismPortal`, see the code snippet below:

```solidity
// Transform the from-address to its alias if the caller is a contract.
address from = msg.sender;
if (msg.sender != tx.origin) {
  from = AddressAliasHelper.applyL1ToL2Alias(msg.sender);
}
```

This is saying “if the call comes from a smart contract, apply the alias”. This results in the [DepositTx](https://github.com/ethereum-optimism/op-geth/blob/ef668c77e03f5401d2882bdfd51cbac8536aa4cd/core/types/deposit_tx.go#L29) having a an aliased `from` field. Contracts on L2 must unalias the `from` field if they want to know what the account of the sender is, meaning that they opt in to receiving messages from L1 contracts.

The same concept applies for L2 to L2 transactions - it is possible that the same account is owned by different entities between different L2s. A difference with the interop scheme is that the top level `transaction.from` on the executing message isn’t necessarily the same entity that creates the initiating message. It is also the case that with our interop scheme the top level call will always be to the `CrossL2Inbox`. This may or may not go through the [`L2ToL2CrossDomainMessenger`](https://specs.optimism.io/interop/predeploys.html#l2tol2crossdomainmessenger) after the `CrossL2Inbox`. When a user wants to do auth, they need to call back to `L2ToL2CrossDomainMessenger` to get the sender of the initiating message. They will also need the chain id of the chain that initiated the message.

### What Aliasing Could Look Like

```solidity
library AddressAliasHelper {
    uint160 constant offset = uint160(0x1111000000000000000000000000000000001111);

    /// @notice Utility function that converts the address in the L1 that submitted a tx to
    /// the inbox to the msg.sender viewed in the L2
    /// @param l1Address the address in the L1 that triggered the tx to L2
    /// @return l2Address L2 address as viewed in msg.sender
    function applyL1ToL2Alias(address l1Address) internal pure returns (address l2Address) {
        unchecked {
            l2Address = address(uint160(l1Address) + offset);
        }
    }

    /// @notice Utility function that converts the msg.sender viewed in the L2 to the
    /// address in the L1 that submitted a tx to the inbox
    /// @param l2Address L2 address as viewed in msg.sender
    /// @return l1Address the address in the L1 that triggered the tx to L2
    function undoL1ToL2Alias(address l2Address) internal pure returns (address l1Address) {
        unchecked {
            l1Address = address(uint160(l2Address) - offset);
        }
    }

    function applyL2ToL2Alias(address localAddress) internal pure returns (address remoteAddress) {
      unchecked {
        remoteAddress = address(uint160(localAddress) + uint160(uint256(keccack(block.chainid))));
      }
    }

    function undoL2ToL2Alias(address remoteAddress, uint256 remoteChainId) internal pure returns (address localAddress) {
      unchecked {
        localAddress = address(uint160(l1Address) - uint160(uint256(keccack(remoteChainId))));
      }
    }
}
```

To ensure that it is not possible to spoof having control of the identity of an account by controlling it on one of many possible domains, some unique information relevant to the chain must be used when aliasing. In the sample implementation, the `block.chainid` is used. This means that on the remote chain, there needs to be knowledge of the chain id of the initiating chain.

The executing message will never assume the identity via `msg.sender` or `tx.origin` since deposit transactions aren’t used and the contract that is being called is at least 1 callframe deep.

This leads me to believe it is far more simple to not alias addresses and instead expose 2 pieces of information - the sender of the cross chain message and the chain id of the originating chain. Using the above address aliasing scheme requires the same information as the remote chain id is required to unalias the contract. Developers will be able to grasp the concept of asserting on sender + chainid more than asserting on unaliased address.

One upside of aliasing is that an aliased owned contract may not be obvious on which domain the owner resides until it is called once. It produces a completely unique account that commits to an account on a particular domain. Developers can build an aliasing scheme on top of the messenger if they so desire.

### User Stories

I am a smart contract developer that has a cross chain app. I have a single contract on OP Mainnet that owns a contract on Base. I want to ensure that only the contract on OP Mainnet can call permissioned functions on the contract on Base. Since there may exist other contracts on other chains that have the same address as my contract on OP Mainnet, I need to check 2 things. The address is correct and the originating domain is correct.

I am a smart contract dev that has a cross chain app with a deterministically deployed contract across all chains. I can be sure that I own the account on all chains since it is deterministically deployed. I just need to check that the address is correct.

tynes:
**Description**

It is now understood that aliasing between L2s is not necessary.
This commit describes why it is safe to not use aliasing and updates
the `L2ToL2CrossDomainMessenger` to also pass through the source
domain's chainid. Making an assertion on both the cross domain message
sender and the source domain's chainid is the same as making an
assertion on an aliased address. It is far more simple to not alias
anything and assert directly on the sender + chainid.

Closes https://github.com/ethereum-optimism/specs/issues/76

### DIFF
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -172,11 +172,18 @@ are prefixed to the abi encoded call.
 An explicit `_destination` chain and `nonce` are used to ensure that the message can only be played on a single remote
 chain a single time. The `_destination` is enforced to not be the local chain to avoid edge cases.
 
+There is no need for address aliasing as the aliased address would need to commit to the source chain's chain id
+to create a unique alias that commits to a particular sender on a particular domain and it is far more simple
+to assert on both the address and the source chain's chain id rather than assert on an unaliased address.
+In both cases, the source chain's chain id is required for security. Executing messages will never be able to
+assume the identity of an account because `msg.sender` will never be the identity that initiated the message,
+it will be the `L2ToL2CrossDomainMessenger` and users will need to callback to get the initiator of the message.
+
 ```solidity
 function sendMessage(uint256 _destination, address _target, bytes calldata _message) external {
     require(_destination != block.chainid);
 
-    bytes memory data = abi.encodeCall(L2ToL2CrossDomainMessenger.relayMessage, (_destination, messageNonce(), msg.sender, _target, _message));
+    bytes memory data = abi.encodeCall(L2ToL2CrossDomainMessenger.relayMessage, (_destination, block.chainid, messageNonce(), msg.sender, _target, _message));
     emit SentMessage(data);
     nonce++;
 }
@@ -194,17 +201,18 @@ It is important to ensure that the source chain is in the dependency set of the 
 it is possible to send a message that is not playable.
 
 ```solidity
-function relayMessage(uint256 _destination, uint256 _nonce, address _sender, address _target, bytes memory _message) external payable {
+function relayMessage(uint256 _destination, uint256 _source, uint256 _nonce, address _sender, address _target, bytes memory _message) external payable {
     require(msg.sender == address(CROSS_L2_INBOX));
     require(_destination == block.chainid);
     require(CROSS_L2_INBOX.origin() == address(this));
     require(_target != address(this));
 
-    bytes32 messageHash = keccak256(abi.encode(_destination, _nonce, _sender, _target, _message));
+    bytes32 messageHash = keccak256(abi.encode(_destination, _source, _nonce, _sender, _target, _message));
     require(sentMessages[messageHash] == false);
 
     assembly {
       tstore(CROSS_DOMAIN_MESSAGE_SENDER_SLOT, _sender)
+      tstore(CROSS_DOMAIN_MESSAGE_SOURCE_SLOT, _source)
     }
 
     bool success = SafeCall.call({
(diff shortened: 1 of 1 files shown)


specs/interop/predeploys.md
```diff
@@ -194,17 +201,18 @@ It is important to ensure that the source chain is in the dependency set of the
 it is possible to send a message that is not playable.
 
 ```solidity
-function relayMessage(uint256 _destination, uint256 _nonce, address _sender, address _target, bytes memory _message) external payable {
+function relayMessage(uint256 _destination, uint256 _source, uint256 _nonce, address _sender, address _target, bytes memory _message) external payable {
```
mds1, line 204: Nit on variable naming, perhaps `_initiating` or `_initiatingChainId` is more consistent than `_source`. Not a blocker though

PR #82, opened 2024-03-05
experimental - deposit queue

tynes:
**Description**

Adds an experimental spec for a deposit queue.

This feature is meant to solve:
- Complexity of `ResourceMetering`
- No deposit transaction frontrunning grief
- Consistent amount of gas used on L1 when depositing

An additional consideration is to be able to handle L3s with equal or higher block time than the L2. (issue currently is that those chains can’t catch up their L1BlockInfo to the latest block). It isn't clear if this should be included as part of this or be its own spec yet

### DIFF
--- a/specs/experimental/deposit-queue.md
+++ b/specs/experimental/deposit-queue.md
@@ -1,0 +1,104 @@
+# Deposit Queue
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Overview](#overview)
+- [Constants](#constants)
+- [Derivation Pipeline](#derivation-pipeline)
+- [ResourceMetering](#resourcemetering)
+- [OptimismPortal](#optimismportal)
+- [SystemConfig](#systemconfig)
+- [L1Attributes](#l1attributes)
+- [Security Considerations](#security-considerations)
+  - [Cheaper to Transact on L2 via L1](#cheaper-to-transact-on-l2-via-l1)
+  - [Size of Deposit Queue](#size-of-deposit-queue)
+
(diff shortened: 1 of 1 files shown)


kaleb-keny:
appreciate pushing for a change on this @tynes  🙏🏼

tynes:
It would be good to also include a nonce with each deposit event so that we don't need to observe all logs that create deposit txs during the fault proof program to be sure that we have not skipped a deposit. It would also be ideal to have a nonce on L2 that increments for each deposit consumed, so that we can look at the L1 nonce and know how many not yet pulled in deposits there are on L2.

We would also want to do the same nonce strategy for the system config `ConfigUpdate` event, its possible that given the static config change in https://github.com/ethereum-optimism/specs/issues/122 we can deprecate `ConfigUpdate` in favor of simply making calls to `L1Block` as a special identity, similar to how custom gas token works

mdehoog:
+1 on including a nonce with each deposit event, and also with config updates. This would allow us to skip processing L1 blocks that don't have any L2-relevant events in them, speeding up derivation and proving.

BlocksOnAChain:
@benjaminion I believe we can move this for Isthmus, since we aren't going to ship it with Holocene.

K-Ho:
> An additional consideration is to be able to handle L3s with equal or higher block time than the L2. (issue currently is that those chains can’t catch up their L1BlockInfo to the latest block)

I would love to know what the scope of this fix would be. If it's not too heavy lift, this would be super high leverage. As we aim to move all chains in the Superchain from 2s -> 1s block times, we need this fix shipped in order to not brick all L3s (which all currently have 1s block times) on L2s that want to move from 2s->1s block times

kaleb-keny:
hey @K-Ho i believe this has taken long enough time to get it fixed that i'm not sure it's worthwhile anymore. The current fix we have for it (bumping gas to crazy levels) more than works, and happy to close the ticket.

PR #83, opened 2024-03-05
fault-proof: Document other bond incentives

Inphi:
Clarify that we assume current bond incentives are sufficient to cover all costs of participating in the game.

Maybe at stage-2, we can revisit this and properly model off-protocol costs to create robust incentives.


ajsutton (approved):
LGTM.

PR #86, opened 2024-03-09
interop: fix sample code for L2ToL2CrossDomainMessenger

0xfuturistic:
on one side the document says that `CrossL2Inbox` cannot call itself through `relayMessage`. However, instead of enforcing that `_target == address(CROSS_L2_INBOX)`, the sample code shows `_target == address(this)`. This PR fixes that.

### DIFF
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -198,7 +198,6 @@ function relayMessage(uint256 _destination, uint256 _nonce, address _sender, add
     require(msg.sender == address(CROSS_L2_INBOX));
     require(_destination == block.chainid);
     require(CROSS_L2_INBOX.origin() == address(this));
-    require(_target != address(CROSS_L2_INBOX));
 
     bytes32 messageHash = keccak256(abi.encode(_destination, _nonce, _sender, _target, _message));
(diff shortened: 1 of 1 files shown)


specs/interop/predeploys.md
```diff
@@ -198,7 +198,7 @@ function relayMessage(uint256 _destination, uint256 _nonce, address _sender, add
     require(msg.sender == address(CROSS_L2_INBOX));
     require(_destination == block.chainid);
     require(CROSS_L2_INBOX.origin() == address(this));
-    require(_target != address(this));
+    require(_target != address(CROSS_L2_INBOX));
```
tynes: The `msg.sender != tx.origin` check achieves this, we do want to make sure that we cannot reenter this function from within itself altho `require(msg.sender == address(CROSS_L2_INBOX)` ensures this so we could probably just delete this check

tynes: Now that the code exists, we could turn this sample implementation into a list of invariants

0xfuturistic: done. just deleted the line!

tynes (approved):

PR #87, opened 2024-03-11
feat: add anchor state registry

smartcontracts:
**Description**
Adds the concept of the Anchor State Registry to the spec for the Fault Dispute Game.

### DIFF
--- a/specs/experimental/fault-proof/stage-one/fault-dispute-game.md
+++ b/specs/experimental/fault-proof/stage-one/fault-dispute-game.md
@@ -10,6 +10,8 @@
   - [PreimageOracle](#preimageoracle)
   - [Execution Trace](#execution-trace)
   - [Claims](#claims)
+  - [Anchor State](#anchor-state)
+  - [Anchor State Registry](#anchor-state-registry)
   - [DAG](#dag)
   - [Subgame](#subgame)
   - [Game Tree](#game-tree)
@@ -26,6 +28,7 @@
   - [Team Dynamics](#team-dynamics)
   - [Game Clock](#game-clock)
   - [Resolution](#resolution)
+  - [Finalization](#finalization)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -79,6 +82,26 @@ trace. A FDG is initialized with an output root that corresponds to the state of
 execution trace subgames at `SPLIT_DEPTH + 1` are initialized with a claim that commits to the entire execution trace
 between two consecutive output roots (a block `n -> n+1` state transition). As we'll see later, there can be multiple
 claims, committing to different output roots and FPVM states in the FDG.
+
+### Anchor State
+
+An anchor state, or anchor output root, is a previous output root that is assumed to be valid. An
+FDG is always initialized with an anchor state and execution is carried out between this anchor
+state and the [claimed output root](#claims). FDG contracts pull their anchor state from the
+[Anchor State Registry](#anchor-state-registry) contract. The initial anchor state for a FDG is the
+genesis state of the L2.
+
+Clients must currently gather L1 data for the window between the anchor state and the claimed
+state. In order to reduce this L1 data requirement, [claims](#claims) about the state of the L2
+become new anchor states when dispute games resolve in their favor. FDG contracts set their anchor
+states at initialization time so that these updates do not impact active games.
+
+### Anchor State Registry
+
+The Anchor State Registry is a registry that maps FDG types to their current [anchor states](#anchor-state).
+The Anchor State Registry is specific to Fault Dispute Game contracts and may not be applicable to
+other types of dispute game contracts that do not have the same concept of state that progresses
+over time.
 
 ### DAG
 
@@ -262,7 +285,7 @@ function addLocalData(uint256 _ident, uint256 _execLeafIdx, uint256 _partOffset)
 
 The `addLocalData` function loads local data into the VM's `PreimageOracle`. This data consists of bootstrap data for
 the program. There are multiple sets of local preimage keys that belong to the `FaultDisputeGame` contract due to the
-ability for players to bisect to any block $n \rightarrow n + 1$ state transition since the configured genesis, the
+ability for players to bisect to any block $n \rightarrow n + 1$ state transition since the configured anchor state, the
 `_execLeafIdx` parameter enables a search for the starting / disputed outputs to be performed such that the contract
 can write to and reference unique local keys in the `PreimageOracle` for each of these $n \rightarrow n + 1$
 transitions.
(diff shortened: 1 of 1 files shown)


specs/experimental/fault-proof/stage-one/fault-dispute-game.md
```diff
@@ -459,3 +476,12 @@ Each move bisects the historical state of L2 and eventually, `MAX_GAME_DEPTH` is
...
 claims to be bonded, while rewarding game winners using the bonds of dishonest claims.
+
+### Finalization
+
+Once the game is resolved, if the claim is shown to be valid, the FDG reports its state to the
+Anchor State Registry. The Anchor State Registry verifies that the request to update an anchor
+state comes from a FDG contract created by the Dispute Game Factory contract, confirms that the
+game resolved in favor of the defender, and confirms that the updated state would be newer than
+the current anchor state. If these conditions are true, the Anchor State Registry updates the
+anchor state for the given game type.
```
ajsutton: Just to confirm my understanding, this is the same logic as in the draft PR where the OptimismPortal tracks the most recent finalized game (ignoring the additional air gap after the game resolves), but we've moved that to a separate contract (which nicely avoids the circular dependency between OptimismPortal and DisputeGameFactory). Is that right?

ajsutton: oh and it's now tracking per game type instead of just the single respected game type.

smartcontracts: Yes, that's correct. I chatted with @clabby and we decided that this was cleaner and avoided adding extra functionality to the `OptimismPortal` contract since it's considered a critical contract.

smartcontracts:
One thing worth noting is that this new contract means that the anchor state can be updated even if the game type being played is not the respected game type. Any suggestions on how this should be handled, if at all? Primary concern would be a situation something along the lines of:

1. DisputeGameFactory owner adds new game type
2. Game type isn't respected so challenger doesn't react to it
3. People start playing this game type and finalize incorrect states
4. Game type becomes respected
5. New games use a bad anchor state

specs/experimental/fault-proof/stage-one/fault-dispute-game.md
tynes: This seems like a sane approach. It could be a clear way to communicate with a list of invariants for state modifications to the contract + pseudocode. It could also be nice to include an interface

specs/experimental/fault-proof/stage-one/fault-dispute-game.md
```diff
@@ -459,3 +479,12 @@ Each move bisects the historical state of L2 and eventually, `MAX_GAME_DEPTH` is
 can be settled conclusively. Dishonest players are disincentivized to participate, via backwards induction,
 as an invalid claim won't remain uncontested. Further incentives can be added to the game by requiring
 claims to be bonded, while rewarding game winners using the bonds of dishonest claims.
+
+### Finalization
+
+Once the game is resolved, if the claim is shown to be valid, the FDG reports its state to the
+Anchor State Registry. The Anchor State Registry verifies that the request to update an anchor
+state comes from a FDG contract created by the Dispute Game Factory contract, confirms that the
+game resolved in favor of the defender, and confirms that the updated state would be newer than
```
Inphi, line 491: > would be newer than

How exactly are two states ordered?

Inphi:
> One thing worth noting is that this new contract means that the anchor state can be updated even if the game type being played is not the respected game type. Any suggestions on how this should be handled, if at all? Primary concern would be a situation something along the lines of:
> 
> 1. DisputeGameFactory owner adds new game type
> 2. Game type isn't respected so challenger doesn't react to it
> 3. People start playing this game type and finalize incorrect states
> 4. Game type becomes respected
> 5. New games use a bad anchor state

I think we shouldn't be concerned with this. Or at the very least, the FDG shouldn't be concerned with the portal or respectability.
A game type should be considered either correct or not. If it yields invalidly finalized claims, then the user shouldn't reuse that game type for the portal. This way the DGF can easily support non-portal usecases.

specs/experimental/fault-proof/stage-one/fault-dispute-game.md
```diff
@@ -80,6 +83,23 @@ execution trace subgames at `SPLIT_DEPTH + 1` are initialized with a claim that
 between two consecutive output roots (a block `n -> n+1` state transition). As we'll see later, there can be multiple
 claims, committing to different output roots and FPVM states in the FDG.
 
+### Anchor State
+
+An anchor state, or anchor state root, is a previous state root that is assumed to be valid. An FDG
```
clabby:
```suggestion
An anchor state, or anchor output root, is a previous output root that is assumed to be valid. An FDG
```

specs/experimental/fault-proof/stage-one/fault-dispute-game.md
```diff
@@ -80,6 +83,23 @@ execution trace subgames at `SPLIT_DEPTH + 1` are initialized with a claim that
...
+is always initialized with an anchor state and execution is carried out between this anchor state
+and the [claimed output root](#claims). FDG contracts pull their anchor state from the
+[Anchor State Registry](#anchor-state-registry) contract. The initial anchor state for a FDG is the
+genesis state of the L2.
+
+Clients must currently gather L1 data for the window between the anchor state and the claimed
+state. In order to reduce this L1 data requirement, [claims](#claims) about the state of the L2
+become new anchor states when they are finalized (assuming they are not challenged). FDG contracts
```
clabby:
```suggestion
become new anchor states when dispute games resolve in their favor. FDG contracts
```

specs/experimental/fault-proof/stage-one/fault-dispute-game.md
clabby: We order ORs by L2 block number. With the registry being generic over game type, we may need custom handlers for each game type?

This could maybe done by having an opaque `bytes` in storage for each game type that holds the calldata payload for the dispute game contracts, to fetch the data necessary to correctly order things. i.e., for FDGs, we'd just store `abi.encode("l2BlockNumber()")`.

clabby (approved):
LGTM at a high level. Curious how we plan to order these games in a generic way over `IDisputeGame`. Maybe the anchor state registry is an `IFaultDisputeGame` specific construct?

specs/experimental/fault-proof/stage-one/fault-dispute-game.md
smartcontracts: I vote that we make this a FaultDisputeGame-specific construct. It doesn't need to be generic for all dispute games.

Inphi (approved):
LGTM!

ajsutton:
In terms of dealing with invalid prostates that may be added, should we allow the guardian to set a prestate manually? That would also allow bootstrapping a new game type without the first game having to be played from genesis.

ajsutton:
Although I guess you could also do that through the fault dispute game contract as well where it still has a hard coded starting block and refuses to use a prestate from the registry prior to that.  That's a lot cleaner than needing to override the registry and avoids a lot of potential race conditions.

ajsutton:
Which summarises to - LGTM!. :)

smartcontracts:
> In terms of dealing with invalid prostates that may be added, should we allow the guardian to set a prestate manually?

Instead of this, we're simply going to have the deployer set the initial states during the initialization process and we'll have the FaultDisputeGame throw if it ever gets a `hash(0)` anchor state from the registry.

Issue #26, opened 2024-02-02
Specs Layout Refactor


tynes:
Right now it is difficult to contribute or consume the specs because of the way that they are laid out. It is not clear where to add new information and there isn't an easy way to know what functionality is associated with what network upgrade.

We should lay out the specs in a way where there is a set of topics and directories for each network upgrade. If a topic is changed as part of a network upgrade, we should include a markdown file named after the topic inside which includes the functionality specific for that network upgrade.

A simplified example below:

```
├── bedrock
│   ├── batcher.md
│   ├── derivation-pipeline.md
│   ├── predeploys.md
│   └── state-transition.md
├── canyon
│   └── state-transition.md
├── delta
│   └── span-batches.md
├── ecotone
│   └── predeploys.md
├── fjord
└── regolith
    └── state-transition.md
```

protolambda:
I really like the per-hardfork categorization. It's consistent with the ethereum L1 consensus-specs, and makes it easy to extend specs incrementally. That said, a type of global overview-doc, with categorized features, and a type of mini-changelog with links to sections of the individual specs, would be really useful for navigation.
Otherwise I would have to search 5 different `state-transition.md` docs until I find the state-transition feature change I am looking for.

PR #88, opened 2024-03-12
feat(engine.queue): consistent safe head progression on failures

hamdiallam:
We can simplify derivation by having the pipeline always ensure safe head progression
when processing payloads pulled from L1.

1. Always Valid (empty deposits-only block)
3. A proper L2 block created from the provided payload.

This simplifies the back & forth between the engine and attributes queue upon failures
when producing batches. Rather than wait for the sequencer window to create empty batches,
we do it right away and move onto the next batch.

### DIFF
--- a/specs/experimental/engine_queue_block_derivation.md
+++ b/specs/experimental/engine_queue_block_derivation.md
@@ -4,44 +4,34 @@
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 **Table of Contents**
 
-- [Overview](#overview)
-- [Motivation](#motivation)
-- [Solution](#solution)
+- [Consistent Safe Head Progression](#consistent-safe-head-progression)
+- [Rationale](#rationale)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-## Overview
+# Consistent Safe Head Progression
 
-The Engine Queue stage pulls the latest derived `PayloadAttributes` to be processed by the execution
-engine. If the execution engine reports an error with the provided payload (i.e invalid transaction
-or state transition in the block), the Engine Queue currently drops the payload without progessing
-the safe head and will attempt to use the next batch available for the same timestamp. If no valid
-batches are found within the sequencing window, the rollup node will create a deposits-only empty
-payload.
-
-## Motivation
-
-In implementation of low-latency interop, we want to minimize the effects of cascading reorgs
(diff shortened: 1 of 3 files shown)


tynes:
Can you make this its own markdown file in the `experimental` section? This helps to keep information more modular, then we can link that doc directly into a network upgrade doc, see https://github.com/ethereum-optimism/specs/issues/26 for context

hamdiallam:
> Can you make this its own markdown file in the `experimental` section? This helps to keep information more modular, then we can link that doc directly into a network upgrade doc, see #26 for context

Moved over to `/experimental/engine_queue_block_derivation.md`

specs/experimental/engine_queue_block_derivation.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Solution
+
+The looping behavior between the engine and attributes queue can be eliminated by always ensuring
+the progression of the safe head when processing payloads. We propose that the engine queue replace
+invalidated payloads with a deposits-only empty one. In other words, when provided `PayloadAttributes`,
+there will always be two possible block candidates:
+
+1. Always Valid (empty deposits-only block)
+2. A proper L2 block created from the provided payload.
```
protolambda:
```suggestion
2. A proper L2 block created from the provided payload attibutes.
```

specs/experimental/engine_queue_block_derivation.md
```diff
@@ -0,0 +1,47 @@
+# Engine Queue Block Derivation
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Overview](#overview)
+- [Motivation](#motivation)
+- [Solution](#solution)
```
protolambda: So far the specs docs are written differently, less so in the form of a proposal. I like the contextual information, but maybe a `Rationale` section at the bottom of the doc that combines the `Overview` and `Motivation`, and then the `Solution` as the primary content, makes more sense for consistency with other specs. We can pivot to this new format though, if we plan to write more docs this way, if things go more in the direction of an EIP type document. cc @tynes

hamdiallam: Removed the Overview and added a Rationale section about the why

tynes:
The complexity here lies in the block hash consistency checks, if we malleate a block then we need a way to signal there was a malleated block and keep the hash consistency

tynes:
Going to close this as @sebastianst now owns this

PR #93, opened 2024-03-15
std-config: initial proposal

mds1:
## Overview

This is a draft proposal defining the Standard Configuration for OP Stack chains. A standard chain does not need to be interoperable, but must support fault proofs and handover their keys. (Read more in [What is the Superchain](https://docs.google.com/presentation/d/1IBHw_KR96DgRaGCvxMSlqMMA3FshbYap9jTQFBO6SXo/edit#slide=id.g269a303f9b1_0_907)).

This PR is largely a port of the [Standard Config Strawman](https://www.notion.so/Standard-Config-Strawman-505874b0b0d24e5f90d47289d8f98485?pvs=4) by @ben-chain. 

## How to Review

There are many outstanding TODOs which we must discuss and finalize before taking this PR out of draft. The asks for reviewers are:
1. Review all requirements.
    1. For requirements that specify a value, leave a comment if you disagree with the value or justification.
    2. For requirements with a TODO, leave a comment proposing a value (there are many of these!)
2. If any requirements are missing, erroneous, unclear, etc. please leave a comment

We should aim avoid breaking changes to these requirements—we don't want a chain that previously met the Standard Configuration requirements to suddenly no longer be considered a standard chain, potentially with no path back in. Therefore we should start by keeping requirements strict, and we can loosen them over time as we learn more.

Tip: When viewing the diff, click the "Display the rich diff" icon to simplify reading of the specs

image

## Path Forward

1. Once merged, we will need to ensure future spec updates also update the standard config, when applicable. 
2. The actual format of this PR (i.e. markdown tables) may not be the final format. We want a human-readable source of truth (what this PR provides) but we also want a machine-readable version for validation scripts in the [superchain-registry](https://github.com/ethereum-optimism/superchain-registry). So the final format ("source of truth markdown, registry parses that", or "source of truth TOML in the registry, specs parse that", or something else) is still TBD


specs/standard-configuration/overview.md
```diff
@@ -0,0 +1,23 @@
...
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+**Table of Contents**
+
+- [Standard OP Stack Configuration](#standard-op-stack-configuration)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+TODO add copy about standard chains / why a chain may want to be a standard chain
```
zchn, line 12: Is this necessary for this PR? Can we leave this to a later PR?

specs/standard-configuration/protocol-parameters.md
```diff
@@ -0,0 +1,41 @@
...
+
+| Protocol Parameter                    | Requirement                                                    | Justification/Notes                                                                                                                                                                      |
+| ------------------------------------- | -------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Fault Proof Window / Challenge Period | 7 days                                                         | High security.                                                                                                                                                                           |
+| Sequencing Window                     | TODO                                                           | High security.                                                                                                                                                                           |
```
zchn, line 17: Asked in #node, https://oplabs-pbc.slack.com/archives/C01HLRH0LKZ/p1710772932465259

specs/standard-configuration/protocol-parameters.md
```diff
@@ -0,0 +1,41 @@
...
+| Chain ID                              | Foundation-approved, non-colliding value [^chain-id]           | Foundation will ensure chains are responsible with their chain IDs until there's a governance process in place.                                                                          |
+| L2 Block Time                         | 2 seconds                                                      | High security & [interoperability](../interop/overview.md) compatibility requirement, until de-risked/solved at app layer.                                                               |
+| Genesis State                         | Only standard predeploys and preinstalls, no additional state  | Homogeneity & standardization.                                                                                                                                                           |
+| Output Frequency                      | TODO this won't exist post-FPAC, should we include it?         | TODO                                                                                                                                                                                     |
```
zchn, line 21: How about `1800 seconds` which is used at least by both OP Mainnet and Mode?

> TODO this won't exist post-FPAC, should we include it?

I think we should focus on what we have *now* in this PR, and have a followup PR at the time when we upgrade OP Mainnet with FP. Too early to make decisions on FP parameters w.r.t. std config.

specs/standard-configuration/protocol-parameters.md
```diff
@@ -0,0 +1,41 @@
...
+| Batch Submission Frequency            | Less than 50% of the sequencing window                         | Batcher needs to fully submit each batch within the sequencing window, so this leaves buffer to account for L1 network congestion and the amount of data the batcher would need to post. |
+| Batch Inbox Address                   | TODO, ideally function of 32 byte chain ID                     | TODO                                                                                                                                                                                     |
```
zchn, line 23: Why do we need to require this? Can we remove this for now? I don't see a benefit or reason to mandate anything about batch inbox.

specs/standard-configuration/protocol-parameters.md
```diff
@@ -0,0 +1,41 @@
...
+| Compression Ratio                     | None                                                           | Safe to set however the entity running the batcher wishes.                                                                                                                               |
+| Fee Vault Contracts                   | Standard revenue share contract deployment [^fee-vault]        | TODO                                                                                                                                                                                     |
```
zchn, line 25: Can we remove this for now and rely on [this process](https://docs.google.com/document/d/1lvoZOCwQvmLSJwApCsMv2dMEwlLsYdpHr7R0IZszZOo/edit#heading=h.kl5u4vro3tn0) to update the config once we make progress here?

specs/standard-configuration/protocol-parameters.md
```diff
@@ -0,0 +1,41 @@
...
+| Resource Config                       | TODO.                                                          | High security, changes here can lead to security vulnerabilities.                                                                                                                        |
```
zchn, line 26: How about just use `20000000,10,8,1000000000,1000000,340282366920938463463374607431768211455` since that's the value for both OP Mainnet and Mode?

specs/standard-configuration/protocol-parameters.md
```diff
@@ -0,0 +1,41 @@
...
+| Target Gas Rate                       | TODO, defined as a function of Resource Config?                | TODO.                                                                                                                                                                                    |
```
zchn, line 27: What are the values for OP Mainnet and Mode?

specs/standard-configuration/protocol-parameters.md
```diff
@@ -0,0 +1,41 @@
...
+| Fee Margin                            | Between 0 and 50%                                              | [Law of Chains](https://github.com/ethereum-optimism/OPerating-manual/blob/main/Law%20of%20Chains.md).                                                                                   |
+| Start Block                           | Any block less than the block where L1 contracts were deployed | Simple clear restriction.                                                                                                                                                                |
+| Gas Limit                             | TODO                                                           | Must have well-tested gaslimit for uptime properties, also dangerously affects EIP1559 parameterization.                                                                                 |
```
zchn, line 30: What are the values for OP Mainnet and Mode?

specs/standard-configuration/protocol-parameters.md
```diff
@@ -0,0 +1,41 @@
...
+| Data Availability                     | Ethereum                                                       | Alt-DA is not yet supported for the standard configuration.                                                                                                                              |
+| Gas Token                             | Ether (ETH)                                                    | Custom gas tokens are not [yet](https://github.com/ethereum-optimism/specs/issues/67) supported.                                                                                         |
+
+[^chain-id]:
+The chain ID must not be present in the [superchain-registry](https://github.com/ethereum-optimism/superchain-registry)
+TODO check other lists such as
```
zchn, line 36: Do we need to address this TODO in this PR? If not, let's remove and track it in an issue instead?

specs/standard-configuration/protocol-parameters.md
```diff
@@ -0,0 +1,41 @@
...
+[ethereum-lists/chains](https://github.com/ethereum-lists/chains) or [chains.json](https://chainid.network/chains.json).
+TODO process for foundation approval?.
```
zchn, line 38: Do we need to address this TODO in this PR? If not, let's remove and track it in an issue instead?

specs/standard-configuration/protocol-parameters.md
```diff
@@ -0,0 +1,41 @@
...
+
+[^fee-vault]:
+The standard revenue share contract is TODO, and the steps to deploy and configure it are TODO.
```
zchn, line 41: Let's remove since we don't have it. And add it by following [this process](https://docs.google.com/document/d/1lvoZOCwQvmLSJwApCsMv2dMEwlLsYdpHr7R0IZszZOo/edit#heading=h.kl5u4vro3tn0) in the future?

specs/standard-configuration/admin-roles.md
```diff
@@ -0,0 +1,26 @@
...
+- [Admin Roles](#admin-roles)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+Admin roles are roles that can upgrade contracts, change role owners, or update protocol parameters.
+These are typically cold wallets not used directly in day to day operations.
+
+The table below defines admin role requirements for an OP Stack chain to be considered a Standard Chain.
+The table references the Optimism Foundation (OF) and the Security Council (SC).
+The Chain Governor is TODO.
```
zchn, line 17: How about "The Chain Governor is defined in the [Law of Chains](https://github.com/ethereum-optimism/OPerating-manual/blob/main/Law%20of%20Chains.md)." ?

specs/standard-configuration/admin-roles.md
```diff
@@ -0,0 +1,26 @@
...
+| -------------------- | ----------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
+| L1 Proxy Admin Owner | 2/2 Safe between OF and SC                                  | Governance-controlled, high security.                                                                                   |
+| L2 Proxy Admin Owner | 2/2 Safe between OF and SC                                  | Governance-controlled, high security.                                                                                   |
+| L1 Proxy Admin       | 2/2 Safe between OF and SC                                  | Governance-controlled, high security.                                                                                   |
+| L2 Proxy Admin       | TODO                                                        | Governance-controlled, high security.                                                                                   |
```
zchn, line 24: How about `Chain Governor for now, and will be eliminated or transitioned to a similar control structure as L1 ProxyAdmin in the future`?

specs/standard-configuration/admin-roles.md
```diff
@@ -0,0 +1,26 @@
...
+| Mint Manager Owner   | TODO not relevant on other chains since no native OP token? | TODO                                                                                                                    |
```
zchn, line 25: How about just remove this from the table?

specs/standard-configuration/servicer-roles.md
```diff
@@ -0,0 +1,23 @@
...
+
+Servicer roles are related to actions taken by Chain Servicers in the
+[Law of Chains](https://github.com/ethereum-optimism/OPerating-manual/blob/main/Law%20of%20Chains.md).
+They are typically hot wallets, as they take active roles in chain progression and are used to
+participate in day-to-day, ongoing operations.
+
+| Configuration Option                       | Requirement                                                      | Justification/Notes |
+| ------------------------------------------ | ---------------------------------------------------------------- | ------------------- |
+| Guardian Account                           | Optimism Foundation Safe                                         | Security            |
+| Batch Submitter Account                    | TODO can we standardize as a deterministic function of chain ID? |                     |
```
zchn: Can we remove this from the table for this PR and worry about it in a future PR?

specs/standard-configuration/servicer-roles.md
```diff
@@ -0,0 +1,23 @@
...
+| Sequencer P2P / Unsafe Head Signer Account | No requirement                                                   |                     |
+| Challenger Account                         | TODO                                                             |                     |
```
zchn: How about also `Optimism Foundation Safe`?

zchn (changes requested):

specs/standard-configuration/protocol-parameters.md
zchn: How about requiring it to be `12h` which is what we currently require?
From Subject Matter Expert: 

> 
It's currently 12h and it's hardcoded in the op-node's [superchain configuration loader](https://github.com/ethereum-optimism/optimism/blob/56f3d1b60b3ba3d7019c159180abf60e11479600/op-node/rollup/superchain.go#L96). The comment above states that it might get configurable in the future, but right now it's hardcoded because all op stack chains with a superchain config use this value. Chains who don't want to use it have to overwrite it with a custom rollup config. Note that its unit is L1-blocks (3600 * 12 / 60 / 60 = 12)

specs/standard-configuration/admin-roles.md
ben-chain: agree.  Better yet:

> The Chain Governor is defined in the Law of Chains, and address verification is implemented as a manual/social-layer check during PR review.

In other words: "validation" of the Chain Governor address should at this point be performed by social verification by us -- e.g. via out of band communication, or confirmation that the PR submitter is a known contributor to the Chain Governor's github org, etc.

specs/standard-configuration/protocol-parameters.md
mds1: We are now tracking this in https://github.com/ethereum-optimism/specs/issues/94. In general we should start with strict requirements and loosen them over time, which would be an argument in favor of not removing this

One benefit @tynes has discussed around standardizing this is to enable counterfactual chains where properties like its batch inbox are known beforehand, so messages can be sent in advance.

specs/standard-configuration/servicer-roles.md
mds1: Ah I meant to put no requirement here, fixed in a2c3f09

specs/standard-configuration/admin-roles.md
ben-chain: Agree, although there's something to do with predeploys that may matter here. (i.e.-- we should make sure to restrict ability to set `0x42000000...42` address to ensure forwards-compatibility)

specs/standard-configuration/overview.md
ben-chain: agree

specs/standard-configuration/protocol-parameters.md
ben-chain: if it's hardcoded, then yes def just stay with that restriction

specs/standard-configuration/protocol-parameters.md
ben-chain: Slipper also requested standardization [here](https://www.notion.so/Standard-Config-Strawman-505874b0b0d24e5f90d47289d8f98485?d=b7fbd86f34bf41fea94ea48121d7c499&pvs=4#c3100479c56d47aeaf563c4001cada7d).  I agree with standardizing this, seems strictly beneficial, but if there are some chains out there which don't currently conform, we should track the work to get them back on standard.  And if that's too much work, drop restriction for now

specs/standard-configuration/protocol-parameters.md
ben-chain: agree

zchn:
removing myself and assigning to @blmalone instead for the review of this PR since @mds1 and @blmalone are working on this together.

blmalone:
@mds1 closing this pr in favor of: https://github.com/ethereum-optimism/specs/pull/134

PR #98, opened 2024-03-19
Non prescriptive terms for standard config introduced

blmalone:
**Description**

As per @mds1's pull request: https://github.com/ethereum-optimism/specs/pull/93, we're in the process of defining what a standard chain means. 

This pull request is meant to introduce the terms in his PR but not be prescriptive about what values they should be. 

**Tests**

N/A

**Additional context**

Configurability Changes based on Ben's Notion document:  https://www.notion.so/Superchain-Segmentation-33a3152ca1dd4e87a40af52e08aae75d 
Relevant slack thread: https://oplabs-pbc.slack.com/archives/C06KERUPJ8H/p1710872632376849 

**Metadata**

N/A


specs/protocol/configurability.md
```diff
@@ -0,0 +1,71 @@
...
+|---------------------------------------|------------------------------------------------------------------------------------------------------------------------------|
+| Batch Inbox address                   | L1 address where calldata/blobs are posted (see [Batcher Transaction](../glossary.md#batcher-transaction)).                         |
+| Batch submission frequency            | Frequency with which batches are submitted to L1 (see [Batcher Transaction](../glossary.md#batcher-transaction)).            |
+| Chain ID                              | Unique ID of Chain used for TX signature validation.                                                                         |
+| Challenge Period                      | Length of time for which an output root can be removed, and for which it is not considered finalized.                                                                                                                                                             |
+| Compression ratio                     | How much compression the batch submitter applies to batches before submission (see [Channel](../glossary.md#channel)).       |
```
protolambda: This is not configurable, but rather a metric of performance of the batcher efficiency. Can you split out metrics in its own table? Or remove this property?

specs/protocol/configurability.md
```diff
@@ -0,0 +1,71 @@
...
+- **Protocol Parameters**: Parameters and properties of the chain that may
+  be set at genesis and fixed for the lifetime of the chain, or may be
+  changeable through a privileged account.
+- **Admin Roles**: These roles can upgrade contracts, change role owners,
+  or update protocol parameters. These are typically cold/multisig wallets not
+  used directly in day to day operations.
+- **Service Roles**: These roles are used to manage the day-to-day
+  operations of the chain, and therefore are often hot wallets.
+
+## Protocol Parameters
```
protolambda: "Superchain target" | Choice of cross-L2 configuration. May be omitted in isolated OP Stack deployments. Includes SuperchainConfig and ProtocolVersions contract addresses."

Or something to that extent should probably be part of the table.

specs/protocol/configurability.md
```diff
@@ -0,0 +1,71 @@
...
+| L1 smart contracts                    | The chain’s L1 smart contracts.                                                                                              |
+| L2 block time                         | Frequency with which blocks are produced as a result of derivation.                                                          |
+| Output frequency                      | Frequency with which output roots are submitted to L1.                                                                       |
+| Resource config                       | Config for the EIP-1559 based curve used for the deposit gas market.                                                         |
+| Sequencing window                     | Maximum allowed batch submission frequency, after which L1 fallback is triggered in derivation.                                                                                                                                                            |
```
protolambda:
```suggestion
| Sequencing window                     | Maximum allowed batch submission gap, after which L1 fallback is triggered in derivation.                                                                                                                                                            |
```

High frequency implies many batch submissions. The fallback kicks in when it's the opposite, i.e. when there aren't any batch submissions for an extended amount of time.

specs/glossary.md
```diff
@@ -483,6 +483,8 @@ When submitted to Ethereum calldata, the batcher transaction's receiver must be
 transaction must also be signed by a recognized batch submitter account. The recognized batch submitter account
 is stored in the [System Configuration][system-config].
 
+TODO: clarify batch submission frequency.
```
tynes: Can we resolve this todo before merge?

specs/protocol/configurability.md
```diff
@@ -0,0 +1,71 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+When deploying the OP Stack software to a production environment,
+certain parameters about the protocol can be configured. These
+configurations can include a variety of parameters which affect the
+resulting properties of the blockspace in question.
+
+There are three categories of OP Stack configuration options:
+
+- **Protocol Parameters**: Parameters and properties of the chain that may
```
tynes: We should differentiate between consensus and policy here. Consensus is enforced while policy includes the set of possible behaviors within consensus. An example of policy is the tx p2p gossip size limit, where consensus will accept a tx of any size given it fits in the block gas limit

specs/protocol/configurability.md
```diff
@@ -0,0 +1,71 @@
...
+  used directly in day to day operations.
+- **Service Roles**: These roles are used to manage the day-to-day
+  operations of the chain, and therefore are often hot wallets.
+
+## Protocol Parameters
+
+| Config Property                       | Description                                                                                                                  |
+|---------------------------------------|------------------------------------------------------------------------------------------------------------------------------|
+| Batch Inbox address                   | L1 address where calldata/blobs are posted (see [Batcher Transaction](../glossary.md#batcher-transaction)).                         |
+| Batch submission frequency            | Frequency with which batches are submitted to L1 (see [Batcher Transaction](../glossary.md#batcher-transaction)).            |
```
tynes: This is policy

specs/protocol/configurability.md
tynes: This is policy

specs/protocol/configurability.md
```diff
@@ -0,0 +1,71 @@
...
+| Challenge Period                      | Length of time for which an output root can be removed, and for which it is not considered finalized.                                                                                                                                                             |
+| Compression ratio                     | How much compression the batch submitter applies to batches before submission (see [Channel](../glossary.md#channel)).       |
+| Data Availability                     | Where L2 transaction calldata is posted.                                                                                     |
+| Fault Proof Window / Challenge Window | Duration network participants have to challenge the integrity of an output root.                                             |
+| Fee margin                            | Markup on transactions compared to the raw L1 data cost.                                                                     |
+| Fee vault contracts                   | Contracts which dictate how user fees are distributed.                                                                       |
```
tynes: The contract addresses themselves are consensus but the bytecode at the 1967 implementations are policy

specs/protocol/configurability.md
```diff
@@ -0,0 +1,71 @@
...
+| L1 smart contracts                    | The chain’s L1 smart contracts.                                                                                              |
+| L2 block time                         | Frequency with which blocks are produced as a result of derivation.                                                          |
+| Output frequency                      | Frequency with which output roots are submitted to L1.                                                                       |
+| Resource config                       | Config for the EIP-1559 based curve used for the deposit gas market.                                                         |
+| Sequencing window                     | Maximum allowed batch submission frequency, after which L1 fallback is triggered in derivation.                                                                                                                                                            |
+| Start block                           | Block at which op-node starts searching for logs from.                                                                       |
```
tynes: This is better defined as the block at which the system config was initialized the first time

specs/protocol/configurability.md
```diff
@@ -0,0 +1,71 @@
...
+
+| Config Property                       | Description                                                                                                                  |
+|---------------------------------------|------------------------------------------------------------------------------------------------------------------------------|
+| L1 Proxy Admin                        | Account authorized to upgrade L1 contracts.                                                                                  |
+| L1 ProxyAdmin owner                   | Account authorized to update the L1 Proxy Admin.                                                                             |
+| L2 Proxy Admin                        | Account authorized to upgrade L2 contracts.                                                                                  |
+| L2 ProxyAdmin owner                   | Account authorized to update the L2 Proxy Admin.                                                                             |
+| MintManager owner                     | Account which controls the MintManager to mint OP tokens.                                                                    |
```
tynes: This is not a predeploy and therefore not present on all chains

specs/protocol/configurability.md
```diff
@@ -0,0 +1,71 @@
...
+| Start block                           | Block at which op-node starts searching for logs from.                                                                       |
+
+## Admin Roles
+
+| Config Property                       | Description                                                                                                                  |
+|---------------------------------------|------------------------------------------------------------------------------------------------------------------------------|
+| L1 Proxy Admin                        | Account authorized to upgrade L1 contracts.                                                                                  |
+| L1 ProxyAdmin owner                   | Account authorized to update the L1 Proxy Admin.                                                                             |
+| L2 Proxy Admin                        | Account authorized to upgrade L2 contracts.                                                                                  |
```
tynes, line 58: What do you mean by "contracts" here? Do you mean contracts owned by L2 Proxy admin? ie all predeploys

specs/glossary.md
```diff
@@ -483,6 +483,8 @@ When submitted to Ethereum calldata, the batcher transaction's receiver must be
...
+
```
protolambda:
```suggestion
## Batch submission frequency

Within the [sequencing-window] constraints the batcher is free to submit L2 blocks to for data-availability at any time. The batcher may be reconfigured with a new policy at any time by its operator. The rollup enforces safety guarantees and liveness through the sequencing window, if the batcher does not submit data within this allotted time.

By submitting new L2 data in smaller more frequent steps, there is less delay in confirmation of the L2 block inputs. This allows verifiers to ensure safety of L2 blocks sooner. This also reduces the time to finality of the data on L1, and thus the time to L2 input-finality.
 
By submitting new L2 data in larger less frequent steps, there is more time to aggregate more L2 data, and thus reduce fixed overhead of the batch-submission work. This can reduce batch-submission costs, especially for lower throughput chains that do not fill data-transactions (typically 128 KB of calldata, or 800 KB of blobdata) as quickly.
```

specs/protocol/configurability.md
protolambda: Also needs a reference to the glossary entry.

specs/glossary.md
protolambda: Needs TOC update.

specs/protocol/configurability.md
```diff
@@ -0,0 +1,76 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+When deploying the OP Stack software to a production environment,
+certain parameters about the protocol can be configured. These
+configurations can include a variety of parameters which affect the
+resulting properties of the blockspace in question. 
+
+There are four categories of OP Stack configuration options:
+
+- **Consensus Parameters**: Parameters and properties of the chain that may
```
blmalone, line 21: @tynes With this definition, it sounds some of the line items in the `Consensus Parameters` table below are out of place now. 

Anything that can be upgraded with admin keys should fall under policy right?

specs/protocol/configurability.md
```diff
@@ -0,0 +1,76 @@
...
+|---------------------------------------|------------------------------------------------------------------------------------------------------------------------------|
+| Batch submitter address               | Account which authenticates new batches submitted to L1 Ethereum.                                                            |
+| Challenger address                    | Account which can delete output roots before challenge period has elapsed.                                                   |
+| Guardian address                      | Account authorized to pause L1 withdrawals from contracts.                                                                   |
+| Proposer address                      | Account which can propose output roots to L1.                                                                                |
+| Sequencer P2P / Unsafe head signer    | Account which authenticates the unsafe/pre-submitted blocks for a chain at the P2P layer.                                    |
+## Policy Parameters
```
blmalone: @tynes new table for policy params

specs/protocol/configurability.md
```diff
@@ -0,0 +1,76 @@
...
+certain parameters about the protocol can be configured. These
+configurations can include a variety of parameters which affect the
+resulting properties of the blockspace in question. 
+
+There are four categories of OP Stack configuration options:
+
+- **Consensus Parameters**: Parameters and properties of the chain that may
+  be set at genesis and fixed for the lifetime of the chain, or may be
+  changeable through a privileged account.
+- **Policy Parameters**: Parameters of the chain that can be changed without breaking consensus. Consensus is enforced while policy includes the set of possible behaviors within consensus. Can be thought of as a superset of consensus.
```
blmalone: @tynes is this more inline with how you were thinking?

specs/protocol/configurability.md
blmalone: The structure of Ben's doc is: https://www.notion.so/Superchain-Segmentation-33a3152ca1dd4e87a40af52e08aae75d

Screenshot 2024-03-20 at 17 42 47

tynes: > Anything that can be upgraded with admin keys should fall under policy right?

Not necessarily - consensus enforces the particular batcher address via the "batcher hash" config. The identity that owns the system config isn't related to consensus, it can be a smart contract or eoa altho we recommend a safe smart contract wallet

specs/protocol/configurability.md
```diff
@@ -0,0 +1,78 @@
...
+certain parameters about the protocol can be configured. These
+configurations can include a variety of parameters which affect the
+resulting properties of the blockspace in question.
+
+There are four categories of OP Stack configuration options:
+
+- **Consensus Parameters**: Parameters and properties of the chain that may
+  be set at genesis and fixed for the lifetime of the chain, or may be
+  changeable through a privileged account.
+- **Policy Parameters**: Parameters of the chain that can be changed without breaking consensus. Consensus is enforced while policy includes the set of possible behaviors within consensus. Can be thought of as a superset of consensus.
```
tynes: Perhaps remove the superset sentence because policy is constrained by consensus, meaning that its a subset. One example of a block builder policy would be "include only 1 tx per sender per block" where consensus allows an arbitrary number of txs per sender per block

specs/protocol/configurability.md
```diff
@@ -0,0 +1,78 @@
...
+| Gas limit                             | L2 block gas limit.                                                                                                          |
+| Gas Token                             | Native token used to pay for gas.                                                                                            |
+| Genesis state                         | Initial state at chain genesis, including code and storage of predeploys (all L2 smart contracts). See [Predeploy](../glossary.md#l2-genesis-block).                                            |
+| L1 smart contracts                    | The chain’s L1 smart contracts.                                                                                              |
+| L2 block time                         | Frequency with which blocks are produced as a result of derivation.                                                          |
+| Output frequency                      | Frequency with which output roots are submitted to L1.                                                                       |
```
tynes: The output frequency is policy. Making it smaller means less time after doing a withdrawal on L2 until you can prove your withdrawal on L1

specs/protocol/configurability.md
tynes: It is important to clarify these as the things that can be changed by anybody and don't impact standardness

specs/protocol/configurability.md
blmalone: The table structure has changed slightly based on Mark's comments. Can you clarify what you mean with the above comment?

protolambda: I am suggesting to add an entry to the table

blmalone: Added superchain target into the consensus parameters table: https://github.com/ethereum-optimism/specs/pull/98/files#diff-f4cc226d6bd3074bf41cc708de428f532feff29c5c7f901b9cc5a5864f8a68b5R50

specs/glossary.md
```diff
@@ -483,6 +484,14 @@ When submitted to Ethereum calldata, the batcher transaction's receiver must be
 transaction must also be signed by a recognized batch submitter account. The recognized batch submitter account
 is stored in the [System Configuration][system-config].
 
+## Batch submission frequency
+
+Within the [sequencing-window] constraints the batcher is free to submit L2 blocks to for data-availability at any time. The batcher may be reconfigured with a new policy at any time by its operator. The rollup enforces safety guarantees and liveness through the sequencing window, if the batcher does not submit data within this allotted time.
```
ben-chain:
```suggestion
Within the [sequencing-window] constraints the batcher is permitted by the protocol to submit L2 blocks to for data-availability at any time. The batcher software allows for dynamic policy configuration by its operator. The rollup enforces safety guarantees and liveness through the sequencing window, if the batcher does not submit data within this allotted time.
```

ben-chain: Bit of a nit, but I think this is a little more neutral/objective way to say it. This could be restricted to only "safe" values (e.g. Slipper's suggestions in notion) in the std-config; so "free to" might be a misnomer in that context.

specs/protocol/configurability.md
```diff
@@ -0,0 +1,80 @@
...
+certain parameters about the protocol can be configured. These
+configurations can include a variety of parameters which affect the
+resulting properties of the blockspace in question.
+
+There are four categories of OP Stack configuration options:
+
+- **Consensus Parameters**: Parameters and properties of the chain that may
+  be set at genesis and fixed for the lifetime of the chain, or may be
+  changeable through a privileged account.
+- **Policy Parameters**: Parameters of the chain that can be changed without breaking consensus. Consensus is enforced while policy includes the set of possible behaviors within consensus. Policy is ultimately constrained by consensus. Changing policy paramters won't impact how an OP chain is classified; a chain can still be considered standard if it has it's own custom policy parameters.
```
ben-chain:
```suggestion
- **Policy Parameters**: Parameters of the chain that might be changed without breaking consensus. Consensus is enforced by the protocol, while policy parameters may be changeable within constraints imposed by consensus.
```

ben-chain: Same kinda note as other language suggeston.  I think the original wording is right most of the time, but strictly speaking, it is possible to impose social restrictions on acceptable policy that are more strict than consensus/protocol.

protolambda (approved):

specs/protocol/configurability.md
```diff
@@ -0,0 +1,80 @@
...
+| Fee margin                            | Markup on transactions compared to the raw L1 data cost.                                                                     |
+| Proxy Fee vault contracts                   | Proxy contracts which point to implementation contracts that dictate how user fees are distributed.                                                                       |
+| Gas limit                             | L2 block gas limit.                                                                                                          |
+| Gas Token                             | Native token used to pay for gas.                                                                                            |
+| Genesis state                         | Initial state at chain genesis, including code and storage of predeploys (all L2 smart contracts). See [Predeploy](../glossary.md#l2-genesis-block).                                            |
+| L1 smart contracts                    | The chain’s L1 smart contracts.                                                                                              |
```
ben-chain: I meant to comment on notion, but I think this entry is too broad/vague.
- I believe there are a set of contracts which are NOT changeable or configurable based on the version. These should not be in the configurability table. For example, changing the L1 bridge code should not be thought of as "configuring" the standard version, it should be thought of creating a new version.
- That above set could be most or all of the L1 contracts, I'm actually not sure if there's anything that does not fulfill the above. Unless @tynes or @protolambda can think of any important contracts which don't meet the above definition, I would propose just removing this entry from the table.

ben-chain (changes requested):
@blmalone this is a great start, thanks for putting together!!  Left some wording comments and a proposal to strike one entry pending others' review.

There is some information not ported over from Notion which I would like to see.  Specifically, can you add the `Administrator` column from the table?  Not opinionated on the name of the column, but for all of these config values, it is important to know who/what the protocol allows to change the config, if any.

Personally, I really liked having the concrete link to code as well, but I think that's a nice-to-have.

specs/glossary.md
```diff
@@ -483,6 +484,22 @@ When submitted to Ethereum calldata, the batcher transaction's receiver must be
 transaction must also be signed by a recognized batch submitter account. The recognized batch submitter account
 is stored in the [System Configuration][system-config].
 
+## Batch submission frequency
+
+Within the [sequencing-window] constraints the batcher is permitted by the protocol to submit L2 blocks to for
+data-availability at any time. The batcher software allows for dynamic policy configuration by its operator.
```
mds1: nit, extra word
```suggestion
Within the [sequencing-window] constraints the batcher is permitted by the protocol to submit L2 blocks for
data-availability at any time. The batcher software allows for dynamic policy configuration by its operator.
```

specs/protocol/configurability.md
```diff
@@ -0,0 +1,81 @@
...
+| Data Availability                     | Where L2 transaction calldata is posted.                                                                                     |                                     |
+| Fault Proof Window / Challenge Window | Duration network participants have to challenge the integrity of an output root.                                             |                                     |
+| [Fee margin](https://github.com/ethereum-optimism/optimism/blob/c927ed9e8af501fd330349607a2b09a876a9a1fb/packages/contracts-bedrock/src/L1/SystemConfig.sol#L281-L283)                            | Markup on transactions compared to the raw L1 data cost.                                                                     | [System Config Owner](#admin-roles)                 |
+| Proxy Fee vault contracts             | Proxy contracts which point to implementation contracts that dictate how user fees are distributed.                          |                                     |
```
mds1: @ben-chain Why are fee vault contracts explicitly called out, but not other L2 predeploys? (this was carried over from your standard config strawman)

specs/protocol/configurability.md
mds1: I think we do still want to include L1 contracts here because there is a corresponding requirement that to be a standard chain you must be using a governance approved version of the L1 contracts.

In other words, the implementations are technically a configurable part of the OP Stack, but any change means you are not a standard chain, and we need to convey that requirement. If we remove this entry from the table we can't convey that.

specs/protocol/configurability.md
ben-chain: @mds1 good callout. @blmalone This can be removed in favor of overall predeploys entry, and then we restrict in standard config.

specs/protocol/configurability.md
ben-chain: @mds1 I follow what you're saying, but I still feel pretty strongly that this is (or should be) a *versioning* difference, not a configuration difference.

I think the right way to convey the requirement is by requiring a version, which implies smart contracts, not by requiring certain bounds on config, which includes contracts.  As you said, "you must be using a governance approved _version_ of the L1 contracts".

It is a little less explicit in this PR's context, but I think it's also a valuable norm to say that changes to L1 contracts are a version difference, not a config difference.

specs/protocol/configurability.md
```diff
@@ -0,0 +1,79 @@
...
+When deploying the OP Stack software to a production environment,
+certain parameters about the protocol can be configured. These
+configurations can include a variety of parameters which affect the
+resulting properties of the blockspace in question.
+
+There are four categories of OP Stack configuration options:
+
+- **Consensus Parameters**: Parameters and properties of the chain that may
+  be set at genesis and fixed for the lifetime of the chain, or may be
+  changeable through a privileged account.
```
ben-chain:
```suggestion
  changeable through a privileged account or protocol upgrade.
```

specs/protocol/configurability.md
```diff
@@ -0,0 +1,79 @@
...
+| [Batch Inbox address](https://github.com/ethereum-optimism/optimism/blob/c927ed9e8af501fd330349607a2b09a876a9a1fb/packages/contracts-bedrock/src/L1/SystemConfig.sol#L176)                   | L1 address where calldata/blobs are posted (see [Batcher Transaction](../glossary.md#batcher-transaction)).                  | [L1 Proxy Admin](#admin-roles)                      |
+| [Chain ID](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/chainids.json)                              | Unique ID of Chain used for TX signature validation.                                                                         |                                     |
+| [Challenge Period](https://github.com/ethereum-optimism/superchain-registry/pull/44)                      | Length of time for which an output root can be removed, and for which it is not considered finalized.                        | [L1 Proxy Admin](#admin-roles)                      |
+| Data Availability                     | Where L2 transaction calldata is posted.                                                                                     |                                     |
```
ben-chain:
```suggestion
```
Can we delete this entry? The DA layer is not configurable currently other than call vs blobdata, which I'll leave another comment on

specs/protocol/configurability.md
```diff
@@ -0,0 +1,79 @@
...
+| [Resource config](https://github.com/ethereum-optimism/optimism/blob/c927ed9e8af501fd330349607a2b09a876a9a1fb/packages/contracts-bedrock/src/L1/SystemConfig.sol#L338-L340)                       | Config for the EIP-1559 based curve used for the deposit gas market.                                                         | [System Config Owner](#admin-roles)                 |
+| Sequencing window                     | Maximum allowed batch submission gap, after which L1 fallback is triggered in derivation.                                    |                                     |
+| [Start block](https://github.com/ethereum-optimism/optimism/blob/c927ed9e8af501fd330349607a2b09a876a9a1fb/packages/contracts-bedrock/src/L1/SystemConfig.sol#L184)                           | Block at which the system config was initialized the first time.                                                             | [L1 Proxy Admin](#admin-roles)                      |
+| Superchain target                     | Choice of cross-L2 configuration. May be omitted in isolated OP Stack deployments. Includes SuperchainConfig and ProtocolVersions contract addresses. |            |
+
+## Policy Parameters
```
ben-chain, line 48: Can we add an entry here for wherever the sequencer can choose between blobdata and calldata?

specs/protocol/configurability.md
```diff
@@ -0,0 +1,79 @@
...
+| Fault Proof Window / Challenge Window | Duration network participants have to challenge the integrity of an output root.                                             |                                     |
+| [Fee margin](https://github.com/ethereum-optimism/optimism/blob/c927ed9e8af501fd330349607a2b09a876a9a1fb/packages/contracts-bedrock/src/L1/SystemConfig.sol#L281-L283)                            | Markup on transactions compared to the raw L1 data cost.                                                                     | [System Config Owner](#admin-roles)                 |
+| [Gas limit](https://github.com/ethereum-optimism/optimism/blob/c927ed9e8af501fd330349607a2b09a876a9a1fb/packages/contracts-bedrock/src/L1/SystemConfig.sol#L298-L300)                             | L2 block gas limit.                                                                                                          | [System Config Owner](#admin-roles)                 |
+| Gas Token                             | Native token used to pay for gas.                                                                                            |                                     |
```
ben-chain:
```suggestion
```
Let's remove this as well, not configurable away from ETH in current protocol version

specs/protocol/configurability.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Policy Parameters
+
+| Config Property                       | Description                                                                                                                  | Administrator                       |
+|---------------------------------------|------------------------------------------------------------------------------------------------------------------------------|-------------------------------------|
+| Batch submission frequency            | Frequency with which batches are submitted to L1 (see [Batcher Transaction](../glossary.md#batcher-transaction)).            |                                     |
+| Compression ratio                     | How much compression the batch submitter applies to batches before submission (see [Channel](../glossary.md#channel)).       |                                     |
+| Implementation Fee vault contracts    | Implementation contracts that sit behind proxies and dictate how user fees are distributed.                                  |                                     |
```
ben-chain:
```suggestion
```

Per @mds1 's comment this should be removed

specs/protocol/configurability.md
```diff
@@ -0,0 +1,79 @@
...
+| Data Availability                     | Where L2 transaction calldata is posted.                                                                                     |                                     |
+| Fault Proof Window / Challenge Window | Duration network participants have to challenge the integrity of an output root.                                             |                                     |
+| [Fee margin](https://github.com/ethereum-optimism/optimism/blob/c927ed9e8af501fd330349607a2b09a876a9a1fb/packages/contracts-bedrock/src/L1/SystemConfig.sol#L281-L283)                            | Markup on transactions compared to the raw L1 data cost.                                                                     | [System Config Owner](#admin-roles)                 |
+| [Gas limit](https://github.com/ethereum-optimism/optimism/blob/c927ed9e8af501fd330349607a2b09a876a9a1fb/packages/contracts-bedrock/src/L1/SystemConfig.sol#L298-L300)                             | L2 block gas limit.                                                                                                          | [System Config Owner](#admin-roles)                 |
```
ben-chain:
```suggestion
```
Let's remove this -- that gaslimit setting is somewhat complex and in practice the target gas rate is downstream of this value AND sequencer behavior being set.  Base is actively working on this.

Since this row is part of the ResourceConfig, I think it's fine to just have an entry for resourceconfig.

specs/protocol/configurability.md
```diff
@@ -0,0 +1,79 @@
...
+| Compression ratio                     | How much compression the batch submitter applies to batches before submission (see [Channel](../glossary.md#channel)).       |                                     |
+| Implementation Fee vault contracts    | Implementation contracts that sit behind proxies and dictate how user fees are distributed.                                  |                                     |
+| [Output frequency](https://github.com/ethereum-optimism/optimism/blob/c927ed9e8af501fd330349607a2b09a876a9a1fb/packages/contracts-bedrock/src/L1/L2OutputOracle.sol#L104)                      | Frequency with which output roots are submitted to L1.                                                                       | [L1 Proxy Admin](#admin-roles)                      |
+
+## Admin Roles
+
+| Config Property                       | Description                                                                                                                  | Administrator                       | Administers                         |
```
ben-chain, line 59: Noting that my requested changes above would need to be reflected in removals here. It would be really nice to have a script to populate the `administers` field, but nice to have

specs/protocol/configurability.md
```diff
@@ -0,0 +1,79 @@
...
+| [Chain ID](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/chainids.json)                              | Unique ID of Chain used for TX signature validation.                                                                         |                                     |
+| [Challenge Period](https://github.com/ethereum-optimism/superchain-registry/pull/44)                      | Length of time for which an output root can be removed, and for which it is not considered finalized.                        | [L1 Proxy Admin](#admin-roles)                      |
+| Data Availability                     | Where L2 transaction calldata is posted.                                                                                     |                                     |
+| Fault Proof Window / Challenge Window | Duration network participants have to challenge the integrity of an output root.                                             |                                     |
```
ben-chain: Duplicative with challenge period, can remove
```suggestion
```

specs/protocol/configurability.md
```diff
@@ -0,0 +1,79 @@
...
+| Superchain target                     | Choice of cross-L2 configuration. May be omitted in isolated OP Stack deployments. Includes SuperchainConfig and ProtocolVersions contract addresses. |            |
+
+## Policy Parameters
+
+| Config Property                       | Description                                                                                                                  | Administrator                       |
+|---------------------------------------|------------------------------------------------------------------------------------------------------------------------------|-------------------------------------|
+| Batch submission frequency            | Frequency with which batches are submitted to L1 (see [Batcher Transaction](../glossary.md#batcher-transaction)).            |                                     |
```
ben-chain: can we specify that the batcher role is the administrator of this and the compression ratio

ben-chain (changes requested):
Shaping up -- few more requested changes (apologies, should've caught some of these sooner)

specs/protocol/configurability.md
blmalone: https://github.com/ethereum-optimism/specs/pull/98/commits/74eb66775dc76013b8288886219dad0185c1f626

specs/protocol/configurability.md
blmalone: Sounds really useful, do we have an existing pattern for using scripts with `.md` files in the specs repo?

specs/protocol/configurability.md
blmalone: https://github.com/ethereum-optimism/specs/pull/98/commits/e2c0da2376b2870f9b182a985d0c7ce64e58e52f

I'm basing this off the docs: https://docs.optimism.io/builders/chain-operators/management/blobs#update-your-batcher-to-post-calldata 

Specifically the batchers env var: `OP_BATCHER_DATA_AVAILABILITY_TYPE`

specs/protocol/configurability.md
ben-chain: no idea xD.  Doesn't need to block this PR.

ben-chain (approved):
Woohoo!

PR #102, opened 2024-03-22
Move Security Council Safe spec to experimental

maurelian:
Makes the following changes: 

1. Moves the Liveness Checking documentation to the experimental folder, since it is not yet in production.
2. Adds a spec for a new "Deputy Guardian" module


specs/experimental/security-council-safe.md
```diff
@@ -23,7 +23,48 @@
...
-## Liveness checking Mechanism
+The Security Council uses a specially extended Safe multisig contract to provide additional security
+guarantees on top of those provided by the Safe contract.
+
+## Deputy guardian module
+
+The Security Council acts as the Guardian, which is authorized to activate the [Superchain Pause](../protocol/superchain-configuration.md#pausability) functionality and for [blacklisting](../experimental/fault-proof/stage-one/bond-incentives.md#authenticated-roles) dispute game contracts.
+
+However the Security Council cannot be expected to react quickly in an emergency situation.
+Therefore the Pausability Pass Through module enables the Security Council to delegate share this
```
mds1: Typo "delegate share"—does it delegate the authority or share it?

specs/experimental/security-council-safe.md
```diff
@@ -23,7 +23,48 @@
...
+
+   /// @dev The address of the account which can pause superchain withdrawals by calling this module
+   function deputyGuardian() external view returns(address);
+
+   /// @dev Calls to the Security Council's `execTransactionFromModule()`, with the arguments
+   ///      necessary to call `pause()` on the `SuperchainConfig` contract.
+   ////     Only the deputy guardian can call this function.
+   function pause() external;
+
+   /// @dev Calls to the Security Council's `execTransactionFromModule()`, with the arguments
```
mds1:
```suggestion
   /// @dev Calls the Security Council Safe's `execTransactionFromModule()`, with the arguments
   ///      necessary to call `pause()` on the `SuperchainConfig` contract.
   ////     Only the deputy guardian can call this function.
   function pause() external;

   /// @dev Calls the Security Council Safe's `execTransactionFromModule()`, with the arguments
```

specs/experimental/security-council-safe.md
```diff
@@ -23,7 +23,48 @@
...
+
+The module has the following minimal interface:
+
+```solidity
+interface DeputyGuardianModule {
+   /// @dev The address of the Security Council Safe
+   function safe() external view returns(address);
+
+   /// @dev The address of the account which can pause superchain withdrawals by calling this module
+   function deputyGuardian() external view returns(address);
```
mds1, line 49: So the SC will only need to give one account the ability to pause? What account is that?

maurelian: In practice it will be the Foundation, I'm never sure how much these specs should focus on describing the setup of OP Mainnet, or just the specific component being described.

specs/experimental/security-council-safe.md
maurelian: Fixed locally

specs/experimental/security-council-safe.md
maurelian: Yes, I went back and forth between them. 

I think 'share' is more accurate though, as both entities will be able to make the call. (I pushed an update to make it just "share").

specs/experimental/security-council-safe.md
mds1: Yea that's fair, I'm not sure if we should mention it explicitly

mds1 (approved):

PR #105, opened 2024-03-25
Fix the count of the game tree items

ZdravkoHr:
**Description**

The game tree is a binary tree, so the number of items it contains depends on its depth, `d`. A binary tree with depth `d` would have $2^d - 1$ items. The current documentation has wrongly calculated it as $2^{d-1}$.

**Tests**

N/A

**Additional context**

N/A

**Metadata**

N/A


ajsutton (approved):

ajsutton:
Actually because `MAX_DEPTH` is 0 based, this winds up as `2^(d+1)-1`. Since that's already in https://github.com/ethereum-optimism/specs/pull/109/ I'll merge that one.  But thank you for submitting this.

PR #108, opened 2024-03-29
interop: logindex wildcard and intents

protolambda:
**Description**

Logindex-wildcard for `Identifier` to not pre-commit to the consumed event being present in a specific place. This allows the block-builder to complete the static declaration of the event.

This is a building block for intents.

### DIFF
--- a/specs/interop/messaging.md
+++ b/specs/interop/messaging.md
@@ -15,6 +15,9 @@
   - [ChainID Invariant](#chainid-invariant)
   - [Only EOA Invariant](#only-eoa-invariant)
   - [Message Expiry Invariant](#message-expiry-invariant)
+- [Special Messages](#special-messages)
+  - [`logIndex` wildcard](#logindex-wildcard)
+    - [Intents](#intents)
 - [Message Graph](#message-graph)
   - [Invalid messages](#invalid-messages)
     - [Block reorgs](#block-reorgs)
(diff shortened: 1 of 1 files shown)


specs/interop/messaging.md
```diff
@@ -162,6 +165,34 @@ The expiry invariant invalidates inclusion of any executing message with
...
 
+## Special Messages
+
+### `logIndex` wildcard
+
+[wildcard]: #logindex-wildcard
+
+The author of an executing-transaction may defer the specification of an exact `logIndex` to the block-builder.
+Deferring the `logIndex` enables the executing side to request a message to be initiated.
+The `logIndex` value of `0x01000000000000000000000000000000000000000000000000000000000000` functions as wildcard.
```
mds1, line 176: This value is only 31 bytes long which I suspect is a typo?

Also how was this value chosen? It doesn't seem very unique, perhaps using `bytes32(uint256(keccak256("interop.logIndexWildcard")) - 1)` is safer (this is the same ERC-1967 slot computation pattern used in contracts). Or even just converting a string like "interop log index wildcard" to hex and padding it

specs/interop/messaging.md
```diff
@@ -162,6 +165,34 @@ The expiry invariant invalidates inclusion of any executing message with
...
+
+The block-builder is responsible for mapping the wildcard `logIndex` to a valid `logIndex` by appending
+it to a list of indices, declared statically in the input data of the last transaction of the block.
+
+This declaring transaction must have a `to` address of `0x42ff000000000000000000000000000000000000`.
```
mds1, line 181: We should have an autogenerated specs page that documents all arbitrary addresses in the 0x42 namespace, or some CI check, to avoid collisions

Edit: probably easier to maintain this manually, and have the CI check ensure that all 0x42 addresses show up on the page

protolambda:
Closing this. The spec is stale due to other interop changes. I might bring back the wildcard logindex idea later however, as I still think it can be very useful.

PR #111, opened 2024-04-02
feat: custom gas token

tynes:
**Description**

Custom gas token spec. Goes with https://github.com/ethereum-optimism/optimism/pull/10011


specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+
+## Constants
+
+| Constant          | Value                           | Description |
+| ----------------- | ------------------------------- | ----------- |
+| `ETHER_TOKEN_ADDRESS` |  `address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE)` | Represents ether for gas paying asset |
+| `DEPOSITOR_ACCOUNT` | `address(0xDeaDDEaDDeAdDeAdDEAdDEaddeAddEAdDEAd0001)` | Account with auth permissions on `L1Block` contract |
+| `GAS_PAYING_TOKEN_SLOT` | `bytes32(uint256(keccak256("opstack.gaspayingtoken")) - 1)` | Storage slot that contains the address and decimals of the gas paying token |
+
+## Properties of a Gas Paying Token
```
geoknee, line 71: These properties could be validated automatically in the `superchain-registry` CI checks.

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+  - [Wrapped Ether](#wrapped-ether)
+- [Future Directions](#future-directions)
+  - [Enabling a L2 Native ERC20 Token to Pay for Gas](#enabling-a-l2-native-erc20-token-to-pay-for-gas)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+Custom gas token is also known as "custom L2 native asset". It allows for an ERC20 token to be collateralized on L1
+to back the native asset on L2. The native asset is used to buy gas which is used to pay for resources on the network
```
mds1: I typically read "collateralized" to mean the the asset is backed by something so to me this reads as if the L1 asset needs to be backed by something on L1. Maybe something like:
```suggestion
Custom gas token is also known as "custom L2 native asset". It allows for an L1-native ERC20 token to collateralize and act as the gas token on L2. The native asset is used to buy gas which is used to pay for resources on the network
```

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
+# Custom Gas Token
```
mds1, line 1: Sometimes we use "custom gas token" and other times "gas paying token". Should we standardize, or have a glossary table defining them and saying they're interchangeable?

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+| `ETHER_TOKEN_ADDRESS` |  `address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE)` | Represents ether for gas paying asset |
+| `DEPOSITOR_ACCOUNT` | `address(0xDeaDDEaDDeAdDeAdDEAdDEaddeAddEAdDEAd0001)` | Account with auth permissions on `L1Block` contract |
+| `GAS_PAYING_TOKEN_SLOT` | `bytes32(uint256(keccak256("opstack.gaspayingtoken")) - 1)` | Storage slot that contains the address and decimals of the gas paying token |
+
+## Properties of a Gas Paying Token
+
+The gas paying token MUST satisfy the standard [ERC20](https://eips.ethereum.org/EIPS/eip-20) interface
+and implementation. It MUST be a L1 contract.
+
+The gas paying token cannot:
```
mds1: Open question to think about: We're trying to automate all checks that verify a chain meets the standard config. Once the custom gas token is part of this, how can we automatically enforce and validate these properties onchain or via offchain checks (in the superchain registry)? 

- Fee on transfer: Simulate a few transfers from random accounts and check before and after balance.
- Rebase: TBD, seems hard to generalize.
- Call hooks on transfer: Can analyze the trace .
- Decimals: Straightforward.
- Balance/allowance modifications:  TBD, seems hard to generalize.

However, if a token knows they are being checked tjese can be gamed unless we did bytecode analysis, which would get very complicated. Additionally, since token and SystemConfig contracts are upgradeable, properties that used to be true can change with an upgrade.

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+
+It MUST be enforced on chain that the token has 18 or less decimals to guarantee no losses of precision when
+depositing a token that has a different number of decimals.
+
+## Configuring the Gas Paying Token
+
+The gas paying token is set within the L1 `SystemConfig` smart contract. This allows for easy access to the information
+required to know if an OP Stack chain is configured to use a custom gas paying token. The gas paying token is set
+during initialization and cannot be modified until another initialization. It is assumed that the chain operator
+will not modify the gas paying token address unless they specifically decide to do so and appropriately handle collateralization.
```
mds1: I agree with SystemConfig as the source of truth for the custom gas token info, but allowing the gas token change feels like it adds a lot of risks and footguns. AFAIK no chain has ever changed their gas token so I'd prefer if we can keep it static per chain ID. We can do this by having the SystemConfig implementation prohibit writing to `GAS_PAYING_TOKEN_SLOT` if the data in that slot is non-zero

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+
+If the gas paying token address is set to the `ETHER_TOKEN_ADDRESS` then the system is configured to use `ether` as the
+native asset on L2. Note that if nothing is stored in state at the `GAS_PAYING_TOKEN_SLOT` then it is considered that
+`ether` is used to pay for gas and getters on smart contracts SHOULD return the `ETHER_TOKEN_ADDRESS`.
+
+When the gas paying token is configured in the `SystemConfig`, it kicks off a special deposit transaction that is from the
+`DEPOSITOR_ACCOUNT` and calls the `L1Block` contract so that it can store the gas paying token address and its number of decimals
```
mds1: Out of scope for this spec, but `L1Block` now stores more than just L1 Block info (gas token info, interop dependency set), so we should probably rename it to `L1Attributes` or similar. However I know we've avoided renaming OptimismPortal since it's a large diff and affects many repos, so we probably don't want to rename this for the same reason

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+
+### OptimismPortal
+
+The `OptimismPortal` is updated with a new interface specifically for depositing custom tokens.
+
+#### `depositERC20Transaction`
+
+The `depositERC20Transaction` is useful for sending custom gas tokens to L2. It is broken out into its own interface
+to maintain backwards compatibility, simplify the implementation and prevent footguns when trying to deposit an
+ERC20 native asset into a chain that uses `ether` as its native asset.
```
mds1: Let's expand on what these footguns are, even if it's just in a footnote. This helps ensure any info/risks in your head are documented to justify the decision and make sure we're all in agreement on the rationale for the proposed design in this spec.

The main pushback I can think of is that UX is probably better for e.g. bridge UI developers if there's only one method they need to call for all gas token deposit transactions. However the flow would be different anyway due to needing approve transactions anyway so this wouldn't really help them simplify much anyway

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+
+```solidity
+function depositERC20Transaction(
+    address _to,
+    uint256 _mint,
+    uint256 _value,
```
mds1, line 196: What is the mint vs. value args? Let's document those here and why we need both

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+### StandardBridge
+
+The `StandardBridge` contracts on both L1 and L2 MUST be aware of the custom gas token. The `ether` specific ABI on the
+`StandardBridge` is disabled when custom gas token is enabled.
+
+This includes the following methods:
+
+- `bridgeETH(uint32,bytes)`
+- `bridgeETHTo(address,uint32,bytes)`
+- `receive()`
```
mds1, line 246:
```suggestion
- `bridgeETH(uint32,bytes)`
- `bridgeETHTo(address,uint32,bytes)`
- `receive()`
- `finalizeBridgeETH(address, address, uint256, bytes)`
```

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+
+The following diagram shows the control flow for when a user attempts to send `ether` through the `StandardBridge`, either
+on L1 or L2.
+
+![](../static/assets/custom-token-bridge-revert.png)
```
mds1: Can we convert all the PNGs to mermaid for maintainability so it's easier to edit if/when needed?

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+
+### CrossDomainMessenger
+
+The `CrossDomainMessenger` contracts on both L1 and L2 MUST not be used to send the native asset between domains.
```
mds1: Any native asset including ETH, or only non-ETH native assets?

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+It is possible to support this in the future, but for now the feature is disabled due to the high probability of footguns.
+Using the `CrossDomainMessenger` with a custom gas token would involve modifying the `sendMessage` implementation.
```
mds1: Similar to above, would be great if we can expand on the risks/justification here

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+
+### L1Block
+
+The `L1Block` contract is upgraded with a permissioned function that allows it to set the `GAS_PAYING_TOKEN_SLOT`.
```
mds1: A new function, or simply extending the existing one? Also per the above comment, if you agree the gas paying token is static, we'll similarly want to ensure the gas token data can only be set once

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+The `DEPOSITOR_ACCOUNT` MUST be the only account that can call the setter function. Any L2 contract that wants
+to learn the address of the gas paying token can call the getter on the `L1Block` contract.
```
mds1: We should define the address and decimal getter interfaces here

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+
+## User Flow
+
+Users should deposit native asset by calling `depositERC20Transaction` on the `OptimismPortal` contract.
```
mds1, line 353: My understanding is that currently ETH (native token) deposits were usually done through the StandardBridge. But now it will be use standard bridge for ETH gas token deposits and OptimismPortal for ERC20 gas token deposits? Feels like they should be the same contract

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+Users must first `approve` the address of the `OptimismPortal` so that the `OptimismPortal` can use
+`transferFrom` to take ownership of the ERC20 asset. If the `transferFrom` fails, a [permit2](https://github.com/Uniswap/permit2)
+`transferFrom` is attempted as a fallback.
```
mds1: I like this idea but would consider removing it from this initial version to shrink scope/risk. A future UX-focused project could add permit/permit2 based deposits separately when desired. Once we add permit2 we can't easily remove it since someone will rely on it. Therefore it may be preferable to wait and see what the industry standardizes around first, because if contract wallets starting gaining traction the need for permit2 goes away

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+It is possible to support this in the future, but for now the feature is disabled due to the high probability of footguns.
+Using the `CrossDomainMessenger` with a custom gas token would involve modifying the `sendMessage` implementation.
+
+### L1Block
+
+The `L1Block` contract is upgraded with a permissioned function that allows it to set the `GAS_PAYING_TOKEN_SLOT`.
+The `DEPOSITOR_ACCOUNT` MUST be the only account that can call the setter function. Any L2 contract that wants
+to learn the address of the gas paying token can call the getter on the `L1Block` contract.
+
+## User Flow
```
mds1, line 321: I would like to see user flows fleshed out a bit more, perhaps in an appendix section since it's peripheral to the spec itself. I'm picturing tables like this:

### When ETH is the Gas Paying Token

| Scenario | Method | Prerequisites |
|----------|--------|---------------|
| ETH Deposits | `SomeContract.someMethod(uint256)` |  |
| ERC20 Deposits |  | Approve `OtherContract` to spend the ERC20 |
| ETH Withdrawals |  |  |
| ERC20 Withdrawals |  |  |

### When an ERC20 Token is the Gas Paying Token

| Scenario | Method | Prerequisites |
|----------|--------|---------------|
| ETH Deposits | `SomeContract.someMethod(uint256)` |  |
| Gas Token ERC20 Deposits |  | Approve `OtherContract` to spend the ERC20 |
| Non-Gas Token ERC20 Deposits |  |  |
| ETH Withdrawals |  |  |
| Gas Token ERC20 Withdrawals |  |  |
| Non-Gas Token ERC20 Withdrawals |  |  |

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+  - L1Block deployment
+  - L2CrossDomainMessenger deployment
+  - L2StandardBridge deployment
+  - Update L1Block Proxy ERC-1967 Implementation Slot
+  - Update L2CrossDomainMessenger Proxy ERC-1967 Implementation Slot
+  - Update L2StandardBridge Proxy ERC-1967 Implementation Slot
+
+### L1Block Deployment
+
+- `from`: `0x4210000000000000000000000000000000000002`
```
mds1, line 413: Where do these `from` addresses come from? I vaguely recall a discussion around this convention

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+- `to`: `null`
```
mds1, line 414: Tangential this spec: Since this is a create deployment, not create2, it seems chains that start with this feature at genesis will have a different L1Block (same for other predeploys below) implementation address than chains which upgrade to this feature. I don't love this property—curious if you have any thoughts here or if this is tracked somewhere? Would be nice to switch all predeploy implementation deployments to just use create2 to get consistency here

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+- `mint`: `0`
+- `value`: `0`
+- `gasLimit`: `375,000`
```
mds1: Should all the gas limits be TODO since gas needed is dependent on size of the contract being deployed?
```suggestion
- `gasLimit`: TODO
```

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+- `data`: TODO
+- `sourceHash`: TODO
+
+### L1Block Proxy Update
+
+- `from`: `0x0000000000000000000000000000000000000000`
```
mds1, line 443: Why are the update transactions from the zero address?

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+- `mint`: `0`
+- `value`: `0`
+- `gasLimit`: `50,000`
+- `data`: TODO
+- `sourceHash`: TODO
+
+## Selection of `ETHER_TOKEN_ADDRESS`
+
+It was decided to use `address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE)` to represent `ether` to push ecosystem standardization
+efforts around using this address to represent `ether` in [DeFi protocols](https://twitter.com/transmissions11/status/1775525755860169010)
```
mds1: This ERC is finalized to define this address for "An address placeholder for ETH when used in the same context as an ERC-20 token"
```suggestion
efforts around using this address to represent `ether` in [DeFi protocols](https://eips.ethereum.org/EIPS/eip-7528)
```

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+
+### OptimismPortal Token Allowance
+
+The `OptimismPortal` makes calls on behalf of users. It is therefore unsafe to be able to call the address of the custom gas
+token itself from the `OptimismPortal` because it would be a simple way to `approve` an attacker's balance and steal the
+entire ERC20 token balance of the `OptimismPortal`.
+
+### Decimal Scaling
+
+A library for scaling `uint256` representations of floating point numbers is used on both ERC20 token deposits and withdrawals.
```
mds1: What library is this?

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+This ensures that from the point of view of the L2 system, it always appears as if the number of tokens that were deposited
+are the same as the native balance. Tokens with less than or 18 decimals are supported, meaning that there is no potential for
+loss of precision due to sending in too small of a unit.
+
+### Interoperability Support
+
+Interop is supported between chains that use a custom gas token. The token address and the number of decimals are legible on
```
mds1: nit to clarify 
```suggestion
Interop is supported between chains even if they use different custom gas tokens. The token address and the number of decimals are legible on
```

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+chain. In the future we may add the ability to poke a chain such that it emits an event that includes the custom gas token address
+and its number of decimals to easily be able to introspect on the native asset of another chain.
```
mds1: If we are certain we're sticking with "events as messages" for interop (compared to e.g. state), what do you think of adding `emit GasPayingTokenSet(address indexed token, uint8 indexed decimals)` now, as part of the L1Block setter? It's very simple to have the event always emitted once when set, and no one ever needs to poke

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+
+### Wrapped Ether
+
+The `WETH9` predeploy at `0x4200000000000000000000000000000000000006` represents wrapped native asset and not wrapped `ether`.
+Portable and fungible `ether` across different domains is left for a future project.
```
mds1: I agree this is the simplest approach, but:
1. Let's add rationale for this decision
2. We should update WETH9 to have a name of `"string.concat("Wrapped ", L1Block.gasPayingTokenName())` and similarly for symbol. This will require name/symbol to be stored in L1Block. Otherwise I think this will cause a lot of confusion for users because all block explorers, wallets, etc. will say they have "Wrapped Ether" when it's really a different token. A static name of "Wrapped Gas Paying Token" is still an improvement, and a smaller diff to implement, but still might be confusing for users who are using AA wallets and don't know what gas is. WETH9 is not proxied, but I think a version of this change is needed regardless.

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,306 @@
...
+
+## Future Directions
+
+### Enabling a L2 Native ERC20 Token to Pay for Gas
+
+This is an experimental feature and it is not clear if it will ever be included in the OP Stack.
```
mds1: In what way is it experimental? I interpreted this section to mean it's a work in progress and may never be completed. If that is correct, what do you think about removing this from the spec and moving this content to an issue? I believe so far the custom gas token demand is all for the L1-based token that's spec'd out, so we could leave this out to keep it more focused, and if we do hear demand for this feature and plan to implement we can add it back to the spec to collect feedback at that time.

specs/experimental/custom-gas-token.md
mds1, line 1: We should take a look at existing custom gas token implementations, like Frax's and others that are out there, and make sure this spec would work for their token with no security or UX hit. 

For example the way Frax's works is: Users send ETH to the portal and the deposit transaction goes through like normal, but they receive frxETH on L2 which is the gas token. Periodically on L1, ETH is swept from the Portal by the FrxEthMinter contract which mints frxETH into the portal 1:1 for ETH. When withdrawing from L2, you always get frxETH.

With this spec the users would have to send ETH to the FrxEhMinter contract directly, which would instead mint frxETH on demand, send it to the portal and deposit it for the user. This seems like it would be ok, but we may want to confirm.

specs/experimental/custom-gas-token.md
tynes: This is a great callout. I don't love calling `ether` a token. This is why the "native asset" language is good. I will add something that clarifies

specs/experimental/custom-gas-token.md
tynes: This decision stems from a lack of standardization on the `initialize` logic. Right now we generally reset everything on an `initialize`. Ideally we move away from the `initialize` pattern in the future because of this and have "versioned storage" with migrations between the versions, where the logic for the storage migrations live in their own contract

tynes: I think it is generally more simple to just set the contract address again in `initialize` which will trigger the setter call on the L1Block contract. Removing that ability adds an extra test case (initialize again, see no change) but it does remove a footgun where it is changed accidentally. Our upgrade tooling should be able to assert that there is no state diff for that slot. Using a 2 step process can still allow for the change in token - just set to `address(0)` using `StorageSetter.setBytes32` and then pass in a new token address into the initializer

tynes: I am fine to update the spec + implementation with this given my thoughts above, although we will have this sort of problem as long as we use `initialize` - how do we maintain consistency and sanity with what slots are set and not set?

specs/experimental/custom-gas-token.md
tynes: I definitely would like to rename it to `L1Attributes` then it would match the "L1 attributes transaction". I think this was just coordination issues with the launch of bedrock

specs/experimental/custom-gas-token.md
tynes: One consideration is that we will likely need to have an "event horizon" where events older than a certain age cannot be used as an initiating message. This is because of memory restrictions in the fault proof. The fault proof needs to reproduce the event via execution, and it would need to walk back a ton of blocks for really old events. Which is why the poke pattern would be required. Not opposed to adding the event now though for clarity, but feels premature to introduce the poking method now

specs/experimental/custom-gas-token.md
tynes: Ok good thinking, I think i just got a little trigger happy when working on this. Will move to an issue

specs/experimental/custom-gas-token.md
tynes: Frax's implementation won't be able to match this 1:1 because their portal can accept both `ether` and frax eth. This sort of pattern is fine for the usecase of the custom gas token being liquid staked, but I won't work for more arbitrary tokens like USDC. I don't really see a way around this unless we did something complex like "send ether to portal, market buy erc20 token with it on uniswap, then deposit that" but this won't work for frax eth so we would need an abstract function that any chain could implement for conversion and that adds a mega headache for standardization process + security

specs/experimental/custom-gas-token.md
tynes: To some degree there will have to be social consensus on these. We can try to automate as much as possible but EVM is permissive and there can always be ways to try to get around automated detection

specs/experimental/custom-gas-token.md
tynes: This is a great callout, will add a section noting this

specs/experimental/custom-gas-token.md
mds1: > Using a 2 step process can still allow for the change in token - just set to `address(0)` using `StorageSetter.setBytes32` and then pass in a new token address into the initializer

True, this is a great counterargument. Backing to up think of my original motivation here, in general the situation is:
- For chains that don't want to be standard chains, they can change their implementations and do whatever they want, so logic preventing this does not matter.
- For chains that do want to be standard chains, the implementation is fixed and logic preventing this helps them avoid mistakes.

So perhaps a potential solution in the initialize method is (pseudocode that ignores `GAS_PAYING_TOKEN_SLOT` for simplicity):
```solidity
function initialize(address _gasPayingToken) external {
  // `gasPayingToken` is the value in storage.
  require(gasPayingToken == address(0) || gasPayingToken == _gasPayingToken);
  gasPayingToken = _gasPayingToken;
}
```

This way, even if chains want to be standard chains but do their upgrades without using our upgrade tooling, they still get guarantees the token address won't change. They can work around this with StorageSetter like you said, but it seems extremely unlikely to *accidentally* do that, so I think something like this would be sufficient.

> set the contract address again in `initialize` which will trigger the setter call on the L1Block contract.

For my own understanding, what's the mechanism through which this works, i.e. how specifically does setting the storage slot to the pre-existing value trigger the L1Block setter call?

specs/experimental/custom-gas-token.md
mds1: Interesting! Did not realize that event horizon requirement. 

> it would need to walk back a ton of blocks for really old events

What does this mean exactly and how does that result in large memory execution compared to more recent logs?

Either way, given that, I agree my suggestion is premature and poking may be preferable. Ideally we have just one way to emit the event for simplicity/fewer code paths. So let's just remember to track this open question somewhere.

specs/experimental/custom-gas-token.md
tynes: We have tried a similar pattern in the past and it adds complexity with downgrades or reinit to the same implementation. We ended up bricking the contract in a way where the source code needed to be modified before a reinitialize was possible 😅 

I am definitely open to following this approach as its semantics are clear but then we will need to be mindful about the semantics forever, which I don't think is that bad in this case. Adding in a gas optimization to not sstore if its the same is also something i would like to avoid

specs/experimental/custom-gas-token.md
tynes: Ok I can expand on the footguns. It would be a large refactor to support the xdm `SendMessage` interface due to the way that the abstractions work. I agree its not ideal

specs/experimental/custom-gas-token.md
tynes: Sure, good call

specs/experimental/custom-gas-token.md
tynes: Its chosen arbitrarily, just needs to be an account with 1) no code and 2) no known private key

specs/experimental/custom-gas-token.md
tynes: This is true and @protolambda had the idea to fix this by defining every L2 contract upgrade via sets of upgrade txs. This would mean the `0xc0d3..` addresses would all be `v1.0.0` and then the genesis generation would apply all of the upgrade txs against the genesis state before spitting out a `genesis.json`. I consider this tech debt, we didn't have a good story for what L2 upgrades should be. To be honest I would like to make the L2 `ProxyAdmin` ownable by the `DEPOSITOR_ACCOUNT` and then define upgrade txs to go through there instead. We would have to have a common address space in the predeploys that we commit to never using that are free for application layer usage and they would need a diff `ProxyAdmin` to manage them

specs/experimental/custom-gas-token.md
tynes: The fault proof program walks backwards from a L2 blockhash that is on L1 until its able to have all of the inputs (L2 block data) to re-execute to recreate the data required to ensure that the proposed commit is "correct". If a log from a really long time ago is used as part of an executing message, then we will need to walk back to the point where that log is created as a side effect of evm execution

specs/experimental/custom-gas-token.md
tynes: People will certainly be unhappy if we do not do this. Previously we didn't want to change the weth implementation due to the desire to just use the same bytecode. I decided on not including this to cut scope. Having some product insight here would be useful cc @K-Ho.  I am open to this change but it just adds extra scope

specs/experimental/custom-gas-token.md
tynes: Perhaps it doesn't make sense to explicitly say library here but the PR has a fully tested library for this purpose

specs/experimental/custom-gas-token.md
tynes: Tracking in https://github.com/ethereum-optimism/protocol-quest/issues/222#issuecomment-2041217618

specs/experimental/custom-gas-token.md
tynes: Doing them through the standard bridge is not best practice, it just wastes gas. The reasoning behind why the standard bridge is generally used was because it just works with the way the indexer is structured now, but the indexer should be architected to read the deposit tx events from the portal and inspect event data directly to do accounting

specs/experimental/custom-gas-token.md
tynes: Ok I am fine with this

specs/experimental/custom-gas-token.md
tynes: Due to the way that the proxies work, you can bypass ownership by using `address(0)`. Otherwise the derivation pipeline would have dynamic data in it, the identity of the owner of the L2 `ProxyAdmin` would need to be stolen and used which could add race conditions and L2 RPC to the fault proof/derivation pipeline so we decided to make it call from `address(0)` so its deterministic

specs/experimental/custom-gas-token.md
tynes: Added a section on automated verification

specs/experimental/custom-gas-token.md
tynes: Added language on this in the overview section

tynes:
I'd like to add a section on "standard config"

edit: complete

specs/experimental/custom-gas-token.md
tynes: Documented in https://github.com/ethereum-optimism/protocol-quest/issues/222#issuecomment-2041250026

specs/experimental/custom-gas-token.md
tynes: Removed mentions, tracking in https://github.com/ethereum-optimism/specs/issues/114

specs/experimental/custom-gas-token.md
tynes: I am tracking this in https://github.com/ethereum-optimism/protocol-quest/issues/222#issuecomment-2041776976

specs/experimental/custom-gas-token.md
tynes: Tracking in https://github.com/ethereum-optimism/protocol-quest/issues/222#issuecomment-2041780190

specs/experimental/custom-gas-token.md
tynes: Will follow up with adding an event that is emitted during the call to set and leave out the ability to poke and have the event emitted again. See https://github.com/ethereum-optimism/protocol-quest/issues/222#issuecomment-2041782114

tynes: Addressed in https://github.com/ethereum-optimism/optimism/pull/10011/commits/d7d8ad71e192b7e995176776d4664a2e56f54e80

specs/experimental/custom-gas-token.md
tynes: This is addressed

specs/experimental/custom-gas-token.md
tynes: Addressed this comment. Would prefer to keep the logic for handling not being able to modify the token on L1 only in case somebody uses the storage setter contract method, would allow consistency, no need to bork that

specs/experimental/custom-gas-token.md
tynes: Tracking in https://github.com/ethereum-optimism/protocol-quest/issues/222#issuecomment-2041812257

specs/experimental/custom-gas-token.md
tynes: Added a comment including this information

specs/experimental/custom-gas-token.md
tynes: Added a comment addressing this

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,363 @@
...
+during initialization and cannot be modified by the `SystemConfig` bytecode. Since the `SystemConfig` is proxied,
+it is always possible to modify the storage slot that holds the gas paying token address directly during an upgrade.
+It is assumed that the chain operator will not modify the gas paying token address unless they specifically decide to
+do so and appropriately handle the consequences of the action.
+
+If the gas paying token address is set to the `ETHER_TOKEN_ADDRESS` then the system is configured to use `ether` as the
+native asset on L2. Note that if nothing is stored in state at the `GAS_PAYING_TOKEN_SLOT` then it is considered that
+`ether` is used to pay for gas and getters on smart contracts SHOULD return the `ETHER_TOKEN_ADDRESS`.
+
+![Custom Token Setup](../static/assets/custom-token-setup.png)
```
tynes: todo: replace this with mermaid

specs/experimental/custom-gas-token.md
tynes: Tracked in https://github.com/ethereum-optimism/protocol-quest/issues/222#issuecomment-2041822395

specs/experimental/custom-gas-token.md
tynes: this has been replaced with mermaid

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,391 @@
...
+in the L1 `SystemConfig` smart contract.
+
+The terms "custom gas token", "gas paying token", "native asset" and "gas paying asset" can all be used interchangebly.
+Ether is not a token.
+
+## Constants
+
+| Constant          | Value                           | Description |
+| ----------------- | ------------------------------- | ----------- |
+| `ETHER_TOKEN_ADDRESS` |  `address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE)` | Represents ether for gas paying asset |
```
0xfuturistic, line 57: `ETHER_TOKEN_ADDRESS` could be misunderstood to mean a token related to eth (i.e., a tokenized instance of ETH like WETH). could `BASE_TOKEN_ADDRESS` or `GAS_TOKEN_ADDRESS` make more sense?

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,391 @@
...
+
+The gas paying token MUST satisfy the standard [ERC20](https://eips.ethereum.org/EIPS/eip-20) interface
+and implementation. It MUST be a L1 contract.
+
+The gas paying token cannot:
+
+- Have a fee on transfer
+- Have rebasing logic
+- Have call hooks on transfer
+- Have more than 18 decimals
```
0xfuturistic: is there a minimum number of decimals?

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,391 @@
...
+- Have out of band methods for modifying balance or allowance
+
+It MUST be enforced on chain that the token has 18 or less decimals to guarantee no losses of precision when
```
0xfuturistic: does it make sense to verify this on-chain? or would that be unnecessary?

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,391 @@
...
+the validity of an ERC20 token.
+
+[erc165]: https://eips.ethereum.org/EIPS/eip-165
+[usdt]: https://forum.openzeppelin.com/t/usdt-locked-in-a-contract-that-has-a-withdraw-token-function/32154/2
+
+## Configuring the Gas Paying Token
+
+The gas paying token is set within the L1 `SystemConfig` smart contract. This allows for easy access to the information
+required to know if an OP Stack chain is configured to use a custom gas paying token. The gas paying token is set
+during initialization and cannot be modified by the `SystemConfig` bytecode. Since the `SystemConfig` is proxied,
```
0xfuturistic: nit: mention that the gas paying token is stored in storage to make clear it's not an immutable

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,391 @@
...
+it is always possible to modify the storage slot that holds the gas paying token address directly during an upgrade.
+It is assumed that the chain operator will not modify the gas paying token address unless they specifically decide to
```
0xfuturistic: how does the system cope with a change of the gas paying token? would anything break? especially curious how users would experience this change. additionally, it would be interesting to explore how an already-deployed chain handles going from eth to custom token

specs/experimental/custom-gas-token.md
tynes: If we do not verify it on chain, then we can have loss of precision for user deposits and they will lose out on a small amount of value that will get stuck in the portal and not be minted on the L2. This is because of the way that the scaling logic works

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,391 @@
...
+### Interoperability Support
+
+Interop is supported between chains even if they use different custom gas tokens. The token address and the number of
+decimals are legible on chain. In the future we may add the ability to poke a chain such that it emits an event that
+includes the custom gas token address and its number of decimals to easily be able to introspect on the native asset
+of another chain.
+
+### Wrapped Ether
+
+The `WETH9` predeploy at `0x4200000000000000000000000000000000000006` represents wrapped native asset and
```
tynes, line 509: We need to update this section potentially

specs/experimental/custom-gas-token.md
tynes: We don't necessarily care about that as a feature when we go to market so while it may be an interesting thing, we can say its out of scope. If a team wants to contribute a how to guide on this then we could accept it in the future

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,391 @@
...
+| `_value` | `uint256` | The value of the deposit transaction, used to transfer native asset that is already on L2 from L1 |
+| `_gasLimit` | `uint64` | The gas limit of the deposit transaction |
+| `_isCreation` | `bool` | Signifies the `_data` should be used with `CREATE` |
+| `_data` | `bytes` | The calldata of the deposit transaction |
+
+#### `depositTransaction`
+
+This function is the hot code path for all usage of the `L1StandardBridge` and `L1CrossDomainMessenger`, so it MUST
+be as backwards compatible as possible. It MUST revert for chains using custom gas token when `ether` is sent with
+the `CALL`. This prevents `ether` from being stuck in the `OptimismPortal`, since it cannot be used to mint native
```
0xfuturistic, line 224: is there a risk of funds becoming stuck if the chain transitions to a custom gas token AFTER an eth deposit but BEFORE the eth deposit is processed in the L2?

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,391 @@
...
+- `sendMessage(address,bytes,uint32)`
+- `relayMessage(uint256,address,address,uint256,uint256,bytes)`
+
+It MAY be safe to not always directly revert in these implementations if the revert happens elsewhere in the callstack.
+
+### SystemConfig
+
+The `SystemConfig` is modified to allow the address of the custom gas paying token to be set during the call
+to `initialize`. When the custom gas token is set, `OptimismPortal.setGasPayingToken` is called. The implementation
+of `initialize` MUST not allow the chain operator to change the address of the custom gas token if it is already set.
```
0xfuturistic: instead of relying on `SystemConfig` making a call to `OptimismPortal.setGasPayingToken`, could `OptimismPortal` read whether a custom gas token is being used by `SystemConfig` and just its behavior automatically? this would allow the contract to internalize the logic inside existing methods (instead of creating new methods to separate logic). no strong feelings about this -- just curious

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,391 @@
...
+---
+flowchart TD
+    A1(User) -->|Attempts to deposit/withdraw ether| A2(StandardBridge or CrossDomainMessenger)
+    A2 -->|Is chain using custom gas token?| A3{SystemConfig or L1Block}
+    A3 --> |Yes| A2
+    A2 --> |Revert| A1
+
+```
+
+Users should withdraw value by calling the `L2ToL1MessagePasser` directly.
```
0xfuturistic: can the withdrawal smart contract release the ERC20 token through `L2ToL1MessagePasser` without any additional logic?

specs/experimental/custom-gas-token.md
0xfuturistic: ok just read the section `Decimal Scaling` and this should be fine

specs/experimental/custom-gas-token.md
tynes: We should think of chains starting always as custom gas token chains and not transitioning between ether and custom gas token

specs/experimental/custom-gas-token.md
tynes: The `setGasPayingToken` function here emits a special deposit tx that sets it in the `L1Block` contract

tynes: The `OptimismPortal` does call to the `SystemConfig` as the source of truth

specs/experimental/custom-gas-token.md
tynes: I added a diagram like this

specs/experimental/custom-gas-token.md
tynes: Yup exactly no new logic necessary

specs/experimental/custom-gas-token.md
tynes: I noted the requirement of the above pattern directly in the spec

specs/experimental/custom-gas-token.md
tynes: We are removing support for decimals other than 18

specs/experimental/custom-gas-token.md
tynes: i will add a comment about it being network specific config

tynes:
We should clarify:

- that the cross domain messenger also has the "get custom gas token api"
- the definition of a chain using custom gas token is the address of the erc20 token being set in the special sys config storage slot
- no enforcement of `msg.value == 0` if `isCustomGasToken` in xdm `relayMessage` because we rely on the initiating side of the bridge to enforce

specs/experimental/custom-gas-token.md
mds1: Noticed that comment was unresolved—did we expand on this somewhere?

specs/experimental/custom-gas-token.md
mds1: Similarly did we expand on this footgun somewhere? Seems like some refactoring might have happened, can't find this sentence

specs/experimental/custom-gas-token.md
mds1, line 1: Extracting https://github.com/ethereum-optimism/specs/pull/111#issuecomment-2057644129 by @tynes into a comment so github forces resolution before merge

> We should clarify:
> 
> - that the cross domain messenger also has the "get custom gas token api"
> - the definition of a chain using custom gas token is the address of the erc20 token being set in the special sys config storage slot
> - no enforcement of `msg.value == 0` if `isCustomGasToken` in xdm `relayMessage` because we rely on the initiating side of the bridge to enforce

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,471 @@
...
+- [Upgrade](#upgrade)
+  - [L1Block Deployment](#l1block-deployment)
+  - [L2CrossDomainMessenger Deployment](#l2crossdomainmessenger-deployment)
+  - [L2StandardBridge Deployment](#l2standardbridge-deployment)
+  - [L1Block Proxy Update](#l1block-proxy-update)
+  - [L2CrossDomainMessenger Proxy Update](#l2crossdomainmessenger-proxy-update)
+  - [L2StandardBridge Proxy Update](#l2standardbridge-proxy-update)
+- [Selection of `ETHER_TOKEN_ADDRESS`](#selection-of-ether_token_address)
+- [Standard Config](#standard-config)
+- [Security Considerations](#security-considerations)
```
mds1, line 40: Not quite a security consideration, maybe fits somewhere else, but we probably should mention something about the fee params and how the transaction fees paid in the custom gas token likely won't cover the L1 posting costs

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,471 @@
+# Custom Gas Token
```
mds1, line 1: Did we ever made a decision around calling this Custom Native Asset instead? I know @maurelian suggested this, and I do agree with it, but don't have a strong opinion. Probably best to go with whatever we think is more intuitive to users/devs

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,471 @@
...
+| `GAS_PAYING_TOKEN_SLOT` | `bytes32(uint256(keccak256("opstack.gaspayingtoken")) - 1)` | Storage slot that contains the address and decimals of the gas paying token |
+| `GAS_PAYING_TOKEN_NAME_SLOT` | `bytes32(uint256(keccak256("opstack.gaspayingtokenname")) - 1)` | Storage slot that contains the ERC20 `name()` of the gas paying token |
+| `GAS_PAYING_TOKEN_SYMBOL_SLOT` | `bytes32(uint256(keccak256("opstack.gaspayingtokensymbol")) - 1)` | Storage slot that contains the ERC20 `symbol()` of the gas paying token |
+
+## Properties of a Gas Paying Token
+
+The gas paying token MUST satisfy the standard [ERC20](https://eips.ethereum.org/EIPS/eip-20) interface
+and implementation. It MUST be a L1 contract.
+
+The gas paying token cannot:
```
mds1: nit
```suggestion
The gas paying token MUST NOT:
```

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,471 @@
...
+- `finalizeBridgeETH(address, address, uint256, bytes)`
+
+The following legacy methods MUST also revert when custom gas token is being used:
+
+- `depositETH(uint32,bytes)`
+- `depositETHTo(address,uint32,bytes)`
+- `finalizeETHWithdrawal(address,address,uint256,bytes)`
+- `withdraw(address,uint256,uint32,bytes)`
+- `withdrawTo(address,address,uint256,uint32,bytes)`
+- `finalizeDeposit(address,address,address,address,uint256,bytes)`
```
mds1, line 260: Why do these ones need to revert? They don't necessarily involve transfer of the gas token right?

Also nit, these 3 methods are only in L2StandardBridge, and 3 above are only in the L1StandardBridge—can we clarify that a bit? We're under a `StandardBridge` header so this is slightly misleading

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,471 @@
...
+Some getters are added to the `SystemConfig`:
+
+```solidity
+function gasPayingToken() public view returns (address, uint8);
+function gasPayingTokenName() public view returns (string);
+function gasPayingTokenSymbol() public view returns (string);
+```
+
+If custom gas token is not used, then `gasPayingToken()` should return `(ETHER_TOKEN_ADDRESS,18)`,
+`gasPayingTokenName` should return `ether` and `gasPayingTokenSymbol` should return `ETH`.
```
mds1: not blocking nit since most token names start with a capital, you can reject if the code diff here is annoying this late
```suggestion
`gasPayingTokenName` should return `Ether` and `gasPayingTokenSymbol` should return `ETH`.
```

mds1: Though one thought here is if we go `ether` then the L2 weth9 predeploy would change to return `Wrapped ether` instead of the current `Wrapped Ether`.

specs/experimental/custom-gas-token.md
tynes: I lean towards renaming it as well but we were unable to get complete consensus on it so its staying as is

specs/experimental/custom-gas-token.md
tynes: Ah you are right cc @0xfuturistic can you double check this is right in the code?

specs/experimental/custom-gas-token.md
tynes: I believe it should say they revert if non zero callvalue

specs/experimental/custom-gas-token.md
0xfuturistic: should we update the specs to match the code (i.e., stick to `Ether`) or update the code to match the specs (i.e., stick to `ether`). imo, the former feels more consistent and would require not adding additional logic to the code (such as a system that capitalized the name when computing `Wrapped + name`)

tynes: It should be updated to be `"Ether"` so its consistent with weth9 behavior

specs/experimental/custom-gas-token.md
0xfuturistic: point 3 would require us to update the code, since currently `msg.value == 0` is being enforced by xdm (cc @tynes)

specs/experimental/custom-gas-token.md
0xfuturistic: what could this text be like? not sure what it could look like @mds1

specs/experimental/custom-gas-token.md
0xfuturistic: leaving this open until there's a conclusion

specs/experimental/custom-gas-token.md
0xfuturistic: originally we had it that the revert would happen for non-zero callvalue, but then we decided to disable the methods that have `ETH` in their names altogether when a custom gas token is used

added the headers for `L1StandardBridge`, `L2StandardBridge`

0xfuturistic:
implemented comments in this PR: https://github.com/ethereum-optimism/specs/pull/138

specs/experimental/custom-gas-token.md
tynes: @0xfuturistic - i already updated the code with this, it was because including that check ate too much gas and caused invariant tests around `callWithMinGas` to fail

specs/experimental/custom-gas-token.md
tynes: We could add a section above security considerations with text like so:

### Fees

The OP Stack natively charges fees in terms of ether due to the fee formula taking into account the basefee and blobbasefee. When a custom gas token is used, fees are paid in the custom gas token but the conversion rate to ether is not taken into account as part of the protocol. It is assumed that the fees will be configured by the chain operator such that the revenue earned in custom gas token can be swapped into ether to pay for posting the data to L1.

specs/experimental/custom-gas-token.md
tynes: I am going to close this, if we want to revisit this we can bring it up on slack

tynes:
Can confirm we are 25 bytes over the codesize limit

specs/experimental/custom-gas-token.md
mds1: Here is the updated part of the spec, annotated with my thoughts/open questions:

---

The following methods MUST revert when custom gas token is being used:

- `bridgeETH(uint32,bytes)`: ✅ Agreed, because this deals with ETH
- `bridgeETHTo(address,uint32,bytes)`: ✅ Agreed, because this deals with ETH
- `receive()`: ✅ Agreed, because this deals with ETH
- `finalizeBridgeETH(address, address, uint256, bytes)`: ✅ Agreed, because this deals with ETH

The following legacy methods MUST also revert when custom gas token is being used:

- In `L1StandardBridge`:
    - `depositETH(uint32,bytes)`: ✅ Agreed, because this deals with ETH
    - `depositETHTo(address,uint32,bytes)`: ✅ Agreed, because this deals with ETH
    - `finalizeETHWithdrawal(address,address,uint256,bytes)`: ✅ Agreed, because this deals with ETH
- In `L2StandardBridge`:
    - `withdraw(address,uint256,uint32,bytes)`: ❓ Unclear on rationale for this one
    - `withdrawTo(address,address,uint256,uint32,bytes)`: ❓ Unclear on rationale for this one
    - `finalizeDeposit(address,address,address,address,uint256,bytes)`: ❓ Unclear on rationale for this one

specs/experimental/custom-gas-token.md
mds1: That looks great to me, thank you @tynes!

specs/experimental/custom-gas-token.md
mds1: > 1. that the cross domain messenger also has the "get custom gas token api"

This means that the `gasPayingToken`, `gasPayingTokenName`, and `gasPayingTokenSymbol` getters also live on the CrossDomainMessenger? If so I don't yet see that in the "Contract Modifications" section

> 2. the definition of a chain using custom gas token is the address of the erc20 token being set in the special sys config storage slot

@tynes What did you mean here? I'm not quite following the "the definition of a chain" part. Do you mean something like "the source of truth for the gas token used by a chain is the address of the erc20 token being set in the special sys config storage slot"? If so I agree we should have a sentence in the spec to be explicit here

Under "Configuring the Gas Paying Token" we have "The gas paying token is set within the L1 `SystemConfig` smart contract", and I think that's the right spot to expand on this.

> 3. no enforcement of msg.value == 0 if isCustomGasToken in xdm relayMessage because we rely on the initiating side of the bridge to enforce

The spec does currently say "The following methods MUST revert when `CALLVALUE` is non zero" then lists the XDM's `relayMessage`. So it sounds like the spec is not consistent with the implementation.

Having a redundant check in `relayMessage` would be nice, but I understand the `callWithMinGas` part is sensitive. So relying on the initiating side to enforce this requirement seems safe as long as we have adequate tests on that requirement

specs/experimental/custom-gas-token.md
tynes: The unclear ones are because they can be used for sending ether between domains. They also work with other tokens which is why it should be clear that they revert when `CALLVALUE` is non zero

specs/experimental/custom-gas-token.md
tynes: Maybe we should define a section on the interface since its in many locations:

- `gasPayingToken`
- `gasPayingTokenName`
- `gasPayingTokenSymbol`
- `isCustomGasToken`

This interface is on:
- `StandardBridge`
- `CrossDomainMessenger`
- `L1Block`
- `SystemConfig`

tynes: > The spec does currently say "The following methods MUST revert when CALLVALUE is non zero" then lists the XDM's relayMessage. So it sounds like the spec is not consistent with the implementation.

If implemented correctly, there can never be a call to `relayMessage` when `CALLVALUE` is non zero. So we can hold the spec true without implementing the check explicitly. Its not ideal to have implicit behavior, so we should note this. Does that sound sane to you @mds1 ?

> Having a redundant check in relayMessage would be nice, but I understand the callWithMinGas part is sensitive. So relying on the initiating side to enforce this requirement seems safe as long as we have adequate tests on that requirement

We do have test coverage for this

specs/experimental/custom-gas-token.md
mds1: I see, so I think it'd be more precise for the spec to read like the below? (maybe a nit here, just want to make sure the spec is clear)

---

The following legacy methods in `L1StandardBridge` MUST revert when custom gas token is being used:

- `depositETH(uint32,bytes)`
- `depositETHTo(address,uint32,bytes)`
- `finalizeETHWithdrawal(address,address,uint256,bytes)`

The following legacy methods in `L2StandardBridge` MUST revert when custom gas token is being used and the `CALLVALUE` is nonzer:
- `withdraw(address,uint256,uint32,bytes)`
- `withdrawTo(address,address,uint256,uint32,bytes)`
- `finalizeDeposit(address,address,address,address,uint256,bytes)`

specs/experimental/custom-gas-token.md
mds1: > Maybe we should define a section on the interface since its in many locations

This makes sense to me, we can define an `IGasToken` interface and in each contract's section, note that they implement the interface

> If implemented correctly, there can never be a call to relayMessage when `CALLVALUE` is non zero. So we can hold the spec true without implementing the check explicitly. Its not ideal to have implicit behavior, so we should note this. Does that sound sane to you @mds1 ?

I see what you're saying. But if implemented correctly, the implementor can never write a passing test for the requirement "`relayMessage` must revert when callvalue is nonzero" because it's impossible to send a message that would get relayed with such a condition. You could assume a bug let it through, and then write the test, but our implementation would fail that test since we aren't implementing the check in the relayMessage

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,488 @@
...
+
+The `StandardBridge` and `CrossDomainMessenger` are symmetric in their APIs between L1 and L2
+meaning that "sending to the other domain" indicates it can be used for both deposits and withdrawals.
+
+### When ETH is the Native Asset
+
+| Scenario | Method | Prerequisites |
+|----------|--------|---------------|
+| Native Asset Send to Other Domain | `L1StandardBridge.bridgeETH(uint32,bytes) payable` | None |
+| Native Asset Send to Other Domain | `L1CrossDomainMessenger.sendMessage(address,bytes,uint32) payable` | None |
```
mds1: Should the second row's scenario be "Native Asset and/or Message Send to Other Domain"? Something to distinguish it's use case from the prior row

specs/experimental/custom-gas-token.md
0xfuturistic: good idea!

specs/experimental/custom-gas-token.md
tynes: We didn't expand on this but basically the footgun is as follows

- on withdrawals for ether, the balance is transferred from portal to messenger to user
- if call fails, ether is stuck in portal, can call `relayMessage` again directly to get it out. nobody can steal it
- for custom gas token on withdrawals, we would need to follow the same pattern, where the token is transferred from the portal to the messenger and then to the user. would need to revert if the transfer fails
- **Need to have different branching logic for what function to call in L1xdm if custom gas token or not, would need to call the `depositERC20Transaction` method**.

To me transferring to the intermediate contract makes no sense and then also the branching logic in the l1xdm for deposits would add a ton of new code to test

specs/experimental/custom-gas-token.md
tynes: This has been resolved

specs/experimental/custom-gas-token.md
tynes: This has been resolved

tynes: > Do you mean something like "the source of truth for the gas token used by a chain is the address of the erc20 token being set in the special sys config storage slot"? If so I agree we should have a sentence in the spec to be explicit here

Yes this is what I mean and the spec now reflects this

specs/experimental/custom-gas-token.md
tynes: This has been addressed. The footgun is simply around being explicit when depositing an ERC20 token, hence the different function name.

specs/experimental/custom-gas-token.md
tynes: Just simplifying the language to say that the messenger can support this functionality in the future if necessary with a redesign. No need to go into additional detail on top of that here

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,501 @@
...
+`gasPayingTokenName` should return `Ether` and `gasPayingTokenSymbol` should return `ETH`.
+
+This interface applies to the following contracts:
+
+- `L1StandardBridge`
+- `L2StandardBridge`
+- `L1CrossDomainMessenger`
+- `L2CrossDomainMessenger`
+- `L1Block`
+- `SystemConfig`
```
mds1: Why this set of contracts? i.e. we can consider deduping to a single source of truth on each chain to reduce the amounts of code/entry points for the gas token functionality
```suggestion
This interface applies to the following contracts:

- `L1Block`
- `SystemConfig`
```

mds1 (approved):
Spec came out great, thank you!

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,501 @@
...
+Custom gas token is also known as "custom L2 native asset". It allows for an L1-native ERC20 token to collateralize
+and act as the gas token on L2. The native asset is used to buy gas which is used to pay for resources on the network
+and is represented by EVM opcodes such as `CALLVALUE` (`msg.value`).
+
+Both the L1 and L2 smart contract systems MUST be able to introspect on the address of the gas paying token to
+ensure the security of the system. The source of truth for the configuration of the address of the token is
+in the L1 `SystemConfig` smart contract.
+
+The terms "custom gas token", "gas paying token", "native asset" and "gas paying asset" can all be used interchangebly.
+Ether is not a token.
```
refcell: This just reads weird to me, can we change this to
```suggestion
Note, Ether is not a **custom** gas token, but may be used to pay gas. More on this in [## Configuring the Gas Paying Token](#-Configuring-the-Gas-Paying-Token).
```

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,501 @@
...
+| `ETHER_TOKEN_ADDRESS` |  `address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE)` | Represents ether for gas paying asset |
+| `DEPOSITOR_ACCOUNT` | `address(0xDeaDDEaDDeAdDeAdDEAdDEaddeAddEAdDEAd0001)` | Account with auth permissions on `L1Block` contract |
+| `GAS_PAYING_TOKEN_SLOT` | `bytes32(uint256(keccak256("opstack.gaspayingtoken")) - 1)` | Storage slot that contains the address and decimals of the gas paying token |
+| `GAS_PAYING_TOKEN_NAME_SLOT` | `bytes32(uint256(keccak256("opstack.gaspayingtokenname")) - 1)` | Storage slot that contains the ERC20 `name()` of the gas paying token |
+| `GAS_PAYING_TOKEN_SYMBOL_SLOT` | `bytes32(uint256(keccak256("opstack.gaspayingtokensymbol")) - 1)` | Storage slot that contains the ERC20 `symbol()` of the gas paying token |
+
+## Properties of a Gas Paying Token
+
+The gas paying token MUST satisfy the standard [ERC20](https://eips.ethereum.org/EIPS/eip-20) interface
+and implementation. It MUST be a L1 contract.
```
refcell: nit
```suggestion
and implementation. It MUST be an L1 contract.
```

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,501 @@
...
+
+The gas paying token MUST NOT:
+
+- Have a fee on transfer
+- Have rebasing logic
```
refcell, line 79: Curious, why not?

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,501 @@
...
+    /// @notice Returns the gas token name.
+    function gasPayingTokenName() external view returns (string memory);
+    /// @notice Returns the gas token symbol.
+    function gasPayingTokenSymbol() external view returns (string memory);
+    /// @notice Returns true if the network uses a custom gas token.
+    function isCustomGasToken() external view returns (bool);
+}
+```
+
+If custom gas token is not used, then `gasPayingToken()` should return `(ETHER_TOKEN_ADDRESS,18)`,
```
refcell: nits
```suggestion
If a custom gas token is not used, then `gasPayingToken()` should return `(ETHER_TOKEN_ADDRESS, 18)`,
```

specs/experimental/custom-gas-token.md
```diff
@@ -0,0 +1,501 @@
...
+information about the gas paying token. The `DEPOSITOR_ACCOUNT` MUST be the only account that can call the
+setter function. This setter is differentiated from the `setL1BlockValues` functions because the gas paying
+token is not meant to be dynamic config whereas the arguments to `setL1BlockValues` are generally dynamic in nature.
+
+Any L2 contract that wants to learn the address of the gas paying token can call the getter on the
+`L1Block` contract.
+
+```solidity
+function setGasPayingToken(address _token, uint8 _decimals, byte32 _name, bytes32 _symbol) external;
+```
```
refcell, line 313: Can we specify the accessor here as well and move the mutator above the mention of the getter?
````suggestion
```solidity
function setGasPayingToken(address _token, uint8 _decimals, byte32 _name, bytes32 _symbol) external;
```

Any L2 contract that wants to learn the address of the gas paying token can call the getter on the
`L1Block` contract.

```solidity
function gasPayingToken() external view returns (address, uint8);
```
````

specs/experimental/custom-gas-token.md
tynes: The bridges and messengers need legibility into whether or not the chain is using custom gas token or not. Previously, they only had a subset of the `IGasToken` interface which gave them just enough info to operate, but it appeared like people wanted the full interface everywhere

specs/experimental/custom-gas-token.md
tynes: This is covered in the `IGasToken` section so i dont think this is necessary

specs/experimental/custom-gas-token.md
tynes: it could undercollatorize the L2 if it rebases down

PR #117, opened 2024-04-08
Specify generic DA commitment

trianglesphere:
**Description**

This specifies a generic DA server commitment. This pushes the logic of decoding it to the DA server. This type is opaque & not inspecting inside the op-node.

Fixes: https://github.com/ethereum-optimism/client-pod/issues/762

### DIFF
--- a/specs/experimental/plasma.md
+++ b/specs/experimental/plasma.md
@@ -88,9 +88,10 @@ on the `commitment_type_byte` where [0, 128) are reserved for official implement
 The `da-service` commitment is as follows: `da_layer_byte ++ len(payload) as uvarint ++ payload`. The DA layer byte
 must be initially restricted to the range `[0, 127)`. This specification will not apportion DA layer bytes, but
 different DA layers should coordinate to ensure that the DA layer bytes do not conflict. The payload length is a
-unsigned 64 bit number, as defined in [protobuf spec]. The payload is a bytestring which is up to the DA layer to
-specify. The DA server should be able to parse the payload, find the data on the DA server, & verify that the data
-returned from the DA server matches what was committed to in the payload.
+unsigned 64 bit number, as defined in [protobuf spec]. We include the payload length so the op-node is able to pull
+out single commitment from a list of commitments that are concatenated together. The payload is a bytestring which
+is up to the DA layer to specify. The DA server should be able to parse the payload, find the data on the DA server,
+& verify that the data returned from the DA server matches what was committed to in the payload.
 
 [protobuf spec]: https://protobuf.dev/programming-guides/encoding/#varints
 
--- a/specs/experimental/plasma.md
+++ b/specs/experimental/plasma.md
@@ -85,15 +85,14 @@ on the `commitment_type_byte` where [0, 128) are reserved for official implement
 | 0                 | `keccak256(tx_payload)`         |
 | 1                 | `da-service`                     |
 
-The `da-service` commitment is as follows: `da_layer_byte ++ len(payload) as uvarint ++ payload`. The DA layer byte
-must be initially restricted to the range `[0, 127)`. This specification will not apportion DA layer bytes, but
-different DA layers should coordinate to ensure that the DA layer bytes do not conflict. The payload length is a
-unsigned 64 bit number, as defined in [protobuf spec]. We include the payload length so the op-node is able to pull
-out single commitment from a list of commitments that are concatenated together. The payload is a bytestring which
-is up to the DA layer to specify. The DA server should be able to parse the payload, find the data on the DA server,
-& verify that the data returned from the DA server matches what was committed to in the payload.
-
-[protobuf spec]: https://protobuf.dev/programming-guides/encoding/#varints
+The `da-service` commitment is as follows: `da_layer_byte ++ payload`.
+The DA layer byte must be initially restricted to the range `[0, 127)`.
+This specification will not apportion DA layer bytes, but different DA layers should coordinate to ensure that the
+DA layer bytes do not conflict. DA Layers can do so in
+[this discussion](https://github.com/ethereum-optimism/specs/discussions/135).
+The payload is a bytestring which is up to the DA layer to specify.
+The DA server should be able to parse the payload, find the data on the DA layer, and verify that the data returned
+from the DA layer matches what was committed to in the payload.
 
 The batcher SHOULD cap input payloads to the maximum L1 tx size or the input will be skipped
 during derivation. See [derivation section](#derivation) for more details.
--- a/specs/experimental/plasma.md
+++ b/specs/experimental/plasma.md
@@ -83,6 +83,16 @@ on the `commitment_type_byte` where [0, 128) are reserved for official implement
 | `commitment_type` | `commitment`                    |
 | ----------------- | ------------------------------- |
 | 0                 | `keccak256(tx_payload)`         |
+| 1                 | `da-server`                     |
+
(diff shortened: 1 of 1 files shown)


specs/experimental/plasma.md
```diff
@@ -83,6 +83,16 @@ on the `commitment_type_byte` where [0, 128) are reserved for official implement
 | `commitment_type` | `commitment`                    |
 | ----------------- | ------------------------------- |
 | 0                 | `keccak256(tx_payload)`         |
+| 1                 | `da-server`                     |
+
+The `da-server` commitment is as follows: `version_byte ++ len(payload) as uvarint ++ payload`. The version byte
```
tynes: I am confused when you say `version_byte` - there is a table above that defines `version_byte`. Can we call it `commitment_version_byte` to make it specific and different from the previous definition?

rac-sri: rather than `da-server`, having something like `da-service` makes things more clear ? Server sounds a bit like a centralised storage place, which more DA layers are not exclusively.

rac-sri: There should also be a flag introduced alongside `usePlasma` or `usePlasma` should be extended to take which commitment type the system should follow. Based on that the Plasma module is initialised, so that there are no issues in op-node integration with the module.

tynes: Should the configuration live in the DA server? It seems like we need some rethinking of how the configuration should happen

trianglesphere: I'm very open to renaming it from `da-server` to `da-service` or something else. The DA server does need to be able to parse the commitment & needs to handle that configuration

trianglesphere: Will rename to try to clarify. We have lots of version bytes so will specify the purpose of each one

rac-sri: I think we should, setup of DA commitment should be encapsulated in `da-server` ( or `da-service` ) .  In favour of that, since different DA solutions can have different parsing logic too.

rac-sri: Just to expand on that, different da will have different payload structure, and ideally the 'GetInput' and 'Set Input' in plasma should be capable of of calling them by having an initialisation step for the different payload structures. That should be the ideal design imo, and currently its easy todo, by adding checks during new plasma type creation simply where the returned type adheres to the 'DAStorage' interface .

Something similar to this ( WIP ) - https://github.com/availproject/optimism-plasma-avail/blob/develop/op-plasma/cli.go#L71 . No additional server required in this way ( how to pass which DA it is ? -  in here i introduced `UseAvailDA` , but we should look at better design patterns here )

Let me know your thoughts!

tuxcanfly: >Should the configuration live in the DA server? It seems like we need some rethinking of how the configuration should happen

It would be ideal if `plasma pkg` implemented a plugin architecture. Each service can provide it's custom flags for configuration which would then be used to initialise the service.

>( how to pass which DA it is ? - in here i introduced UseAvailDA , but we should look at better design patterns here )

We can add a plasma flag that specifies which plugin to use which would then initialise the plugin with its custom flags and use it for the da service.

We can look at popular ORM implementations for inspiration as this sounds like a very similar problem.

specs/experimental/plasma.md
```diff
@@ -83,6 +83,16 @@ on the `commitment_type_byte` where [0, 128) are reserved for official implement
...
 | ----------------- | ------------------------------- |
 | 0                 | `keccak256(tx_payload)`         |
+| 1                 | `da-service`                     |
+
+The `da-service` commitment is as follows: `da_layer_byte ++ len(payload) as uvarint ++ payload`. The DA layer byte
+must be initially restricted to the range `[0, 127)`. This specification will not apportion DA layer bytes, but
+different DA layers should coordinate to ensure that the DA layer bytes do not conflict. The payload length is a
+unsigned 64 bit number, as defined in [protobuf spec]. The payload is a bytestring which is up to the DA layer to
+specify. The DA server should be able to parse the payload, find the data on the DA server, & verify that the data
+returned from the DA server matches what was committed to in the payload.
```
tuxcanfly: Would this handled similarly in the `DataAvailabilityChallenge` contract? It would be ideal if the challenge resolution can be left to the implementation as well, because we can have compact proofs instead of having to submit the entire blob.

rach-id: 👍 
It would be nice since there are constructions where proving data unavailability can be done without sending the actual blob. Also, proving invalid state transitions can be reduced to providing a single share. 
So, it would be good if the resolution is implementation specific

specs/experimental/plasma.md
rac-sri: is there any decision made on this ?

specs/experimental/plasma.md
trianglesphere: We do want to let the DA commitment resolution be handled by the contract, but this is focusing on just the commitment <-> da server portion right now.

specs/experimental/plasma.md
trianglesphere: I think there is some confusion on the location of the DA server. The da-server uses the interface `get/set`. The exact interface may need to change a little bit as we pull in different layers, but op-node should only ever need the DA server URL

specs/experimental/plasma.md
tuxcanfly: Got it, thanks.

specs/experimental/plasma.md
```diff
@@ -83,6 +83,17 @@ on the `commitment_type_byte` where [0, 128) are reserved for official implement
 | `commitment_type` | `commitment`                    |
 | ----------------- | ------------------------------- |
 | 0                 | `keccak256(tx_payload)`         |
+| 1                 | `da-service`                     |
+
+The `da-service` commitment is as follows: `da_layer_byte ++ len(payload) as uvarint ++ payload`. The DA layer byte
+must be initially restricted to the range `[0, 127)`. This specification will not apportion DA layer bytes, but
+different DA layers should coordinate to ensure that the DA layer bytes do not conflict. The payload length is a
+unsigned 64 bit number, as defined in [protobuf spec]. We include the payload length so the op-node is able to pull
+out single commitment from a list of commitments that are concatenated together. The payload is a bytestring which
```
axelKingsley: Why are there multiple concatenated commits in the first place?

trianglesphere: That doesn't exist yet, but if we wanted to submit multiple commitments in the same L1 transaction, we would be able to. We would have to define a new L1 txn format, so could remove this data length byte

specs/experimental/plasma.md
```diff
@@ -83,6 +83,17 @@ on the `commitment_type_byte` where [0, 128) are reserved for official implement
 | `commitment_type` | `commitment`                    |
 | ----------------- | ------------------------------- |
 | 0                 | `keccak256(tx_payload)`         |
+| 1                 | `da-service`                     |
+
+The `da-service` commitment is as follows: `da_layer_byte ++ len(payload) as uvarint ++ payload`. The DA layer byte
+must be initially restricted to the range `[0, 127)`. This specification will not apportion DA layer bytes, but
+different DA layers should coordinate to ensure that the DA layer bytes do not conflict. The payload length is a
```
tynes: How will DA layers be able to coordinate such that they do not conflict? I lean towards either having a table in this doc or linking someplace where it can be discussed, ie a github issue in this repo

trianglesphere: I added a link to a github discussion thread

tynes (approved):

PR #120, opened 2024-04-10
Add SystemConfig values to Configurability Definitions

blmalone:
As per slack message: https://oplabs-pbc.slack.com/archives/C01HLRH0LKZ/p1712783478152239?thread_ts=1712781208.269489&cid=C01HLRH0LKZ 

Do you think that we need to include other System config variables? https://specs.optimism.io/protocol/system_config.html


ben-chain:
> Do you think that we need to include other System config variables?

Good call, yes I think all should be included. We should aim for this to be exhaustive. Will wait on that to approve.  If we don't want to duplicate info, linking out in the style of the predeploys is also fine, but we should recover all detail in the standard bounds PR, same as I've requested for predeploys there.

ben-chain (approved):

PR #123, opened 2024-04-11
draft(fps): Move LPPs to blobs

clabby:
## Overview

Begins to rework the "Large Preimage Proposals" spec to describe a system that accepts preimage data in blobspace, rather than making the data available in block bodies. [EIP-3074](https://eips.ethereum.org/EIPS/eip-3074) was considered for inclusion in the upcoming Prague hardfork, which breaks the large preimage proposals route's assumption on the intermediate SHA3 state commitments always being available within block bodies.

### How does EIP-3074 break large preimage proposals?

[EIP-3074](https://eips.ethereum.org/EIPS/eip-3074) introduces `AUTH` and `AUTHCALL`, allowing for EOAs to be the `msg.sender` of callframes beyond a depth of 0. The off-chain challenge agent currently has assumptions of the input data to `addLeavesLPP` always being available in block bodies, and as such we have a check that asserts `tx.origin == msg.sender`. This can no longer be trusted, since an EOA can submit an `AUTHCALL` to the `PreimageOracle`'s `addLeavesLPP` function for their own wallet in a lower callframe.

In order to correctly monitor and potentially challenge these proposals, the off-chain challenge agent must be able to reconstruct the full merkle tree for all large preimage proposals off-chain. If this data is possibly available within memory during execution, which will be possible after EIP-3074's inclusion, this task becomes more difficult for the challenge agent to handle.

### Why blobs?

The [KZG Point Evaluation precompile](https://eips.ethereum.org/EIPS/eip-4844#point-evaluation-precompile) allows us to create an "opening" (read: for simpletons like myself, it's just verifying an inclusion proof) on one or many field elements within a blob. This effectively behaves like our old incremental merkle tree design, with the added benefit of us not having to perform the merkleization within the contract. Instead, the verifier is a precompile, and the "merkleization" process is handled by the protocol.

When challenging, the off-chain challenge agent can supply all necessary field elements, the points at which they exist within their respective blobs, and the index(es) within the tracking array in the `PreimageOracle` that stored all versioned hashes, in ascending order, as they were posted. After the data is proven to be included at the correct location, the keccak dispute stays the same (absorb post state's input data into prestate, permute, check resulting post state against the claim). Blob data is guaranteed to be available for up to 18 days, in a location easily accessible to the off-chain challenge agent. In addition, large preimage disputes resolve in a bounded time, typically 1-2 days. Blobs are available for 18 days, so data availability is guaranteed.

### Open Questions

* How do we determine true preimage length? Previously, we padded the user's input data on-chain, which allowed us to determine the true length of the preimage. We will not be able to do this anymore and will require the field elements to include the padding, unless we get a bit clever with the intermediate state format.
    * Possibly track the final block absorbed & make a special case where we pad it, omitting padding from the data available in the blob?
* Do we want to batch point evaluations together? How complex is it to handle this?
    * Right now, it seems easier to just prove one field element's inclusion at a time.


specs/experimental/fault-proof/stage-one/fault-dispute-game.md
```diff
@@ -314,38 +314,49 @@ Large preimage proposals allow for submitters to stream in a large preimage over
...
-
-    leaf = keccak256(abi.encodePacked(input, blockIndex, stateCommitment));
-}
-```
+**Large Preimage Proposals**
+
+Large preimage proposals allow for submitters to stream in a large preimage over multiple transactions, along-side
+commitments to the intermediate state of the `keccak256` function after absorbing/permuting the $1088$ bit block.
+This data is available to off-chain challenge agents in blobs, and each blob contains up to $738$ intermediate SHA3
+commitments. The data is laid out as follows:
```
ajsutton, line 322: I'm a little unclear exactly how this layout maps to the field elements in blobs. Ideally we'd use the same encoding as for derivation since we already have code to support that (https://specs.optimism.io/protocol/derivation.html#blob-encoding) and then once the data is converted from field elements to its continuous byte-string I assume we then read 176 byte chunks of it and extract the keccak block and post-state commitment as per the diagram.

ajsutton: Also, to reassemble the data correctly, the challenger needs to know the sender and uuid the data is related to.  Previously this could be retrieved from the tx signer and uuid in the call data but it will now have to be either recorded by the contract or included in the blob data (and then verified by the contract).

The challenger also needs to find the _beacon_ block the blob is from, either by beacon block root or slot.  The execution block timestamp can be converted to beacon slot (so could load the execution block header by block number to get the timestamp).

The challenger will also need a way to determine which transaction in the block actually posted the content - we need to avoid the challenger being fooled by a tx with valid preimage data that didn't ever call the actual preimage oracle.

ajsutton: Ah, block number/hash/timestamp and versioned kzg commitment the data was read from would be enough to identify blobs that were actually read - we'd just have to load all blobs from the block and we can then filter by commitment.

Still need a way to identify the sender and uuid the blob data was for though.

Inphi: We can maintain a mapping of blobhashes to uuids. This can be updated in the contract whenever a state commitment in the blob is added. The `BLOBHASH` opcode can be used to retrieve every single hash included in the tx.

ajsutton: In looking at it further I came around to the idea of just adjusting `mapping(address => mapping(uint256 => uint64[])) public proposalBlocks;` which currently maps `sender -> uuid -> block_number` to map to a `block_number, blob_hash` tuple.  We can find all the entries in that mapping by iterating the `proposals` array so it should be a pretty minimal change and gives us all the required info.

tynes:
This proposal has been replaced by using logs instead? We should be able to close this @clabby ?

clabby:
Yep! Closing.

PR #124, opened 2024-04-11
fix: batch inbox address is not dynamic right now.

blmalone:
We don't want to be implying that some fields are dynamic when they currently aren't. 

e.g. Batch Inbox Address. 

All other parameters that are static should be labeled as such: 
- Chain ID
- Genesis State
- Sequencing Window
- Superchain Target

Internal slack [thread](https://oplabs-pbc.slack.com/archives/C01HLRH0LKZ/p1712788865979079).


ben-chain:
Can we just put "Static" for all the administrator-less values?

blmalone:
@ben-chain https://github.com/ethereum-optimism/specs/pull/124/commits/cf24e58b4ab88f1f496d19f0cfa659f173d8f531 

Is this what you mean?

ben-chain (approved):
Yes!

PR #136, opened 2024-04-17
DA Server Spec

trianglesphere:
**Description**

This expands the description of the DA server & makes some modifications to the plasma spec to better work with different DA layers.

### DIFF
--- a/specs/experimental/plasma.md
+++ b/specs/experimental/plasma.md
@@ -88,9 +88,10 @@ on the `commitment_type_byte` where [0, 128) are reserved for official implement
 The `da-service` commitment is as follows: `da_layer_byte ++ len(payload) as uvarint ++ payload`. The DA layer byte
 must be initially restricted to the range `[0, 127)`. This specification will not apportion DA layer bytes, but
 different DA layers should coordinate to ensure that the DA layer bytes do not conflict. The payload length is a
-unsigned 64 bit number, as defined in [protobuf spec]. The payload is a bytestring which is up to the DA layer to
-specify. The DA server should be able to parse the payload, find the data on the DA server, & verify that the data
-returned from the DA server matches what was committed to in the payload.
+unsigned 64 bit number, as defined in [protobuf spec]. We include the payload length so the op-node is able to pull
+out single commitment from a list of commitments that are concatenated together. The payload is a bytestring which
+is up to the DA layer to specify. The DA server should be able to parse the payload, find the data on the DA server,
+& verify that the data returned from the DA server matches what was committed to in the payload.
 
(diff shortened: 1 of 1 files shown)


specs/experimental/plasma.md
```diff
@@ -32,12 +32,13 @@ chain derivation must be reset, omitting the input data when rederiving therefor
...
 
-Input data is uploaded to the storage layer via plain HTTP calls to the DA storage service.
-This service is horizontally scalable and concerned with replicating the data across preferred storage layers
-such as IPFS or any S3 compatible storage. Input data is content addressed by its hash in the request url
-so responses are easily cachable.
+Input data is uploaded to the storage layer via plain HTTP calls to the DA server.
+
+This service is responsible to interacting with the Data Availability Layer (DA layer).
+The layer could be a content addressable storage layer like IPFS or any S3 compatible storage.
+These layers use `commitment_type 0` and can pre-compute the commitment prior to posting the data.
```
tynes: We haven't yet defined the commitment types if reading from top of doc

specs/experimental/plasma.md
```diff
@@ -51,6 +52,18 @@ Any DA provider can implement the following endpoints to receive and serve input
     200 OK
   ```
```
tynes: Can you elaborate on why there are 2 ways to do it, if both most be implemented, why one would be implemented over the other

specs/experimental/plasma.md
tynes: Perhaps worth breaking out sub headers for the different commitment types and have specific information under each sub header

specs/experimental/plasma.md
trianglesphere: Will do. The first way is backwards compat with S3 DA server, the second is because with blockchain like DA, you can't know the commitment ahead of time

specs/experimental/plasma.md
trianglesphere: Would linking forwards also work here? It may also make sense to reflow the entire docs to first talk about commitments, then the DA server, then derivation b/c each builds on the last

tynes: I think your idea about changing the order makes the most sense

tynes: Up to you if you want to do that as part of this PR

specs/experimental/plasma.md
tynes: Thank you!

specs/experimental/plasma.md
tynes: Ah i see it was just a small diff

tynes (approved):

PR #139, opened 2024-04-19
feat/custom gas token: diego's changes

0xfuturistic:


specs/experimental/custom-gas-token.md
```diff
@@ -209,16 +225,18 @@ The following methods MUST revert when custom gas token is being used:
...
 - `finalizeBridgeETH(address, address, uint256, bytes)`
 
-The following legacy methods MUST also revert when custom gas token is being used:
+The following legacy methods in `L1StandardBridge` MUST revert when custom gas token is being used:
+
+- `depositETH(uint32,bytes)`
+- `depositETHTo(address,uint32,bytes)`
+- `finalizeETHWithdrawal(address,address,uint256,bytes)`
+
+The following legacy methods in `L2StandardBridge` MUST also revert when custom gas token is being used and the
```
tynes, line 248: nit: small typo, `nonzero`

specs/experimental/custom-gas-token.md
```diff
@@ -141,6 +141,22 @@ flowchart LR
 
 ## Contract Modifications
 
+### Interface
```
tynes: I think @mds1 suggested that we call this interface `IGasToken`, we could make this `## IGasToken Interface`

and add a little description about the interface, like why it exists (encapsulates shared interface)

We should expand on the arguments + return types of the interface, it could be good just to do a solidity block and define the interface itself. We want to describe what each function in the interface does

specs/experimental/custom-gas-token.md
```diff
@@ -141,6 +141,22 @@ flowchart LR
...
+
+The following methods make up the gas paying token interface:
+
+- `gasPayingToken`
+- `gasPayingTokenName`
+- `gasPayingTokenSymbol`
+- `isCustomGasToken`
+
+This interface is on
```
tynes: nit: trailing `:`

specs/experimental/custom-gas-token.md
```diff
@@ -234,6 +252,8 @@ The following methods MUST revert when `CALLVALUE` is non zero:
 - `relayMessage(uint256,address,address,uint256,uint256,bytes)`
```
tynes: We should remove `relayMessage` from this list and instead follow up saying that it should be impossible to call with non zero callvalue with custom gas token due to the enforcement in `sendMessage`, framing it as an implicit invariant is better than explicit since we don't explicitly enforce it like @mds1 suggested

specs/experimental/custom-gas-token.md
0xfuturistic: should we add `OptimismPortal` to the contracts that use the interface? it uses `gasPayingToken()` but no the other methods. still, it might be useful to just have this interface for all contracts affected by the custom gas token

0xfuturistic: (otherwise, I just implemented the changes)

tynes (approved):
Great changes!

PR #143, opened 2024-04-22
fix: Inconsistent Chain ID Format

refcell:
**Description**

Small fix to consistently use `chain id` colloquially rather than variable or constant expressions.


sebastianst (approved):

tynes (approved):

tynes:
This also needs an update for `ci-builder` that is used in ci to get the one with pnpm@9

PR #146, opened 2024-04-23
fjord: Add constant maximum sequencer drift spec

sebastianst:
**Description**

Adds spec to make `max_sequencer_drift` a constant in Fjord.

**Metadata**

- Fixes https://github.com/ethereum-optimism/protocol-quest/issues/232

### DIFF
--- a/specs/fjord/derivation.md
+++ b/specs/fjord/derivation.md
@@ -1,20 +1,50 @@
-# L2 Chain Derivation
+# Fjord L2 Chain Derivation Changes
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 **Table of Contents**
 
-- [Architecture](#architecture)
-  - [L2 Chain Derivation Pipeline](#l2-chain-derivation-pipeline)
-    - [Batch Queue](#batch-queue)
+- [Protocol Parameter Changes](#protocol-parameter-changes)
+  - [Constant Maximum Sequencer Drift](#constant-maximum-sequencer-drift)
+    - [Rationale](#rationale)
+    - [Security Considerations](#security-considerations)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-# Architecture
+# Protocol Parameter Changes
 
-## L2 Chain Derivation Pipeline
+The following table gives an overview of the changes in parameters.
 
-### Batch Queue
+| Parameter | Pre-Fjord (default) value | Fjord value | Notes |
+| --------- | ------------------------- | ----------- | ----- |
+| `max_sequencer_drift` | 600 | 1800 | Was a protocol parameter since Bedrock. Now becomes a constant. |
 
-With Fjord, the `max_sequencer_drift` parameter becomes a constant of value
-`1800` _seconds_, translating to a fixed maximum sequencer drift of 30 minutes.
+## Constant Maximum Sequencer Drift
+
+With Fjord, the `max_sequencer_drift` parameter becomes a constant of value `1800` _seconds_,
+translating to a fixed maximum sequencer drift of 30 minutes.
+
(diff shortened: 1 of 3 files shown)


specs/fjord/derivation.md
```diff
@@ -0,0 +1,20 @@
...
+    - [Batch Queue](#batch-queue)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Architecture
+
+## L2 Chain Derivation Pipeline
+
+### Batch Queue
+
```
tynes, line 21: This seems like it would be perfect for a constants table. It would also be amazing to include information like explicitly saying it was configurable before, tradeoffs of smaller/larger values and why this value was chosen

sebastianst: @tynes I wasn't sure how much background info we actually want to give in the spec, so I rather started short to only state the most relevant information to implementers of what actually changes. But I'm happy to add this information if we want to give more context for the change.

Regarding a `constants` table, we don't have that yet in the main spec (`protocol/` folder). Where would you add this table? If I add it here, it would be a table with only a single entry.

Having said that, we might want to tweak two further constants in Fjord, namely `MaxChannelBankSize` and `MaxRLPBytesPerChannel`. @trianglesphere has requested to change these as well.

tynes: This highlights the problem of defining what a good spec is and scaling this information. I personally lean towards including rationale in the spec so that we can scale the knowledge, which is why i always put "security considerations" sections in specs that i work on

tynes: We don't yet have a constants table but its something we have been slowly moving towards :) curious how you feel about it, I think a table with 1 value is fine for now so it can be added to in the future

sebastianst: @tynes I agree, it's a good idea to but more context into the specs. It makes them easier to understand! Added two sections on rationale and security considerations.

I also added tables to the Fjord derivation spec, as well as added one to the main derivation spec. I believe it's nice for implementers of the spec to have a table of protocol parameters and constants that are affected by protocol upgrades at least.

specs/protocol/derivation.md
```diff
@@ -188,6 +189,15 @@ first block of the epoch appears in the very last L1 block of the window. Note t
...
 without deriving blocks from them.
 
+## Protocol Parameters and Constants
+
+The following table gives an overview of some protocol parameters and constants, and how they are
+affected by protocol upgrades.
+
+| Parameter/Constant | Bedrock (default) value | Latest (default value) | Changes |
+| ------------------ | ----------------------- | ---------------------- | ------- |
+| `max_sequencer_drift` | 600 | 1800 | [Fjord](../fjord/derivation.md#constant-maximum-sequencer-drift) |
```
trianglesphere: We have a couple more hardcoded parameters in https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/params.go
Max Seq Drift actually comes from the [rollup config](https://github.com/ethereum-optimism/optimism/blob/66a21a9a17972369cf964ff603db0ef0837dcdd4/op-node/rollup/types.go#L53-L125)

sebastianst: @trianglesphere Yeah I meant to differentiate between changeable protocol _parameters_ and unchangeable _constants_, to collect both in that table because, like in the case of the seq drift, we change between parameters and constants. Do we use the word _parameter_ for something unchangeable?

I could change the first column to just read `Parameter` and then have a `Notes` column that explains whether this is a const, or configurable per rollup config, or per L1 SystemConfig, ...

specs/protocol/derivation.md
```diff
@@ -188,6 +189,15 @@ first block of the epoch appears in the very last L1 block of the window. Note t
 applies to _block_ derivation. Sequencer batches can still be derived and tentatively queued
 without deriving blocks from them.
 
+## Protocol Parameters
```
tynes, line 192: I think this section is nice, we will have to remember to update it as more things are added

tynes (approved):

trianglesphere (approved):

PR #147, opened 2024-04-23
fix: Design Goal Verbiage

refcell:
**Description**

Cleans up the top level spec `Design Goals` wording to be more inclusive to the OP Stack and less Optimism-specific.


tynes (approved):

tynes:
Great refactor :D

PR #149, opened 2024-04-24
specs: interop 3074 updates

tynes:
**Description**

Updates the interop specs for 3074. This removes the "only eoa" invariant
and redefines an executing message to be an event emitted by the `CrossL2Inbox`

### DIFF
--- a/specs/interop/messaging.md
+++ b/specs/interop/messaging.md
@@ -112,6 +112,9 @@ Both the block builder and the verifier use this information to ensure that all 
 The executing message is verified by checking if there is an existing initiating-message
 that originates at [`Identifier`] with matching [Message Payload](#message-payload).
 
+Since an executing message is defined by a log, it means that reverting calls to the `CrossL2Inbox`
+do not count as executing messages.
+
 ## Messaging Invariants
 
 - [Timestamp Invariant](#timestamp-invariant): The timestamp at the time of inclusion of the executing message MUST
--- a/specs/interop/messaging.md
+++ b/specs/interop/messaging.md
@@ -13,7 +13,6 @@
 - [Messaging Invariants](#messaging-invariants)
   - [Timestamp Invariant](#timestamp-invariant)
   - [ChainID Invariant](#chainid-invariant)
-  - [Only EOA Invariant](#only-eoa-invariant)
   - [Message Expiry Invariant](#message-expiry-invariant)
 - [Message Graph](#message-graph)
   - [Invalid messages](#invalid-messages)
@@ -89,7 +88,7 @@ exact state of the block templates between multiple chains together.
 
 [log]: https://github.com/ethereum/go-ethereum/blob/5c67066a050e3924e1c663317fd8051bc8d34f43/core/types/log.go#L29
 
-Each Log (also known as `event` in solidity) forms an initiating message.
+Each [Log][log] (also known as `event` in solidity) forms an initiating message.
 The raw log data froms the [Message Payload](#message-payload).
 
 Messages are *broadcast*: the protocol does not enshrine address-targeting within messages.
@@ -101,8 +100,12 @@ An initiating message may be executed many times: no replay-protection is enshri
 
 ### Executing Messages
 
-All the information required to satisfy the invariants MUST be included in the calldata
-of the function that is used to execute messages.
+An executing message is represented by the [ExecutingMessage event][event] that is emitted by
+the `CrossL2Inbox` predeploy. This event is coupled to a `CALL` with the payload that is emitted
+within the event, allowing introspection of the data without needing to do call tracing.
+All of the information required to satisfy the invariants MUST be included in this event.
+
+[event]: ./predeploys.md#executingmessage-event
 
 Both the block builder and the verifier use this information to ensure that all system invariants are held.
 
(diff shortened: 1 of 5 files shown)


specs/interop/predeploys.md
```diff
@@ -214,15 +232,15 @@ function relayMessage(uint256 _destination, uint256 _source, uint256 _nonce, add
...
 
+    sentMessages[messageHash] = true;
+
     bool success = SafeCall.call({
        _target: _target,
        _value: msg.value,
        _calldata: _message
     });
 
     require(success);
```
0xfuturistic, line 243: should the call revert or split into an if clause based on the success of the call and emit two different events? 
something like this:
```
        bool success = _callWithAllGas(_target, _message);

        if (success) {
            successfulMessages[messageHash] = true;
            emit RelayedMessage(messageHash);
        } else {
            emit FailedRelayedMessage(messageHash);
        }
```
it's what we have in the code right now, so want to double check if we should stick to this or change it to match this part of the spec

0xfuturistic, line 245: should the call revert if the target call fails or just emit a different event? 
something like this:
```
        bool success = _callWithAllGas(_target, _message);

        if (success) {
            successfulMessages[messageHash] = true;
            emit RelayedMessage(messageHash);
        } else {
            emit FailedRelayedMessage(messageHash);
        }
```
it's what we have in the code right now, so want to double check if we should stick to this or change it to match this part of the spec (reverting instead of emitting events)

tynes: good question, i think that the actual implementation is better, this is meant to be a naive pseudocode kind of implementation, so no need to modify anything

specs/interop/predeploys.md
```diff
@@ -214,15 +232,15 @@ function relayMessage(uint256 _destination, uint256 _source, uint256 _nonce, add
       tstore(CROSS_DOMAIN_MESSAGE_SOURCE_SLOT, _source)
     }
 
+    sentMessages[messageHash] = true;
```
0xfuturistic, line 237: should we keep this before the call? it's not stored in transient storage, so we don't use reentrancy protection

specs/interop/messaging.md
```diff
@@ -260,7 +246,7 @@ The graph is bounded in 4 ways:
   as per the [ChainID invariant](#chainid-invariant).
 - Every block cannot depend on future blocks, as per the [Timestamp invariant](#timestamp-invariant).
 - Every block has a maximum gas limit, an intrinsic cost per transaction,
-  and thus a maximum inward degree of dependencies, as per the [Only-EOA invariant](#only-eoa-invariant)
+  and thus a maximum inward degree of dependencies
```
mds1: nit, other bullets end with a period
```suggestion
  and thus a maximum inward degree of dependencies.
```

specs/interop/overview.md
```diff
@@ -13,15 +13,15 @@ network upgrade will be included in this document in the future.
 | Source Chain        | A blockchain that includes an initiating message                                                    |
 | Destination Chain   | A blockchain that includes an executing message                                                     |
 | Initiating Message  | An event emitted from a source chain                                                                |
-| Executing Message   | A transaction submitted to a destination chain that corresponds to an initiating message            |
+| Executing Message   | An event emitted from a destination chain that includes an initiating message                       |
```
mds1: this feels a bit more precise
```suggestion
| Executing Message   | An event emitted from a destination chain's `CrossL2Inbox` that includes an initiating message                       |
```

specs/interop/sequencer.md
```diff
@@ -30,11 +30,17 @@ without adding additional state-transition complexity or cross-chain synchronici
...
+executing messages.
 
 ### Static analysis
 
-The block builder SHOULD use static analysis on executing messages to determine the dependency of the message.
+Note that static analysis is not always reliable, but it is far faster than having to perform
+execution to get the data required to validate an executing message.
+
+The block builder SHOULD use static analysis when possible on executing messages to determine
+the dependency of the message.
```
mds1: When a transaction has a top level `to` address of the `CrossL2Inbox`, the call that it executes could result in reentering the `CrossL2Inbox` to relay a second executing message, and you cannot statically know this will occur. So you can use static analysis to identify invalid transactions and discard them to avoid executing those, but must execute them to confirm validity. Is that correct?

protolambda:
Edge-cases I think we should clarify in the spec, now that executing messages are also log events:

1) Events of executing messages can be consumed as initiating messages
2) Executing messages can be self-referential to their own event as initiating message
3) There can be loops between L2 chains with these self-referential messages, since there is no total ordering between L2s

And we may want message-hashing to have a version-byte and distinguish messages that are executing others, so we can detect these edge-cases without loading the full message. It'll also be more forward-compatible to possible interop changes or extensions.

specs/interop/sequencer.md
tynes: That is actually a good point, this leads me to believe that static analysis can never work. Thank you for pointing this out

tynes:
> And we may want message-hashing to have a version-byte and distinguish messages that are executing others, so we can detect these edge-cases without loading the full message. It'll also be more forward-compatible to possible interop changes or extensions.

Per https://github.com/ethereum-optimism/design-docs/pull/9, the versioned hash is an offchain optimization. We could specify its usage to make client diversity of the superchain backend easier, but its not part of the core protocol

specs/interop/predeploys.md
tynes: This is a simple example and doesn't necessarily need to be the full implementation but instead is meant to illustrate how the system works. Perhaps making it be python would make that more clear.

This follows the [checks, effects, interactions pattern](https://docs.soliditylang.org/en/v0.6.11/security-considerations.html#use-the-checks-effects-interactions-pattern) which is generally considered best practice but has been supplemented with [FREI-PI](https://www.nascent.xyz/idea/youre-writing-require-statements-wrong)

PR #150, opened 2024-04-24
EL Chain ID

trianglesphere:
**Description**

Specify that the EL also has a key.


clabby:
cc @emhane

specs/protocol/rollup-node-p2p.md
```diff
@@ -111,6 +111,19 @@ The discovery process in Optimism is a pipeline of node records:
...
+The Ethereum Node Record (ENR) for an Optimism execution node must contain the following values, identified by unique keys:
+
+- An OpStack (`opel` field) L2 network identifier.
+
+The EL uses a different key from the CL in order to stop EL and CL nodes from connecting to each other.
+
+The `opel` value is encoded as a single RLP `bytes` value, the concatenation of:
+
+- chain ID (`unsigned varint`)
+- fork ID (`unsigned varint`)
```
emhane: do you mind using the EL standard? https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2124.md
in reth we have this function for all chain specs https://github.com/paradigmxyz/reth/blob/652be135c795e7a8518078de31f8b51e219f331e/crates/primitives/src/chain/spec.rs#L769-L774

sebastianst: Sounds like a good idea! I'm wondering what fork ID we use for the CL case right now? It's not further specified in section "Consesus Layer Structure" above.

emhane: @trianglesphere it is only the chain id being used right and fork ID is set to zero?

trianglesphere: As implemented, fork ID/version is zero for `opstack`: https://github.com/ethereum-optimism/optimism/blob/ba174f4d5f4020ff16298fefd86b55a29d4724a9/op-node/p2p/discovery.go#L79-L82

We can consider setting version to non-zero with Fjord

emhane: @sebastianst @trianglesphere discv5 uses a list of fork IDs for the `"eth"` field. doing the same thing for `"opel"` would make sense. cc @mattsse <https://github.com/paradigmxyz/reth/pull/7900>. from specs <https://github.com/ethereum/devp2p/blob/master/enr-entries/eth.md#entry-format>

tynes:
@protolambda What do you think about this?

protolambda (commented):
LGTM, but we should rename the doc to `p2p.md` or separate the execution-layer part out so it's in `exec-engine.md` instead of `rollup-node-p2p.md`

emhane:
we went ahead and optimistically implemented this. so, every op-reth node is being built with `'opel': [eip-2124-fork-id]` kv-pair in its ENR

specs/protocol/exec-engine.md
```diff
@@ -483,3 +484,12 @@ For the Ecotone upgrade, this entails that:
 [engine_getPayloadV3]: https://github.com/ethereum/execution-apis/blob/a0d03086564ab1838b462befbc083f873dcf0c0f/src/engine/cancun.md#engine_getpayloadv3
 [HEX value encoding]: https://eth.wiki/json-rpc/API#hex-value-encoding
 [JSON-RPC-API]: https://github.com/ethereum/execution-apis
+
+## P2P Modifications
+
+The Ethereum Node Record (ENR) for an Optimism execution node must contain an `opel` key-value pair where the key is
+`opel` and the value is a [EIP-2124](https://eips.ethereum.org/EIPS/eip-2124) fork id.
+
+TODO: Do we actually want to just use chainID + use the native EIP-2124 checking to differentiate based on the chain?
```
tynes: @trianglesphere It seems like this was already implemented in reth, is this todo still relevant?

trianglesphere: can just skip it.

specs/protocol/exec-engine.md
```diff
@@ -483,3 +484,12 @@ For the Ecotone upgrade, this entails that:
...
+
```
trianglesphere:
```suggestion

```

tynes (approved):

ajsutton (approved):

tynes (approved):

tynes:
We need an approval from @sebastianst or @axelKingsley

sebastianst (approved):

PR #163, opened 2024-04-30
Add DGM security properties

maurelian:
Adds a section on security properties to the Deputy Guardian Module specs. This already exists for the Liveness extensions.

### DIFF
--- a/specs/experimental/security-council-safe.md
+++ b/specs/experimental/security-council-safe.md
@@ -24,8 +24,10 @@
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-The Security Council uses a specially extended Safe multisig contract to provide additional security
-guarantees on top of those provided by the Safe contract.
+The Security Council (at
+[eth:0xc2819DC788505Aac350142A7A707BF9D03E3Bd03](https://etherscan.io/address/0xc2819DC788505Aac350142A7A707BF9D03E3Bd03))
+uses a specially extended Safe multisig contract to provide additional security guarantees on top of
+those provided by the Safe contract.
 
 ## Deputy guardian module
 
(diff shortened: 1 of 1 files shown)


specs/experimental/security-council-safe.md
```diff
@@ -77,6 +79,18 @@ interface DeputyGuardianModule {
 For simplicity, the `DeputyGuardianModule` module does not have functions for updating the `safe` and
 `deputyGuardian` addresses. If necessary these can be modified by swapping out with a new module.
 
+### Deputy Guardian Module Security Properties
+
+The following security properties must be upheld by the `DeputyGuardianModule`:
+
+1. The module must correctly enforce access controls so that only the Deputy Guardian can call state
+   modifying functions.
```
mds1: Should we specify that these "state modifying functions" are all external to the Safe, and the deputy guardian should not be able to modify the state of the Safe itself or any of its extensions?

mds1: Or more narrow: DGM only allows the deputy guardian to perform some or all of the actions that the regular guardian can take. It must not enable the deputy guardian to perform actions the guardian cannot perform

maurelian: Interesting. My intent here was to just to indicate that all state changing functions on the DGM itself should have auth checks so that only the DeputyGuardian could call them. 

Based on your comment I've expanded this item into 3 which I think aligns with both of our intents, except the 'some or all' part, ie. I don't think we would want a future situation where there exist a Guardian authed function which the `DeputyGuardian` cannot also call.

```
1. The module must correctly enforce access controls so that only the Deputy Guardian can call state
   modifying functions on the `DeputyGuardianModule`.
1. The module must be able to cause the Safe to make calls to all of the functions which the
   Guardian role is authorized to make.
1. The module must not be able to cause the Safe to make calls to functions which the Guardian role
   is not authorized to make.
```

mds1 (approved):

PR #164, opened 2024-04-30
Update configurability.md

sbvegan:
**Description**

Updating the units to L1 and L2 blocks, which are the parameters used to configure these parameters. Also mentioning that chain operators can submit batches in smaller intervals.

**Tests**

Please describe any tests you've added. If you've added no tests, or left important behavior untested, please explain why not.

**Additional context**

Add any other context about the problem you're solving.

**Metadata**

- Fixes #[Link to Issue]


blmalone (approved):
LGTM

PR #145, opened 2024-04-22
fix: Modularize Superchain Upgrades

tynes:
**Description**

Modularizes superchain upgrades, splitting each upgrade into it's own section.

PR #168, opened 2024-05-01
Fjord: update Fjord spec for brotli compression feature

roberto-bayardo:
**Description**

Update spec for the Fjord upgrade's introduction of versioned channel_encodings & Brotli-based channel compression.

### DIFF
--- a/specs/fjord/derivation.md
+++ b/specs/fjord/derivation.md
@@ -8,6 +8,7 @@
   - [Constant Maximum Sequencer Drift](#constant-maximum-sequencer-drift)
     - [Rationale](#rationale)
     - [Security Considerations](#security-considerations)
+  - [Brotli Channel Compression](#brotli-channel-compression)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -48,3 +49,25 @@ be in the same epoch as the the last pre-Fjord blocks, even if these blocks woul
 have these L1-origin timestamps according to pre-Fjord rules. So the same L1 timestamp would be
 shared within a pre- and post-Fjord mixed epoch. This is considered a feature and is not considered
 a security issue.
+
+# Brotli Channel Compression
+
+[legacy-channel-format]: ../protocol/derivation.md#channel-format
+
+Fjord introduces a new versioned channel encoding format to support alternate compression
+algorithms, with the [legacy channel format][legacy-channel-format] remaining supported. The
+versioned format is as follows:
+
+```text
+channel_encoding = `channel_version_byte ++ compress(rlp_batches)`
+```
+
+The `channel_version_byte` must never have its 4 lower order bits set to `0b1000 = 8` or `0b1111 =
+15`, which are reserved for usage by the header byte of zlib encoded data (see page 5 of
+[RFC-1950][rfc1950]). This allows a channel decoder to determine if a channel encoding is legacy or
+versioned format by testing for these bit values. If the channel encoding is determined to be
+versioned format, the only valid `channel_version_byte` is `1`, which indicates `compress()` is the
+Brotli compression algorithm (as specified in [RFC-7932][rfc7932]) with no custom dictionary.
+
+[rfc7932]: https://datatracker.ietf.org/doc/html/rfc7932
+[rfc1950]: https://www.rfc-editor.org/rfc/rfc1950.html
--- a/specs/fjord/derivation.md
+++ b/specs/fjord/derivation.md
@@ -8,7 +8,7 @@
   - [Constant Maximum Sequencer Drift](#constant-maximum-sequencer-drift)
     - [Rationale](#rationale)
     - [Security Considerations](#security-considerations)
-  - [Brotli Channel Compression](#brotli-channel-compression)
+- [Brotli Channel Compression](#brotli-channel-compression)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
--- a/specs/fjord/overview.md
+++ b/specs/fjord/overview.md
@@ -17,6 +17,5 @@ This document is not finalized and should be considered experimental.
 
 ## Consensus Layer
 
-- [Constant maximum sequencer drift](./derivation.md#batch-queue)
-
-- [Channel encoding format](./derivation.md#channel-format)
+- [Constant maximum sequencer drift](./derivation.md#constant-maximum-sequencer-drift)
+- [Brotli channel compression](./derivation.md#brotli-channel-compression)
--- a/specs/fjord/overview.md
+++ b/specs/fjord/overview.md
@@ -18,3 +18,5 @@ This document is not finalized and should be considered experimental.
 ## Consensus Layer
 
 - [Constant maximum sequencer drift](./derivation.md#batch-queue)
+
+- [Channel encoding format](./derivation.md#channel-format)
--- a/specs/protocol/derivation.md
+++ b/specs/protocol/derivation.md
@@ -373,28 +373,34 @@ where:
 - `is_last` is a single byte with a value of 1 if the frame is the last in the channel, 0 if there are frames in the
   channel. Any other value makes the frame invalid (it must be ignored by the rollup node).
 
-[batcher-spec]: batching.md
-
 ### Channel Format
 
-A channel is encoded as `channel_encoding`, defined as:
+[channel-format]: #channel-format
+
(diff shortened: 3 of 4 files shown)


specs/protocol/derivation.md
```diff
@@ -363,29 +363,47 @@ where:
...
+```text
+channel_encoding = `version_byte ++ compress(rlp_batches)`
+```
+
+For the versioned format, the `version_byte` must never have its 4 lower order bits set to 0x1000,
+as those bits indicate the channel encoding is of legacy type by definition of the zlib encoding
+header.
+
+As of the Fjord upgrade, the only valid `version_byte` is `1`, which indicates `compress()` is the
+Brotli compression algorithm (as specified in [RFC-7932][rfc7932]) with no custom dictionary.
```
cody-wang-cb: Wondering if we should also update the `Channel Reader (Batch Decoding)` section to mention how we check for zlib vs brotli?

roberto-bayardo: Hmm, that section only seemed to reference the batch-format section which does seem incomplete.  I added a reference to the channel-encoding format as well. Should be clear from the way we note the 0x0001 low order bits can be used for disambiguation how one would determine zlib vs brotli?

cody-wang-cb: Yeah sounds good

cody-wang-cb (approved):
LGTM

specs/protocol/derivation.md
```diff
@@ -363,29 +363,49 @@ where:
...
+
+```text
+channel_encoding = `version_byte ++ compress(rlp_batches)`
+```
+
+For the versioned format, the `version_byte` must never have its 4 lower order bits set to 0x1000,
+as those bits indicate the channel encoding is of legacy type by definition of the zlib encoding
+header.
+
+As of the Fjord upgrade, the only valid `version_byte` is `1`, which indicates `compress()` is the
```
sebastianst: This may read as if `zlib` is not allowed post-Fjord, but we would still allow it if `version_byte & 0xf == 8` and in this case treat the version byte as the first byte of the compressed zlib stream.

specs/protocol/derivation.md
```diff
@@ -363,29 +363,49 @@ where:
...
 [rfc1950]: https://www.rfc-editor.org/rfc/rfc1950.html
 
+After the Fjord upgrade, there are two possible channel encoding formats. The legacy format,
+which is exactly as above, and a versioned format:
+
+```text
+channel_encoding = `version_byte ++ compress(rlp_batches)`
+```
+
+For the versioned format, the `version_byte` must never have its 4 lower order bits set to 0x1000,
```
sebastianst: I think using `0x...` here is misleading, because it's not in hex, but in binary notation. So maybe better
```suggestion
For the versioned format, the `version_byte` must never have its 4 lower order bits set to `0b1000 = 8`,
```

We should also elaborate on why this selects for zlib and maybe reference Page 5 from the zlib rfc that explains that this is set to 8. Maybe we should also forbid 15 since it's reserved in the zlib spec? But maybe that's too paranoid.

specs/protocol/derivation.md
```diff
@@ -363,29 +363,49 @@ where:
...
+
+```text
+channel_encoding = zlib_compress(rlp_batches)
+```
+
+where zlib_compress is the ZLIB algorithm (as specified in [RFC-1950][rfc1950]) with no dictionary.
 
 [rfc1950]: https://www.rfc-editor.org/rfc/rfc1950.html
 
+After the Fjord upgrade, there are two possible channel encoding formats. The legacy format,
```
sebastianst: Note that we use a new folder layout in the specs. This should primarily go into `specs/fjord/derivation.md`. We could then here just reference it.

Ideally you could base this PR onto https://github.com/ethereum-optimism/specs/pull/146 which will be merged very soon. The `fjord/overview.md` also needs a Brotli entry.

The specs are further modularized in this direction with https://github.com/ethereum-optimism/specs/pull/145 which is probably also merged soon.

specs/protocol/derivation.md
```diff
@@ -363,29 +363,49 @@ where:
...
+which is exactly as above, and a versioned format:
+
+```text
+channel_encoding = `version_byte ++ compress(rlp_batches)`
```
sebastianst: `version_byte` may be ambiguous and is already used in the frame spec. Maybe we name it something like `compr_version` or `compr_version_byte`?

specs/protocol/derivation.md
```diff
@@ -424,6 +445,10 @@ where:
 [RLP format]: https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/
 [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
 
+The Delta upgrade introduced an additional batch type, [span batches][span-batches].
+
+[span-batches]: span-batches.md
```
sebastianst, line 446: Similarly to how span batches are referenced here, we should also reference the Fjord Brotli spec from the main derivation spec file.

sebastianst (changes requested):
Thank you! Would be great to rebase onto https://github.com/ethereum-optimism/specs/pull/146

specs/protocol/derivation.md
roberto-bayardo: Hmm, technically there is no version byte if it's the legacy format, which is why I explicitly note that no version byte can have 0b1000 lower order bits.   Let me see if I can clarify...

specs/protocol/derivation.md
roberto-bayardo: changed to channel_version_byte

specs/protocol/derivation.md
roberto-bayardo: ok, might be best then to wait for those to merge and then I can rebase if it will happen soon?

specs/protocol/derivation.md
roberto-bayardo: reworded, and added 15 to reserved bit list.

specs/protocol/derivation.md
roberto-bayardo: reworded a bit, does that address the concern?

specs/protocol/derivation.md
sebastianst: #146 got merged, I'm checking in on #145

specs/protocol/derivation.md
sebastianst: yep thanks

specs/protocol/derivation.md
sebastianst: #145 won't be merged soon. Please still move this content to `specs/fjord/derivation.md` and already mention in `fjord/overview.md`, since we already use this layout for Fjord.

specs/protocol/derivation.md
roberto-bayardo: done

specs/protocol/derivation.md
roberto-bayardo: done

specs/fjord/derivation.md
```diff
@@ -48,3 +49,26 @@ be in the same epoch as the the last pre-Fjord blocks, even if these blocks woul
 have these L1-origin timestamps according to pre-Fjord rules. So the same L1 timestamp would be
 shared within a pre- and post-Fjord mixed epoch. This is considered a feature and is not considered
 a security issue.
+
+## Channel Format
```
sebastianst:
```suggestion
# Brotli Channel Compression
```
or it's part of section "Protocol Parameter Changes". Maybe also a more catchy title :D

roberto-bayardo: changed, and updated the ToC reference accordingly.

specs/fjord/overview.md
```diff
@@ -18,3 +18,5 @@ This document is not finalized and should be considered experimental.
 ## Consensus Layer
 
 - [Constant maximum sequencer drift](./derivation.md#batch-queue)
+
+- [Channel encoding format](./derivation.md#channel-format)
```
sebastianst: Just realize that both references are broken
```suggestion
- [Constant maximum sequencer drift](./derivation.md#constant-maximum-sequencer-drift)
- [Brotli channel compression](./derivation.md#brotli-channel-compression)
```

sebastianst (changes requested):

sebastianst (approved):

trianglesphere (approved):

Hahaheehaw:
great work!

PR #170, opened 2024-05-03
Cleanup Span Batch Limits

trianglesphere:
**Description**

This modifies how the span batch limits are specified. Previously it was using a protocol constant which was denominated in bytes (MAX_RLP_BYTES_PER_CHANNEL) to limit the total size, the number of blocks, the number of transactions per block, and the total number of transactions.

This PR modifies the spec to split out MAX_RLP_BYTES_PER_CHANNEL and MAX_SPAN_BATCH_ELEMENT_COUNT. This is not a consensus change because the values are the same. In addition, this removes the size check from the span batch validation. This is also not a consensus change until the Fjord hard fork because MAX_RLP_BYTES_PER_CHANNEL is checked prior to decoding the span batch. After Fjord, the behavior will change when the MAX_RLP_BYTES_PER_CHANNEL constant is increased.

### DIFF
--- a/specs/protocol/derivation.md
+++ b/specs/protocol/derivation.md
@@ -199,7 +199,7 @@ protocol upgrades.
 | `max_sequencer_drift` | 600 | 1800 | [Fjord](../fjord/derivation.md#constant-maximum-sequencer-drift) | Changed from a chain parameter to a constant with Fjord. |
 | `MAX_RLP_BYTES_PER_CHANNEL` | 10,000,000 | 100,000,000 | [Fjord](../fjord/derivation.md#increasing-max_rlp_bytes_per_channel-and-max_channel_bank_size) | Constant increased with Fjord. |
 | `MAX_CHANNEL_BANK_SIZE` | 100,000,000 | 1,000,000,000 | [Fjord](../fjord/derivation.md#increasing-max_rlp_bytes_per_channel-and-max_channel_bank_size) | Constant increased with Fjord. |
-| `MAX_SPAN_BATCH_ELEMENT_COUNT` | 10,000,000 | 10,000,000 | | Number of elements |
+| `MAX_SPAN_BATCH_ELEMENT_COUNT` | 10,000,000 | 10,000,000 | Effectively introduced in [Fjord](../fjord/derivation.md#increasing-max_rlp_bytes_per_channel-and-max_channel_bank_size)| Number of elements |
 
 ---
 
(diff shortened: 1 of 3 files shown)


specs/protocol/span-batches.md
```diff
@@ -133,16 +133,15 @@ tx_sigs ++ tx_tos ++ tx_datas ++ tx_nonces ++ tx_gases ++ protected_bits`
...
 
-Total size of encoded span batch is limited to `MAX_SPAN_BATCH_SIZE` (currently 10,000,000 bytes,
-equal to `MAX_RLP_BYTES_PER_CHANNEL`). Therefore every field size of span batch will be implicitly limited to
-`MAX_SPAN_BATCH_SIZE` . There can be at least single span batch per channel, and channel size is limited
-to `MAX_RLP_BYTES_PER_CHANNEL` and you may think that there is already an implicit limit. However, having an explicit
-limit for span batch is helpful for several reasons. We may save computation costs by avoiding malicious input while
-decoding. For example, let's say bad batcher wrote span batch which `block_count = max.Uint64`. We may early return
-using the explicit limit, not trying to consume data until EOF is reached. We can also safely preallocate memory for
-decoding because we know the upper limit of memory usage.
+Total size of encoded span batch is limited to `MAX_RLP_BYTES_PER_CHANNEL` which is defined in the
```
sebastianst: Maybe we take the opportunity to clean up grammar a bit
```suggestion
The total size of an encoded span batch is limited to `MAX_RLP_BYTES_PER_CHANNEL`, which is defined in the
```

specs/protocol/derivation.md
```diff
@@ -199,6 +199,7 @@ protocol upgrades.
 | `max_sequencer_drift` | 600 | 1800 | [Fjord](../fjord/derivation.md#constant-maximum-sequencer-drift) | Changed from a chain parameter to a constant with Fjord. |
 | `MAX_RLP_BYTES_PER_CHANNEL` | 10,000,000 | 100,000,000 | [Fjord](../fjord/derivation.md#increasing-max_rlp_bytes_per_channel-and-max_channel_bank_size) | Constant increased with Fjord. |
 | `MAX_CHANNEL_BANK_SIZE` | 100,000,000 | 1,000,000,000 | [Fjord](../fjord/derivation.md#increasing-max_rlp_bytes_per_channel-and-max_channel_bank_size) | Constant increased with Fjord. |
+| `MAX_SPAN_BATCH_ELEMENT_COUNT` | 10,000,000 | 10,000,000 | | Number of elements |
```
sebastianst: Maybe also say something like "Effectively introduced with Fjord" and link to the `Span Batch Size Limits` section? It may also make sense to add one sentence to the Fjord specs section about increasing the `MAX_RLP_BYTES_PER_CHANNEL` constant, that this new constant takes effect with it. This may make it easier to implement the Fjord diff in existing implementations.

specs/protocol/span-batches.md
```diff
@@ -133,16 +133,15 @@ tx_sigs ++ tx_tos ++ tx_datas ++ tx_nonces ++ tx_gases ++ protected_bits`
...
+[Protocol Parameters table](#protocol-parameters).
+This is done at the channel level rather than at the span batch level.
+
+In addition to the byte limit, the number of blocks, transactions per block, and total transactions is limited
```
sebastianst, line 142: If we already limit the number of blocks and total transactions by this limit, don't we implicitly already limit the transactions per block with it? So we could drop mentioning it explicitly.

sebastianst: The limiting of the number of blocks and total number of transactions already implies that limit on the transaction count per block. So mentioning the "transactions per block" is more of an implementation hint.

protolambda (approved):

sebastianst (approved):

PR #173, opened 2024-05-07
hardforks: next 2

tynes:
**Description**

Create proposals for the names of the next 2 hardforks with what they
could include. This is to help with planning. Basically we have 1
hardfork that contains a routine upgrade of the smart contracts,
this is relatively low lift and can include additional features if
there is bandwidth. After that we plan for an interop focused hardfork.
This enables us to implement the hardfork features behind feature
flags such that we do not need a feature branch.


ajsutton (approved):
In terms of feature toggling in code, I'd recommend using a feature specific method rather than `IsGranite` though - the feature specific method can just delegate straight to `IsGranite` but it means we can easily change the hard fork that a feature is scheduled for if needed without risk of missing some cases.  It's also easy to inline the feature specific method if we want to after it ships and the hard fork is fixed.

refcell (approved):
SGTM

sebastianst:
> In terms of feature toggling in code, I'd recommend using a feature specific method rather than `IsGranite` though - the feature specific method can just delegate straight to `IsGranite` but it means we can easily change the hard fork that a feature is scheduled for if needed without risk of missing some cases. It's also easy to inline the feature specific method if we want to after it ships and the hard fork is fixed.

@ajsutton I described a similar pattern in [this issue](https://github.com/ethereum-optimism/client-pod/issues/702) a while back. I will showcase this with the sequencer drift change for Fjord, and can then write a short design doc to describe the pattern for future use.

sebastianst (approved):

ajsutton:
> @ajsutton I described a similar pattern in https://github.com/ethereum-optimism/client-pod/issues/702 a while back. I will showcase this with the sequencer drift change for Fjord, and can then write a short design doc to describe the pattern for future use.

@sebastianst Yep, I couldn't remember where I'd seen it before but that's exactly what I had in mind when suggesting it.

protolambda (approved):
I am fine with the previous dubbed "Grove" / "Horizon" names as well as the new "Granite" / "Holocene".

PR #176, opened 2024-05-09
fault-proof: Specify L2 Block Number Challenge

Inphi:
Specify the mechanics of the L2 block number challenge in the FDG.


ajsutton (approved):
LGTM.

Issue #122, opened 2024-04-11
Potential Research Area: Separate Dynamic and Static L1 Attributes Values


tynes:
As of Ecotone, the L1 Attributes transaction includes the following information:

- baseFeeScalar
- blobBaseFeeScalar
- sequenceNumber
- l1BlockTimestamp
- l1BlockNumber  
- basefee
- blobBaseFee
- l1BlockHash
- batcherHash

Not all of this information changes every L2 block, which is the reasoning behind [Diff Based L1 Attributes Transactions](https://github.com/ethereum-optimism/specs/blob/main/specs/experimental/diff-based-attributes-tx.md). Some fields are guaranteed to change every L2 block like the sequence number, some fields change every time the L1 origin updates (L1 blocknumber, L1 timestamp, etc) and some fields only change based on chain operator input (batcher hash). 

For the fields that change automatically based on the L1 origin, we can call those dynamic values. For the values that only change based on chain operator input, we can call those static values.

The custom gas token spec adds a new static value which is the address of the ERC20 that represents the gas paying asset for the L2. A new design pattern is followed where the value is set in the system config and then it triggers a special deposit transaction that is able to set the address directly in the L2's L1Block predeploy. This pattern can be made more generic to work with all static values.

The next time that we need to add a static value to the L1 Attributes transaction, we should:
- Make a generic pipeline that is very similar to the way that custom gas token deposits its information into L2
- Remove all static values from the L1 attributes transaction and instead allow them to be configured "just in time" by the generic pipeline

This would allow us to easily allow for the configuration of the batch inbox for example

tynes:
Much of the design has already been fleshed out for interop. The high level is there is a new setter function on the `OptimismPortal` for spoofing the depositor address via deposit txs that are able to call the `L1Block` contract to set important values. This could be "upstreamed" into the mainline contracts as part of holocene to reduce the diff for the interop contracts upgrade.

See the following links: 
- https://github.com/ethereum-optimism/optimism/blob/5b7d2b98b109302b1356eee18ebf0dd970b47bb4/packages/contracts-bedrock/src/L2/L1BlockInterop.sol#L14
- https://github.com/ethereum-optimism/optimism/blob/5b7d2b98b109302b1356eee18ebf0dd970b47bb4/packages/contracts-bedrock/src/L1/SystemConfigInterop.sol#L29
[...]

PR #179, opened 2024-05-09
interop: add separation for static vs. dynamic values in L1Attributes

0xfuturistic:
context: https://github.com/ethereum-optimism/specs/issues/122

### DIFF
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -280,3 +280,11 @@ with `ADD_DEPENDENCY` or `REMOVE_DEPENDENCY`, respectively.
 ## Security Considerations
 
 TODO
+
+# Appendix
+
+## Static Values
+
+Static values are values that only change based on the chain operator's input, such as the gas paying token and the
+dependency set. This contrasts with dynamic values, which change automatically based on the L1 origin's updates, such
+as the sequence number and the L1 block number.
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -234,16 +234,18 @@ getters.
 
 ## L1Block
 
-| Constant | Value                                        |
-|----------|----------------------------------------------|
-| Address  | `0x4200000000000000000000000000000000000015` |
+| Constant            | Value                                        |
+|---------------------|----------------------------------------------|
+| Address             | `0x4200000000000000000000000000000000000015` |
+| `DEPOSITOR_ACCOUNT` | `0xDeaDDEaDDeAdDeAdDEAdDEaddeAddEAdDEAd0001` |
+
 
 ### Static Configuration
 
 The `L1Block` contract MUST include method `setConfig(ConfigType, bytes)` for setting the system's static values, which
 are defined as values that only change based on the chain operator's input. This function serves to reduce the size of
 the L1 Attributes transaction, as well as to reduce the need to add specific one off functions. It can only be called by
-the special depositor account.
+`DEPOSITOR_ACCOUNT`.
 
 The `ConfigType` enum is defined as follows:
 
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -22,8 +22,6 @@
   - [Static Configuration](#static-configuration)
   - [Dependency Set](#dependency-set)
 - [Security Considerations](#security-considerations)
-- [Appendix](#appendix)
-  - [Static Values](#static-values)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -247,8 +247,9 @@ The interop upgrade block itself MUST include a call to `setL1BlockValuesEcotone
 
 ### L1Attributes
 
-The L1 Atrributes transaction is updated to include the dependency set. Since the dependency set is dynamically sized,
-a `uint8` "interopSetSize" parameter prefixes tightly packed `uint256` values that represent each chain id.
+The L1 Attributes transaction is updated to only include dynamic values. Static values, such as `batcherHash` and
+the interop dependency set, are set through the `setConfig` method.
+
 
 | Input arg         | Type                     | Calldata bytes          | Segment |
 |-------------------|--------------------------|-------------------------|---------|
@@ -261,9 +262,6 @@ a `uint8` "interopSetSize" parameter prefixes tightly packed `uint256` values th
 | basefee           | uint256                  | 36-67                   | 2       |
 | blobBaseFee       | uint256                  | 68-99                   | 3       |
 | l1BlockHash       | bytes32                  | 100-131                 | 4       |
-| batcherHash       | bytes32                  | 132-163                 | 5       |
-| interopSetSize    | uint8                    | 164-165                 | 6       |
-| chainIds          | uint256\[interopSetSize] | 165-(32*interopSetSize) | 6+      |
 
 ## Security Considerations
 
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -254,11 +254,9 @@ enum ConfigType {
 }
 ```
 
-The second argument of `setConfig` is the encoded value being set:
+For `GAS_PAYING_TOKEN`, let `token` be the gas paying token's address (type `address`), `decimals` be the token's decimals (type `uint8`),  `name` be the token's name (type `bytes32`), and `symbol` be the token's symbol (type `bytes32`). The value passed to `setConfig` as the second argument would then be `abi.encode(token, decimals, name, symbol)`.
 
-- For `GAS_PAYING_TOKEN`, this value contains the address of the gas paying token, its decimals, name, and symbol.
-
-- For `ADD_DEPENDENCY` and `REMOVE_DEPENDENCY`, this value contains the chain id being added or removed.
+On the other hand, for `ADD_DEPENDENCY` and `REMOVE_DEPENDENCY`, let `chainId` be the chain id intended to be added or removed from the dependency set. The value passed to `setConfig` as the second argument would be `abi.encode(chainId)`.
 
 Calls to `setConfig` MUST originate from `SystemConfig` and are forwarded to `L1Block` by `OptimismPortal`.
 
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -242,9 +242,9 @@ getters.
 
 ### Static Configuration
 
-The `L1Block` contract MUST include method `setConfig(ConfigType, bytes)` for setting static values that are used by the
-system. This function serves to reduce the size of the L1 Attributes transaction, as well as to reduce the need to add
-specific one off functions.
+The `L1Block` contract MUST include method `setConfig(ConfigType, bytes)` for setting the system's static values, which
+are defined as values that only change based on the chain operator's input. This function serves to reduce the size of
+the L1 Attributes transaction, as well as to reduce the need to add specific one off functions.
 
 The `ConfigType` enum is defined as follows:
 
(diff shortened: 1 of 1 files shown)


specs/interop/predeploys.md
```diff
@@ -237,18 +237,34 @@ getters.
 |----------|----------------------------------------------|
 | Address  | `0x4200000000000000000000000000000000000015` |
 
-The `L1Block` contract is updated to include the set of allowed chains. The L1 Attributes transaction
-sets the set of allowed chains. The `L1Block` contract MUST provide a public getter to check if a particular
-chain is in the dependency set called `isInDependencySet(uint256)`. This function MUST return true when
-the chain's chain id is passed in as an argument.
+The `L1Block` contract MUST include method `setConfig(ConfigType, bytes)` for setting static values that are used
```
tynes: I think that this should be in its own subsection

specs/interop/predeploys.md
```diff
@@ -237,18 +237,34 @@ getters.
...
 
 ### L1Attributes
 
-The L1 Atrributes transaction is updated to include the dependency set. Since the dependency set is dynamically sized,
-a `uint8` "interopSetSize" parameter prefixes tightly packed `uint256` values that represent each chain id.
+The L1 Attributes transaction is updated to only include dynamic values. Static values, such as `batcherHash` and
+the interop dependency set, are set through the `setConfig` method.
 
 | Input arg         | Type                     | Calldata bytes          | Segment |
 |-------------------|--------------------------|-------------------------|---------|
```
tynes: I am noticing that the fee scalars are also static config, we could simplify this more by adding enum fields for them

tynes: This would effectively replace https://github.com/ethereum-optimism/specs/blob/main/specs/experimental/diff-based-attributes-tx.md

The node implementation section is important to be mindful of as it would also apply to this change

0xfuturistic: good idea!

specs/interop/predeploys.md
tynes: If you made a subsection specific about the `setConfig` function then that would be great. It should include why it exists (reduces size of L1 attributes tx, reduces need to add specific one off functions)

After thinking through the upgrade process for this, I recommend that we don't include batcher hash or any of the other existing static config types yet. If we just do gas paying token + add + remove, I think it will be most simple for the upgrade. Otherwise we will have a really gnarly upgrade. This gives us the main benefit of not needing to include the full dep set into the L1 attributes tx every time

tynes: The subsection should also include the definition of the `bytes` for each `ConfigType` option

specs/interop/predeploys.md
tynes: If we also use the static path for the fee scalars, then we will need to guarantee that we can always update them, meaning that we would need to stage the upgrades such that the L2 hardforks and adds the new path for setting them but the L1 contracts don't call them yet, then we would need to update the L1 contracts to be able to call them, then we would need to update the L2 contract to only use the static setter path

tynes: That is too big of a decision to make in this PR, it would require a design doc and presenting it at the design review session, so i recommend we just stick with custom gas token + interop related features for now and revisit the idea later

tynes: I think the best next steps are to not modify the L1 attributes tx at all from ecotone and still follow the old flow of doing the updates for the old static attributes but have the `L1Block` contract ready with the new setters for each of the static fields. Then we can have a smooth transition and deprecate the old path. This removes a synchronous upgrade requirement

0xfuturistic: good call! I agree with everything

0xfuturistic: since L1 Attributes tx is not modified, dropping the section in about it

specs/interop/predeploys.md
0xfuturistic: good idea!

specs/interop/predeploys.md
```diff
@@ -237,33 +238,37 @@ getters.
...
+enum ConfigType {
+    GAS_PAYING_TOKEN,
+    ADD_DEPENDENCY,
+    REMOVE_DEPENDENCY
+}
+```
+
+The second argument of `setConfig` is the encoded value being set:
+
+- For `GAS_PAYING_TOKEN`, this value contains the address of the gas paying token, its decimals, name, and symbol.
```
tynes: This is not thoroughly specified. There are hundreds of ways to serialize this information. Can you be very precise in how its serialized?

specs/interop/predeploys.md
```diff
@@ -237,33 +238,37 @@ getters.
...
+
+```solidity
+enum ConfigType {
+    GAS_PAYING_TOKEN,
+    ADD_DEPENDENCY,
+    REMOVE_DEPENDENCY
+}
+```
+
+The second argument of `setConfig` is the encoded value being set:
```
tynes: It would be ideal if this was done in a markdown table

specs/interop/predeploys.md
0xfuturistic: yes -- totally agree

specs/interop/predeploys.md
0xfuturistic: great idea!

0xfuturistic: i just pushed a few changes. realized that I could use markdown too late, but I think the current approach isn't too bad. could you check it out?

0xfuturistic: i just pushed a few changes. realized that I could use markdown too late, but I think the current approach isn't too bad. could you check it out @tynes?

tynes: I think it would go best in an markdown table with the column containing a specific encoding, ie `abi.encode(...)`

specs/interop/predeploys.md
```diff
@@ -237,33 +238,39 @@ getters.
...
-| l1BlockNumber     | uint64                   | 28-35                   |         |
-| basefee           | uint256                  | 36-67                   | 2       |
-| blobBaseFee       | uint256                  | 68-99                   | 3       |
-| l1BlockHash       | bytes32                  | 100-131                 | 4       |
-| batcherHash       | bytes32                  | 132-163                 | 5       |
-| interopSetSize    | uint8                    | 164-165                 | 6       |
-| chainIds          | uint256\[interopSetSize] | 165-(32*interopSetSize) | 6+      |
+### setConfig
+
+The `L1Block` contract MUST include method `setConfig(ConfigType, bytes)` for setting static values that are used by the
```
tynes: Its not clear what a static value is from this and I am not sure if its super important to the main part of the spec. Defining static values could be its own subsection at the end of this doc

specs/interop/predeploys.md
```diff
@@ -237,33 +238,39 @@ getters.
...
-| sequenceNumber    | uint64                   | 12-19                   |         |
-| l1BlockTimestamp  | uint64                   | 20-27                   |         |
-| l1BlockNumber     | uint64                   | 28-35                   |         |
-| basefee           | uint256                  | 36-67                   | 2       |
-| blobBaseFee       | uint256                  | 68-99                   | 3       |
-| l1BlockHash       | bytes32                  | 100-131                 | 4       |
-| batcherHash       | bytes32                  | 132-163                 | 5       |
-| interopSetSize    | uint8                    | 164-165                 | 6       |
-| chainIds          | uint256\[interopSetSize] | 165-(32*interopSetSize) | 6+      |
+### setConfig
```
tynes: The title `setConfig` doesn't give good context. I recommend we use a different name for this section

specs/interop/predeploys.md
0xfuturistic: added a subsection at the end of this doc (in a new section `Appendix`). lmk if it makes sense to define an Appendix in this context

specs/interop/predeploys.md
0xfuturistic: renamed to `Static Configuration`

specs/interop/predeploys.md
tynes: Could you link on the first time "static values" is mentioned to the part of the appendix where its defined?

specs/interop/predeploys.md
```diff
@@ -237,34 +240,60 @@ getters.
...
+
+Calls to `setConfig` MUST originate from `SystemConfig` and are forwarded to `L1Block` by `OptimismPortal`.
+
+### Dependency Set
+
+`L1Block` is updated to include the set of allowed chains. These chains are added and removed through `setConfig` calls
+with `ADD_DEPENDENCY` or `REMOVE_DEPENDENCY`, respectively.
+
+`L1Block` MUST provide a public getter to check if a particular chain is in the dependency set called
+`isInDependencySet(uint256)`. This function MUST return true when the chain's chain id is passed in as an argument.
```
tynes: Can we also add another function `dependencySet()([]uint256)` as a way to fetch all of the dependencies?

specs/interop/predeploys.md
tynes: Adding an appendix makes sense altho we don't have that section elsewhere in the specs. Usually its best practice to follow the same standards of the repo to keep things consistent

0xfuturistic: just we just remove appendix and add the subsection for static values at the end? it may not be clear enough that it's not part of the main spec, but maybe this is good enough for the sake of consistency

specs/interop/predeploys.md
0xfuturistic: good idea!

specs/interop/predeploys.md
tynes: I think just defining it inline where its first mentioned could be the most simple

tynes:
needs a lint

specs/interop/predeploys.md
```diff
@@ -237,33 +238,55 @@ getters.
...
-| l1BlockNumber     | uint64                   | 28-35                   |         |
-| basefee           | uint256                  | 36-67                   | 2       |
-| blobBaseFee       | uint256                  | 68-99                   | 3       |
-| l1BlockHash       | bytes32                  | 100-131                 | 4       |
-| batcherHash       | bytes32                  | 132-163                 | 5       |
-| interopSetSize    | uint8                    | 164-165                 | 6       |
-| chainIds          | uint256\[interopSetSize] | 165-(32*interopSetSize) | 6+      |
+### Static Configuration
+
+The `L1Block` contract MUST include method `setConfig(ConfigType, bytes)` for setting the system's static values, which
```
tynes, line 244: This is missing details on auth - can anybody call it? I think it should only be the depositor account

0xfuturistic: good catch!

specs/interop/predeploys.md
```diff
@@ -237,33 +238,56 @@ getters.
...
+are defined as values that only change based on the chain operator's input. This function serves to reduce the size of
+the L1 Attributes transaction, as well as to reduce the need to add specific one off functions. It can only be called by
```
tynes, line 246: The special depositor account needs to be defined

tynes: It is important when working on a spec to think about all of the information that is required for somebody else to implement the core features of it. Right now its not possible to know what the depositor account is. There is a table already specifically for `L1Block` constants

specs/interop/predeploys.md
```diff
@@ -237,33 +238,56 @@ getters.
...
+
+- `token` is the gas paying token's address (type `address`)
+
+- `decimals` is the gas paying token's decimals (type `uint8`)
+
+- `name` is the gas paying token's name (type `bytes32`)
+
+- `symbol` is the gas paying token's symbol (type `bytes32`)
+
+- `chainId` is the chain id intended to be added or removed from the dependency set
```
tynes: This is missing a type

tynes:
Linting is failing

tynes (approved):

PR #184, opened 2024-05-10
Adding RollbackInbox and upgrading Messenger & Bridge

skeletor-spaceman:
**Description**

Adds a way for the `CrossDomainMessenger` to "send-back" failed messages.
These failed messages are stored on the origin domain in a `RollbackInbox` contract.
This enables smart contracts to handle "failed" messages.
e.g. the `StandardBridge` can safely recover stuck funds by validating that a message was indeed NOT successful on the other domain.

For this proposal, we've also prepared a detailed Notion document, found [here](https://defi-wonderland.notion.site/RollbackInbox-V1-c5dc38b62cf64b51807427012c991bb3), that goes through the rationale and need of the `RollbackInbox` and accompanying functionality on the `CrossDomainMessenger` &  `StandardBridge`.

We also believe this would be an important primitive for Interop.

Would love to hear your thoughts and any feedback you might have on this proposal :)

**Tests**

N/A

**Additional context**

https://defi-wonderland.notion.site/RollbackInbox-V1-c5dc38b62cf64b51807427012c991bb3
L2ToL2CrossDomainMessenger alternative: https://github.com/ethereum-optimism/specs/pull/203

**Metadata**

N/A

This proposal was developed as an internal research project at @defi-wonderland lead by @0xng.

### DIFF
--- a/specs/protocol/bridges.md
+++ b/specs/protocol/bridges.md
@@ -6,6 +6,7 @@
 
 - [Overview](#overview)
 - [Token Depositing](#token-depositing)
+- [ERC20 Unlocking](#erc20-unlocking)
 - [Upgradability](#upgradability)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
@@ -29,15 +30,19 @@ interface StandardBridge {
     event ERC20BridgeInitiated(address indexed localToken, address indexed remoteToken, address indexed from, address to, uint256 amount, bytes extraData);
     event ETHBridgeFinalized(address indexed from, address indexed to, uint256 amount, bytes extraData);
     event ETHBridgeInitiated(address indexed from, address indexed to, uint256 amount, bytes extraData);
+    event ERC20Unlocked(address indexed localToken, address indexed remotetoken, address indexed from, uint256 amount, bytes32 messageHash);
 
     function bridgeERC20(address _localToken, address _remoteToken, uint256 _amount, uint32 _minGasLimit, bytes memory _extraData) external;
     function bridgeERC20To(address _localToken, address _remoteToken, address _to, uint256 _amount, uint32 _minGasLimit, bytes memory _extraData) external;
     function bridgeETH(uint32 _minGasLimit, bytes memory _extraData) payable external;
     function bridgeETHTo(address _to, uint32 _minGasLimit, bytes memory _extraData) payable external;
     function deposits(address, address) view external returns (uint256);
+    function processedMessages(bytes32 _messageHashes) view external returns (uint256);
+    function unlockERC20s(address _localToken, address _remoteToken, address _from, address _to, uint256 _amount, bytes calldata _extraData, uint240 _nonce, uint32 _minGasLimit) external;
     function finalizeBridgeERC20(address _localToken, address _remoteToken, address _from, address _to, uint256 _amount, bytes memory _extraData) external;
     function finalizeBridgeETH(address _from, address _to, uint256 _amount, bytes memory _extraData) payable external;
     function messenger() view external returns (address);
+    function ROLLBACK_INBOX() view external returns (address);
     function OTHER_BRIDGE() view external returns (address);
 }
 ```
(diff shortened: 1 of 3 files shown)


specs/protocol/messengers.md
```diff
@@ -79,6 +84,16 @@ then waits for the finalization window to pass, and then finalizes the withdrawa
...
 `L1CrossDomainMessenger` to finalize the withdrawal.
 
+## Rolling a Message Back
+
+The `sendHashToRollbackInbox` function is used to send the hashes of failed
+and unsuccesful messages to the `ROLLBACK_INBOX` contract on the other domain
+after a given delay, ensuring the message corresponding to the rolled back hash
+is never processed in the current domain. The main purpose of this function
+is to inform the other domain of failed messages so contracts living in
+this other domain can handle the message failure. This allows functionalities
```
tynes, line 94: Do you see any other use cases besides unlocking stuck ERC20s? Trying to get a sense of what the major use cases built on this would be

specs/protocol/bridges.md
```diff
@@ -6,6 +6,7 @@
 
 - [Overview](#overview)
 - [Token Depositing](#token-depositing)
+- [ERC20 Unlocking](#erc20-unlocking)
```
tynes, line 9: We use to have this built directly into the standard bridge, see https://github.com/ethereum-optimism/optimism-legacy/blob/8205f678b7b4ac4625c2afe351b9c82ffaa2e795/packages/contracts/contracts/L2/messaging/L2StandardBridge.sol#L163

specs/protocol/messengers.md
skeletor-spaceman: that's a great question, yes.
we think this will be a great primitive for handling the not-so-happy path of cross-chain actions.
another example would be a cross-chain(x) swap, where the origin.amm would forward and save an identifier of the user xAction.
which, if the xMessage fails, will be used to continue and gracefully handle the reverted xAction, returning the funds or an LP position to the swapper.

tynes:
While [$320,000](https://dune.com/defi_wonderland/gg) worth of ether is a good amount of money to be stuck, we should ask the question of whether or not this is the biggest fish to fry. Would love to hear some feedback from users that are asking for this feature right now.

I also wonder if there are simpler ways to solve the problem, like reintroducing the [check](https://github.com/ethereum-optimism/optimism-legacy/blob/8205f678b7b4ac4625c2afe351b9c82ffaa2e795/packages/contracts/contracts/L2/messaging/L2StandardBridge.sol#L163) we had in the legacy version of the contracts or changing the semantics around the `minGasLimit`. The maintainers of the EVM recommend against gas introspection and want to remove the ability to do that in the future, we learned the hard way through the complexity that was introduced by having gas introspection through `minGasLimit`. It was originally introduced to make withdrawal relay networks more safe, but it turned out that the withdrawal relay networks were never built in practice. I lean towards solutions that remove the `minGasLimit` (example: introduce new style of withdrawal message without `minGasLimit`, ignore `minGasLimit` on legacy withdrawals) as it reduces complexity while solving problems

specs/protocol/bridges.md
skeletor-spaceman: this is great! this was our first approach to try to fix the problem.
but we thought that if we were to modify the core contracts, we would rather see a future-proof feature being added than a patch for a specific use-case.
while also keeping at a minimum the required modifications to the current codebase.

skeletor-spaceman:
@tynes I think this might not be a "big" fish to fry, but it's definitely an interesting one, as it aligns us on a solid standard for cross-chain messaging, which will become the bloodstream of the Superchain.
also, we'll work on a quick Spike/R&D task, that would complement this RFC, to check how could we remove the `minGasLimit` overhead, and what the side-effects might be.

tynes:
> @tynes I think this might not be a "big" fish to fry, but it's definitely an interesting one, as it aligns us on a solid standard for cross-chain messaging, which will become the bloodstream of the Superchain. also, we'll work on a quick Spike/R&D task, that would complement this RFC, to check how could we remove the `minGasLimit` overhead, and what the side-effects might be.

cc @clabby who has thought about this problem a bunch

PR #185, opened 2024-05-10
specs: reorganize into new format

tynes:
**Description**

Reorganizes the specs into a folder based approach.
This does not modify `fjord`, it will need to be migrated
in a future PR. The main problem that this solves is an
encapsulation of features by network upgrade.

Each network upgrade should contain documents that
pertain to the specific features that network upgrade
introduced. This PR exists to unblock things like https://github.com/ethereum-optimism/specs/pull/173
where we begin to work on the next set of network upgrades.

This commit is not meant to be perfect and we will need to continue
to refine the layout of the specs into the future.

This is a replacement for https://github.com/ethereum-optimism/specs/pull/145.
Ideally we can merge this pretty quickly so that it doesn't
face risk of becoming stale.


trianglesphere:
lint is failing on multiple blank lines (should be a quick fix)

trianglesphere (approved):

ajsutton (approved):

PR #186, opened 2024-05-10
feat: Specify max L2 block gas limit

clabby:
## Overview

Specifies that the max gas limit in the `SystemConfig` is `200_000_000`

### DIFF
--- a/specs/protocol/system_config.md
+++ b/specs/protocol/system_config.md
@@ -104,6 +104,9 @@ The gas limit of the L2 blocks is configured through the system config.
 Changes to the L2 gas limit are fully applied in the first L2 block with the L1 origin that introduced the change,
 as opposed to the 1/1024 adjustments towards a target as seen in limit updates of L1 blocks.
 
+The gas limit may not be set to a value larger than `200_000_000`. This is to ensure that the L2 blocks are fault
+provable and of reasonable size to be processed by the client software. Over time, this value will be increased.
+
 ### `unsafeBlockSigner` (`address`)
 
 Blocks are gossiped around the p2p network before they are made available on L1.


clabby:
* **#186** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/specs/186?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tynes:
Ideally the rationale is included, ie ensure that its fault provable

tynes (approved):

mds1:
Rationale for 200M per @ajsutton:

> Primarily it was based on the fact that we previously fault proved many blocks during FPA and now even using the fullest blocks we can find executing the block itself is under a minute in cannon. Trawling through a full sequence window of L1 blocks to find the batch data we need takes 3-4 hours :cry:
> So we have a ton of scope to increase gas limits and no scope to increase sequencing window. That said we haven’t done worst case analysis of this which is why the 200m is still reasonably conservative compared to the amount of gas we have executed in cannon in some tests. A single call to some if the precompiles was enough to be not fault provable for example and we had to go via the preimage oracle to skirt around that (we now just call the L1 precompile instead of running it in cannon).
> 
> And yeah the interop fault proof is designed such that the gas limit is per chain. We only ever execute a single block from a single chain in each cannon run.
> The thing we do need to sort out is the cross checking of interop messages - we can’t afford to walk back a lot of blocks of each chain. Those limits are very very tight and we’re going to have to come up with something if we want to fault prove executing messages that reference initiating messages older than a few hours I suspect (depending on number of chains in the interop set). There are some options there but need to work through the details.

PR #188, opened 2024-05-15
Minor edit for clarity - 'used by claimed by' -> 'used by'

NiloCK:
Edited sentence was:

> Picking a high identifier minimizes the risk that the identifier will be used be claimed by another transaction type on the L1 chain...

Now:

> Picking a high identifier minimizes the risk that the identifier will be used by another transaction type on the L1 chain...


tynes (approved):

tynes:
Thank you!

PR #190, opened 2024-05-15
Update verifier.md for typo

toolchainx:
typo, should be unsafe here


tynes (approved):
Good catch, thank you

PR #193, opened 2024-05-16
feat/interop: add reverts for manipulating dependency set

0xfuturistic:
these reverts specify what happens when trying to add a chain id when the dependency set is at it's maximum, and trying to remove the current chain's chain id or a non-dependency from the dependency set.

### DIFF
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -300,7 +300,10 @@ Calls to `setConfig` MUST originate from `SystemConfig` and are forwarded to `L1
 ### Dependency Set
 
 `L1Block` is updated to include the set of allowed chains. These chains are added and removed through `setConfig` calls
-with `ADD_DEPENDENCY` or `REMOVE_DEPENDENCY`, respectively.
+with `ADD_DEPENDENCY` or `REMOVE_DEPENDENCY`, respectively. The maximum size of the dependency is `type(uint256).max`, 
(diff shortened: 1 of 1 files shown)


specs/interop/predeploys.md
```diff
@@ -300,7 +300,10 @@ Calls to `setConfig` MUST originate from `SystemConfig` and are forwarded to `L1
 ### Dependency Set
 
 `L1Block` is updated to include the set of allowed chains. These chains are added and removed through `setConfig` calls
-with `ADD_DEPENDENCY` or `REMOVE_DEPENDENCY`, respectively.
+with `ADD_DEPENDENCY` or `REMOVE_DEPENDENCY`, respectively. The maximum size of the dependency is `type(uint256).max`,
```
tynes: When you say "the maximum size of the dependency" what do you mean exactly? iirc there is a chain id definition in the interop specs which defines it as a `uint256` meaning that a chain id can be any valid `uint256`

tynes:
Good addition, just one clarification needs to be made

specs/interop/predeploys.md
0xfuturistic: oh I meant to say the dependency _set_ size. thanks for noticing

0xfuturistic: addressed

0xfuturistic: also corrected the dependency set size from `type(uint256).max` to `type(uint8).max`

tynes (approved):

PR #194, opened 2024-05-16
standard predeploy management

tynes:
**Description**

This is a proposal for predeploy upgrade management to be included
in the granite network upgrade.

This aims to solve scaling the stage 1 rollup status to many chains in the superchain

### DIFF
--- a/specs/protocol/granite/predeploys.md
+++ b/specs/protocol/granite/predeploys.md
@@ -1,4 +1,4 @@
-# Predeploys
+# Protocol Predeploys
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
@@ -8,9 +8,115 @@
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
+[stage-1]: https://ethereum-magicians.org/t/proposed-milestones-for-rollups-taking-off-training-wheels/11571#stage-1-limited-training-wheels-3
+
+## Overview
+
+The OP Stack has a permissive management system for L2 predeploys that makes it
+difficult for an OP Stack chain to be considered a [Stage 1 Rollup][stage-1].
+This upgrade is meant to solve the problem of chain operators being able to
+easily break invariants important to the system by being able to arbitrarily
+modify the predeploys themselves.
+
+The `ProxyAdmin` on L2 is upgraded to the `L2ProxyAdmin`, which removes
+the code for handling the L1 legacy proxies and delineates cleanly between
+system ownership and chain operator ownership of different predeploys.
+
+## Constants
+
+| Constant          | Value                           | Description |
+| ----------------- | ------------------------------- | ----------- |
+| `PREDEPLOY_COUNT` |  2048 | The total number of predeploys on L2 |
+| `L2ProxyAdmin` address | `0x4200000000000000000000000000000000000018` | The predeploy of the `ProxyAdmin` on L2 |
+| `DEPOSITOR_ACCOUNT` | `address(0xDeaDDEaDDeAdDeAdDEAdDEaddeAddEAdDEAd0001)` | Account that represents the system |
+| `COLLECTIVE_ACCOUNT` | `address(0xDeaDDEaDDeAdDeAdDEAdDEaddeAddEAdDEAd0002)` | Account that represents the collective |
+| `CHAIN_ACCOUNT` | `address(0xDeaDDEaDDeAdDeAdDEAdDEaddeAddEAdDEAd0003)` | Account that represents the chain operator |
+
+## System Predeploys
+
+System predeploys are predeploys that can only be modified via protocol upgrades. This means
+that the chain governance cannot modify them at the application layer. These predeploys
+are important for maintaining protocol level invariants and therefore the security of
+the system.
+
+TODO: define range of system predeploys
+
+## Userland Predeploys
+
+Userland predeploys are predeploys that can be modified by governance. They should not
+be responsible for managing the security of the system, meaning that if governance was
+to go rogue, then the system should still be considered secure.
+
+Userland predeploys are broken down into two categories, collective governed and chain
+governed.
+
+TODO: define range of userland predeploys, both collective and chain
+
+### Collective Governed
+
+The collective governed userland predeploys are managed at the superchain level. This means
+that the governance of the superchain as a whole is responsible for upgrading these predeploys.
+
+### Chain Governed
+
+The chain governed userland predeploys are managed on a per chain level. This is a commitment
+that these predeploys will never be used by the protocol or superchain governance and can be
+safely used by the chain operator for whatever purposes they desire.
+
+## L2ProxyAdmin
+
+The `L2ProxyAdmin` is a minimal proxy admin that is the ERC-1967 admin for each of the predeploys.
+For any permissioned call, the required caller is different depending on the address that is being
+called.
+
+To modify the implementation:
+- For system predeploys, it MUST be the `DEPOSITOR_ACCOUNT`
+- For collective governed predeploys, it MUST be the `COLLECTIVE_ACCOUNT`
+- For chain governed predeploys, it MUST be the `CHAIN_ACCOUNT`
+
+None of these accounts have known private keys, therefore they must come from specially crafted
+`TransactionDeposited` events that come from the `OptimismPortal`.
+
+## OptimismPortal
+
+A method for emitting special `TransactionDeposited` events is added to the `OptimismPortal`.
+Authorization must be set up such that the only the `SystemConfig` can result in a deposit
+transaction from the `CHAIN_ACCOUNT` and only the `SuperchainConfig` can result in a deposit
+transaction from the `COLLECTIVE_ACCOUNT`.
+
+## SuperchainConfig
+
+A new method is added that can call an `OptimismPortal` that can only be called by the superchain
+governance system.
+
+## Network Upgrade Transactions
+
+TODO: perhaps break this section into its own document
+
 The granite network upgrade includes a routine upgrade
 of all L2 predeploys using system transactions. The exact
 and upgrade transactions themselves will be finalized in
 the future. This ensures that the L1 and L2 pairs of contracts
 (CrossDomainMessengers and StandardBridges) don't drift too
 far in their versioning.
+
+## Security Considerations
+
+### Why put ownership delineation in the `L2ProxyAdmin`?
+
+This prevents the need for tons of upgrade transactions that change the existing `admin`
+values that are set in the predeploys.
+
+### Why use deposit transactions for this management?
+
+This allows the hardfork to be uniform across all chains in the superchain. The config
+values that are different live on L1, otherwise we would have to hardfork in a way
+that enshrines a particular address (which may be different per L2 network).
+This also prevents the need to maintain balances on every chain in the superchain.
+
+### Is this susceptible to deposit transaction griefing?
+
+No, if it follows the same pattern used by the custom gas token upgrade where the
+deposit transaction path skips the gas burn and emits a special "system deposit transaction"
+that originates on L1. There is precedent for this kind of pattern based on the custom
+gas token feature.


specs/protocol/granite/predeploys.md
```diff
@@ -8,9 +8,115 @@
...
+## OptimismPortal
+
+A method for emitting special `TransactionDeposited` events is added to the `OptimismPortal`.
+Authorization must be set up such that the only the `SystemConfig` can result in a deposit
+transaction from the `CHAIN_ACCOUNT` and only the `SuperchainConfig` can result in a deposit
+transaction from the `COLLECTIVE_ACCOUNT`.
+
+## SuperchainConfig
+
+A new method is added that can call an `OptimismPortal` that can only be called by the superchain
```
maurelian, line 89: Which optimism portal, all of them?

specs/protocol/granite/predeploys.md
```diff
@@ -8,9 +8,115 @@
...
+- For chain governed predeploys, it MUST be the `CHAIN_ACCOUNT`
+
+None of these accounts have known private keys, therefore they must come from specially crafted
+`TransactionDeposited` events that come from the `OptimismPortal`.
+
+## OptimismPortal
+
+A method for emitting special `TransactionDeposited` events is added to the `OptimismPortal`.
+Authorization must be set up such that the only the `SystemConfig` can result in a deposit
+transaction from the `CHAIN_ACCOUNT` and only the `SuperchainConfig` can result in a deposit
```
maurelian, line 84: This phrasing is unclear to me. Is the SuperchainConfig forwarding calls? If so, from whom?

specs/protocol/granite/predeploys.md
```diff
@@ -8,9 +8,115 @@
...
+the code for handling the L1 legacy proxies and delineates cleanly between
+system ownership and chain operator ownership of different predeploys.
+
+## Constants
+
+| Constant          | Value                           | Description |
+| ----------------- | ------------------------------- | ----------- |
+| `PREDEPLOY_COUNT` |  2048 | The total number of predeploys on L2 |
+| `L2ProxyAdmin` address | `0x4200000000000000000000000000000000000018` | The predeploy of the `ProxyAdmin` on L2 |
+| `DEPOSITOR_ACCOUNT` | `address(0xDeaDDEaDDeAdDeAdDEAdDEaddeAddEAdDEAd0001)` | Account that represents the system |
```
tynes, line 31: The `DEPOSITOR_ACCOUNT` would only be able to be assumed by the `op-node` under this model

specs/protocol/granite/predeploys.md
```diff
@@ -8,9 +8,115 @@
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
```
mds1, line 9: TOC is outdated

specs/protocol/granite/predeploys.md
```diff
@@ -8,9 +8,115 @@
...
+
+The OP Stack has a permissive management system for L2 predeploys that makes it
+difficult for an OP Stack chain to be considered a [Stage 1 Rollup][stage-1].
+This upgrade is meant to solve the problem of chain operators being able to
+easily break invariants important to the system by being able to arbitrarily
+modify the predeploys themselves.
+
+The `ProxyAdmin` on L2 is upgraded to the `L2ProxyAdmin`, which removes
+the code for handling the L1 legacy proxies and delineates cleanly between
+system ownership and chain operator ownership of different predeploys.
```
mds1, line 23: Does system ownership mean who can upgrade contracts? If so we might want to come up with a new name for this role, based on https://specs.optimism.io/protocol/configurability.html#admin-roles — ProxyAdmin Owner is a bit too technical for the role name, and system owner is too similar to SystemConfig owner

specs/protocol/granite/predeploys.md
```diff
@@ -8,9 +8,115 @@
...
+
+## Constants
+
+| Constant          | Value                           | Description |
+| ----------------- | ------------------------------- | ----------- |
+| `PREDEPLOY_COUNT` |  2048 | The total number of predeploys on L2 |
```
mds1, line 29: Nit on phrasing (open to alternatives), since we don't have 2048 active predeploys
```suggestion
| `PREDEPLOY_COUNT` |  2048 | The total number of predeploy accounts reserved on L2 |
```

specs/protocol/granite/predeploys.md
```diff
@@ -8,9 +8,115 @@
...
+Userland predeploys are predeploys that can be modified by governance. They should not
+be responsible for managing the security of the system, meaning that if governance was
+to go rogue, then the system should still be considered secure.
+
+Userland predeploys are broken down into two categories, collective governed and chain
+governed.
+
+TODO: define range of userland predeploys, both collective and chain
+
+### Collective Governed
```
mds1, line 55: This is a new term, so we might need some standardization of role names to make sure the [configurability spec](https://specs.optimism.io/protocol/configurability.html#admin-roles), [docs](https://docs.optimism.io/chain/security/privileged-roles), [law of chains](https://github.com/ethereum-optimism/OPerating-manual/blob/main/Law%20of%20Chains.md), and this all align.

specs/protocol/granite/predeploys.md
```diff
@@ -8,9 +8,115 @@
...
+
+The collective governed userland predeploys are managed at the superchain level. This means
+that the governance of the superchain as a whole is responsible for upgrading these predeploys.
```
mds1, line 58: This sounds the same to me as System Predeploys, i.e. both can only be upgraded by a fork that goes through governance

specs/protocol/granite/predeploys.md
```diff
@@ -8,9 +8,115 @@
...
+
+### Chain Governed
+
+The chain governed userland predeploys are managed on a per chain level. This is a commitment
+that these predeploys will never be used by the protocol or superchain governance and can be
+safely used by the chain operator for whatever purposes they desire.
```
mds1, line 64: I like the distinction between chain governed and governance governed—but what predeploys that we currently have would fall into this category? If we don't currently have any, we should just remove this for now and simplify the spec

specs/protocol/granite/predeploys.md
```diff
@@ -8,9 +8,115 @@
...
+TODO: perhaps break this section into its own document
+
 The granite network upgrade includes a routine upgrade
 of all L2 predeploys using system transactions. The exact
 and upgrade transactions themselves will be finalized in
 the future. This ensures that the L1 and L2 pairs of contracts
 (CrossDomainMessengers and StandardBridges) don't drift too
 far in their versioning.
+
+## Security Considerations
```
mds1, line 103: I want to also flesh out the alternate architecture that we've talked about and discuss a tradeoff between the two. The counter-proposal would be something like:
- Make the L1 ProxyAmin, and the L1 ProxyAdminOwner, [discoverable onchain](https://github.com/ethereum-optimism/client-pod/issues/699).
- Push this to L2 via an L1Block deposit transaction.
- The L2ProxyAdmin has two owners:
    1. `AddressAliasHelper.applyL1ToL2Alias(L1Block.l1ProxyAdminOwner())`.
    2. The `0xDeaD...` address.
- This enables two ways of upgrading system predeploys (*"system predeploys" might mean "system predeploys + collective governed" using the terms in this PR, but those details are not crucial to this comment's discussion*):
    - **Approach 1**: A hard fork inserts a deposit transaction from `0xDeaD...`. This is similar to the approach described in this spec. We use `0xDeaD...` instead of the aliased L1 ProxyAdminOwner address so the inserted transaction works on any OP Stack fork, not just standard chains with the same L1 ProxyAdminOwner
    - **Approach 2**: The L1ProxyAdmin owner can send a regular deposit transaction to upgrade predeploys.
        - To upgrade 1 predeploy on 1 chain, send 1 deposit transaction.
        - To upgrade `n` predeploys on `m` chains, send `n*m` deposit transactions.
        - This lets predeploys be upgraded atomically across L2s, up to the block gas limit.
        - Since L1ProxyAdmin owner is usually a safe, we can write a helper contract that it can call to to help enforce identical upgrades of the same predeploy set accross chains.

Support for Approach 2 has two main benefits which this PR's proposal does not easily support:
1. If we ever have an incident response scenario where we must upgrade L2 predeploys, it will be a very big effort to quietly coordinate a hardfork among so many actors (each standard chain, RPC providers, other infra providers, etc). With the counter-proposal above, it's significantly easier to coordinate incident response as it can be done with a much smaller set of actors, i.e. the L1ProxyAdmin Owner signers (for standard chains, the 2/2 between the SC and Foundation) 
3. Some L2 predeploys need to be initialized with chain-specific values (such L2CrossDomainMessenger needing the address of the L1CrossDomainMessenger). Having to hardcode these addresses in fork code feels clunkier than passing them in as calldata to an L1 transaction, and having initialization params in the fork code doesn't scale well for chains that are not included in that fork logic (e.g. OP Stack forks)

What are downsides of this alternate approach?  Here's what I can think of:
- Having two ways to upgrade contracts is not conventional, and may be confusing. However, because both are ultimately deposit transactions, they should look the same onchain (i.e. both emit an `Upgrade` event), so I think this should be ok.
- A malicious L1 ProxyAdminOwner can quietly upgrade predeploys to malicious code, whereas a hardfork is noisier and more detectable. This seems ok to me, the L1 ProxyAdminOwner is an extremely powerful role, and they can already insert malicious code on L1 by upgrading those implementations.
- Maybe others?

Very much open to discussion on these two ideas (and any others) to make sure we pick the best approach.

mds1 (commented):
Thank you for writing this up! I left a few small comments, and one larger comment discussing what an alternate architecture might look like

specs/protocol/granite/predeploys.md
blmalone: Might be useful to have the possible range in here too. 

```
0x4200000000000000000000000000000000000000 - 0x4200000000000000000000000000000000000800
```

or maybe it's `0x42000000000000000000000000000000000007FF` ?

specs/protocol/granite/predeploys.md
mds1: As I reread your spec, I think it actually may be more similar to my approach then I initially thought, but I was confused by questions like this: https://github.com/ethereum-optimism/specs/pull/194/files#r1604144867

So in your proposal, IIUC: we split up predeploys into a few categories. Some can only be upgraded via hard fork, but others are upgraded via L1 deposit transactions which can be sent by a privileged user?

Then the (simplified) diff between our proposals then is:
- In yours, a given predeploy can be upgraded by hard fork OR by deposit transaction
- In mine, a given predeploy can be upgraded by hard fork AND by deposit transaction

Is that correct?

tynes: Appreciate you taking the time to read through the spec and leave thoughtful comments

> Is that correct?

Yes this is correct. I lean towards making it an explicit OR so that a norm is created around predeploy upgrades being explicit hardforks that require node software upgrades.

I hear your point regarding possible hacks, in the OR scenario it is much more difficult to "stealth" upgrade the code to a non vulnerable version of the code. There exists some exploits that can only be fixed this way but its not immediately obvious to me what an exploit like this would be in practice. No tokens are locked on L2, perhaps if there was an exploit that unlocked the withdrawing (locked) ether? We would need to pause withdrawals on L1 in that situation to prevent the attack and then perform a contract upgrade.

I lean towards having pause buttons at the right level of abstraction rather than instant upgrades of the predeploys. For interop, having the superchain config or the system config be able to pause L2 to L2 messaging makes sense. That is where a vulnerability would be, like an infinite mint.

> Some L2 predeploys need to be initialized with chain-specific values (such L2CrossDomainMessenger needing the address of the L1CrossDomainMessenger). Having to hardcode these addresses in fork code feels clunkier than passing them in as calldata to an L1 transaction, and having initialization params in the fork code doesn't scale well for chains that are not included in that fork logic (e.g. OP Stack forks)

Maybe i am missing something, but I did try to avoid any sort of "different logic in hardfork" sort of design. Ideally the same hardfork applies on all chains and the network specific information is pushed into L2 from L1. This is why the concept of proxy admin owner goes away and it becomes an account with no know private key, so the L1 contracts can be configured with the keys that differ between networks

specs/protocol/granite/predeploys.md
tynes: This explicitly exists to address your desire for an AND approach

tynes: Perhaps not obvious from the spec, but the `op-node` can freely assume the identity of any account when crafting deposit transactions since they have no signature. The collective governed userland predeploys are meant to be upgraded by the foundation but could also be upgraded by `op-node` since it has ultimate sudo power. I would prefer to set the precedent that the `op-node` doesn't do that but it could in theory

specs/protocol/granite/predeploys.md
tynes: I would prefer to define all existing predeploys as chain governed but would be open for discussion on it. I believe that the `L2ToL1MessagePasser` MUST be chain governed at the least, as well as upcoming interop factory contracts to guarantee solvency of superchain erc20 tokens

If there is a strong argument for particular predeploys being categorized in a specific way, I am open to hearing the argument, but I think given the types described in this document, we have a good framework for categorizing them

specs/protocol/granite/predeploys.md
tynes: I am not super opinionated on naming, open to use whatever names make sense if you all have strong opinions

specs/protocol/granite/predeploys.md
tynes: System ownership means that only the `op-node` can create a deposit tx that can upgrade the predeploy

specs/protocol/granite/predeploys.md
tynes: The `SuperchainConfig` becomes `Ownable` and then has a function like:

```
function upgrade(address _portal, address _proxy, address _implementation) onlyOwner
```

This may want to be `upgradeToAndCall`, need to think through the abstraction a bit more. It would call the `OptimismPortal` function:

```
function upgrade(address _proxy, address _implementation) onlySuperchainConfig
```

which then emits a `TransactionDeposited` event that calls the `L2ProxyAdmin`

The `SystemConfig` would also gain a function similar to the `SuperchainConfig.upgrade`.

Its also possible to gate `OptimismPortal.upgrade` with `onlySystemConfigOrSuperchainConfig` and then that modifies the `from` in the `TransactionDeposited` event

Open for discussion on the exact ABI naming in the contracts, just want something simple.

specs/protocol/granite/predeploys.md
tynes: This is noted a bit in https://github.com/ethereum-optimism/specs/pull/194#discussion_r1605529214 but the idea is that the `SuperchainConfig` is now `Ownable` and in practice it would be the foundation owning it

specs/protocol/granite/predeploys.md
mds1: > perhaps if there was an exploit that unlocked the withdrawing (locked) ether? We would need to pause withdrawals on L1 in that situation to prevent the attack and then perform a contract upgrade.
>
> I lean towards having pause buttons at the right level of abstraction rather than instant upgrades of the predeploys. For interop, having the superchain config or the system config be able to pause L2 to L2 messaging makes sense. That is where a vulnerability would be, like an infinite mint.

That's a good example of an exploit that would fall into this category. In general I agree with having pauses at the right levels of abstraction, and think we should do that regardless of the decision here.

> Maybe i am missing something, but I did try to avoid any sort of "different logic in hardfork" sort of design. Ideally the same hardfork applies on all chains and the network specific information is pushed into L2 from L1. This is why the concept of proxy admin owner goes away and it becomes an account with no know private key, so the L1 contracts can be configured with the keys that differ between networks

Sounds like we agree on the goal here, but I don't yet fully understand how your solution achieves it—can you use the L2xDM initializer (which needs the L1xDM address, different per chain) as an example of what the fork transaction would look like? I think you're saying all L1 values that are needed by L2 initializers would also become part of L1Block so they can get pushed and read from storage on L2, instead of being passed as calldata in the fork upgrade transaction?

tynes: Ah good call, I was not properly handling the `initialize` with network specific config case. Perhaps a "L2 MCP" could solve it, altho it would increase the scope a bit. Just going to explore what it looks like now as an exercise.

We have the `L1Block` contract mirror the `SystemConfig` on L1 and hold network specific config. We gate updates to these config values based on the same actors defined in this PR. We move away from `initialize` patterns in the L2 predeploys (its more complicated when it requires execution to create the L2 genesis).

We put these network specific config in unstructured storage, so its easy to automate checks to assert what the values are. The L2xDM then calls to the `L1Block` contract to get the values.

Designing an upgrade to such a system would be a bit of a pain, I think it would need to be done in 3 steps, one where the L2 `L1Block` ABI is introduced but not used and then the L1 `SystemConfig` ABI is introduced and called (resulting in L2 storage being set, but not used) then upgrades to the L2 predploys so that they start referencing the L2 storage in the `L1Block` contract.

This generalizes the pattern in custom gas token and creates a standard API for managing config on L2. This might be able to get us to a world where the genesis file for each chain is the same, given what hardforks they have activated at genesis.

tynes:
This is also known as "MCP L2"

tynes:
Closing this as it is out of date

PR #195, opened 2024-05-17
Gas Limit - No higher than 200,000,000 gas

blmalone:
Referencing @clabby's PR: https://github.com/ethereum-optimism/specs/pull/186/files


clabby (approved):
Great spot - thanks!

geoknee:
We should make the change here as well https://github.com/ethereum-optimism/superchain-registry/blob/d080a69a08edccb3910389c331bd05d099f4693b/validation/standard/standard-config-mainnet.toml#L26-L27

sebastianst (approved):

PR #200, opened 2024-05-20
Remove resource config from standard config spec

blmalone:
Should target post-L2Beat-Stage 1. 

Resource config is now going to be hardcoded and updatable by L1PAO.

> remove the setResourceConfig setter to prevent the [System Config Owner](https://docs.optimism.io/chain/security/privileged-roles#system-config-owner), from misconfiguring the resource config and preventing force inclusions from L1 (e.g. by setting the maximum deposit gas too low). The resource config can now only be changed by upgrading the SystemConfig contract which is controlled by the [L1 Proxy Admin](https://docs.optimism.io/chain/security/privileged-roles#l1-proxy-admin) (a 2-of-2 multisig with the Optimism Foundation and Security Council).

### DIFF
--- a/specs/protocol/configurability.md
+++ b/specs/protocol/configurability.md
@@ -5,7 +5,6 @@
 **Table of Contents**
 
 - [Consensus Parameters](#consensus-parameters)
-  - [Resource Config](#resource-config)
 - [Policy Parameters](#policy-parameters)
 - [Admin Roles](#admin-roles)
 - [Service Roles](#service-roles)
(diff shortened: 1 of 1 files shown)


tynes (approved):
I think we have the default values elsewhere in the specs and eventually I want to remove the resource config completely via [deposit queue](https://github.com/ethereum-optimism/specs/pull/82)

tynes:
Looks like this needs a rebase

PR #203, opened 2024-05-22
Adding returnedMessages primitive into L2ToL2CrossDomainMessenger

skeletor-spaceman:
**Description**

Adds a way for the `L2ToL2CrossDomainMessenger` to "send-back" failed messages.
These failed messages are stored on the origin domain's `L2ToL2CrossDomainMessenger`.
This enables smart contracts to handle "failed" messages.
e.g. a future `L2toL2StandardBridge` will be able to safely recover stuck funds by validating that a message was indeed NOT successful on the other domain.

For this proposal, we've also prepared a detailed Notion document, found [here](https://defi-wonderland.notion.site/L2-L2-Stuck-Messages-71a4cdb5a06c48fb94d0b98d89c46a03), that goes through the rationale and need of the `returnedMessages` mapping and accompanying functionality and required modifications.
We've also prepared an example of the [technical PR](https://github.com/defi-wonderland/optimism/pull/2)

We believe this would be an important primitive for Interop and the canonical bridge.

Would love to hear your thoughts and any feedback you might have on this proposal :)

**Tests**

N/A

**Additional context**

https://defi-wonderland.notion.site/L2-L2-Stuck-Messages-71a4cdb5a06c48fb94d0b98d89c46a03
previous PR: https://github.com/ethereum-optimism/specs/pull/184

**Metadata**

N/A

This proposal was developed as an internal research project at @defi-wonderland lead by @0xng.


tynes:
Just to be clear that the `StandardBridge` and the `L2ToL2CrossDomainMessenger` are in different bridge systems. The L1<>L2 bridging contracts are strictly different than the L2<>L2 bridging contracts, so a change to the `L2ToL2CrossDomainMessenger` wouldn't impact the L1<>L2 `StandardBridge` flow

specs/interop/predeploys.md
```diff
@@ -254,6 +279,62 @@ Note that the `relayMessage` function is `payable` to enable relayers to earn in
...
 getters.
 
+#### Sending Back Messages Hashes
+
+When sending back a message that failed to be relayed on the destination chain
+to the source chain, it's crucial to ensure the message can only be sent back
+to the `L2ToL2CrossDomainMessenger` contract in its source chain.
+
+```solidity
+function sendMessageHashBack(uint256 _messageSource, uint256 _nonce, address _sender, address _target, bytes calldata _message) external nonReentrant {
```
tynes: So we need the preimages instead of just being able to pass the digest so that we can check that its not sending to self chain?

The invariant is held within

```solidity
function sendMessage(uint256 _destination, address _target, bytes calldata _message) external {
    require(_destination != block.chainid);
    ...
```

So i think we should be able to just pass the message hash and can save on calldata and on chain hashing

skeletor-spaceman:
> Just to be clear that the `StandardBridge` and the `L2ToL2CrossDomainMessenger` are in different bridge systems. 

oh nice catch, since we don't yet have a `L2toL2StandardBridge`, I mentioned it as an example, but I'll add a clarification as to not get things confused with the L1<>L2 flow.

specs/interop/predeploys.md
```diff
@@ -254,6 +279,62 @@ Note that the `relayMessage` function is `payable` to enable relayers to earn in
...
+    if (_source == block.chainid) revert MessageSourceSameChain();
+
+    bytes32 messageHash = keccak256(abi.encode(block.chainid, _messageSource, _nonce, _sender, _target, _message));
+
+    if (successfulMessages[messageHash]) revert MessageAlreadyRelayed();
+    if (block.timestamp <  failedMessages[messageHash] + RETURN_DELAY) revert DelayHasNotEnsued();
```
tynes: What is the specific purpose of the return delay? Its defined in an invariant but there is no rationale for why it exists. What would happen if it didn't exist? What are the tradeoffs with a larger/smaller value?

specs/interop/predeploys.md
```diff
@@ -159,6 +163,23 @@ as well as domain binding, ie the executing transaction can only be valid on a s
...
+- The sent-back message MUST not be relayable after being sent back
+
+### `receiveMessageHash` Invariants
+
+- Only callable by the `CrossL2Inbox`
+- The message source MUST be `block.chainid`
+- The sender MUST be `address(L2ToL2CrossDomainMessenger)`
+- The `Identifier.origin` MUST be `address(L2ToL2CrossDomainMessenger)`
+- The `Identifier.chainId` MUST be `_destination`
+
```
tynes, line 181: If a message is sent back, we would want an invariant such that it cannot be replayed again?

specs/interop/predeploys.md
```diff
@@ -254,6 +279,62 @@ Note that the `relayMessage` function is `payable` to enable relayers to earn in
...
+    if (CrossL2Inbox(Predeploys.CROSS_L2_INBOX).origin() != Predeploys.L2_TO_L2_CROSS_DOMAIN_MESSENGER) {
+        revert CrossL2InboxOriginNotL2ToL2CrossDomainMessenger();
+    }
+
+    if (CrossL2Inbox(Predeploys.CROSS_L2_INBOX).chainid() != _messageDestination) {
+        revert CrossL2InboxDestinationDoesntMatch();
+    }
+
+    returnedMessages[_messageHash] = block.timestamp;
```
tynes: `returnedMessages` isn't defined with a description - can you describe what is in it and any auth/invariants involved with what defines something that can be set inside of it?

specs/interop/predeploys.md
skeletor-spaceman: return delay is there to allow for replay-ability of the message over that time window.
if it wasn't there, there could be a griefing attack where a malicious actor will execute a message with minimum gas, forcing a revert, and then send that message back to the origin domain.
the current time window has been picked at random, we have not done any research on either the best static or dynamic delay for the system.
do you have any other solution in mind besides a time-window? we believe for the MVP this should work ok-ish.

specs/interop/predeploys.md
tynes: I like the idea of calling this function `revert(bytes32)` or something more minimal

Another thought is that there is no auth on this so anybody can call it on behalf of somebody else, there is a grief here and the `RETURN_DELAY` is the only thing that gives time for replayability

specs/interop/predeploys.md
skeletor-spaceman: yes! this is done by setting the message as successful on the reverted domain `successfulMessages[messageHash] = true`.
we have the invariant: `- The sent-back message MUST not be relayable after being sent back` under `sendMessageHashBack Invariants`

specs/interop/predeploys.md
```diff
@@ -245,7 +266,11 @@ function relayMessage(uint256 _destination, uint256 _source, uint256 _nonce, add
        _calldata: _message
     });
 
-    require(success);
+    if (!success) {
+      if (failedMessages[messageHash] == 0) {
```
tynes: This branch needs to exist so that somebody cannot replay over and over again, making the timestamp in storage go up and up, preventing a rollback. Can you note this?

specs/interop/predeploys.md
```diff
@@ -254,6 +279,62 @@ Note that the `relayMessage` function is `payable` to enable relayers to earn in
...
+```solidity
+function sendMessageHashBack(uint256 _messageSource, uint256 _nonce, address _sender, address _target, bytes calldata _message) external nonReentrant {
+    if (_source == block.chainid) revert MessageSourceSameChain();
+
+    bytes32 messageHash = keccak256(abi.encode(block.chainid, _messageSource, _nonce, _sender, _target, _message));
+
+    if (successfulMessages[messageHash]) revert MessageAlreadyRelayed();
+    if (block.timestamp <  failedMessages[messageHash] + RETURN_DELAY) revert DelayHasNotEnsued();
+
+    returnableMessageHashes[messageHash] = 0;
```
```suggestion
```
I this is outdated and should be removed.

specs/interop/predeploys.md
```suggestion
      // Will set the timestamp only once on the *first* failed execution to protect against extending the time-window and blocking the rollback.
      if (failedMessages[messageHash] == 0) {
```

specs/interop/predeploys.md
0xng: The preimage was required to know in which chain the initiating message originated, so the `source` argument was key here. The reason is that without it someone could send back the digest to the wrong chain. Hashing the preimage gives us a way to know the source corresponds to the resulting digest/`messageHash`. 

However, we agree that `revert(bytes32)` will be a lot cleaner, so we just pushed a change that makes it work by having a `FailedMessage` struct containing the `source` chain id of the failed message, as well as the timestamp in which it first failed to be relayed.

Yes, the reason behind the `RETURN_DELAY` is to give enough time to replay to the users so griefing is not possible immediately. We didn't a good way to add auth to this function without adding a parameter to the encoding of the `messageHash`, or without adding more complexity.

We also added more descriptions and comments in the spec, let us know if you would like them to be ordered/styled different.

specs/interop/predeploys.md
```diff
@@ -254,6 +288,77 @@ Note that the `relayMessage` function is `payable` to enable relayers to earn in
...
+It's also important to ensure only the hashes of messages that were initiated
+in this chain are accepted and that they come from the correct destination
+chain id.
+
+If all checks have been successful, the message has is stored in the
+`returnedMessages` mapping. This enables smart contracts to read from it and
+check whether a message failed or not, and handle the failing case accordingly.
+
+```solidity
+function receiveMessageHash(uint256 _messageSource, uint256 _messageDestination, address _sender, bytes32 _messageHash) external {
```
tynes: I would like to think about the naming here a bit more, just so its easy to explain and think about. There are 2 functions that need to be called in order. First you `revert`, then you `....`.

Some words that come to mind:
- `rollback`
- `revertFrom`
- `revertBy`
- `reverted`

Open to changing the `revert` function name, just would like both of them to be concise and play well together so its easy to understand conceptually what is going on. Perhaps `revert` and `reverted` (which has EVM connotations which isn't ideal) or `rollback` and `rolledback`  ? This implies a present and past tense with the language

tynes: Also, the interface needs to be updated based on the updates to `revert`

specs/interop/predeploys.md
```diff
@@ -254,6 +288,77 @@ Note that the `relayMessage` function is `payable` to enable relayers to earn in
...
+    if (_messageSource != block.chainid) revert IncorrectMessageSource();
```
tynes, line 343: I think its possible to remove some of these invariants since they are enforced on the remote side. We need the `messageSource` check, do we need all of the other ones?

specs/interop/predeploys.md
```diff
@@ -254,6 +288,77 @@ Note that the `relayMessage` function is `payable` to enable relayers to earn in
...
+        revert CrossL2InboxOriginNotL2ToL2CrossDomainMessenger();
+    }
+
+    if (CrossL2Inbox(Predeploys.CROSS_L2_INBOX).chainid() != _messageDestination) {
+        revert CrossL2InboxDestinationDoesntMatch();
+    }
+
+    returnedMessages[_messageHash] = block.timestamp;
+
+    emit MessageHashReceived(_messageHash);
```
tynes: Can we update this event name to match the `revert` or `rollback` naming scheme?

specs/interop/predeploys.md
```diff
@@ -254,6 +288,77 @@ Note that the `relayMessage` function is `payable` to enable relayers to earn in
...
+
+```solidity
+function revert(bytes32 _messageHash) external nonReentrant {
+    if (successfulMessages[messageHash]) revert MessageAlreadyRelayed();
+
+    (uint256 messageTimestamp, uint256 messageSource) = failedMessages[messageHash];
+
+    // Return delay is necessary to avoid griefing attacks by providing users with time to
+    // replay their transactions give this function has no auth
+    if (block.timestamp <  messageTimestamp + RETURN_DELAY) revert DelayHasNotEnsued();
```
tynes: There is a constants table above, could we define the value of `RETURN_DELAY` in it?

specs/interop/predeploys.md
```diff
@@ -245,7 +267,19 @@ function relayMessage(uint256 _destination, uint256 _source, uint256 _nonce, add
...
     });
 
-    require(success);
+    if (success) {
+        successfulMessages[messageHash] = true;
+        delete failedMessages[messageHash];
+        emit RelayedMessage(messageHash);
+    } else {
+        // Will set the timestamp only once on the *first* failed execution to protect against extending
+        // the time-window and blocking the rollback.
```
tynes: Is it ok if we move the inline code comments outside of the code block? Just to optimize the codeblock for less stuff going on and have the human readable text outside of it

tynes:
I am generally in favor of this approach and would be open to getting it merged, I think it can be a powerful primitive. It doesn't solve the problem of sending to a chain not in the dependency set but that might be impossible to solve (?)

Left a set of comments that are generally minor/aesthetic

specs/interop/predeploys.md
skeletor-spaceman: I think both revert and rollback might be a bit misleading, since we are not changing any origin state back.
the function name should convey that a message wont be re-playable anymore, and that it's hash has been sent to it's origin domain.
the receiving domain function name should be easy to derive from the above.
naming is hard x.x

tynes: I agree that naming is hard and I don't mean to nit, just would like cohesive/simple naming. Maybe we could use the interface:

- `expire(bytes32)`
- `expired(bytes32 hash, uint256 chainid)`

If we call the constant `EXPIRY_WINDOW` then it conveys a concept of the failed cross chain messages expiring if they fail and some amount of time passes. You would call `expire` on the remote domain which would then call `expired` on the local domain. I think you need the `chainid` to `require(chainid == block.chainid)` for security but need to double check to be sure

skeletor-spaceman: `expire` is perfect, nice one!

0xng: Agree `expire` is great. `expired` I like as well, but it feels more like a `view` function you can query instead of something that updates storage. `storeExpired` makes more sense to me, but I have no strong opinion on this. And please do nitpick, it has improved the code a lot.

For the security measures of `expired` , I agree with only having `chainId`. I think we can safely remove `sender`. `chainid` seems necessary to me to ensure only messages originated in that chain can be stored there. `destination` not needed due to what you mentioned, that it's checked on `sendMessage`.

Here's how the code would look like, what do you think?

```solidity
  /// @notice Sends a failed message hash back to the source chain if the expiry window has ensued.
  /// @param _expiredHash The hash of the expired message to be returned.
  function expire(bytes32 _expiredHash) external nonReentrant {
      if (successfulMessages[_expiredHash]) revert MessageAlreadyRelayed();

      (uint256 messageTimestamp, uint256 messageSource) = failedMessages[_expiredHash];

      if (block.timestamp <  messageTimestamp + EXPIRY_WINDOW) revert ExpiryWindowHasNotEnsued();

      delete failedMessages[_expiredHash];
      successfulMessages[_expiredHash] = true;

      bytes memory data = abi.encodeCall(
          L2ToL2CrossDomainMessenger.expired,
          (_expiredHash, messageSource)
      );
      emit SentMessage(data);
  }

  /// @notice Receives the expired hash of a message that could not be relayed on the destination chain.
  /// @param _messageSource Chain ID of the original message's source chain. Should be this chain.
  /// @param _expiredHash   Message hash of the message that failed to be relayed on the destination chain.
  function expired(bytes32 _expiredHash, uint256 _messageSource) external {
      if (_messageSource != block.chainid) revert IncorrectMessageSource();
      if (msg.sender != Predeploys.CROSS_L2_INBOX) revert ExpiredMessageCallerNotCrossL2Inbox();

      if (CrossL2Inbox(Predeploys.CROSS_L2_INBOX).origin() != Predeploys.L2_TO_L2_CROSS_DOMAIN_MESSENGER) {
          revert CrossL2InboxOriginNotL2ToL2CrossDomainMessenger();
      }

      expiredMessages[_expiredHash] = block.timestamp;

      emit MessageHashExpired(_expiredHash);
  }
```

tynes: Ah i see what you mean about `expired` seeming like a `view` function, I agree with that. We could make it `sendExpire` + `relayExpire` so its like `sendMessage` and `relayMessage`

0xng: Added the changes with that nomenclature, I think is very clean now. The only thing we are missing is defining the value of `EXPIRY_WINDOW`. I added it to the constants table with a `TBD`. Is this something you may have more insight on? Too large of a value may be bad for UX. Too small of a value may not give enough time for the chain state to allow a message to be successfully relayed.

Also, do you think it makes sense for this to be constant, over it having a setter in case it needs to be adjusted?

specs/interop/predeploys.md
```diff
@@ -254,6 +287,68 @@ Note that the `relayMessage` function is `payable` to enable relayers to earn in
...
+
+```solidity
+function relayExpire(bytes32 _expiredHash, uint256 _messageSource) external {
+    if (_messageSource != block.chainid) revert IncorrectMessageSource();
+    if (msg.sender != Predeploys.CROSS_L2_INBOX) revert ExpiredMessageCallerNotCrossL2Inbox();
+
+    if (CrossL2Inbox(Predeploys.CROSS_L2_INBOX).origin() != Predeploys.L2_TO_L2_CROSS_DOMAIN_MESSENGER) {
+        revert CrossL2InboxOriginNotL2ToL2CrossDomainMessenger();
+    }
+
```
tynes, line 350: We should add a check that it hasn't been expired already by reverting if the `expired[_hash] != 0`

specs/interop/predeploys.md
```diff
@@ -146,6 +150,7 @@ properties about the `_msg`.
 |-------------------|----------------------------------------------|
 | Address           | `0x4200000000000000000000000000000000000023` |
 | `MESSAGE_VERSION` | `uint256(0)`                                 |
+| `EXPIRY_WINDOW`   | `TBD`                                        |
```
tynes: Can we define this as 2 hours? Not sure what the best time should be to be honest, we could always modify it in the future

specs/interop/predeploys.md
tynes: This could result in subtle bugs for app devs building on top

specs/interop/predeploys.md
```diff
@@ -245,7 +269,16 @@ function relayMessage(uint256 _destination, uint256 _source, uint256 _nonce, add
        _calldata: _message
     });
```
tynes, line 272: I don't love how much branching logic this adds, perhaps there could be a way to simplify it, I try to keep nested branching to a minimum but it does express what we want and isn't too bad

0xng: hmm, a possible way of reducing branching is doing something along these lines:

```solidity
if (success) {
   successfulMessages[messageHash] = true;
   delete failedMessages[messageHash];
   emit RelayedMessage(messageHash);
   return;
}

if (failedMessages[messageHash].timestamp == 0) {
   failedMessages[messageHash] = FailedMessage({timestamp: block.timestamp, sourceChainId: _source});
}      
emit FailedRelayedMessage(messageHash);
```

But I will also think if this entire logic can be simplified with a more general approach. What do you think of the code above? If I go full branchless, it reduces clarity/readability as well.

tynes: I am curious about your branchless approach

specs/interop/predeploys.md
tynes: If you could add this check to the invariants that would be great too

specs/interop/predeploys.md
tynes: I do like this logic a bit more, perhaps put the `(!success) { ... return }` case first and have the success case under that, i think that will further improve readability

0xng: Added. My memory failed with the branchless approach. I remembered you could simplify branches into mathematical conditions of the sort `a * (condition for a) + b * (condition for b) + ....` but this doesn't work here due to the nature of the operations

tynes (approved):

PR #213, opened 2024-05-29
plasma mode: Properly spec finalization

trianglesphere:
**Description**

Fix the finalization description. See also the design doc for how it should work: https://github.com/ethereum-optimism/design-docs/blob/main/protocol/plasma-mode.md#finalization

### DIFF
--- a/specs/experimental/plasma.md
+++ b/specs/experimental/plasma.md
@@ -281,17 +281,14 @@ and users must use the “finalized” label for a guarantee that their state ca
 With Plasma mode on, the engine queue does receives finality signals from the L1 RPC AND
 from the DA manager that keeps track of challenges.
 The DA manager maintains an internal state of all the input commitments in the current `challengeWindow`
-as they are validated by the derivation pipeline. In addition, it filters challenge events to calculate
-when commitments are challenged/resolved and elect the next finalized L1 block such that:
+as they are validated by the derivation pipeline.
 
-```python
-if active_challenges_count > 0
-    challenge = findOldestActiveChallenge(active_challenges)
-    finality_delay = (challenge.block_number - challenge.commitment_block_number) + resolve_window + 1
-    l1_finalized_block_number = min(latest_l1_block_number - finality_delay, finalized_l1_block_number)
-else 
-    l1_finalized_block_number = min(latest_l1_block_number - challenge_window, finalized_l1_block_number)
-```
+The L2 chain can be marked as finalized when the L1 block in which commitments can no longer be invalidated
+becomes finalized. Previously L2 blocks are fully derived from an L1 block when the batch is fully available.
+Plasma commitments are only partially derived when the commitment is found in a L1 block ("plasma safe"), and
+the become fully derived when that commitment can no longer be invalidated ("safe"). Then finalization can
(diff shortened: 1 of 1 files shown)


specs/experimental/plasma.md
```diff
@@ -281,17 +281,14 @@ and users must use the “finalized” label for a guarantee that their state ca
...
-    l1_finalized_block_number = min(latest_l1_block_number - finality_delay, finalized_l1_block_number)
-else 
-    l1_finalized_block_number = min(latest_l1_block_number - challenge_window, finalized_l1_block_number)
-```
+as they are validated by the derivation pipeline.
+
+The L2 chain can be marked as finalized when the L1 block in which commitments can no longer be invalidated
+becomes finalized. Previously L2 blocks are fully derived from an L1 block when the batch is fully available.
+Plasma commitments are only partially derived when the commitment is found in a L1 block ("plasma safe"), and
+the become fully derived when that commitment can no longer be invalidated ("safe"). Then finalization can
```
ajsutton:
```suggestion
then become fully derived when that commitment can no longer be invalidated ("safe"). Then finalization can
```

specs/experimental/plasma.md
```diff
@@ -281,17 +281,14 @@ and users must use the “finalized” label for a guarantee that their state ca
...
+proceed as normal: when the L1 block than an L2 block is fully derived from becomes finalized, the L2 block
```
ajsutton:
```suggestion
proceed as normal: when the L1 block that an L2 block is fully derived from becomes finalized, the L2 block
```

specs/experimental/plasma.md
```diff
@@ -281,17 +281,14 @@ and users must use the “finalized” label for a guarantee that their state ca
...
-    finality_delay = (challenge.block_number - challenge.commitment_block_number) + resolve_window + 1
-    l1_finalized_block_number = min(latest_l1_block_number - finality_delay, finalized_l1_block_number)
-else 
-    l1_finalized_block_number = min(latest_l1_block_number - challenge_window, finalized_l1_block_number)
-```
+as they are validated by the derivation pipeline.
+
+The L2 chain can be marked as finalized when the L1 block in which commitments can no longer be invalidated
+becomes finalized. Previously L2 blocks are fully derived from an L1 block when the batch is fully available.
+Plasma commitments are only partially derived when the commitment is found in a L1 block ("plasma safe"), and
```
ajsutton: Is partially derived the right term here or would it be better to say they're optimistically derived? ie the full block and state is derived, it just hasn't completed all the validation checks yet so is still optimistic.

ajsutton (approved):
Couple of typos and a possible wording change but feel free to ignore that.

specs/experimental/plasma.md
trianglesphere: Yea, optimistically derived is good term. Realistically we need to introduce "plasma safe" and we're marking blocks as actual "safe" too early right now.

trianglesphere: Though I guess usually if we have the data, we should be able to win any challenge made against the commitment so it's pretty close to safe if we are also running a challenger agent + have stored the data somewhere.

PR #214, opened 2024-05-29
feat/governance: create gov-token

0xfuturistic:
create the specs for the Governance Token per https://github.com/voteagora/governor/blob/main/src/lib/OptimismToken.sol#L1869


specs/governance/gov-token.md
```diff
@@ -0,0 +1,93 @@
...
+
+The specific implementation details of the voting and delegation logic are inherited from the `ERC20Votes` contract. The
+`GovernanceToken` contract focuses on integrating this functionality with the ERC20 token standard and adding the minting
+capability.
+
+### Token Minting
+
+The contract MUST have a `mint` function with external visibility that allows the contract owner to mint new tokens to an
+arbitrary address. This function MUST only be called by the contract owner, as enforced by the `onlyOwner` modifier inherited
+from the `Ownable` contract. When tokens are minted, the voting power of the recipient address MUST be updated accordingly.
```
tynes: It could be good to state that the token is owned by the `MintManager` but we don't need to go into details about the implementation of that contract here

specs/governance/gov-token.md
```diff
@@ -0,0 +1,93 @@
...
+### Checkpoints
+
+The contract keeps a history of each account's voting power using checkpoints.
+Checkpoints are recorded whenever tokens are transferred, minted, or burned, and when delegation occurs.
+Each checkpoint represents the voting power of an address at a specific block number.
+The contract MUST maintain a mapping of addresses to their checkpoint history. Additionally, the contract MUST
+provide functions `numCheckpoints` and `checkpoints` to allow retrieving the number of checkpoints for an account
+and the details of a specific checkpoint.
+
+### Overrides
```
tynes: I think that you can delete this section as its very implementation specific and not really important about the way that it works. You could imagine its written in a different language that doesn't have overrides

specs/governance/gov-token.md
```diff
@@ -0,0 +1,93 @@
...
+- `_afterTokenTransfer`: Called after each token transfer to update the voting power. Concretely, the contract uses
+  the `ERC20Votes` implementation to override the `_afterTokenTransfer` function.
+- `_mint`: Called when new tokens are minted to update the voting power and total supply. The contract uses the
+  `ERC20Votes` implementation to override the `_mint` function.
+- `_burn`: Called when tokens are burned to update the voting power and total supply. The contract uses the `ERC20Votes`
+  implementation to override the `_burn` function.
+
+### Supply Cap
+
+The total token supply is capped to `2^208^ - 1` to prevent overflow risks in the voting system.
```
tynes: We are on an older version of the oz contracts before they added custom errors

specs/governance/gov-token.md
```diff
@@ -0,0 +1,93 @@
...
+### Voting Power
+
+Each token corresponds to one unit of voting power.
+By default, token balance does not account for voting power. To have their voting power counted, token holders MUST delegate
+their voting power to an address (can be their own address).
+The contract MUST offer public accessors for quering voting power, as outlined below.
+
+#### Queries
+
+- The `getVotes` function MUST return the current voting power of an address.
```
tynes: It could be good to use syntax like `getVotes()(uint256)` to denote the return type

specs/governance/gov-token.md
```diff
@@ -0,0 +1,93 @@
...
+arbitrary address. This function MUST only be called by the contract owner, as enforced by the `onlyOwner` modifier inherited
+from the `Ownable` contract. When tokens are minted, the voting power of the recipient address MUST be updated accordingly.
+
+### Token Burning
+
+The contract MUST allow token holders to burn their own tokens using the inherited `burn` function from `ERC20Burnable`.
+When tokens are burned, the total supply and the holder's voting power MUST be reduced accordingly.
+
+### Voting Power
+
```
tynes, line 73: What happens to voting power when a user transfers their tokens that they delegated?

specs/governance/gov-token.md
```diff
@@ -0,0 +1,93 @@
...
+### Token Minting
+
+The contract MUST have a `mint` function with external visibility that allows the contract owner to mint new tokens to an
+arbitrary address. This function MUST only be called by the contract owner, as enforced by the `onlyOwner` modifier inherited
+from the `Ownable` contract. When tokens are minted, the voting power of the recipient address MUST be updated accordingly.
+
+### Token Burning
+
+The contract MUST allow token holders to burn their own tokens using the inherited `burn` function from `ERC20Burnable`.
```
tynes: Is `burn(uint256)` part of the public ABI?

specs/governance/gov-token.md
```diff
@@ -0,0 +1,93 @@
...
+  - [Delegation](#delegation)
+  - [Checkpoints](#checkpoints)
+  - [Overrides](#overrides)
+  - [Supply Cap](#supply-cap)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+| Constant          | Value                                        |
```
tynes: nit: `Constants`

tynes:
Can you double check this against the version of oz that is used in the monorepo? Its commit hash `ecd2ca2c`

specs/governance/gov-token.md
```diff
@@ -0,0 +1,93 @@
...
+
+#### Queries
+
+- The `getVotes` function MUST return the current voting power of an address.
+- The `getPastVotes` function MUST allow querying the voting power of an address at a specific block number in the past.
+- The `getPastTotalSupply` function MUST return the total voting power at a specific block number in the past.
+
+### Delegation
+
+Vote power can be delegated either by calling the `delegate` function directly or by providing a signature to be used
```
tynes: Its not clear how to call `delegate` here. What arguments does it take?

specs/governance/gov-token.md
```diff
@@ -0,0 +1,93 @@
...
+
+Delegation is recorded in checkpoints, which store the voting power of each address at specific points in time.
+
+### Checkpoints
+
+The contract keeps a history of each account's voting power using checkpoints.
+Checkpoints are recorded whenever tokens are transferred, minted, or burned, and when delegation occurs.
+Each checkpoint represents the voting power of an address at a specific block number.
+The contract MUST maintain a mapping of addresses to their checkpoint history. Additionally, the contract MUST
+provide functions `numCheckpoints` and `checkpoints` to allow retrieving the number of checkpoints for an account
```
tynes: Its not clear what a checkpoint looks like in practice. Is there a checkpoint type? What do these functions accept and return?

specs/governance/gov-token.md
0xfuturistic: I just checked and our current version of oz contracts seems to include custom errors, at least in `Governor`. could you elaborate more?

specs/governance/gov-token.md
0xfuturistic: yes, because the contract inherits form `ERC20Burnable`, which is where `burn(uint256)` is defined: https://github.com/voteagora/governor/blob/main/src/lib/OptimismToken.sol#L550-L552

specs/governance/gov-token.md
0xfuturistic: it's unaffected. the tokens' voting power becomes independent of the originator in that case

0xfuturistic: still, i added more detail

specs/governance/gov-token.md
```diff
@@ -0,0 +1,101 @@
...
+
+```solidity
+mapping(address => Checkpoint[]) private _checkpoints;
+```
+
+An account's checkpoints' `fromBlock` values are strictly increasing, which means checkpoints at the same block number get
+overwritten by the new one, ensuring there's always at most one checkpoint per block number. This is enforced by the `_writeCheckpoint`
+function.
+
+Additionally, the contract MUST provide functions `numCheckpoints` and `checkpoints` to allow retrieving the number of
```
cairoeth: which contract does this line refer to? `GovernanceToken` doesn't provide `numCheckpoints` and `checkpoints` because they are already inherited from `ERC20Votes`

specs/governance/gov-token.md
```diff
@@ -0,0 +1,101 @@
...
+### Token Minting
+
+The contract MUST have a `mint` function with external visibility that allows the contract owner to mint new tokens to an
+arbitrary address. This function MUST only be called by the contract owner, the `MintManager`, as enforced by the
+`onlyOwner` modifier inherited from the `Ownable` contract. When tokens are minted, the voting power of the recipient
+address MUST be updated accordingly.
+
+### Token Burning
+
+The contract MUST allow token holders to burn their own tokens using the inherited `burn` function from `ERC20Burnable`.
```
cairoeth: `burnFrom(address,uint256)` can also be used for token burning, per https://github.com/voteagora/governor/blob/main/src/lib/OptimismToken.sol#L565C14-L569

specs/governance/gov-token.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Delegation
+
+Vote power can be delegated either by calling the `delegate(address)` function directly (to delegate as the `msg.sender`)
+or by providing a signature to be used with `function delegateBySig(address, uint256, uint256, uint8, bytes32, bytes32)`,
+as inherited from `ERC20Votes`.
+
+The delegation is recorded in a checkpoint. When a token holder delegates their voting power, the delegated address receives
+the voting power corresponding to the token holder's balance. These tokens become independent of the user's balance, so their
+transfers do not affect the voting power of the delegated address. The delegated address can further transfer these tokens,
```
cairoeth: `transfers do not affect the voting power of the delegated address` afaik this is not correct per https://github.com/voteagora/governor/blob/main/src/lib/OptimismToken.sol#L1705-L1718

example: if Alice delegates to Bob, and then Alice transfers tokens to Charlie, Charlie now holds the voting power (so Bob has no voting power). If Charlie has an active delegatee, then after the token transfer, the voting power will be moved to Charlie's delegatee.

cairoeth: `The delegated address can further transfer these tokens,` if the `delegated address` is referring to the delegatee (user that receives voting power), they can't transfer the tokens because they don't have hold them.

specs/governance/gov-token.md
cairoeth: the current [GovernanceToken](https://github.com/voteagora/governor/blob/main/src/lib/OptimismToken.sol) implementation uses v4.5.0, which doesn't use/include custom errors. op monorepo is using v4.7.3

specs/governance/gov-token.md
```diff
@@ -0,0 +1,101 @@
...
+## Overview
+
+| Constants    | Value                                        |
+|--------------|----------------------------------------------|
+| Address      | `0x4200000000000000000000000000000000000042` |
+| Token name   | `Optimism`                                   |
+| Token symbol | `OP`                                         |
+
+`GovernanceToken` is an [ERC20](https://eips.ethereum.org/EIPS/eip-20) token contract that inherits from `ERC20Burnable`,
+`ERC20Votes`, and `Ownable`. It allows token holders to delegate their voting power to other addresses, enabling a representative
```
cairoeth, line 26: it may be useful to mention somewhere `ERC20Permit`, which is inherited from `ERC20Votes`, and the [`permit` function](https://github.com/voteagora/governor/blob/main/src/lib/OptimismToken.sol#L1126-L1134) as a way of approving/giving allowance.

specs/governance/gov-token.md
cairoeth: related: https://github.com/ethereum-optimism/specs/pull/214#discussion_r1621481742

specs/governance/gov-token.md
0xfuturistic: ok cool. that makes sense

specs/governance/gov-token.md
0xfuturistic: you're right!

kansas1295 (commented):
🔥🔥🔥

specs/governance/gov-token.md
```diff
@@ -0,0 +1,95 @@
...
+
+- The `getVotes()(uint256)` function MUST return the current voting power of an address.
+- The `getPastVotes(address,uint256)(uint256)` function MUST allow querying the voting power of an address at a specific
+  block number in the past.
+- The `getPastTotalSupply(uint256)(uint256)` function MUST return the total voting power at a specific block number in
+  the past.
+
+### Delegation
+
+Vote power can be delegated either by calling the `delegate(address)` function directly (to delegate as the `msg.sender`)
```
tynes, line 89: I am curious how the voting power is computed when there is a change - does it basically look at the latest `Checkpoint` and then add/subtract from that?

tynes (approved):

specs/governance/gov-token.md
0xfuturistic: exactly. this is where it applies the delta onto the weight of the last checkpoint: https://github.com/voteagora/governor/blob/main/src/lib/OptimismToken.sol#L1760

0xfuturistic: I updated the specs with more detail on this

PR #219, opened 2024-05-31
Let CrossDomainMessenger refund the failed & non-replayable message's ETH back

devc32494:
### Purpose

Let `CrossDomainMessenger` send-back the failed & non-replayable message’s ETH to the sender on the source chain. 

### Description

The cross domain messengers provide a higher level API for sending cross domain messages. When it receives a message from a source contract/chain, it triggers `relayMessage` to relay the user message with the ETH being transferred.

When `CrossDomainMessenger.relayMessage` is called, the possible outcomes are:
- (1) the transaction succeeds with the successful relay with `successfulMessages[mshHash] = True`
- (2) the transaction succeeds with the unsuccessful relay with `failedMessages[msgHash] = True` so that it is replayable
- (3) the transaction reverts

The OP’s smart contract suite handles the (1) and (2) cases well. However, for (3) where the message has failed to be relayed and won’t be replayable, any ETH being delivered is stuck forever.

### An example of unsuccessful & non-replayable message

While the best practice is to prevent these unsuccessful & non-replayable messages from being issued from the origin contract and chain, there have been missed cases due to the nature of cross-chain environments where the contexts between the origin and destination chains are disconnected.

The `L2CrossDomainMessenger.sendMessage` encodes the user's message from L2 to L1 as the withdrawal by calling `L2ToL1MessagePasser.initiateWithdrawal`. Then, the message passer relays the message to the corresponding cross domain messenger deployed at the target L1.

While users are expected to send the message via the proxy contract (`0x4200…0007`), a user can send a message from L2 to L1 via the implementation contract without being reverted. Then, the relay of the message on the destination chain will revert due to the failed sanity check at `CrossDomainMessenger.relayMessage`since the`sender` is not the whitelisted system address and`msg.value` is non-zero value.

Let's look into the smart contract code where the sanity check fails in the above case. The if-else branch statement on `_isOtherMessenger()` at `CrossDomainMessenger.relayMessage`. It performs sanity checks: (1) if the `sender` is whitelisted, (2) if it is a valid call for replay. However, as mentioned above, there have been cases when (1) the `sender` is not whitelisted system address (because `sender` is the implementation contract, not the proxy) and (2) the call is not for replay. The aforementioned example falls on it and the revert leads to the ETH getting stuck.

```
function relayMessage(
    uint256 _nonce,
    address _sender,
    address _target,
    uint256 _value,
    uint256 _minGasLimit,
    bytes calldata _message
)
    external
    payable
{
  ...

  if (_isOtherMessenger()) {
      // These properties should always hold when the message is first submitted (as
      // opposed to being replayed).
      assert(msg.value == _value);
      assert(!failedMessages[versionedHash]);
  } else {
      require(msg.value == 0, "CrossDomainMessenger: value must be zero unless message is from a system address");

      require(failedMessages[versionedHash], "CrossDomainMessenger: message cannot be replayed");
  }
  
  ...
}
```

### Proposed Smart Contract Implementation

The if-else branch now considers the unsuccessful & nonreplayable message and falls back into `sendMessage` back to the sender on the source chain.

```
function relayMessage(
    uint256 _nonce,
    address _sender,
    address _target,
    uint256 _value,
    uint256 _minGasLimit,
    bytes calldata _message
)
    external
    payable
{
  ...

  if (_isOtherMessenger()) {
      // These properties should always hold when the message is first submitted (as
      // opposed to being replayed).
      assert(msg.value == _value);
      assert(!failedMessages[versionedHash]);
  } else if (failedMessages[versionedHash]) {
      require(msg.value == 0, "CrossDomainMessenger: value must be zero unless message is from a system address");
  } else {
      // If the message was not from the valid system address nor it is the failed message, just return the fund back
      // fallback to sendMessage call to refund and return shortly after
      _sendMessage(_sender, "", uint32(RELAY_RESERVED_GAS));
      emit RefundedRelayMessage(versionedHash);
      return;
  }
  
  ...
}
```

Here is the link to the PoC implementation PR https://github.com/ethereum-optimism/optimism/pull/10675
The relevant tests will be updated after further discussion!

### DIFF
--- a/specs/protocol/messengers.md
+++ b/specs/protocol/messengers.md
@@ -34,6 +34,7 @@ The base `CrossDomainMessenger` interface is:
 
 ```solidity
 interface CrossDomainMessenger {
+    event RefundedRelayMessage(bytes32 indexed msgHash);
     event FailedRelayedMessage(bytes32 indexed msgHash);
     event RelayedMessage(bytes32 indexed msgHash);
     event SentMessage(address indexed target, address sender, bytes message, uint256 messageNonce, uint256 gasLimit);
@@ -67,8 +68,9 @@ interface CrossDomainMessenger {
 The `sendMessage` function is used to send a cross domain message. To trigger
 the execution on the other side, the `relayMessage` function is called.
 Successful messages have their hash stored in the `successfulMessages` mapping
-while unsuccessful messages have their hash stored in the `failedMessages`
-mapping.
+while unsuccessful but replayable messages have their hash stored in the `failedMessages`
+mapping. For the unsuccessful and non-replayable messages, their whole ETH amounts are
+sent back to the sender on the source chain.
 
 The user experience when sending from L1 to L2 is a bit different than when
 sending a transaction from L2 to L1. When going into L1 from L2, the user does


tynes:
It is possible to replay a failed message by calling `relayMessage` again, the ether that is stuck in the contract can be sent out that way, so I don't completely understand what this is solving for. These contracts are extremely difficult to get right when modifying them, so any changes will require a lengthy review and security process. I believe that the message replayability problem is already solved

seongyun-ko:
> It is possible to replay a failed message by calling `relayMessage` again, the ether that is stuck in the contract can be sent out that way, so I don't completely understand what this is solving for. These contracts are extremely difficult to get right when modifying them, so any changes will require a lengthy review and security process. I believe that the message replayability problem is already solved

Correct, the CrossDomainMessenger can replay the message. However it is possible only when the message's `failedMessages[msgHash]` was properly updated to `True` during its first run. The concerned case here is when `relayMessage` call was reverted during its first run.

When `CrossDomainMessenger.relayMessage` is called, the possible outcomes are:
(1) the transaction succeeds with the successful relay with successfulMessages[mshHash] = True
(2) the transaction succeeds with the unsuccessful relay with failedMessages[msgHash] = True so that it is replayable
(3) the transaction reverts

When it reverts, it does not update `failedMessages` mapping and the message is not replayable, sadly.

devc32494:
> It is possible to replay a failed message by calling `relayMessage` again, the ether that is stuck in the contract can be sent out that way, so I don't completely understand what this is solving for. These contracts are extremely difficult to get right when modifying them, so any changes will require a lengthy review and security process. I believe that the message replayability problem is already solved

Correct, the `CrossDomainMessenger` can replay the message. However it is possible only when the message's `failedMessages[msgHash]` was properly updated to `True` during its first run. The concerned case here is when relayMessage call was reverted during its first run.

When `CrossDomainMessenger.relayMessage` is called, the possible outcomes are:
(1) the transaction succeeds with the successful relay with `successfulMessages[mshHash] = True`
(2) the transaction succeeds with the unsuccessful relay with `failedMessages[msgHash] = True` so that it is replayable
(3) the transaction reverts

When it reverts, it does not update `failedMessages` mapping and the message is not replayable, sadly.

tynes:
> > It is possible to replay a failed message by calling `relayMessage` again, the ether that is stuck in the contract can be sent out that way, so I don't completely understand what this is solving for. These contracts are extremely difficult to get right when modifying them, so any changes will require a lengthy review and security process. I believe that the message replayability problem is already solved
> 
> Correct, the `CrossDomainMessenger` can replay the message. However it is possible only when the message's `failedMessages[msgHash]` was properly updated to `True` during its first run. The concerned case here is when relayMessage call was reverted during its first run.
> 
> When `CrossDomainMessenger.relayMessage` is called, the possible outcomes are: (1) the transaction succeeds with the successful relay with `successfulMessages[mshHash] = True` (2) the transaction succeeds with the unsuccessful relay with `failedMessages[msgHash] = True` so that it is replayable (3) the transaction reverts
> 
> When it reverts, it does not update `failedMessages` mapping and the message is not replayable, sadly.

Ah yes, so there is an invariant such that either `failedMessages` or `successfulMessages` are populated. There should be no cases in which this doesn't happen. If you can produce a foundry based poc that does show this is possible, then it would be covered by our bug bounty.

Under the invariant that it cannot happen, we should not add code to handle the case that it does happen, instead we should focus on preserving the invariant

devc32494:
> > > It is possible to replay a failed message by calling `relayMessage` again, the ether that is stuck in the contract can be sent out that way, so I don't completely understand what this is solving for. These contracts are extremely difficult to get right when modifying them, so any changes will require a lengthy review and security process. I believe that the message replayability problem is already solved
> > 
> > 
> > Correct, the `CrossDomainMessenger` can replay the message. However it is possible only when the message's `failedMessages[msgHash]` was properly updated to `True` during its first run. The concerned case here is when relayMessage call was reverted during its first run.
> > When `CrossDomainMessenger.relayMessage` is called, the possible outcomes are: (1) the transaction succeeds with the successful relay with `successfulMessages[mshHash] = True` (2) the transaction succeeds with the unsuccessful relay with `failedMessages[msgHash] = True` so that it is replayable (3) the transaction reverts
> > When it reverts, it does not update `failedMessages` mapping and the message is not replayable, sadly.
> 
> Ah yes, so there is an invariant such that either `failedMessages` or `successfulMessages` are populated. There should be no cases in which this doesn't happen. If you can produce a foundry based poc that does show this is possible, then it would be covered by our bug bounty.
> 
> Under the invariant that it cannot happen, we should not add code to handle the case that it does happen, instead we should focus on preserving the invariant


yes, this is intended to be protected by the invariant you described. However, the scenario that it does not cover is when users by mistakes interacted with the implementation contract, not the proxy. This (https://sepolia-optimism.etherscan.io/tx/0xf9e3d1dcad305974fd43e19153ff084f328e613ba37a0c10835185fca8419c78#eventlog) is an example `sendMessage` from OP L2 to L1 via the implementation contract of `CrossDomainMessenger`. When these messages arrive L1, the call to `CrossDomainMessenger.relayMessage` will revert making the fund stuck. 

While this is a clear mistake by the user and the best practice is to prevent these unsuccessful & non-replayable messages from being issued from the origin contract and chain (e.g., the txn should revert if it is being sent from the implementation contract), it is worth of noting that there have been missed cases as such due to the nature of cross-chain environments where the contexts between the origin and destination chains are not shared directly.

The proposed solution will allow those fund to get refunded to L2s! 

What do you think about the solution here? Appreciate your inputs and glad to hear any idea on better ways of handling these cases!

devc32494:
Do you have some comments or feedbacks? @tynes 
If there is any suggestion in other directions, I can check them as well

tynes:
> However, the scenario that it does not cover is when users by mistakes interacted with the implementation contract, not the proxy. This (https://sepolia-optimism.etherscan.io/tx/0xf9e3d1dcad305974fd43e19153ff084f328e613ba37a0c10835185fca8419c78#eventlog) is an example sendMessage from OP L2 to L1 via the implementation contract of CrossDomainMessenger. When these messages arrive L1, the call to CrossDomainMessenger.relayMessage will revert making the fund stuck.

I do not think its a good idea to add this complexity to the bridge to handle this case. We do not explicitly write the contracts to handle this kind of user error, this is the same sort of user error as sending eth to the wrong address.

By adding this in retroactively, its bailing out users that may have stuck funds when the responsible thing to do is test ahead of time on a testnet

tynes:
I also have no idea who you are @devc32494, you could be trying to add a subtle backdoor into the bridge. You made a fresh account specific for this PR.

mds1:
> While users are expected to send the message via the proxy contract (`0x4200…0007`), a user can send a message from L2 to L1 via the implementation contract without being reverted. Then, the relay of the message on the destination chain will revert due to the failed sanity check at `CrossDomainMessenger.relayMessage`since the`sender` is not the whitelisted system address and`msg.value` is non-zero value.

A simpler solution to this problem that does not touch critical code paths would be to add an `onlyDelegatecall` modifier to the implementation contracts

tynes:
> > While users are expected to send the message via the proxy contract (`0x4200…0007`), a user can send a message from L2 to L1 via the implementation contract without being reverted. Then, the relay of the message on the destination chain will revert due to the failed sanity check at `CrossDomainMessenger.relayMessage`since the`sender` is not the whitelisted system address and`msg.value` is non-zero value.
> 
> A simpler solution to this problem that does not touch critical code paths would be to add an `onlyDelegatecall` modifier to the implementation contracts

Because the call is async (between domains), I don't think this will work? The funds would still be bricked

devc32494:
> > > While users are expected to send the message via the proxy contract (`0x4200…0007`), a user can send a message from L2 to L1 via the implementation contract without being reverted. Then, the relay of the message on the destination chain will revert due to the failed sanity check at `CrossDomainMessenger.relayMessage`since the`sender` is not the whitelisted system address and`msg.value` is non-zero value.
> > 
> > 
> > A simpler solution to this problem that does not touch critical code paths would be to add an `onlyDelegatecall` modifier to the implementation contracts
> 
> Because the call is async (between domains), I don't think this will work? The funds would still be bricked

I believe the suggestion from @mds1 aims to prevent future issues by disallowing users from interacting with the implementation contract using the onlyDelegateCall modifier.

I appreciate your feedback, @tynes. I understand your concern, and I choose to remain anonymous since the withdrawal process is permission-less.

Regarding the strict testing requirements, I fully understand and agree with your point. However, considering that Optimism is one of the most mature protocols in the field, I think it is worth noting that such issues are bound to occur at even small portions (but with increasing number of cases as more use-cases are added). We can do:
- enhance the protocol to prevent similar issues from happening in the future (as suggested by @mds1).
- develop possible routes to unstuck funds, provided these actions align with the constitution.

I have also reviewed the new spec #203, which added support for returning failed messages. Unfortunately, we cannot utilize this feature as it is on L2-to-L2 message transfers, and our message won't be recorded in the `failedMessage` mapping due to the failed invariant

If you have any feedback on potential directions, we would be happy to explore them!

PR #222, opened 2024-06-04
superchain: add Fjord to list of superchain upgrades

sebastianst:
add optimistic Fjord upgrades


axelKingsley:
Should this just be queued to merge until the gov-vote is done? Probably can move forward before the veto period (since, upon veto we'd just revert this), but we'll have to come in and remove the `[1]` later anyway.

sebastianst:
> Should this just be queued to merge until the gov-vote is done? Probably can move forward before the veto period (since, upon veto we'd just revert this), but we'll have to come in and remove the `[1]` later anyway.

I find it already valuable to mention the optimistic activation dates, given that we pull them into the implementations already.  Will leave the decision to other reviewers @protolambda @tynes

protolambda (approved):

PR #224, opened 2024-06-06
fault-proofs: Move fault proofs specs out of experimental

ajsutton:
**Description**

With fault proofs having been approved via governance, move the specs out of experimental.

Adds redirects so links should keep working.


refcell (approved):
Looks good, but shouldn't we have plurality and call it "fault-proofs" instead of "fault-proof" for future proof stack additions?

ajsutton:
I think strictly we have a single fault proof system.  There are multiple implementations of the FP program, FP VM and in future possibly dispute games, but they ultimately work together as a single fault proof system.

specs/SUMMARY.md
```diff
@@ -36,14 +36,14 @@
...
 - [Experimental]()
-  - [Fault Proof](./experimental/fault-proof/index.md)
-    - [Cannon Fault Proof VM](./experimental/fault-proof/cannon-fault-proof-vm.md)
-    - [Stage One Decentralization](./experimental/fault-proof/stage-one/index.md)
-      - [Dispute Game Interface](./experimental/fault-proof/stage-one/dispute-game-interface.md)
-      - [Fault Dispute Game](./experimental/fault-proof/stage-one/fault-dispute-game.md)
-        - [Honest Challenger](./experimental/fault-proof/stage-one/honest-challenger-fdg.md)
-        - [Bond Incentives](./experimental/fault-proof/stage-one/bond-incentives.md)
-      - [Bridge Integration](./experimental/fault-proof/stage-one/bridge-integration.md)
+  - [Fault Proof](fault-proof/index.md)
```
mbaxter: Looks like this is still nested under `[Experimental]()`?

mbaxter (approved):

specs/SUMMARY.md
ajsutton: Good spot, I didn't even know that list was there (GoLand just updated all the URLs for me).  Fixed now.

mslipper (approved):

sebastianst (approved):

PR #233, opened 2024-06-11
gov-token: implement advanced delegation

### LINKED ISSUE #10863: Governance: Specify Advanced Delegation in GovToken
## Definition of Done

The specs include a spec for partial delegation, this should be possible after all of the parties have agreed on the design-doc in a design review meeting.

0xfuturistic:
this implements the contents of the design doc for advanced delegation: https://github.com/ethereum-optimism/design-docs/blob/091b151428d8c52f683f8c589116a7d055d6246d/governance/advanced-delegation.md

### DIFF
--- a/specs/governance/gov-token.md
+++ b/specs/governance/gov-token.md
@@ -5,7 +5,7 @@
 **Table of Contents**
 
 - [Overview](#overview)
-  - [Checkpoints](#checkpoints)
+  - [Hook-based Integration with Alligator](#hook-based-integration-with-alligator)
   - [Token Minting](#token-minting)
   - [Token Burning](#token-burning)
   - [Voting Power](#voting-power)
@@ -24,72 +24,87 @@
 
 `GovernanceToken` is an [ERC20](https://eips.ethereum.org/EIPS/eip-20) token contract that inherits from `ERC20Burnable`,
 `ERC20Votes`, and `Ownable`. It allows token holders to delegate their voting power to other addresses, enabling a representative
-voting system.
+voting system. The contract integrates with the `Alligator` contract through a hook-based approach to enable advanced delegation
+functionality, including partial delegation and subdelegations.
 
-The specific implementation details of the voting and delegation logic are inherited from the `ERC20Votes` contract. The
-`GovernanceToken` contract focuses on integrating this functionality with `ERC20Burnable` and adding the minting
-capability. EIP-712 permit functionality is also inherited from `ERC20Permit`, which is imported by `ERC20Votes`.
+### Hook-based Integration with Alligator
 
-### Checkpoints
+The `GovernanceToken` contract integrates with the `Alligator` contract through a hook-based approach to enable advanced
+delegation features. The `_afterTokenTransfer` function in the `GovernanceToken` is modified to call the `afterTokenTransfer`
+function in the `Alligator` contract, allowing the `Alligator` to consume the hooks and update its delegation and checkpoint
+mappings accordingly.
 
-`ERC20Votes` uses checkpoints to track voting power history. A checkpoint is a "snapshot" of the voting power of an
-address at a specific block number.
+To ensure that token transfers can continue even if the call to the Alligator fails, a try-catch mechanism is implemented
+in the `_afterTokenTransfer` function. If the call to the Alligator's `afterTokenTransfer` function fails, the token transfer
+will still be completed, and an event will be emitted to indicate the failure. This prevents the token transfer from being
+blocked due to issues with the Alligator contract.
 
 ```solidity
-struct Checkpoint {
-    uint32 fromBlock;
-    uint224 votes;
+function _afterTokenTransfer(address from, address to, uint256 amount) internal override {
+    super._afterTokenTransfer(from, to, amount);
+    try IAlligator(alligatorAddress).afterTokenTransfer(from, to, amount) {
+        // Call successful, no action needed
+    } catch {
+        // Call failed, emit an event to indicate the failure
+        emit AlligatorCallFailed(from, to, amount);
+    }
 }
 ```
 
-Checkpoints are organized as a one-to-one mapping of user addresses to append-only lists of checkpoints. When a
-user is involved in a token transfer, mint, burn, or vote delegation, a new checkpoint is added to their corresponding
-list of checkpoints, reflecting the user's updated voting power at that block number.
+All delegation-related state, including the `delegates`, `checkpoints`, and `numCheckpoints` mappings, is shifted from the
+`GovernanceToken` to the `Alligator` contract. The `Alligator` treats the original checkpoints from the `GovernanceToken`
+as a starting point for its own checkpoints.
 
-```solidity
-mapping(address => Checkpoint[]) private _checkpoints;
-```
-
-For any checkpoint list, the `fromBlock` value of each checkpoint is strictly increasing as a factor of the checkpoint's
-index in the list. Additionally, if a checkpoint is created when there's already a checkpoint at the same block number,
-the new checkpoint replaces the old one.
+Basic delegation functions in the `GovernanceToken`, such as `delegate` and `delegateBySig`, are modified to forward the
+calls to the `Alligator` contract using an advanced delegation rule that mimics basic delegation.
 
 ### Token Minting
 
 `GovernanceToken` MUST have a `mint(address,uint256)` function with external visibility that allows the contract owner
-to mint an arbitrary number of new tokens to an specific address. This function MUST only be called by the contract
+to mint an arbitrary number of new tokens to a specific address. This function MUST only be called by the contract
 owner, the `MintManager`, as enforced by the `onlyOwner` modifier inherited from the `Ownable` contract. When tokens
-are minted, the voting power of the recipient address MUST be updated accordingly. The total token supply is capped to
-`2^208^ - 1` to prevent overflow risks in the voting system. If the total supply exceeds this limit,
-`_mint(address,uint256)`, as inherited from `ERC20Votes`, MUST revert.
+are minted, the voting power of the recipient address MUST be updated accordingly in the `Alligator` contract. The total
+token supply is capped to `2^208^ - 1` to prevent overflow risks in the voting system. If the total supply exceeds this
+limit, `_mint(address,uint256)`, as inherited from `ERC20Votes`, MUST revert.
 
 ### Token Burning
 
 The contract MUST allow token holders to burn their own tokens using the inherited `burn(uint256)` or
 `burnFrom(address,uint256)` functions inherited from `ERC20Burnable`. When tokens are burned, the total supply and the
-holder's voting power MUST be reduced accordingly.
+holder's voting power MUST be reduced accordingly in the `Alligator` contract.
 
 ### Voting Power
 
 Each token corresponds to one unit of voting power.
 By default, token balance does not account for voting power. To have their voting power counted, token holders MUST delegate
 their voting power to an address (can be their own address).
-The contract MUST offer public accessors for quering voting power, as outlined below.
+The contract MUST offer public accessors for querying voting power, as outlined below.
 
 #### Queries
 
-- The `getVotes()(uint256)` function MUST return the current voting power of an address.
+- The `getVotes()(uint256)` function MUST retrieve the current voting power of an address from the `Alligator` contract.
 - The `getPastVotes(address,uint256)(uint256)` function MUST allow querying the voting power of an address at a specific
-  block number in the past.
+  block number in the past from the `Alligator` contract.
 - The `getPastTotalSupply(uint256)(uint256)` function MUST return the total voting power at a specific block number in
-  the past.
+  the past from the `Alligator` contract.
 
 ### Delegation
 
 Vote power can be delegated either by calling the `delegate(address)` function directly (to delegate as the `msg.sender`)
 or by providing a signature to be used with function `delegateBySig(address,uint256,uint256,uint8,bytes32,bytes32)`,
-as inherited from `ERC20Votes`.
+as inherited from `ERC20Votes`. These functions are modified to forward the calls to the `Alligator` contract using an
+advanced delegation rule that mimics basic delegation.
 
-The delegation is recorded in a checkpoint. When a token holder delegates their voting power, the delegated address receives
-the voting power corresponding to the token holder's balance. Under token transfers, the voting power associated with the
-tokens is moved from the delegate of the sender to that of the recipient.
+Advanced delegation features, such as partial delegation and subdelegations, are handled by the `Alligator` contract.
+Partial delegation allows delegators to distribute their voting power among multiple delegatees in a fractional manner,
+while subdelegations enable delegatees to further delegate their received voting power to other delegatees based on predefined
+rules and constraints.
+
+The `Alligator` contract maintains the necessary invariants, such as preventing circular delegation chains, ensuring vote
+weight consistency, and managing checkpoints. It is incorporated as a predeploy of the OP stack to avoid manual deployments
+across the Superchain.
+
+By incorporating the try-catch mechanism in the `_afterTokenTransfer` function, the GovernanceToken contract ensures that
+token transfers can continue smoothly even if the Alligator contract encounters issues. This provides a robust fallback
+mechanism to maintain the functionality of the token while allowing the advanced delegation features to be handled
+separately by the Alligator contract.
--- a/specs/governance/gov-token.md
+++ b/specs/governance/gov-token.md
@@ -63,9 +63,9 @@ calls to the `Alligator` contract using an advanced delegation rule that mimics 
 `GovernanceToken` MUST have a `mint(address,uint256)` function with external visibility that allows the contract owner
 to mint an arbitrary number of new tokens to a specific address. This function MUST only be called by the contract
 owner, the `MintManager`, as enforced by the `onlyOwner` modifier inherited from the `Ownable` contract. When tokens
-are minted, the voting power of the recipient address MUST be updated accordingly in the `Alligator` contract via the transfer hook. The total
-token supply is capped to `2^208^ - 1` to prevent overflow risks in the voting system. If the total supply exceeds this
-limit, `_mint(address,uint256)`, as inherited from `ERC20Votes`, MUST revert.
+are minted, the voting power of the recipient address MUST be updated accordingly in the `Alligator` contract via the 
+transfer hook. The total token supply is capped to `2^208^ - 1` to prevent overflow risks in the voting system. If the
+total supply exceeds this limit, `_mint(address,uint256)`, as inherited from `ERC20Votes`, MUST revert.
 
 ### Token Burning
 
--- a/specs/governance/gov-token.md
+++ b/specs/governance/gov-token.md
@@ -63,7 +63,7 @@ calls to the `Alligator` contract using an advanced delegation rule that mimics 
 `GovernanceToken` MUST have a `mint(address,uint256)` function with external visibility that allows the contract owner
 to mint an arbitrary number of new tokens to a specific address. This function MUST only be called by the contract
 owner, the `MintManager`, as enforced by the `onlyOwner` modifier inherited from the `Ownable` contract. When tokens
-are minted, the voting power of the recipient address MUST be updated accordingly in the `Alligator` contract via the 
+are minted, the voting power of the recipient address MUST be updated accordingly in the `Alligator` contract via the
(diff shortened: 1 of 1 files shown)


specs/governance/gov-token.md
```diff
@@ -24,72 +24,87 @@
...
-struct Checkpoint {
-    uint32 fromBlock;
-    uint224 votes;
+function _afterTokenTransfer(address from, address to, uint256 amount) internal override {
+    super._afterTokenTransfer(from, to, amount);
+    try IAlligator(alligatorAddress).afterTokenTransfer(from, to, amount) {
+        // Call successful, no action needed
+    } catch {
+        // Call failed, emit an event to indicate the failure
+        emit AlligatorCallFailed(from, to, amount);
```
cairoeth: Return bomb safe call:
```suggestion
    bool success;
    bytes memory data = abi.encodeWithSelector(IAlligator.afterTokenTransfer.selector, from, to, amount);
    assembly ("memory-safe") {
        success := call(gas(), alligatorAddress, 0, add(data, 0x20), mload(data), 0, 0)
    }
    
    if (!success) emit AlligatorCallFailed(from, to, amount);
```

specs/governance/gov-token.md
```diff
@@ -24,72 +24,87 @@
...
 ### Token Minting
 
 `GovernanceToken` MUST have a `mint(address,uint256)` function with external visibility that allows the contract owner
-to mint an arbitrary number of new tokens to an specific address. This function MUST only be called by the contract
+to mint an arbitrary number of new tokens to a specific address. This function MUST only be called by the contract
 owner, the `MintManager`, as enforced by the `onlyOwner` modifier inherited from the `Ownable` contract. When tokens
-are minted, the voting power of the recipient address MUST be updated accordingly. The total token supply is capped to
-`2^208^ - 1` to prevent overflow risks in the voting system. If the total supply exceeds this limit,
-`_mint(address,uint256)`, as inherited from `ERC20Votes`, MUST revert.
+are minted, the voting power of the recipient address MUST be updated accordingly in the `Alligator` contract. The total
```
cairoeth: should we specify that the mints and burns are updated in the Alligator via the transfer hooks? (minting transfers from `address(0)`, and burns transfers to `address(0)`)

specs/governance/gov-token.md
```diff
@@ -24,72 +24,87 @@
...
+rules and constraints.
+
+The `Alligator` contract maintains the necessary invariants, such as preventing circular delegation chains, ensuring vote
+weight consistency, and managing checkpoints. It is incorporated as a predeploy of the OP stack to avoid manual deployments
+across the Superchain.
+
+By incorporating the try-catch mechanism in the `_afterTokenTransfer` function, the GovernanceToken contract ensures that
+token transfers can continue smoothly even if the Alligator contract encounters issues. This provides a robust fallback
+mechanism to maintain the functionality of the token while allowing the advanced delegation features to be handled
+separately by the Alligator contract.
```
cairoeth: i think we can remove this as we already have it [here](https://github.com/ethereum-optimism/specs/pull/233/files#diff-e1c4b822c42f7fba380f47d85e0f5636325095de27c085d526bd8bb4429f1dc3R37-R40)

specs/governance/gov-token.md
```diff
@@ -24,72 +24,82 @@
...
 
-The specific implementation details of the voting and delegation logic are inherited from the `ERC20Votes` contract. The
-`GovernanceToken` contract focuses on integrating this functionality with `ERC20Burnable` and adding the minting
-capability. EIP-712 permit functionality is also inherited from `ERC20Permit`, which is imported by `ERC20Votes`.
+### Hook-based Integration with Alligator
 
-### Checkpoints
+The `GovernanceToken` contract integrates with the `Alligator` contract through a hook-based approach to enable advanced
+delegation features. The `_afterTokenTransfer` function in the `GovernanceToken` is modified to call the `afterTokenTransfer`
+function in the `Alligator` contract, allowing the `Alligator` to consume the hooks and update its delegation and checkpoint
```
mds1:
```suggestion
function in the `Alligator` contract, allowing the `Alligator` contract to consume the hooks and update its delegation and checkpoint
```

specs/governance/gov-token.md
```diff
@@ -24,72 +24,82 @@
...
+functionality, including partial delegation and subdelegations.
 
-The specific implementation details of the voting and delegation logic are inherited from the `ERC20Votes` contract. The
-`GovernanceToken` contract focuses on integrating this functionality with `ERC20Burnable` and adding the minting
-capability. EIP-712 permit functionality is also inherited from `ERC20Permit`, which is imported by `ERC20Votes`.
+### Hook-based Integration with Alligator
 
-### Checkpoints
+The `GovernanceToken` contract integrates with the `Alligator` contract through a hook-based approach to enable advanced
+delegation features. The `_afterTokenTransfer` function in the `GovernanceToken` is modified to call the `afterTokenTransfer`
```
mds1: Since this is a spec, can we be more specific than "advanced delegation features" and say specifically what the features are?

specs/governance/gov-token.md
```diff
@@ -24,72 +24,82 @@
...
+function in the `Alligator` contract, allowing the `Alligator` to consume the hooks and update its delegation and checkpoint
+mappings accordingly.
 
-`ERC20Votes` uses checkpoints to track voting power history. A checkpoint is a "snapshot" of the voting power of an
-address at a specific block number.
+To ensure that token transfers can continue even if the call to the Alligator fails, a try-catch mechanism is implemented
+in the `_afterTokenTransfer` function. If the call to the Alligator's `afterTokenTransfer` function fails, the token transfer
+will still be completed, and an event will be emitted to indicate the failure. This prevents the token transfer from being
+blocked due to issues with the Alligator contract.
```
mds1: We should avoid leaking implementation details into the spec. In this case, whether we use a try/catch vs. a low-level call is an implementation detail, and what we actually care about is that `GovernanceToken` can continue execution even if a call to `Alligator` reverts. 

```suggestion
If the call to the Alligator's `afterTokenTransfer` function fails, the token transfer
must still be completed, and an `AlligatorCallFailed` event will be emitted to indicate the failed call. This prevents the token transfer from being
blocked due to issues with the Alligator contract.
```

specs/governance/gov-token.md
```diff
@@ -24,72 +24,82 @@
...
+function _afterTokenTransfer(address from, address to, uint256 amount) internal override {
+    bytes memory data = abi.encodeWithSelector(IAlligator.afterTokenTransfer.selector, from, to, amount);
+    bool success;
+    assembly ("memory-safe") {
+        success := call(gas(), alligatorAddress, 0, add(data, 0x20), mload(data), 0, 0)
+    }
+    
+    if (!success) emit AlligatorCallFailed(from, to, amount);
 }
 ```
```
mds1: In the same spirit of not leaking implementation details I would delete this Solidity block, I don't think it adds value and it makes the spec harder to maintain. I would instead suggest adding a sentence defining the event signature

---

To expand a bit so I don't think we've discussed this much, the rationale for "not leaking implementation details into the spec" is:
- Simplify spec maintenance: By specifying as little as possible to implement the requirements, we make it easier to keep the spec up to date by having less code to maintain. The code in this spec is not tested or audited, and we may end up tweaking it.
- Readability: The less code, the easier it is for a less-technical reader to understand the spec and ensure it's sound.
- Testing: The implementer of the spec has flexibility around the final result, as long as it meets the requirements. When implementing the spec, we write tests for all the requirements in here, but we wouldn't write a test to e.g. ensure we're using assembly since that's an implementation detail and not a requirement

specs/governance/gov-token.md
```diff
@@ -24,72 +24,82 @@
...
 
-Checkpoints are organized as a one-to-one mapping of user addresses to append-only lists of checkpoints. When a
-user is involved in a token transfer, mint, burn, or vote delegation, a new checkpoint is added to their corresponding
-list of checkpoints, reflecting the user's updated voting power at that block number.
+All delegation-related state, including the `delegates`, `checkpoints`, and `numCheckpoints` mappings, is shifted from the
+`GovernanceToken` to the `Alligator` contract. The `Alligator` treats the original checkpoints from the `GovernanceToken`
+as a starting point for its own checkpoints.
```
mds1: > The `Alligator` treats the original checkpoints from the `GovernanceToken` as a starting point for its own checkpoints.

Can we expand on this? It's not clear to me what this looks like, e.g. what happens to all existing delegation. It sounds like a lot of state needs to be migrated from `GovernanceToken` to `Alligator`?

specs/governance/gov-token.md
```diff
@@ -24,72 +24,82 @@
...
 
-```solidity
-mapping(address => Checkpoint[]) private _checkpoints;
-```
-
-For any checkpoint list, the `fromBlock` value of each checkpoint is strictly increasing as a factor of the checkpoint's
-index in the list. Additionally, if a checkpoint is created when there's already a checkpoint at the same block number,
-the new checkpoint replaces the old one.
+Basic delegation functions in the `GovernanceToken`, such as `delegate` and `delegateBySig`, are modified to forward the
+calls to the `Alligator` contract using an advanced delegation rule that mimics basic delegation.
```
mds1: Are there others than `delegate` and `delegateBySig`? If so let's include those, if not we can rephrase to something like
```suggestion
The `delegate` and `delegateBySig` functions in the `GovernanceToken` are modified to forward the
calls to the `Alligator` contract which implements the required delegation logic.
```

specs/governance/gov-token.md
```diff
@@ -24,72 +24,82 @@
...
 
 `GovernanceToken` MUST have a `mint(address,uint256)` function with external visibility that allows the contract owner
-to mint an arbitrary number of new tokens to an specific address. This function MUST only be called by the contract
+to mint an arbitrary number of new tokens to a specific address. This function MUST only be called by the contract
 owner, the `MintManager`, as enforced by the `onlyOwner` modifier inherited from the `Ownable` contract. When tokens
-are minted, the voting power of the recipient address MUST be updated accordingly. The total token supply is capped to
-`2^208^ - 1` to prevent overflow risks in the voting system. If the total supply exceeds this limit,
-`_mint(address,uint256)`, as inherited from `ERC20Votes`, MUST revert.
+are minted, the voting power of the recipient address MUST be updated accordingly in the `Alligator` contract via the
+transfer hook. The total token supply is capped to `2^208^ - 1` to prevent overflow risks in the voting system. If the
```
mds1:
```suggestion
`afterTokenTransfer` hook. The total token supply is capped to `2^208^ - 1` to prevent overflow risks in the voting system. If the
```

specs/governance/gov-token.md
```diff
@@ -24,72 +24,82 @@
...
+total supply exceeds this limit, `_mint(address,uint256)`, as inherited from `ERC20Votes`, MUST revert.
 
 ### Token Burning
 
 The contract MUST allow token holders to burn their own tokens using the inherited `burn(uint256)` or
 `burnFrom(address,uint256)` functions inherited from `ERC20Burnable`. When tokens are burned, the total supply and the
-holder's voting power MUST be reduced accordingly.
+holder's voting power MUST be reduced accordingly in the `Alligator` contract via the transfer hook.
```
mds1:
```suggestion
holder's voting power MUST be reduced accordingly in the `Alligator` contract via the `afterTokenTransfer` hook.
```

specs/governance/gov-token.md
```diff
@@ -24,72 +24,82 @@
...
-  the past.
+  the past from the `Alligator` contract.
 
 ### Delegation
 
 Vote power can be delegated either by calling the `delegate(address)` function directly (to delegate as the `msg.sender`)
 or by providing a signature to be used with function `delegateBySig(address,uint256,uint256,uint8,bytes32,bytes32)`,
-as inherited from `ERC20Votes`.
+as inherited from `ERC20Votes`. These functions are modified to forward the calls to the `Alligator` contract using an
+advanced delegation rule that mimics basic delegation.
```
mds1: Above we also said "using an advanced delegation rule that mimics basic delegation", but that feels vague to me so I've been deleting it in my suggestions. Feel free to rephrase as appropriate!
```suggestion
as inherited from `ERC20Votes`. These functions are modified to forward the calls to the `Alligator` contract which implements the required logic.
```

specs/governance/gov-token.md
```diff
@@ -24,72 +24,82 @@
...
+
+Advanced delegation features, such as partial delegation and subdelegations, are handled by the `Alligator` contract.
```
mds1:
```suggestion
Partial delegation and subdelegations are handled by the `Alligator` contract.
```

specs/governance/gov-token.md
```diff
@@ -24,72 +24,82 @@
...
+Partial delegation allows delegators to distribute their voting power among multiple delegatees in a fractional manner,
+while subdelegations enable delegatees to further delegate their received voting power to other delegatees based on predefined
+rules and constraints.
 
-The delegation is recorded in a checkpoint. When a token holder delegates their voting power, the delegated address receives
-the voting power corresponding to the token holder's balance. Under token transfers, the voting power associated with the
-tokens is moved from the delegate of the sender to that of the recipient.
+The `Alligator` contract maintains the necessary invariants, such as preventing circular delegation chains, ensuring vote
+weight consistency, and managing checkpoints. It is incorporated as a predeploy of the OP stack to avoid manual deployments
+across the Superchain.
```
mds1: Is there going to be another spec for `Alligator`? There should be, since it will become a predeploy. Related, the design doc had a lot of invariants that we should ensure are captured in the spec somewhere.

specs/governance/gov-token.md
```diff
@@ -24,72 +24,74 @@
...
-address at a specific block number.
+If the call to the Alligator's `afterTokenTransfer` function fails, the token transfer must still be completed, and an
+`AlligatorCallFailed` event will be emitted to indicate the failed call.
 
 ```solidity
-struct Checkpoint {
-    uint32 fromBlock;
-    uint224 votes;
-}
+emit AlligatorCallFailed(from, to, amount);
```
mds1: I named the params `transferFrom` and `transferTo` to make it clear it's the accounts involved in the token transfer, instead of being interpreted as the from and to address of the failed call
```suggestion
event AlligatorCallFailed(address indexed transferFrom, address indexed transferTo, uint256 amount);
```

specs/governance/gov-token.md
0xfuturistic: yes, there's going to be another spec for `Alligator`, which are tracked by these issues: https://github.com/ethereum-optimism/protocol-quest/issues/297 & https://github.com/ethereum-optimism/protocol-quest/issues/298

mds1: Perfect, thanks! Will resolve this, since the invariants were all properties of Alligator

specs/governance/gov-token.md
0xfuturistic: good idea with this!

specs/governance/gov-token.md
```diff
@@ -24,72 +24,74 @@
...
-    uint32 fromBlock;
-    uint224 votes;
-}
+event AlligatorCallFailed(address indexed transferFrom, address indexed transferTo, uint256 amount);
 ```
 
-Checkpoints are organized as a one-to-one mapping of user addresses to append-only lists of checkpoints. When a
-user is involved in a token transfer, mint, burn, or vote delegation, a new checkpoint is added to their corresponding
-list of checkpoints, reflecting the user's updated voting power at that block number.
+This prevents the token transfer from being blocked due to issues with the Alligator contract.
```
mds1: What happens to delegations in this case? It feels like this would result in broken invariants. For example, Alice has 100 tokens and delegates them all to Bob. Alice transfers all 100 tokens to Clara, but the call to Alligator fails. Now, Clara has 100 tokens and no voting weight, and Bob has 100 votes he's not entitled to, because Clara should have those votes.

If we are ok with this as currently spec'd, what is the way to recover? Presumably we'd need a method for Clara to try delegating again so she can reclaim her voting weight?

My initial reaction to this problem is that instead the transfer should revert if the call reverts, to ensure the token holders always have voting power. If calls to Alligator are reverting, that is a high severity bug that we would prioritize deploying a fix for

specs/governance/gov-token.md
cairoeth: specified the features/use cases in https://github.com/ethereum-optimism/specs/pull/233/commits/674dc7c8c240d870bb1963d8c9b62cb3113b1e1a.

mds1: That looks great, thank you!

specs/governance/gov-token.md
cairoeth: https://github.com/ethereum-optimism/specs/pull/233/commits/8def8fa0af9746ed8f4f6535edcfb9cb98b8668a. lmk if it's clearer!

specs/governance/gov-token.md
```diff
@@ -24,72 +24,75 @@
...
-user is involved in a token transfer, mint, burn, or vote delegation, a new checkpoint is added to their corresponding
-list of checkpoints, reflecting the user's updated voting power at that block number.
+This prevents the token transfer from being blocked due to issues with the `Alligator` contract.
 
-```solidity
-mapping(address => Checkpoint[]) private _checkpoints;
-```
+All delegation-related state, including the `delegates`, `checkpoints`, and `numCheckpoints` mappings, is gradually
+shifted from the `GovernanceToken` to the `Alligator` contract through transactions that call the `Alligator`'s hook
+(i.e., transfers). In the hook, the `Alligator` should check if the `to` and `from` addresses have been migrated.
```
mds1: major nit cause this is a random thing I know about 😅

here it should be "e.g." instead of "i.e.", since the former means "for example" and is non-comprehensive and the latter means "that is" and denotes equivalence
```suggestion
(e.g. transfers). In the hook, the `Alligator` should check if the `to` and `from` addresses have been migrated.
```

specs/governance/gov-token.md
mds1: I see, so is this all correct?
1. Alligator always check it's own state first. If not empty, use that. If empty, fetch state from GovernanceToken and set that in its own storage
2. The existing data in GovernanceToken storage will not be cleared and will exist forever
3. Getter methods in GovernanceToken will be changed to not reference GovernanceToken storage vars directly, and instead call alligator to ask for the data. Per (1), Alligator may then call back into GovernanceToken storage
4. We need to make sure there are no public vars exposed in GovernanceToken, as this is a potential way users with existing integrations may continue to read stale data. We should think about any others

specs/governance/gov-token.md
cairoeth: yup, good catch!

specs/governance/gov-token.md
cairoeth: yes, correct on all points

mds1: Awesome, thanks. Let's make sure these points are all clear in the spec (maybe it is, I will re-review later)

specs/governance/gov-token.md
0xfuturistic: implemented! feel free to take a look now :)

specs/governance/gov-token.md
cairoeth: clarified point 3 (and in effect point 1) in 42b51914c0e3dcee164c934986fbd2dfff0e6698

mds1: Thank you!

specs/governance/gov-token.md
mds1: Looks great, thank you!

specs/governance/gov-token.md
```diff
@@ -24,72 +24,74 @@
...
-struct Checkpoint {
-    uint32 fromBlock;
-    uint224 votes;
-}
-```
+All delegation-related state, including the `delegates`, `checkpoints`, and `numCheckpoints` mappings, is gradually
+shifted from the `GovernanceToken` to the `Alligator` contract through transactions that call the `Alligator`'s hook
+(e.g. transfers). In the hook, the `Alligator` should check if the `to` and `from` addresses have been migrated.
+If an address hasn't been migrated, the `Alligator` should save a copy of the `GovernanceToken`'s mappings for that
+address in its state. When reading delegation data in the `Alligator` for a delegator that hasn't been migrated,
```
mds1: nit
```suggestion
If an address hasn't been migrated, the `Alligator` should write the data from the `GovernanceToken`'s mappings for that
address to its own state. When reading delegation data in the `Alligator` for a delegator that hasn't been migrated,
```

mds1 (approved):

tynes (approved):

PR #234, opened 2024-06-13
Update typos in bond-incentives.md

michalsidzej:
**Description**

Update small typos in specs.


ajsutton:
The original is correct.  `iff` means if and only if.

PR #237, opened 2024-06-14
fjord: Fix FastLZ cost specs to also scale estimated size

sebastianst:
**Description**

Seems there was an inconsistency around the estimated size being a scaled value.

### DIFF
--- a/specs/fjord/exec-engine.md
+++ b/specs/fjord/exec-engine.md
@@ -24,8 +24,8 @@ The L1 cost is computed as:
 
 ```pseudocode
 l1FeeScaled = baseFeeScalar*l1BaseFee*16 + blobFeeScalar*l1BlobBaseFee
-estimatedSize = max(minTransactionSize, intercept + fastlzCoef*fastlzSize)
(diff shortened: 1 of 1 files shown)


roberto-bayardo:
LGTM, though I worry "scaled" implies it's been scalad with the "scalar" not just scaled to correct the units.

ajsutton (approved):

sebastianst:
> LGTM, though I worry "scaled" implies it's been scalad with the "scalar" not just scaled to correct the units.

@roberto-bayardo I agree; I just followed the pattern for the already existing parameter name `l1FeeScaled` that the `Scaled` suffix means scaled (by 1e6). Open to other suggestions to better distinguish between 1e6 scaling and scaled by the scalar values. I will merge this now so that the spec is at least correct again, and we can think about better wording later.

roberto-bayardo:
OK I thought the "scaled" suffix meant it had been scaled by the scalars, not 1e6 :)

I'll take a crack at clarifying.

PR #247, opened 2024-06-18
governance: create `GovernanceDelegation` specs

### LINKED ISSUE #10848: GovernanceDelegation: Specify Hook Call
# Definition of Done

The testing should ensure to cover the invariants defined in the spec

### LINKED ISSUE #11211: GovToken: Specify `GovernanceDelegation`

0xfuturistic:
this PR adds the specs for the `GovernanceDelegation` contracts, which works in conjunction with the `GovernanceToken` contract and has been described in https://github.com/ethereum-optimism/design-docs/pull/29

### DIFF
--- a/specs/experimental/alligator.md
+++ b/specs/experimental/alligator.md
@@ -20,6 +20,7 @@
     - [`getVotes`](#getvotes)
     - [`getPastVotes`](#getpastvotes)
     - [`getPastTotalSupply`](#getpasttotalsupply)
+    - [`migrated`](#migrated)
   - [Events](#events)
     - [`DelegationCreated`](#delegationcreated)
     - [`DelegationsCreated`](#delegationscreated)
@@ -36,6 +37,7 @@
   - [Partial Delegations](#partial-delegations)
   - [Absolute & Relative Delegations](#absolute--relative-delegations)
   - [Differences](#differences)
+- [Delegation Validation](#delegation-validation)
 - [Security Considerations](#security-considerations)
   - [Dependence on Alligator](#dependence-on-alligator)
   - [Connection with GovernanceToken](#connection-with-governancetoken)
@@ -145,11 +147,13 @@ function _delegate(address _delegator, Delegation[] memory _newDelegations) inte
 
 The `_delegate` function MUST first check that the length of `_newDelegations` does not exceed `MAX_SUBDELEGATIONS`.
 Then, it MUST calculate the voting power of all active delegations of the `_delegator`, and calculate the voting
-power adjustements for the new delegatee set, according to `_newDelegations`. Afterwards, the function MUST update
+power adjustements for the new delegatee set using `_newDelegations`. Afterwards, the function MUST update
 the voting power of the current and new set of delegatees.
 
-The function MUST then sort the new delegatee set in descending order, and store the array the `_delegations` mapping.
-While sorting, the function MUST emit a `DelegateVotesChanged` event for each change in delegatee voting power.
+The function MUST then sort the new delegatee set in descending order, and store the delegations as part of the
+`_delegations` mapping. While sorting, the function MUST check for the validity of the delegations, as specified
+in the [Delegation Validation](#delegation-validation) section. Aditionally, the function MUST emit a
+`DelegateVotesChanged` event for each change in delegatee voting power.
 
 ### Getters
 
@@ -177,8 +181,8 @@ function numCheckpoints(address _account) external view returns (uint32)
 
 #### `delegates`
 
-Retrieves the delegations of a given user address. This function is intended to be used by the
-`GovernanceToken` contract to maximize for backwards compatibility.
+Retrieves the delegations of a given user address, sorted in descending order by voting power. 
+This function is intended to be used by the `GovernanceToken` contract to maximize for backwards compatibility.
 
 ```solidity
 function delegates(address _account) external view returns (Delegation[] memory)
@@ -206,6 +210,14 @@ Retrieves the total supply of the `GovernanceToken` at a given block.
 
 ```solidity
 function getPastTotalSupply(uint256 _blockNumber) external view returns (uint256)
+```
+
+#### `migrated`
+
+Returns the migration status of an account — `True` if the account has been migrated, `False` otherwise.
+
+```solidity
+function migrated(address _account) public view returns (bool)
 ```
 
 ### Events
@@ -391,6 +403,17 @@ flowchart TD
     A3 --> |No, use its state| A2
     A2 --> |Return data| A1
 ```
+
+## Delegation Validation
+
+When applying a new delegation set as part of the [`_delegate`](#_delegate) function, the `Alligator` MUST check that:
+
+- The length of the `_newDelegations` array DOES NOT exceed `MAX_SUBDELEGATIONS`.
+- The sum of `amount` in relative delegations DOES NOT exceed the `DENOMINATOR`, and each `amount` is greater than 0.
+- The sum of `amount` in absolute delegations DOES NOT exceed the total voting power of the `_delegator`.
+- The new delegatee set is sorted in descending order by voting power.
+
+If any of the above conditions are not met, the `_delegate` function MUST revert with a specific error message.
 
 ## Security Considerations
 
--- a/specs/experimental/alligator.md
+++ b/specs/experimental/alligator.md
@@ -19,21 +19,21 @@
     - [`getVotes`](#getvotes)
     - [`getPastVotes`](#getpastvotes)
     - [`getPastTotalSupply`](#getpasttotalsupply)
-    - [`subdelegations`](#subdelegations)
   - [Events](#events)
-    - [`Subdelegation`](#subdelegation)
-    - [`Subdelegations`](#subdelegations)
+    - [`DelegationCreated`](#delegationcreated)
+    - [`DelegationsCreated`](#delegationscreated)
     - [`DelegateVotesChanged`](#delegatevoteschanged)
 - [Storage](#storage)
 - [Types](#types)
-  - [`SubdelegationRule`](#subdelegationrule)
+  - [`Delegation`](#delegation)
   - [`AllowanceType`](#allowancetype)
+  - [`DelegationAdjustment`](#delegationadjustment)
+  - [`Op`](#op)
 - [Backwards Compatibility](#backwards-compatibility)
 - [Migration](#migration)
 - [User Flow](#user-flow)
-  - [Partial delegations](#partial-delegations)
-  - [Constrained delegations](#constrained-delegations)
-  - [Redelegations](#redelegations)
+  - [Partial Delegations](#partial-delegations)
+  - [Absolute & Relative Delegations](#absolute--relative-delegations)
   - [Differences](#differences)
 - [Security Considerations](#security-considerations)
   - [Dependence on Alligator](#dependence-on-alligator)
@@ -75,20 +75,15 @@ function subdelegate(address _delegatee, Delegation calldata _delegation) extern
 This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for `msg.sender`
 and `_delegatee` addresses.
 
+<!-- TODO: specify call to `_delegate` -->
 
-Before updating the subdelegation, the `subdelegate` function MUST check the validity of the subdelegation rule. Specifically,
-the function MUST check that the `allowance` field of the rule does not exceed the total voting power of the delegator
-if the `allowanceType` is `Absolute`.
-
-When updating the subdelegation, the `subdelegate` function MUST override any previous subdelegation of the `msg.sender`
-to the `_delegatee`. Afterwards, this function MUST emit a `Subdelegation` event with the given function parameters.
+At the end, the `subdelegate` function MUST emit a `DelegationCreated` event with the given function parameters.
 
 #### `subdelegateFromToken`
 
-Allows delegation of token voting power from an address (delegator) to another address (delegatee) using a subdelegation
-rule that performs 100% delegation, mimicking the behavior of the `ERC20Votes`'s `delegate` function for backwards
-compatibility. This function MUST only be callable by the `GovernanceToken` contract as part of its `delegate` and
-`delegateBySig` functions.
+Delegates 100% of the token voting power of an address (delegator) to another address (delegatee), mimicking the behavior
+of the `ERC20Votes`'s `delegate` function for backwards compatibility. This function MUST only be callable by the
+`GovernanceToken` contract as part of its `delegate` and `delegateBySig` functions.
 
 ```solidity
 function subdelegateFromToken(address _delegator, address _delegatee) external
@@ -97,32 +92,23 @@ function subdelegateFromToken(address _delegator, address _delegatee) external
 This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for the `_delegator`
 and `_delegatee` addresses.
 
-When updating the subdelegation, the `subdelegateFromToken` function MUST override any previous subdelegation of the
-`_delegator` to the `_delegatee`. Afterwards, this function MUST emit a `Subdelegation` event with the given function
-parameters.
+<!-- TODO: specify call to `_delegate` -->
+
+At the end, the `subdelegateFromToken` function MUST emit a `DelegationCreated` event with the given function parameters.
 
 #### `subdelegateBatched`
 
-Allows batch subdelegation of token voting power to multiple addresses with specified subdelegation rules. This
-function is intended to be called by users.
+Delegates voting power to multiple addresses (delegatees) using the delegation array. This function is intended to be
+called by users.
 
 ```solidity
-function subdelegateBatched(address[] calldata _delegatees, SubdelegationRules[] calldata _rules) external
+function subdelegateBatched(Delegation[] calldata _delegations) external
 ```
 
 This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for `msg.sender`
-and all `_delegatees`.
+and every delegatee address in the `_delegations` array.
 
-This function MUST check that the length of `_delegatees` and `_rules` are equal. If the lengths are not equal, it MUST
-revert with an error.
-
-The `subdelegateBatched` function MUST iterate over each pair of `_delegatees` address and `_rules` subdelegation rule.
-At every iteration, the function MUST check the validity of the subdelegation rule, and migrate the delegatee address
-if it has not been migrated. The function MUST then update the subdelegation. Specifically for validation,
-the function MUST check that the `allowance` field of the rule does not exceed the total voting power of the delegator'
-if the `allowanceType` is `Absolute`.
-
-Afterwards, the `subdelegateBatched` function MUST emit a `Subdelegation` event with the given function parameters.
+At the end, the `subdelegateBatched` function MUST emit a `DelegationsCreated` event with the given function parameters.
 
 #### `afterTokenTransfer`
 
@@ -133,18 +119,16 @@ by the `GovernanceToken` contract.
 function afterTokenTransfer(address _from, address _to, uint256 _amount) external
 ```
 
-The `Alligator` MUST check if the `_from` or `_to` addresses have been migrated by checking the `migrated` mapping
-from its [storage](#storage). If either address has not been migrated, the `Alligator` MUST copy the delegation
-and checkpoint data from the token contract to its own state. After copying the data, the `Alligator` MUST update
-the `migrated` mapping to reflect that the address has been migrated.
+This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for the `_from`
+and `_to` addresses.
 
-The `afterTokenTransfer` function MUST update the voting power of the `_from` and `_to` addresses with the same logic
-of the `GovernanceToken`.
+<!-- TODO: call `_moveVotingPower` -->
 
 #### `migrateAccounts`
 
-Migrates the delegation state of given accounts from the `GovernanceToken` to the `Alligator` contract. This function
-MUST iterate over the list of `_accounts` addresses and only migrate the account if it has not been migrated yet.
+Migrates the delegation state of the given accounts from the `GovernanceToken` to the `Alligator` contract. This
+function MUST iterate over the list of `_accounts` addresses and apply the logic specified in the
+[Migration](#migration) section.
 
 ```solidity
 function migrateAccounts(address[] calldata _accounts) external
@@ -176,11 +160,11 @@ function numCheckpoints(address _account) external view returns (uint32)
 
 #### `delegates`
 
-Retrieves the delegatee with the highest voting power for a given user address. This function
-is intended to be used by the `GovernanceToken` contract to maximize for backwards compatibility.
+Retrieves the delegations of a given user address. This function is intended to be used by the
+`GovernanceToken` contract to maximize for backwards compatibility.
 
 ```solidity
-function delegates(address _account) external view returns (address)
+function delegates(address _account) external view returns (Delegation[] memory)
 ```
 
 #### `getVotes`
@@ -207,52 +191,44 @@ Retrieves the total supply of the `GovernanceToken` at a given block.
 function getPastTotalSupply(uint256 _blockNumber) external view returns (uint256)
 ```
 
-#### `subdelegations`
-
-Retrieves the subdelegations for a given user and delegatee.
-
-```solidity
-function subdelegations(address _account, address _delegatee) external view returns (SubdelegationRules memory)
-```
-
 ### Events
 
-#### `Subdelegation`
+#### `DelegationCreated`
 
-MUST trigger when an account subdelegates voting power to another address (delegatee).
+MUST trigger when an account delegates voting power to another address (delegatee).
 
 ```solidity
-event Subdelegation(address indexed account, address indexed delegatee, SubdelegationRules rule, uint256 previousVotingPower, uint256 newVotingPower)
+event DelegationCreated(address indexed account, Delegation delegation);
 ```
 
-#### `Subdelegations`
+#### `DelegationsCreated`
 
-MUST trigger when an account subdelegates voting power to multiple addresses (delegatees).
+MUST trigger when an account delegates voting power to multiple addresses (delegatees).
 
 ```solidity
-event Subdelegations(address indexed account, address[] delegatee, SubdelegationRules[] rules, uint256 previousVotingPower, uint256 newVotingPower)
+event DelegationsCreated(address indexed account, Delegation[] delegations);
 ```
 
 #### `DelegateVotesChanged`
 
-MUST trigger every time the voting power of a user changes, including when a token transfer occurs or a subdelegation
-is updated.
+MUST trigger every time the voting power of a user changes, including when a token transfer occurs or when a
+delegation is created.
 
 ```solidity
-event DelegateVotesChanged(address indexed delegate, uint256 previousBalance, uint256 newBalance)
+event DelegateVotesChanged(address indexed delegatee, uint256 previousBalance, uint256 newBalance)
 ```
 
 ## Storage
 
-The `Alligator` contract MUST be able to store subdelegation rules and checkpoints. These storage variables MUST be
+The `Alligator` contract MUST be able to store delegations and checkpoints. These storage variables MUST be
 defined as in the `GovernanceToken` and use the same types:
 
 ```solidity
 // Addresses that had their delegation state migrated from the `GovernanceToken` to the `Alligator`.
 mapping(address => bool) public migrated;
 
-// Subdelegation rules for an account and delegatee.
-mapping(address => mapping(address => SubdelegationRules)) internal _subdelegations;
+// Voting power delegations of an account.
+mapping(address => Delegation[]) internal _delegations;
 
 // Checkpoints of voting power for an account.
 mapping(address => ERC20Votes.Checkpoint[]) internal _checkpoints;
@@ -265,32 +241,27 @@ ERC20Votes.Checkpoint[] internal _totalSupplyCheckpoints;
 
 The `Alligator` contract MUST define the following types:
 
-### `SubdelegationRule`
+### `Delegation`
 
-Subdelegation rules define the parameters and constraints for delegated voting power, encapsulated in the following
-struct:
+`Delegation` defines the parameters for voting power delegations, encapsulated in the following struct:
 
 ```solidity
-struct SubdelegationRule {
-    uint256 maxRedelegations;
-    uint256 notValidBefore;
-    uint256 notValidAfter;
-    AllowanceType allowanceType;
-    uint256 allowance;
+struct Delegation {
+  AllowanceType allowanceType;
+  address delegatee;
+  uint256 amount;
 }
 ```
 
 | Name                     | Type            | Description                                                             |
 |--------------------------|-----------------|-------------------------------------------------------------------------|
-| `maxRedelegations`       | `uint256`       | Maximum number of times the delegated votes can be redelegated.         |
-| `notValidBefore`         | `uint256`       | Timestamp after which the delegation is valid.                          |
-| `notValidAfter`          | `uint256`       | Timestamp before which the delegation is valid.                         |
 | `allowanceType`          | `AllowanceType` | Type of allowance (e.g., absolute or relative).                         |
-| `allowance`              | `uint256`       | Amount of votes delegated, denominated in the token's decimals.         |
+| `delegatee`              | `address`       | The address of the delegatee receveing the voting power.                |
+| `amount`                 | `uint256`       | Amount of votes delegated, denomination depending on `allowanceType`.   |
 
 ### `AllowanceType`
 
-Subdelegations can have different types of allowances, represented with:
+Delegations can either be absolute or relative, represented by the following:
 
 ```solidity
 enum AllowanceType {
@@ -301,13 +272,45 @@ enum AllowanceType {
 
 | Name        | Number | Description                                                                                 |
 |-------------|--------|---------------------------------------------------------------------------------------------|
-| `Absolute`  | `0`    | The amount of votes delegated is fixed.                                                     |
-| `Relative`  | `1`    | The amount of votes delegated is relative to the total amount of votes the delegator has.   |
+| `Absolute`  | `0`    | The amount of votes delegated is fixed and denominated in the `GovernanceToken`'s decimals  |
+| `Relative`  | `1`    | The amount of votes delegated is relative and denominated in percentages.                   |
+
+### `DelegationAdjustment`
+
+`DelegationAdjustment` act as temporary storage when adjusting the delegation state of an account. This struct is defined as:
+
+```solidity
+struct DelegationAdjustment {
+  address delegatee;
+  uint208 amount;
+}
+```
+
+| Name        | Type      | Description                                                    |
+|-------------|-----------|----------------------------------------------------------------|
+| `delegatee` | `address` | The address for the delegation adjustment.                     |
+| `amount`    | `uint208` | The amount of voting power to add or remove.                   |
+
+### `Op`
+
+`Op` indicate the type of operation to perform when adjusting the delegation state of an account. This enum is defined as:
+
+```solidity
+enum Op {
+  ADD,
+  SUBTRACT
+}
+```
+
+| Name        | Numbe  | Description                                                    |
+|-------------|--------|----------------------------------------------------------------|
+| `ADD`       | `0`    | Add the amount of voting power when adjusting.                 |
+| `SUBSTRACT` | `1`    | Substract the amount of voting power when adjusting.           |
 
 ## Backwards Compatibility
 
 The `Alligator` contract ensures backwards compatibility by allowing the migration of delegation state from the
-token contract.
+`GovernanceToken`.
 
 ## Migration
 
@@ -318,40 +321,27 @@ and checkpoint data from the token contract to its own state. After copying the 
 
 ## User Flow
 
-The following sections highlight the use cases that MUST be supported by subdelegations, and the difference for basic
+The following sections highlight the use cases that MUST be supported by the `Alligator`, and the difference for basic
 delegations made from the `GovernanceToken` contract.
 
-### Partial delegations
+### Partial Delegations
 
-Users MUST be able to perform partial delegations of their voting power to another address. Additionally, the `Alligator`'s
-subdelegation rules MUST allow users to perform subdelegations for relative and absolute amounts of voting power. Absolute
-amounts MUST be denominated in the token's decimals, whereas relative amounts MUST be denominated in percentages.
+Users MUST be able to perform partial delegations of their voting power to another address. This must be supported for
+both absolute and relative amounts.
 
-### Constrained delegations
+### Absolute & Relative Delegations
 
-Users MUST be able to perform time & block constrained delegations. The subdelegation rules MUST include optional
-`notValidBefore` and `notValidAfter` fields to allow users to specify the time range in which a delegation is valid, both
-denominated as timestamps.
-
-### Redelegations
-
-Users MUST be able to redelegate their voting power to another address. If a user delegates their voting power to another
-address, this second address can further delegate the voting power to a third address. The first delegator MUST be able to
-limit the number of redelegations that can be performed for their delegated voting power by setting the `maxRedelegations`
-field in the subdelegation rule.
+Users MUST be able to delegate their voting power to another address with an absolute or relative amount.
 
 ### Differences
 
-The main difference for delegations made from the `GovernanceToken` contract is basic delegations are encapsulated as
-subdelegations and forwarded to the `Alligator` contract. Basic delegation can be achieved with a subdelegation rule
-such as:
+The main difference for delegations made from the `GovernanceToken` contract is that basic delegations are encapsulated in
+a [`Delegation`](#delegation) struct and forwarded to the `Alligator` contract. Basic delegation can be achieved with a
+`Delegation` such as:
 
 ```solidity
-SubdelegationRules({
-  maxRedelegations: 0,
-  blocksBeforeVoteCloses: 0,
-  notValidBefore: 0,
-  notValidAfter: 0,
+Delegation({
+  delegatee: ...,
   allowanceType: AllowanceType.Relative,
   allowance: 10e4 // 100%
 })
@@ -362,9 +352,9 @@ The following diagram shows the sequence of a basic delegation performed from th
 ```mermaid
 sequenceDiagram
     User ->> GovernanceToken: call `delegate` or `delegateBySig`
-    GovernanceToken ->> Alligator: call `subdelegate` with basic delegation rule
+    GovernanceToken ->> Alligator: call `subdelegateFromToken`
     Alligator -->> GovernanceToken: migrate if user hasn't been migrated
-    Alligator ->> Alligator: update subdelegation
+    Alligator ->> Alligator: apply basic delegation
 ```
 
 Once a user has been migrated to the `Alligator`, the `GovernanceToken` MUST always use the `Alligator`'s delegation
--- a/specs/experimental/alligator.md
+++ b/specs/experimental/alligator.md
@@ -11,7 +11,7 @@
     - [`subdelegateFromToken`](#subdelegatefromtoken)
     - [`subdelegateBatched`](#subdelegatebatched)
     - [`afterTokenTransfer`](#aftertokentransfer)
-    - [`migrateAccount`](#migrateaccount)
+    - [`migrateAccounts`](#migrateaccounts)
   - [Getters](#getters)
     - [`checkpoints`](#checkpoints)
     - [`numCheckpoints`](#numcheckpoints)
@@ -76,8 +76,8 @@ and checkpoint data from the token contract to its own state. After copying the 
 the `migrated` mapping to reflect that the address has been migrated.
 
 Before updating the subdelegation, the `subdelegate` function MUST check the validity of the subdelegation rule. Specifically,
-the function MUST check that the `allowance` field of the rule does not exceed the total voting power of the delegator if the
-`allowanceType` is `Absolute`.
+the function MUST check that the `allowance` field of the rule does not exceed the total voting power of the delegator
+if the `allowanceType` is `Absolute`.
 
 When updating the subdelegation, the `subdelegate` function MUST override any previous subdelegation of the `msg.sender`
 to the `_delegatee`. Afterwards, this function MUST emit a `Subdelegation` event with the given function parameters.
@@ -122,8 +122,8 @@ revert with an error.
 The `subdelegateBatched` function MUST iterate over each pair of `_delegatees` address and `_rules` subdelegation rule.
 At every iteration, the function MUST check the validity of the subdelegation rule, and migrate the delegatee address
 if it has not been migrated. The function MUST then update the subdelegation. Specifically for validation,
-the function MUST check that the `allowance` field of the rule does not exceed the total voting power of the delegator if the
-`allowanceType` is `Absolute`.
+the function MUST check that the `allowance` field of the rule does not exceed the total voting power of the delegator'
+if the `allowanceType` is `Absolute`.
 
 Afterwards, the `subdelegateBatched` function MUST emit a `Subdelegation` event with the given function parameters.
 
--- a/specs/experimental/alligator.md
+++ b/specs/experimental/alligator.md
@@ -29,6 +29,7 @@
   - [`SubdelegationRule`](#subdelegationrule)
   - [`AllowanceType`](#allowancetype)
 - [Backwards Compatibility](#backwards-compatibility)
+- [Migration](#migration)
 - [User Flow](#user-flow)
   - [Partial delegations](#partial-delegations)
   - [Constrained delegations](#constrained-delegations)
@@ -48,8 +49,8 @@
 |----------|----------------------------------------------|
 | Address  | `0x4200000000000000000000000000000000000043` |
 
-The `Alligator` contract implements subdelegations for the [`GovernanceToken`](gov-token.md). Subdelegations enable
-advanced delegation use cases, such as partial, time-constrained & block-based delegations, and relative & fixed allowances.
+The `Alligator` contract implements advanced delegation for the [`GovernanceToken`](gov-token.md). Advance delegation allows
+for partial and relative delegations of voting power.
 
 The `Alligator` contract migrates the delegation state from the `GovernanceToken` to itself
 through a hook-based approach. Specifically, the `GovernanceToken` calls the `Alligator` contract's
@@ -64,17 +65,16 @@ been migrated to the `Alligator` contract, the contract copies the address' chec
 
 #### `subdelegate`
 
-Allows subdelegation of token voting power to another address (delegatee) with a specified subdelegation rule. This function
+Delegates voting power to another address (delegatee) with the given delegation parameters. This function
 is intended to be called by users that require advanced delegation of the `GovernanceToken`.
 
 ```solidity
-function subdelegate(address _delegatee, SubdelegationRule _rule) external
+function subdelegate(address _delegatee, Delegation calldata _delegation) external
 ```
 
-This function MUST check if the `msg.sender` or `_delegatee` addresses have been migrated by checking the `migrated` mapping
-from its [storage](#storage). If either address has not been migrated, the `Alligator` MUST copy the delegation
-and checkpoint data from the token contract to its own state. After copying the data, the `Alligator` MUST update
-the `migrated` mapping to reflect that the address has been migrated.
+This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for `msg.sender`
+and `_delegatee` addresses.
+
 
 Before updating the subdelegation, the `subdelegate` function MUST check the validity of the subdelegation rule. Specifically,
 the function MUST check that the `allowance` field of the rule does not exceed the total voting power of the delegator
@@ -94,10 +94,8 @@ compatibility. This function MUST only be callable by the `GovernanceToken` cont
 function subdelegateFromToken(address _delegator, address _delegatee) external
 ```
 
-This function MUST check if the `_delegator` or `_delegatee` addresses have been migrated by checking the `migrated` mapping
-from its [storage](#storage). If either address has not been migrated, the `Alligator` MUST copy the delegation
-and checkpoint data from the token contract to its own state. After copying the data, the `Alligator` MUST update
-the `migrated` mapping to reflect that the address has been migrated.
+This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for the `_delegator`
+and `_delegatee` addresses.
 
 When updating the subdelegation, the `subdelegateFromToken` function MUST override any previous subdelegation of the
 `_delegator` to the `_delegatee`. Afterwards, this function MUST emit a `Subdelegation` event with the given function
@@ -112,10 +110,8 @@ function is intended to be called by users.
 function subdelegateBatched(address[] calldata _delegatees, SubdelegationRules[] calldata _rules) external
 ```
 
-This function MUST check if the `msg.sender` has been migrated by checking the `migrated` mapping from its [storage](#storage).
-If it has not been migrated, the `Alligator` MUST copy the delegation and checkpoint data from the token contract
-to its own state. After copying the data, the `Alligator` MUST update the `migrated` mapping to reflect that the address
-has been migrated.
+This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for `msg.sender`
+and all `_delegatees`.
 
 This function MUST check that the length of `_delegatees` and `_rules` are equal. If the lengths are not equal, it MUST
 revert with an error.
@@ -312,6 +308,13 @@ enum AllowanceType {
 
 The `Alligator` contract ensures backwards compatibility by allowing the migration of delegation state from the
 token contract.
+
+## Migration
+
+All write functions in the `Alligator` MUST check if the users interacting with it have been migrated by checking the
+`migrated` mapping from its [storage](#storage). If a user has not been migrated, the `Alligator` MUST copy the delegation
+and checkpoint data from the token contract to its own state. After copying the data, the `Alligator` MUST update the
+`migrated` mapping to reflect that the address has been migrated.
 
 ## User Flow
 
--- a/specs/experimental/alligator.md
+++ b/specs/experimental/alligator.md
@@ -12,6 +12,7 @@
     - [`subdelegateBatched`](#subdelegatebatched)
     - [`afterTokenTransfer`](#aftertokentransfer)
     - [`migrateAccounts`](#migrateaccounts)
+    - [`_delegate`](#_delegate)
   - [Getters](#getters)
     - [`checkpoints`](#checkpoints)
     - [`numCheckpoints`](#numcheckpoints)
@@ -69,13 +70,12 @@ Delegates voting power to another address (delegatee) with the given delegation 
 is intended to be called by users that require advanced delegation of the `GovernanceToken`.
 
 ```solidity
-function subdelegate(address _delegatee, Delegation calldata _delegation) external
+function subdelegate(Delegation calldata _delegation) external
 ```
 
 This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for `msg.sender`
-and `_delegatee` addresses.
-
-<!-- TODO: specify call to `_delegate` -->
+and the `_delegatee` address in the delegation. Afterwards, the function MUST call the [`_delegate`](#_delegate) with
+the `msg.sender` and `_newDelegations` parameters.
 
 At the end, the `subdelegate` function MUST emit a `DelegationCreated` event with the given function parameters.
 
@@ -90,9 +90,8 @@ function subdelegateFromToken(address _delegator, address _delegatee) external
 ```
 
 This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for the `_delegator`
-and `_delegatee` addresses.
-
-<!-- TODO: specify call to `_delegate` -->
+and `_delegatee` addresses. Afterwards, the function MUST call the [`_delegate`](#_delegate) with the `_delegator` and
+basic delegation parameters.
 
 At the end, the `subdelegateFromToken` function MUST emit a `DelegationCreated` event with the given function parameters.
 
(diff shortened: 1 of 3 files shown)


tynes:
Its not possible to look at this spec and implement the contract based on it. The point of a spec is to define ahead of time what the expected behavior is. There is no ABI and its not clear exactly what it should be doing. The spec should be clear enough such that its possible for the code reviewer to match the spec against the implementation. Its good to think in terms of state machines + invariants when writing a spec.

0xfuturistic:
> Its not possible to look at this spec and implement the contract based on it. The point of a spec is to define ahead of time what the expected behavior is. There is no ABI and its not clear exactly what it should be doing. The spec should be clear enough such that its possible for the code reviewer to match the spec against the implementation. Its good to think in terms of state machines + invariants when writing a spec.

sorry -- that was too early of a draft. just added info from the actual implementation!

specs/governance/alligator.md
```diff
@@ -0,0 +1,114 @@
...
+- [Subdelegations](#subdelegations)
+  - [Subdelegation Rules](#subdelegation-rules)
+- [`afterTokenTransfer`](#aftertokentransfer)
+- [View functions](#view-functions)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+The `Alligator` contract implements advanced delegation features for the `GovernanceToken` contract. It is based on
```
tynes: What is AlligatorOP?

specs/governance/alligator.md
```diff
@@ -0,0 +1,114 @@
...
+the Alligator V2 contract, with the following modifications from AlligatorOP:
+
+- uses hashed proxy rules to reduce calldata size
```
tynes: What is a proxy rule?

specs/governance/alligator.md
```diff
@@ -0,0 +1,114 @@
...
+  - [Subdelegation Rules](#subdelegation-rules)
+- [`afterTokenTransfer`](#aftertokentransfer)
+- [View functions](#view-functions)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+The `Alligator` contract implements advanced delegation features for the `GovernanceToken` contract. It is based on
+the Alligator V2 contract, with the following modifications from AlligatorOP:
```
tynes: What are advanced delegation features?

specs/governance/alligator.md
```diff
@@ -0,0 +1,114 @@
...
+
+The `Alligator` contract integrates with the `GovernanceToken` contract through a hook-based approach. The
+`GovernanceToken` contract calls the `Alligator` contract's `afterTokenTransfer` function after a token transfer. This
+allows the `Alligator` contract to consume the hooks and update its delegation and checkpoint mappings accordingly.
+If either of the addresses passed as arguments to `afterTokenTransfer`, `from` and `to`, have not been migrated,
+`Alligator` copies the address's delegation and checkpoint data from the `GovernanceToken` contract to its own state.
+
+## Governor
+
+The `Alligator` contract MUST provide a method for validating subdelegation rules and casting a vote on the governor.
```
tynes: What is subdelegation? How does it work?

specs/governance/alligator.md
```diff
@@ -0,0 +1,114 @@
...
+It should also provide methods for variants such as for
+
+- casting votes with reason
```
tynes: How does voting work?

cairoeth: updated content, but voting remains the same/unaffected (done via the Governor contract)

specs/governance/alligator.md
```diff
@@ -0,0 +1,61 @@
...
+- [Token Transfer Hook](#token-transfer-hook)
+- [View functions](#view-functions)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+The `Alligator` contract implements subdelegations that can be used by tokens contracts that inherit
+[`GovernanceToken`](gov-token.md). Subdelegations allow for advanced delegation use cases, such as partial,
+time-constrained & block-based delegations, relative & fixed allowances, and custom rules. The `Alligator` contract
```
tynes: Alligator only requires migrating delegation state from the token contract for backwards compatibility, this isn't clear here. Its important to call that out because fresh chains don't need that feature

specs/governance/alligator.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Subdelegations
+
+To support subdelegations for multiple token contracts, the `Alligator` contract MUST implement storage references to
+store delegations and checkpoints for addresses. Specifically, the `Alligator` contract MUST implement the equivalent
+of the following mappings from `GovernanceToken` contract: `_delegates`, `_checkpoints`, `_totalSupplyCheckpoints`.
+
+The contract MUST also provide methods for subdelegation voting power to another address. Concretely, the `Alligator`
+contract MUST implement a `subdelegate` function which takes a `to` address, and a subdelegation rule. A subdelegation
+rule MUST contain the following fields:
```
tynes: This is good laying out the definitions of the inputs, it could be good to include the specific function signature for these functions before defining their inputs so its very obvious what the ABI is in the form of `subdelegate(to, ...)()`. If a `rule` is a struct then the struct should be defined as well

specs/governance/alligator.md
tynes: I think you should move all of the backwards compat stuff out of a the overview and into a "Backwards Compatibility" section

specs/governance/alligator.md
```diff
@@ -0,0 +1,61 @@
...
+
+- `maxRedelegations`: The maximum number of times the delegated votes can be redelegated.
+- `blocksBeforeVoteCloses`: The number of blocks before the vote closes that the delegation is valid.
+- `notValidBefore`: The timestamp after which the delegation is valid.
+- `notValidAfter`: The timestamp before which the delegation is valid.
+- `allowanceType` The type of allowance. If `Absolute`, the amount of votes delegated is fixed. If `Relative`, the
```
tynes: Are the only 2 types absolute and relative?

specs/governance/alligator.md
```diff
@@ -0,0 +1,61 @@
...
+## Token Transfer Hook
+
+The `Alligator` contract MUST provide a method, referred to as `afterTokenTransfer`, that can be called by the token
+contract's `_afterTokenTransfer` function. This method should update its delegation and checkpoint mappings for the
+address of the token contract. Specifically, if the `to` or `from` addresses of the transfer have not been migrated,
+the `Alligator` MUST migrate it by copying the delegation and checkpoint data from the token contract to its own state.
+
+## View functions
+
+`Alligator` MUST provide methods to get the delegation state for a token contract address. Specifically, the `Alligator`
```
tynes: It would be good to define the getter functions here

specs/governance/alligator.md
```diff
@@ -0,0 +1,61 @@
...
+- `notValidBefore`: The timestamp after which the delegation is valid.
+- `notValidAfter`: The timestamp before which the delegation is valid.
+- `allowanceType` The type of allowance. If `Absolute`, the amount of votes delegated is fixed. If `Relative`, the
+  amount of votes delegated is relative to the total amount of votes the delegator has.
+- `allowance`: The amount of votes delegated, denominated in the token contract's decimals.
+
+The `Alligator` contract MUST provide a method to subdelegate in batches, called `subdelegateBatched`. This function
+should take an array of `to` addresses and an array of subdelegation rules, and call the `subdelegate` function for each
+pair of `to` address and subdelegation rule. `Alligator` MUST also provide a method to subdelegate votes using a
+signature, called `subdelegateBySig`. This function should take the same arguments as `subdelegate`, plus a signature
```
tynes: the full ABI for `subdelegateBySig` would be good to add

tynes:
See the way that https://github.com/ethereum-optimism/specs/pull/257 lays out the functions in subsections for a good example on how to structure the spec in a more uniform way with the rest of the specs

specs/governance/alligator.md
0xfuturistic: all good ideas @tynes!

specs/governance/alligator.md
0xfuturistic: existing contracts only define absolute & relative. do we need anything else? If not, I'm not sure it would be worth it to add more allowance types under the scope of this PR

specs/governance/alligator.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+Calls the `subdelegate` function for each pair of `_to` address and subdelegation rule.
+
+#### `subdelegateBySig`
+
+Allows subdelegation of voting power using a signature.
+
+```solidity
+subdelegateBySig(address _to, SubdelegationRule _rule, bytes _signature)
```
tynes: How is the signature encoded? ie v, r, s?

specs/governance/alligator.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+Takes the same arguments as `subdelegate`, plus a signature of the previous parameters.
+
+#### `afterTokenTransfer`
```
tynes: Which accounts can call this?

specs/governance/alligator.md
```diff
@@ -0,0 +1,144 @@
...
+```solidity
+afterTokenTransfer(address _from, address _to, uint256 _amount)
+```
+
+If the `_to` or `_from` addresses have not been migrated, `Alligator` migrates by copying the delegation and checkpoint
+data from the token contract to its own state.
+
+### Getters
+
+The output for these functions is conditional on whether the user address has been migrated or not. Concretely, the
```
tynes: How does it know if there was a migration already?

specs/governance/alligator.md
```diff
@@ -0,0 +1,144 @@
...
+Allows subdelegation of voting power to another address with specified subdelegation rules.
+
+```solidity
+subdelegate(address _to, SubdelegationRule _rule)
+```
+
+A subdelegation rule is an instance of the following struct:
+
+```solidity
+struct SubdelegationRule {
```
tynes: It could be good to have a `## Types` section that includes the subdelegation rule + checkpoint with descriptions of each field in them outside of the code block

specs/governance/alligator.md
cairoeth: removing this function as it's a duplicate of the `delegatebySig` function from the `GovernanceToken` (inherited from `ERC20Votes`)

specs/governance/alligator.md
cairoeth: updated

specs/governance/alligator.md
cairoeth: added the migration mapping in the storage section, which is a getter to check if a user has been migrated or not.

specs/governance/alligator.md
cairoeth: agree, section added.

specs/governance/alligator.md
```diff
@@ -0,0 +1,179 @@
...
+
+#### `subdelegate`
+
+Allows subdelegation of token voting power to another address with specified subdelegation rules. This function
+is inteded to be called by users.
+
+```solidity
+subdelegate(address _token, address _delegatee, SubdelegationRule _rule)
+```
+
```
tynes: I am confused by the existence of 2 `subdelegate` functions with the exact same function signature

specs/governance/alligator.md
tynes: We should say something like "This function MUST only be callable by the GovernanceToken contract"

specs/governance/alligator.md
```diff
@@ -0,0 +1,179 @@
...
+#### `afterTokenTransfer`
+
+Updates delegation and checkpoint mappings for the address of a token contract after a transfer. This function
+MUST be called by the `_afterTokenTransfer` function in the token contract.
+
+```solidity
+afterTokenTransfer(address _from, address _to, uint256 _amount)
+```
+
+If the `_to` or `_from` addresses have not been migrated, `Alligator` migrates by copying the delegation and checkpoint
```
tynes: How do you know if the address was migrated?

specs/governance/alligator.md
```diff
@@ -0,0 +1,179 @@
...
+
+```solidity
+getVotingPower(address _token, address _user, uint256 _blockNumber) returns (uint256)
+```
+
+## Storage
+
+The `Alligator` contract MUST be able to store delegations and checkpoints for multiple token contracts.
+These storage variables MUST be defined in the same way as in the token contract:
+
```
tynes: Its a bit easier to read when the descriptions are outside of the code block, see https://github.com/ethereum-optimism/specs/pull/247#discussion_r1681672402

cairoeth: yup, updated with same style as in https://github.com/ethereum-optimism/specs/blob/main/specs/interop/messaging.md#message-identifier

specs/governance/alligator.md
cairoeth: clarified that the contract must check against the `migrated` mapping, and update it after migrating an address.

specs/governance/alligator.md
```diff
@@ -0,0 +1,207 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+| Constant | Value                                        |
+|----------|----------------------------------------------|
+| Address  | `0x4200000000000000000000000000000000000043` |
+
+The `Alligator` contract implements subdelegations for the [`GovernanceToken`](gov-token.md). Subdelegations enable
```
tynes: Are any of these advanced delegation use cases documented anywhere? It would be really helpful if they were documented in this spec

specs/governance/alligator.md
```diff
@@ -0,0 +1,207 @@
...
+advanced delegation use cases, such as partial, time-constrained & block-based delegations, and relative & fixed allowances.
+
+The `Alligator` contract migrates the delegation state from the [`GovernanceToken`](gov-token.md) to itself
```
tynes: I think that all of the backwards compat language should go in a "backwards compatibility" section rather than making it core to the spec of how the alligator contract works

specs/governance/alligator.md
```diff
@@ -0,0 +1,207 @@
...
+#### `subdelegate`
+
+Allows subdelegation of token voting power to another address (delegatee) with specified subdelegation rules. This function
+is inteded to be called by users that require advanced delegation of the [`GovernanceToken`](gov-token.md).
+
+```solidity
+subdelegate(address _delegatee, SubdelegationRule _rule)
+```
+
+#### `subdelegateFromToken`
```
tynes: Now that `subdelegate` has had its function signature changed, do we need to name the other function `subdelegateFromToken` or can it be an overloaded `subdelegate`?

specs/governance/alligator.md
```diff
@@ -0,0 +1,207 @@
...
+
+#### `subdelegateBatched`
+
+Allows batch subdelegation of token voting power to multiple addresses with specified subdelegation rules. This
+function is intended to be called by users.
+
+```solidity
+subdelegateBatched(address[] _delegatees, SubdelegationRule[] _rules)
+```
+
```
tynes: What happens if the length of the input args don't match? Is it supposed to interpret pairs at the same indices? Is this approach better than adding a new type `BatchDelegation` which includes a delegate + rule?

specs/governance/alligator.md
```diff
@@ -0,0 +1,207 @@
...
+Calls the `subdelegate` function for each pair of `_delegatees` address and subdelegation rule.
+
+#### `afterTokenTransfer`
+
+Updates the voting power of two addresses (`_from` and `_to`) after a token transfer. This function MUST
+be called by the `_afterTokenTransfer` function in the [`GovernanceToken`](gov-token.md).
+
+```solidity
+afterTokenTransfer(address _from, address _to, uint256 _amount)
```
tynes: We should also be sure to add language about the risk of this reverting, ie it would prevent `GovToken` from being transferred

specs/governance/alligator.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+The `Alligator` MUST check if the `_from` or `_to` addresses have been migrated by checking the `migrated` mapping
+from its [storage](#storage). If either address has not been migrated, the `Alligator` MUST copy the delegation
+and checkpoint data from the token contract to its own state. After copying the data, the `Alligator` MUST update
+the `migrated` mapping to reflect that the address has been migrated.
+
+### Getters
+
+For backwards compatibility, the `Alligator` MUST implement all public getter functios of the
```
tynes: typo: `functions`

specs/governance/alligator.md
```diff
@@ -0,0 +1,207 @@
...
+[`GovernanceToken`](gov-token.md) related to delegation and voting power. These functions MUST be used by the
```
tynes: Do each of these getters check the `migrated` mapping and then forward to `GovToken` if not migrated?

specs/governance/alligator.md
```diff
@@ -0,0 +1,207 @@
...
+The `Alligator` contract migrates the delegation state from the [`GovernanceToken`](gov-token.md) to itself
+through a hook-based approach. Specifically, the [`GovernanceToken`](gov-token.md) calls the `Alligator` contract's
+`afterTokenTransfer` function after a token transfer. This enables the `Alligator` contract to consume the hook and update
+its delegation and checkpoint mappings accordingly. If either address involved in the transfer (`_from_` or `_to`) has not
+been migrated to the `Alligator` contract, the contract copies the address' checkpoint data from the
+[`GovernanceToken`](gov-token.md) to its own state.
+
+## Interface
+
+### Core Functions
```
tynes: I notice that no events are defined here, are there events that are emitted as part of these functions? If so, they should be specified. The specific reason is to make it easy for data people that want to understand how to go about indexing the data

specs/governance/alligator.md
cairoeth: imo it's more clear if it's named `subdelegateFromToken` because of the limitation that it can only be called from the `GovernanceToken`

specs/governance/alligator.md
cairoeth: If the lengths mismatch, it should revert. Updated the content for this function to clarify this. I think it would be easier for external UIs to use this instead of having a `BatchDelegation` type, but no strong opinion

specs/governance/alligator.md
cairoeth: not in the `Alligator` -- the getters will always fetch from its storage. the `GovToken` has the check, so im wondering if we should also do that in the `Alligator` so that the behavior match

specs/governance/alligator.md
cairoeth: totally. added an `Events` section, and specified for each function what event it should emit.

specs/governance/alligator.md
cairoeth: added in `Security Considerations` section

specs/governance/alligator.md
cairoeth: added a section with the user flow and usecases.

specs/governance/alligator.md
stepandel: I think we should do on Alligator as well. Otherwise, the voting power / delegation state reflected on the `Alligator` won't be accurate when an address hasn't been `migrated`

cairoeth: agreed. updated in https://github.com/ethereum-optimism/specs/pull/247/commits/7752d7d2abc4040f3cbd4104ace81be69067bbf1

specs/governance/alligator.md
```diff
@@ -0,0 +1,389 @@
...
+#### `subdelegateFromToken`
+
+Allows subdelegation of token voting power from an address (delegator) to another address (delegatee) with a specified
+subdelegation rule. This function is intended to be called by the [`GovernanceToken`](gov-token.md) contract as part
+of its `delegateBySig` function. To ensure backwards compatibility in the [`GovernanceToken`](gov-token.md), the
+subdelegation rule MUST be 100% delegation, mimicking the behavior of the [`GovernanceToken`](gov-token.md)'s `delegate`
+function.
+
+```solidity
+function subdelegateFromToken(address _delegator, address _delegatee, SubdelegationRule _rule) external
```
stepandel: In case we want to change the default rule set for the `subdelegateFromToken`, although quite unlikely, it would be easier to upgrade `Alligator` to change this default. Having `subdelegateFromToken(address _delegator, address _delegatee)`  + set default rule set on `Alligator` instead of `GovernanceToken` would make our life easier in this case

specs/governance/alligator.md
```diff
@@ -0,0 +1,389 @@
...
+
+MUST trigger when an account subdelegates voting power to another address (delegatee).
+
+```solidity
+event Subdelegation(address indexed account, address indexed delegatee, SubdelegationRules rule)
+```
+
+#### `Subdelegations`
+
+MUST trigger when an account subdelegates voting power to multiple addresses (delegatees).
```
stepandel: would this be emitting the requested `subdelegations`  or the `subdelegations` state after the changes have been applied?

specs/governance/alligator.md
```diff
@@ -0,0 +1,389 @@
...
+
+The `Alligator` contract MUST be able to store subdelegation rules and checkpoints. These storage variables MUST be
+defined as in the [`GovernanceToken`](gov-token.md) and use the same types:
+
+```solidity
+// Addresses that had their delegation state migrated from the `GovernanceToken` to the `Alligator`.
+mapping(address => bool) public migrated;
+
+// Subdelegation rules for an account and delegatee.
+mapping(address => mapping(address => SubdelegationRules)) internal _subdelegations;
```
stepandel: I think we need a redundant backwards `subdelegations` mapping here as well in order to calculate voting power
`mapping(address => mapping(address => SubdelegationRules)) internal _subdelegationsTo;`
`mapping(address => mapping(address => SubdelegationRules)) internal _subdelegationsFrom;`

`voting_power = sum(subdelegationsTo) - sum(subdelegatesionsFrom)`

specs/governance/alligator.md
tynes: I see that the spec has been updated to show the revert with an error, thank you!

specs/governance/alligator.md
```diff
@@ -0,0 +1,390 @@
...
+and checkpoint data from the token contract to its own state. After copying the data, the `Alligator` MUST update
+the `migrated` mapping to reflect that the address has been migrated.
+
+The `afterTokenTransfer` function MUST update the voting power of the `_from` and `_to` addresses with the same logic
+of the `GovernanceToken`.
+
+#### `migrateAccount`
+
+Migrates the delegation state of a given account from the `GovernanceToken` to the `Alligator` contract. This function
+MUST only migrate the `_account` if it has not been migrated yet.
```
tynes: Thoughts on having this accept an array of address? Generally we want to keep a small ABI but it would be much easier to pass thru an array rather than using multicall

specs/governance/alligator.md
```diff
@@ -0,0 +1,390 @@
...
+
+```solidity
+function migrateAccount(address _account) external
+```
+
+### Getters
+
+For backwards compatibility, the `Alligator` MUST implement all public getter functions of the
+`GovernanceToken` related to delegation and voting power. These functions MUST be used by the
+`GovernanceToken` when an account has been been migrated to the `Alligator` contract. Otherwise,
```
tynes: So the gov token will ask the alligator if its been migrated? The idea here is that the gov token and alligator should be consistent with these getter functions

specs/governance/alligator.md
tynes: Based on the mermaid flow, I agree with the idea of removing the `SubdelegationRule` from the function signature and just using a default one. A user would then call `subdelegate` directly on alligator to update the rule

specs/governance/alligator.md
```diff
@@ -0,0 +1,390 @@
...
+  allowanceType: AllowanceType.Relative,
+  allowance: 10e4 // 100%
+})
+```
+
+The following diagram shows the sequence of a basic delegation performed from the `GovernanceToken` contract.
+
+```mermaid
+sequenceDiagram
+    User ->> GovernanceToken: call `delegate` or `delegateBySig`
```
tynes: Would this be a delegate to alligator?

cairoeth: user calls the `delegate` function in the govtoken, which it then calls the `subdelegateFromToken` function in the alligator

specs/governance/alligator.md
```diff
@@ -0,0 +1,390 @@
...
+`afterTokenTransfer` function after a token transfer. This enables the `Alligator` contract to consume the hook and update
+its delegation and checkpoint mappings accordingly. If either address involved in the transfer (`_from_` or `_to`) has not
+been migrated to the `Alligator` contract, the contract copies the address' checkpoint data from the
+`GovernanceToken` to its own state.
+
+## Interface
+
+### Core Functions
+
+#### `subdelegate`
```
tynes: Does this require the user to first delegate to alligator?

specs/governance/alligator.md
cairoeth: yeah i think it would be more useful, tho don't we will often use this function

specs/governance/alligator.md
```diff
@@ -0,0 +1,390 @@
...
+```solidity
+function subdelegate(address _delegatee, SubdelegationRule _rule) external
+```
+
+This function MUST check if the `msg.sender` or `_delegatee` addresses have been migrated by checking the `migrated` mapping
+from its [storage](#storage). If either address has not been migrated, the `Alligator` MUST copy the delegation
+and checkpoint data from the token contract to its own state. After copying the data, the `Alligator` MUST update
+the `migrated` mapping to reflect that the address has been migrated.
+
+Before updating the subdelegation, the `subdelegate` function MUST check the validity of the subdelegation rule.
```
tynes: What is the definition of validity in this case?

specs/governance/alligator.md
```diff
@@ -0,0 +1,390 @@
...
+
+When updating the subdelegation, the `subdelegate` function MUST override any previous subdelegation of the `msg.sender`
+to the `_delegatee`. Afterwards, this function MUST emit a `Subdelegation` event with the given function parameters.
```
chuxinh: Is it possible for the sub-delegation event or some other event to also emit the resulted delegate vote change as well? 

From what I learned previously, there's a need to construct subdelegation / partial delegation from authority of chains, which is probably an overkill from the data angle, similar to `DelegateVotesChanged` in the current OP token setup

specs/governance/alligator.md
```diff
@@ -0,0 +1,390 @@
...
+
+Before updating the subdelegation, the `subdelegateFromToken` function MUST check the validity of the subdelegation rule.
+
+When updating the subdelegation, the `subdelegateFromToken` function MUST override any previous subdelegation of the
+`_delegator` to the `_delegatee`. Afterwards, this function MUST emit a `Subdelegation` event with the given function
+parameters.
+
+#### `subdelegateBatched`
+
+Allows batch subdelegation of token voting power to multiple addresses with specified subdelegation rules. This
```
chuxinh: Is the above subdelegation rule 100% delegation to another address? How does this work under this batching function?

specs/governance/alligator.md
```diff
@@ -0,0 +1,390 @@
...
+This function MUST check that the length of `_delegatees` and `_rules` are equal. If the lengths are not equal, it MUST
+revert with an error.
+
+The `subdelegateBatched` function MUST iterate over each pair of `_delegatees` address and `_rules` subdelegation rule.
+At every iteration, the function MUST check the validity of the subdelegation rule, and migrate the  delegatee address
+if it has not been migrated. The function MUST then update the subdelegation.
+
+Afterwards, the `subdelegateBatched` function MUST emit a `Subdelegation` event with the given function parameters.
+
+#### `afterTokenTransfer`
```
chuxinh, line 129: yea reiterating from the above and wonder if it's possible to emit events that reflect the updated voting power post subdelegation

specs/governance/alligator.md
```diff
@@ -0,0 +1,390 @@
...
+
+#### `Subdelegations`
+
+MUST trigger when an account subdelegates voting power to multiple addresses (delegatees).
+
+```solidity
+event Subdelegations(address indexed account, address[] delegatee, SubdelegationRules[] rules)
+```
+
+#### `DelegateVotesChanged`
```
chuxinh: This should be the one i've been talking about. The current version does not incorporate changes from the existing alligator contract

specs/governance/alligator.md
cairoeth: users don't need to delegate to alligator

specs/governance/alligator.md
cairoeth: yes, that's possible -- I'll update this to add the voting power balance parameter in the events.

specs/governance/alligator.md
cairoeth: this function accepts a list of subdelegation rules, which can be of any type. the subdelegation of 100% delegation is only for the `subdelegateFromToken` to mimick the current behavior of delegations in the token.

this batching function will apply each pair of delegatee and subdelegation rule.

specs/governance/alligator.md
cairoeth: @chuxinh do you prefer to have two events when a subdelegation is applied (so the `Subdelegation` and `DelegateVotesChanged` emitted at the same time) or only one event (for example `Subdelegation`, which includes the same parameters as `DelegateVotesChanged` -- new balance and old balance). in this second case, the `DelegateVotesChanged` would only be triggered for token transfers and not when a subdelegation rule is applied/updated.

specs/governance/alligator.md
cairoeth: updating events to reflect new voting power balances, similar to https://github.com/ethereum-optimism/specs/pull/247#discussion_r1689859919

specs/governance/alligator.md
cairoeth: updated for the second approach: https://github.com/ethereum-optimism/specs/pull/247/commits/88c104f1a1ce5af225d224cbe324c0f63bcd8a5e

specs/governance/alligator.md
cairoeth: https://github.com/ethereum-optimism/specs/pull/247/commits/88c104f1a1ce5af225d224cbe324c0f63bcd8a5e

specs/governance/alligator.md
cairoeth: https://github.com/ethereum-optimism/specs/pull/247/commits/88c104f1a1ce5af225d224cbe324c0f63bcd8a5e

specs/governance/alligator.md
cairoeth: updated to specify that it should check that the `allowance` amount in the rule is not more than the voting power of the delegator if the `allowanceType` of the rule is `Absolute`: https://github.com/ethereum-optimism/specs/pull/247/commits/3beb12172e8620d36eca2847c5c6186f220ce0af

specs/experimental/alligator.md
```diff
@@ -0,0 +1,401 @@
...
+been migrated to the `Alligator` contract, the contract copies the address' checkpoint data from the
+`GovernanceToken` to its own state.
+
+## Interface
+
+### Core Functions
+
+#### `subdelegate`
+
+Allows subdelegation of token voting power to another address (delegatee) with a specified subdelegation rule. This function
```
smartcontracts: Originally I was going to nitpick the use of the word "delegatee" instead of "delegate" but I think it's better in this context because the noun and the verb for "delegate" are the same which makes the code more confusing. Just writing this down for posterity, "delegatee" while less widely used is probably better from a code perspective.

specs/experimental/alligator.md
```diff
@@ -0,0 +1,401 @@
...
+```solidity
+function subdelegate(address _delegatee, SubdelegationRule _rule) external
+```
+
+This function MUST check if the `msg.sender` or `_delegatee` addresses have been migrated by checking the `migrated` mapping
+from its [storage](#storage). If either address has not been migrated, the `Alligator` MUST copy the delegation
+and checkpoint data from the token contract to its own state. After copying the data, the `Alligator` MUST update
+the `migrated` mapping to reflect that the address has been migrated.
+
+Before updating the subdelegation, the `subdelegate` function MUST check the validity of the subdelegation rule. Specifically,
```
smartcontracts: Are there any other validity rules?

specs/experimental/alligator.md
```diff
@@ -0,0 +1,401 @@
...
+the function MUST check that the `allowance` field of the rule does not exceed the total voting power of the delegator
+if the `allowanceType` is `Absolute`.
+
+When updating the subdelegation, the `subdelegate` function MUST override any previous subdelegation of the `msg.sender`
+to the `_delegatee`. Afterwards, this function MUST emit a `Subdelegation` event with the given function parameters.
+
+#### `subdelegateFromToken`
```
smartcontracts: I think the function name here is a bit confusing and could maybe be refined to `subdelegateByGovToken` or something similar.

specs/experimental/alligator.md
```diff
@@ -0,0 +1,401 @@
...
+```solidity
+function subdelegateBatched(address[] calldata _delegatees, SubdelegationRules[] calldata _rules) external
+```
+
+This function MUST check if the `msg.sender` has been migrated by checking the `migrated` mapping from its [storage](#storage).
+If it has not been migrated, the `Alligator` MUST copy the delegation and checkpoint data from the token contract
+to its own state. After copying the data, the `Alligator` MUST update the `migrated` mapping to reflect that the address
+has been migrated.
+
+This function MUST check that the length of `_delegatees` and `_rules` are equal. If the lengths are not equal, it MUST
```
smartcontracts: Any particular reason for this vs nested struct?

specs/experimental/alligator.md
smartcontracts: It would probably be worth adding a separate section called "Subdelegation Rule Validation" that explains the validation rules. You can then link to that section here which will make it easier to update if we ever add to those rules.

specs/experimental/alligator.md
```diff
@@ -0,0 +1,401 @@
...
+
+Retrieves the number checkpoints for a given user address.
+
+```solidity
+function numCheckpoints(address _account) external view returns (uint32)
+```
+
+#### `delegates`
+
+Retrieves the delegatee with the highest voting power for a given user address. This function
```
smartcontracts: I would recommend adding a note here that explains the behavior of the original function in the `GovernanceToken` contract and why this updated behavior was chosen.

specs/experimental/alligator.md
```diff
@@ -0,0 +1,401 @@
...
+```
+
+## Storage
+
+The `Alligator` contract MUST be able to store subdelegation rules and checkpoints. These storage variables MUST be
+defined as in the `GovernanceToken` and use the same types:
+
+```solidity
+// Addresses that had their delegation state migrated from the `GovernanceToken` to the `Alligator`.
+mapping(address => bool) public migrated;
```
smartcontracts: I would recommend clearly defining this as a public getter within the Getters section.

specs/experimental/alligator.md
```diff
@@ -0,0 +1,401 @@
...
+```solidity
+function afterTokenTransfer(address _from, address _to, uint256 _amount) external
+```
+
+The `Alligator` MUST check if the `_from` or `_to` addresses have been migrated by checking the `migrated` mapping
+from its [storage](#storage). If either address has not been migrated, the `Alligator` MUST copy the delegation
+and checkpoint data from the token contract to its own state. After copying the data, the `Alligator` MUST update
+the `migrated` mapping to reflect that the address has been migrated.
+
+The `afterTokenTransfer` function MUST update the voting power of the `_from` and `_to` addresses with the same logic
```
smartcontracts: Logic will need to be updated because of the more advanced delegation logic. We should make the update logic explicit.

specs/experimental/alligator.md
```diff
@@ -0,0 +1,401 @@
...
+
+Users MUST be able to perform time & block constrained delegations. The subdelegation rules MUST include optional
+`notValidBefore` and `notValidAfter` fields to allow users to specify the time range in which a delegation is valid, both
+denominated as timestamps.
+
+### Redelegations
+
+Users MUST be able to redelegate their voting power to another address. If a user delegates their voting power to another
+address, this second address can further delegate the voting power to a third address. The first delegator MUST be able to
+limit the number of redelegations that can be performed for their delegated voting power by setting the `maxRedelegations`
```
smartcontracts: Is the convention to set this value to `type(maxRedelgations).max` to represent "infinite redelegations"?

specs/experimental/alligator.md
```diff
@@ -0,0 +1,401 @@
...
+### Dependence on Alligator
+
+As the `GovernanceToken` depends on the `Alligator` contract, the `Alligator` contract MUST be implemented so that it
+minimizes the risk of unexpected reverts during the transfer hook call. If the `Alligator` contract reverts,
+`GovernanceToken` transfers will be blocked. Additionally, the `GovernanceToken` MUST always use the `Alligator`'s
+delegation state if a user has been migrated.
+
+### Connection with GovernanceToken
+
+Similarly, the `Alligator` MUST always be in sync with the `GovernanceToken` contract via token transfers. If the
```
smartcontracts: Would be great if we could more clearly define what "in sync" means

specs/experimental/alligator.md
cairoeth: no particular reason before, but now due to the logic needed to re-calculate tokens (we need to use lists) we need to have the delegatee address inside the subdelegation rule

specs/experimental/alligator.md
tynes: Similarly to this suggestion, we could have a single paragraph that describes the following:

> This function MUST check if the `_delegator` or `_delegatee` addresses have been migrated by checking the `migrated` mapping
from its [storage](#storage). If either address has not been migrated, the `Alligator` MUST copy the delegation
and checkpoint data from the token contract to its own state. After copying the data, the `Alligator` MUST update
the `migrated` mapping to reflect that the address has been migrated.

Then link the paragraph in the multiple places where needed rather than copy/paste the text

specs/experimental/alligator.md
```diff
@@ -0,0 +1,401 @@
...
+Retrieves the number checkpoints for a given user address.
+
+```solidity
+function numCheckpoints(address _account) external view returns (uint32)
+```
+
+#### `delegates`
+
+Retrieves the delegatee with the highest voting power for a given user address. This function
+is intended to be used by the `GovernanceToken` contract to maximize for backwards compatibility.
```
tynes: Why is this `delegates()(address)` rather than `delegates()([]address)`? Given that it is plural

cairoeth: yes, good catch -- should return an array

smartcontracts: Original function doesn't return an array - if this is meant to replace the original function then this is correct. It's just a quirk of the naming of `ERC20Votes`.

specs/experimental/alligator.md
cairoeth: removed redelegations as it was out of scope

specs/experimental/alligator.md
cairoeth: added the `Delegation Validation` and `Migration` sections.

specs/experimental/alligator.md
cairoeth: wdyt of `delegateByToken` instead of `delegateByGovToken`? latter feels a bit too long. (we are now using `delegations` instead of `subdelegations`)

specs/experimental/alligator.md
cairoeth: per discord chat, `delegates` now returns an array of delegations -- this was updated.

specs/experimental/alligator.md
cairoeth: updated this section in the `afterTokenTransfer` to specify that voting powers must be adjusted, and what state needs to be updated.

specs/experimental/alligator.md
```diff
@@ -0,0 +1,444 @@
...
+```
+
+## Storage
+
+The `Alligator` contract MUST be able to store delegations and checkpoints. These storage variables MUST be
+defined as in the `GovernanceToken` and use the same types:
+
+```solidity
+// The maximum number of delegations allowed.
+uint256 public constant MAX_DELEGATIONS = 100;
```
tynes: These would be good to go in the constants table where the token address is defined

cairoeth: we are getting the token address from the Predeploys lib, so should we create a constants section in these specs with the `MAX_DELEGATIONS` and `DENOMINATOR` constants?

specs/governance/alligator.md
chuxinh: Got it, so basically you can have any combination of subdelegation you'd like to. In terms of the amount, does it also have to be 100% in the batch, or it can be lower (just for curiosity)

specs/governance/alligator.md
chuxinh: Thanks for the clarification here! Let me know if my understanding here is correct
`DelegateVotesChanged` which is the current event emitted in the OP token contract will continue emitting in the new interop OP standard when there's token transfer. And in the current spec, `Subdelegation` will only be emitted when it's a subdelegation but not a token transfer. So in order to figure out a delegate's actual voting power, it needs to be an add up from both events?

chuxinh: Also i see that there are two events, one called `Subdelegation` and the other is `Subdelegations`, given that `Subdelegations` seems to emit for batch subdelegation, what's it representing for before and after balance? 

Also wanna make sure event naming are clear enough with no duplications because previously for Alligator V5, third party providers such as Dune had to create custom decoding due to poor event naming. 
There are two events named `subdelegations` and both of these events have a field named `subdelegationsRules`

specs/experimental/alligator.md
```diff
@@ -0,0 +1,444 @@
+# Alligator
```
smartcontracts: I would highly recommend that we call this GovernanceDelegation or similar. Otherwise we'll be stuck with Alligator forever and it'll probably just confuse people.

specs/experimental/alligator.md
smartcontracts: Yeah we typically reserve a section for constants with a table like

| Name | Value |
| --------------- | ----- |
| `MY_CONSTANT` | 123 |

specs/governance/alligator.md
cairoeth: it can be any amount, as long as the sum of the %s doens't exceed 100

specs/governance/alligator.md
cairoeth: `DelegateVotesChanged` will be emitted every time the voting power of a user changes -- this includes for both token transfers and subdelegation updates.

so, for a token transfer, it will only emit `DelegateVotesChanged` because the subdelegations per se haven't changed.

when a subdelegation is created/updated/removed, it will event the `Subdelegation` event and `DelegateVotesChanged` for every voting power change. this allows to always have visibility of voting power changes, for both token transfers and subdelegation updates.

specs/experimental/alligator.md
cairoeth: updated https://github.com/ethereum-optimism/specs/pull/247/commits/fef1220fb763c6f0e25b2d71f5f76ba21391445f

specs/experimental/alligator.md
cairoeth: added in https://github.com/ethereum-optimism/specs/pull/247/commits/fef1220fb763c6f0e25b2d71f5f76ba21391445f

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,449 @@
...
+
+Delegates voting power to another address (delegatee) with the given delegation parameters. This function
+is intended to be called by users that require advanced delegation of the `GovernanceToken`.
+
+```solidity
+function delegate(Delegation _delegation) external
+```
+
+This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for `msg.sender`
+and the `_delegatee` address in the delegation. Afterwards, the function MUST call the [`_delegate`](#_delegate) with
```
smartcontracts:
```suggestion
and the `_delegatee` address in the delegation. Afterwards, the function MUST call the [`_delegate`](#_delegate) function with
```

Brief suggestion. Also, as an aside, this is probably good enough for now but in the future it's typically simpler if we define the behavior of these functions in terms of their effects rather than the specific function calls they make to achieve those effects. In this case I believe that the function definition for `_delegate` does a lot of the heavy lifting on that front so it's fine but just useful to keep in mind for future specs.

Edit: I wrote the above comment before reading the rest of the spec, remainder of the spec does a really good job at this, I don't think it makes a huge difference so no need to change the language here.

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,449 @@
...
+Delegates 100% of the token voting power of an address (delegator) to another address (delegatee), mimicking the behavior
+of the `ERC20Votes`'s `delegate` function for backwards compatibility. This function MUST be callable only by the
+`GovernanceToken` contract as part of its `delegate` and `delegateBySig` functions.
+
+```solidity
+function delegateFromToken(address _delegator, address _delegatee) external
+```
+
+This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for the `_delegator`
+and `_delegatee` addresses. Afterwards, the function MUST call the [`_delegate`](#_delegate) with the `_delegator` and
```
smartcontracts:
```suggestion
and `_delegatee` addresses. Afterwards, the function MUST call the [`_delegate`](#_delegate) function with the `_delegator` and
```

Probably same thing in other places too, will let you handle that.

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,449 @@
...
+
+Migrates the delegation state of the given accounts from the `GovernanceToken` to the `GovernanceDelegation` contract. This
+function MUST iterate over the list of `_accounts` addresses and apply the logic specified in the
+[Migration](#migration) section.
+
+```solidity
+function migrateAccounts(address[] calldata _accounts) external
+```
+
+#### `_delegate`
```
smartcontracts, line 144: Yeah basic feedback on this for future ref would just be that the spec should really cover external interface and expected function behavior. In some sense what this section is really doing is providing common function behavior for a number of external functions. Readers of the spec shouldn't really care that those functions specifically call an internal function called `_delegate` but they *do* care that those external functions implement the function behavior being described here. I think this is perfectly fine for now but in a future spec cleanup we can probably streamline this a bit.

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,449 @@
...
+
+Delegates voting power from the `_delegator` to the specified delegates in `_newDelegations`. This function MUST
+override any existing delegations of `_delegator`.
+
+```solidity
+function _delegate(address _delegator, Delegation[] memory _newDelegations) internal virtual
+```
+
+The `_delegate` function MUST first check that the length of `_newDelegations` does not exceed `MAX_DELEGATIONS`.
```
smartcontracts, line 153: We should be considering the worst-case impact of an unoptimized version of this update functionality. We probably want to run the numbers on the total cost to update voting power for someone with the max number of delegations and see that it doesn't revert, at a minimum.

We should also then figure out how this would impact people trying to transfer OP to other parties since I could somewhat troll you by creating a large number of complex delegations that you're forced to pay for. I could even frontrun you, increase the number of delegations, and then cause some transaction of yours to revert because you're updating my vote tally. It may be worth only updating the vote tally of the sender and not the recipient to avoid this case.

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,449 @@
...
+function numCheckpoints(address _account) external view returns (uint32)
+```
+
+#### `delegates`
+
+Retrieves the delegations of a given user address sorted in descending order by voting power.
+This function is intended to be used by the `GovernanceToken` contract to maximize for backwards compatibility.
+
+```solidity
+function delegates(address _account) external view returns (Delegation[] memory)
```
smartcontracts: Technically not backwards compatible, for the sake of compatibility, maybe we make this match the original function signature and introduce a new function that returns the full array? Just in the case that someone wants to point their old code at the Alligator and still have it work.

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,449 @@
...
+
+#### `DelegationCreated`
+
+MUST trigger when an account delegates voting power to another address (delegatee).
+
+```solidity
+event DelegationCreated(address indexed account, Delegation delegation);
+```
+
+#### `DelegationsCreated`
```
smartcontracts: Worth asking data team if having a separate event is better or easier to just emit N events

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,449 @@
...
+  uint208 amount;
+}
+```
+
+| Name        | Type      | Description                                                    |
+|-------------|-----------|----------------------------------------------------------------|
+| `delegatee` | `address` | The address for the delegation adjustment.                     |
+| `amount`    | `uint208` | The amount of voting power to add or remove.                   |
+
+### `Op`
```
smartcontracts: Where is this used?

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,449 @@
...
+  Relative
+}
+```
+
+| Name        | Number | Description                                                                                 |
+|-------------|--------|---------------------------------------------------------------------------------------------|
+| `Absolute`  | `0`    | The amount of votes delegated is fixed and denominated in the `GovernanceToken`'s decimals  |
+| `Relative`  | `1`    | The amount of votes delegated is relative and denominated in percentages.                   |
+
+### `DelegationAdjustment`
```
smartcontracts: Where is this used?

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,449 @@
...
+reverts, `GovernanceToken` transfers will be blocked. Additionally, the `GovernanceToken` MUST always use the
+`GovernanceDelegation`'s delegation state if a user has been migrated.
+
+### Connection with GovernanceToken
+
+Similarly, the `GovernanceDelegation` MUST always process token transfers from the `GovernanceToken` contract correctly to
+stay in sync with token balances. If the `GovernanceDelegation` contract is not in sync with the `GovernanceToken` contract,
+the voting power of users MAY be incorrect or outdated.
+
+## Future Considerations
```
smartcontracts: I think we can drop this section, we'll be working on that spec right after this one anyway.

specs/governance/alligator.md
chuxinh: That makes sense. I think it's okay to keep it for now. so for someone who does not care about if the change is coming from subdelegation or not, they can continue using `DelegateVotesChanged`.

Also curious how it's treating the historical records since at the moment, we cannot technically derive the actual voting power without computing the subdelegation rule. How does it work to derive the voting power balance when this goes live?

specs/experimental/gov-delegation.md
cairoeth: @chuxinh any thoughts on this? we could have just one event called `DelegationsCreated`

cairoeth: updated to just use one event as they are very similar -- it will make it easier to index and for the implementation too.

specs/experimental/gov-delegation.md
tynes: Agreed with this, defining an internal function like this is in the weeds with implementation details

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,440 @@
...
+```
+
+| Name        | Number | Description                                                                                 |
+|-------------|--------|---------------------------------------------------------------------------------------------|
+| `Absolute`  | `0`    | The amount of votes delegated is fixed and denominated in the `GovernanceToken`'s decimals  |
+| `Relative`  | `1`    | The amount of votes delegated is relative and denominated in percentages.                   |
+
+### `DelegationAdjustment`
+
+`DelegationAdjustment` act as temporary storage when adjusting the delegation state of an account. This struct is defined
```
smartcontracts: If this is an internal type it's best not to define it. We only need to define things that are part of the public interface.

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,440 @@
...
+
+Retrieves the voting power of a given user address at a block.
+
+```solidity
+function getPastVotes(address _account, uint256 _blockNumber) external view returns (uint256)
+```
+
+#### `getPastTotalSupply`
+
+Retrieves the total supply of the `GovernanceToken` at a given block.
```
smartcontracts, line 214: Let's specify that it's an L1 block number

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,440 @@
...
+
+#### `delegations`
+
+Retrieves the delegations of a given user address sorted in descending order by voting power.
+
+```solidity
+function delegations(address _account) public view returns (Delegation[] memory)
+```
+
+### Events
```
smartcontracts, line 236: If delegations are keyed by L1 block number then we should include the L1 block number as part of these events. L2 block number can already be retrieved from events by default but I suspect having the L1 block number in there will be useful too.

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,440 @@
...
+`afterTokenTransfer` function after a token transfer. This enables the `GovernanceDelegation` contract to consume
+the hook and update its delegation and checkpoint mappings accordingly. If either address involved in the transfer
+(`_from_` or `_to`) has not been migrated to the `GovernanceDelegation` contract, the contract copies the address'
+checkpoint data from the `GovernanceToken` to its own state.
+
+## Interface
+
+### Core Functions
+
+#### `delegate`
```
smartcontracts, line 69: Does this function overwrite the set of delegations or add to it?

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,440 @@
...
+override any existing delegations of `_delegator`.
+
+```solidity
+function _delegate(address _delegator, Delegation[] memory _newDelegations) internal virtual
+```
+
+The `_delegate` function MUST first check that the length of `_newDelegations` does not exceed `MAX_DELEGATIONS`.
+Then, it MUST calculate the voting power of all active delegations of the `_delegator`, and calculate the voting
+power adjustments for the new delegatee set using `_newDelegations`. Afterwards, the function MUST update
+the voting power of the current and new set of delegatees. Throughout this process, the function MUST use the
```
smartcontracts: Use of `DelegationAdjustment` and `Op` here seems like an implementation detail and can probably be removed, unless that's part of the public interface. From what I can see this isn't exposed publicly and therefore can be removed.

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,440 @@
...
+`DelegationAdjustment` and `Op` types to store voting power adjustments and the operation to perform (add or
+subtract voting power).
+
+The function MUST then sort the new delegatee set in descending order, and store the delegations as part of the
```
smartcontracts, line 158: Just to confirm, this function overwrites all existing delegations, correct?

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,440 @@
...
+    A2 --> |Return data| A1
+```
+
+## Delegation Validation
+
+When applying a new delegation set as part of the [`_delegate`](#_delegate) function, the `GovernanceDelegation` MUST check
+that:
+
+- The length of the `_newDelegations` array DOES NOT exceed `MAX_DELEGATIONS`.
+- The sum of `amount` in relative delegations DOES NOT exceed the `DENOMINATOR`, and each `amount` is greater than 0.
```
smartcontracts, line 380: Can you have both relative and absolute delegations in the same set? If so, what is the invariant that must hold globally?

specs/experimental/gov-delegation.md
cairoeth: @smartcontracts i think would be part of the interop specs

specs/experimental/gov-delegation.md
cairoeth: overwites, added clarification

specs/experimental/gov-delegation.md
cairoeth: correct, it's commented on L142 as well.

specs/experimental/gov-delegation.md
cairoeth: yep, you can have both in a set.

updated to clarify invariant:

> If a delegation set includes both absolute and relative delegations, the relative delegations MUST be applied first and the absolute delegations second with the remaining voting power. Therefore, the remaining voting power MUST be greater than or equal than the sum of the absolute delegations.

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,407 @@
...
+## Overview
+
+| Constant | Value                                        |
+|----------|----------------------------------------------|
+| Address  | `0x4200000000000000000000000000000000000043` |
+
+The `GovernanceDelegation` contract implements advanced delegation for the [`GovernanceToken`](gov-token.md).
+Advanced delegation allows for partial and relative delegations of voting power.
+
+The `GovernanceDelegation` contract migrates the delegation state from the `GovernanceToken` to itself
```
tynes: I still don't think that this is so important that it needs to be highlighted in the first paragraph. It won't make sense in the context of new chains in the interop world. It only makes sense in context of backwards compatibility. I think that an overview of what advanced delegation is makes more sense in this section

cairoeth: agreed, updated the section in https://github.com/ethereum-optimism/specs/pull/247/commits/5997c515116651381b65acc4173b648657b9c296

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,406 @@
...
+#### `_delegate`
+
+Delegates voting power from the `_delegator` to the specified delegates in `_newDelegations`. This function MUST
+override any existing delegations of `_delegator`.
+
+```solidity
+function _delegate(address _delegator, Delegation[] memory _newDelegations) internal virtual
+```
+
+The `_delegate` function MUST first check that the length of `_newDelegations` does not exceed `MAX_DELEGATIONS`.
```
tynes, line 153: `MAX_DELEGATIONS` should be defined in the constants table

specs/experimental/gov-delegation.md
tynes: Why does L1 block number make more sense here? Isn't it an L2 native token?

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,406 @@
...
+// Total supply checkpoints of the GovernanceToken.
+ERC20Votes.Checkpoint[] internal _totalSupplyCheckpoints;
+```
+
+## Constants
+
+The `GovernanceDelegation` contract MUST define the following constants:
+
+| Name                     | Value     | Description                                              |
+|--------------------------|-----------|----------------------------------------------------------|
```
tynes: Noticing there are 2 constants tables, one at the top and one here

tynes: We should combine these two tables

specs/experimental/gov-delegation.md
cairoeth: yep we already have it defined here https://github.com/ethereum-optimism/specs/blob/5997c515116651381b65acc4173b648657b9c296/specs/experimental/gov-delegation.md#constants, is this what you refer to?

specs/experimental/gov-delegation.md
cairoeth: so combining the storage table and the constants table into one?

cairoeth: ohh i think maybe you mean this one in the overview section https://github.com/ethereum-optimism/specs/blob/5997c515116651381b65acc4173b648657b9c296/specs/experimental/gov-delegation.md#overview?

specs/experimental/gov-delegation.md
chuxinh: missed this comment. Are you referring to the above two events? I see the difference is that one is delegating to one address and the other is multiple ones. If there's no issue of combining them that should be fine to merge into one.

Also is this only created when they first delegate or every time?

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,406 @@
...
+
+This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for `msg.sender`
+and the `_delegatee` address in the delegation. Afterwards, the function MUST call the [`_delegate`](#_delegate) function
+with the `msg.sender` and `_newDelegations` parameters.
+
+At the end, the `delegate` function MUST emit a `DelegationCreated` event with the given function parameters.
+
+#### `delegateFromToken`
+
+Delegates 100% of the token voting power of an address (delegator) to another address (delegatee), mimicking the behavior
```
smartcontracts, line 87: Minor comment, work explicitly calling out that this overwrites all existing delegations.

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,406 @@
...
+
+This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for the `_delegator`
+and `_delegatee` addresses. Afterwards, the function MUST call the [`_delegate`](#_delegate) function with the `_delegator`
+and basic delegation parameters.
+
+At the end, the `delegateFromToken` function MUST emit a `DelegationCreated` event with the given function parameters.
+
+#### `delegateBatched`
+
+Delegates voting power to multiple addresses (delegatees) using the delegation array. This function is intended to be
```
smartcontracts, line 104: Worth calling out that this overwrites existing delegations.

specs/experimental/gov-delegation.md
smartcontracts: Going to resolve this since it's a discussion best left for the OP interop design doc. Problem statement is that w need to be able to coordinate voting power from different chains which will have different L2 block numbers at the same time. We therefore need to either use timestamp or L1 block number. There are arguments for each. Since this is not relevant to this specific spec, I will close this comment for now.

specs/experimental/gov-delegation.md
smartcontracts: I agree, closing this comment for now and will discuss in the OP interop design doc.

specs/experimental/gov-delegation.md
smartcontracts: Yes better to combine everything into the constants table at the beginning of the doc since that's our current format.

cairoeth: updated in https://github.com/ethereum-optimism/specs/pull/247/commits/b7288ea73d8ce7eaa0606d9aba624e0557cd213f

specs/experimental/gov-delegation.md
```diff
@@ -0,0 +1,402 @@
...
+- [Delegation Validation](#delegation-validation)
+- [Security Considerations](#security-considerations)
+  - [Dependence on GovernanceDelegation](#dependence-on-governancedelegation)
+  - [Connection with GovernanceToken](#connection-with-governancetoken)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+The `GovernanceDelegation` contract implements advanced delegation for the [`GovernanceToken`](gov-token.md).
```
tynes: This link is broken

tynes (approved):

PR #249, opened 2024-06-18
config: update batch inbox address convention to support 256 bit chain IDs

### LINKED ISSUE #94: Compatibility between interop chain ID and batch inbox convention

mds1:
Closes https://github.com/ethereum-optimism/specs/issues/94, see the discussion in that issue for context


specs/protocol/configurability.md
```diff
@@ -39,7 +39,7 @@ These requirements are currently a draft, pending governance approval.
...
-| [Batch Inbox address](https://github.com/ethereum-optimism/optimism/blob/c927ed9e8af501fd330349607a2b09a876a9a1fb/packages/contracts-bedrock/src/L1/SystemConfig.sol#L176)                   | L1 address where calldata/blobs are posted (see [Batcher Transaction](../glossary.md#batcher-transaction)).                  | Static | Current convention is `0xff000...000{chainId}`. | Follow convention if possible. A future hardfork plans to migrate Batch Inbox Addresses to better support `uint256` sized chain IDs. |
+| [Batch Inbox address](https://github.com/ethereum-optimism/optimism/blob/c927ed9e8af501fd330349607a2b09a876a9a1fb/packages/contracts-bedrock/src/L1/SystemConfig.sol#L176)                   | L1 address where calldata/blobs are posted (see [Batcher Transaction](../glossary.md#batcher-transaction)).                  | Static | Current convention is <code>versionByte &vert;&vert; keccak256(bytes32(chainId))[:19]</code>, where <code>&vert;&vert;</code> denotes concatenation, `versionByte` is `0x00`, and `chainId` is a `uint256`. | It is recommended, but not required, to follow this convention. |
```
mds1, line 42: Have to use `<code>` to write the pipes since otherwise it wouldn't renderly properly in a table, see https://github.com/rust-lang/mdBook/issues/637

tynes:
My one comment is that we do not follow the convention ourselves and it differs from the convention that we previously used

tynes (approved):

PR #254, opened 2024-06-21
Adding messageHash to ExecutingMessage

0xng:
**Description**

This PR introduces a small change to the ExecutingMessage event in the `CrossL2Inbox` predeploy. Instead of emitting the entire message, it emits the hash of it as was suggested by @skeletor-spaceman.

**Additional context**

This change, although small, may affect other specs. I was not certain whether to include changes to multiple specs here or to have people more familiar with other specs push the respective changes (should this change be accepted) as well.

As an example, the sequencer.md file contains pseudocode showing how to check the existence of a log based on the Identifier. That code deserializes the `log.data` to get the entire message. After this change, it will only get the `messageHash`, and that code's further check should most likely compare the `messageHash` with the hash of `encode(log[0])`


tynes:
This seems like a great improvement, going to want to double check with @ajsutton to make sure this doesn't impact him

ajsutton:
The database always works on payload hash anyway so I don't see any issues with this.

tynes:
We will need to update this section too: https://github.com/ethereum-optimism/specs/blob/main/specs/interop/sequencer.md#direct-dependency-confirmation

The `abi.decode` part will need to decode a hash instead of the message

specs/interop/predeploys.md
```diff
@@ -94,19 +94,21 @@ The `ExecutingMessage` event represents an executing message. It MUST be emitted
 to `executeMessage`.
 
 ```solidity
-event ExecutingMessage(bytes,bytes);
+event ExecutingMessage(bytes,bytes32);
 ```
```
tynes, line 98: Thoughts on making this `event ExecutingMessage(bytes32,bytes)` to have the fixed size in the front? We could also make it `indexed` if we want but then would need to modify https://github.com/ethereum-optimism/specs/blob/main/specs/interop/sequencer.md#direct-dependency-confirmation more to handle that, ie pull the hash from the topics[1] rather than decoding the log.data for it

0xng: The order change is much cleaner and makes the decoding way more intuitive in my opinion. I have not added the `indexed` keyword yet as the decoding seems straightforward, but I can add it if needed and change both the solidity pseudocode and python's.

Should we also change `logHash` to `msgHash` in the solidity pseudocode? We are using `msg` in all other places instead of `log`.

tynes: My personal style is shorter names like just `hash` but not super opinionated, i think @mds1 would like more descriptive names, if `msg` is used everywhere then `msgHash` could make sense

mds1: Yea my ideal is short + descriptive and I think `logHash` and `msgHash` both fit that with only an extra 3 chars over `hash`. Also not super opinionated here so happy to defer :)

ajsutton (commented):
For the record, this makes sense to me. I'm not aware of any reason we'd need or want the claimed payload off-chain. The original log is still available via the info in the Identifier if the message is valid, and I don't see any reason to care about the original claimed payload if the message is invalid.

Will leave it to @tynes to give final approval though since he has more context on this than I do.

ajsutton:
Can we/should we use this PR to switch to outputting the `Identifier` as a struct rather than first serialising it to bytes? Because the field isn't indexed it should work fine and makes parsing the event much simpler and easier to test that its in sync with the contracts (see also https://github.com/ethereum-optimism/optimism/pull/11104)

specs/interop/predeploys.md
```diff
@@ -94,19 +94,21 @@ The `ExecutingMessage` event represents an executing message. It MUST be emitted
 to `executeMessage`.
 
 ```solidity
-event ExecutingMessage(bytes,bytes);
+event ExecutingMessage(bytes32,bytes);
```
mds1: What is the rationale for this change? It saves a little bit of gas but adds an extra RPC query for consumers of this to find the original log data, as they must look it up from the `Identifier`

specs/interop/predeploys.md
```diff
@@ -133,7 +135,7 @@ function executeMessage(Identifier calldata _id, address _target, bytes calldata
 
     require(success);
 
-    emit ExecutingMessage(abi.encode(_id), _msg);
+    emit ExecutingMessage(keccak256(_msg), abi.encode(_id));
```
mds1: Agree with @ajsutton above that we should just emit the `Identifier` struct directly instead of ABI encoding it to bytes first

specs/interop/predeploys.md
mds1: nit but let's name the arguments in this event definition for clarity/explicitness

skeletor-spaceman: this is a good question @mds1 .
the rationale is the following:
the `ExecutingMessage` is designed to be used by the Interop service, which does not really care about the plain data.
for other off-chain agents wanting to consume the `bytes _data` argument, they can, as you said, do 1 RPC call to the Identifier origin domain, or they could also trace the tx (which is more complex).
remember data could get quite big, since it can be dynamically generated on-chain.

you raise a good point which is "how can we make off-chain devs life easier when having to deal with multiple RPCs", especially when we might end up with bundled txs that reference multiple Identifiers from different domains.
@fainashalts will be able to provide more insight on this last q' :)

specs/interop/predeploys.md
```diff
@@ -94,19 +94,20 @@ The `ExecutingMessage` event represents an executing message. It MUST be emitted
 to `executeMessage`.
 
 ```solidity
-event ExecutingMessage(bytes,bytes);
+event ExecutingMessage(bytes32 msgHash, Identifier identifier);
```
tynes: Small nit: I would like to make this indexed, then this is good to merge

specs/interop/predeploys.md
tynes: I ended up calling it `msgHash` in https://github.com/ethereum-optimism/optimism/pull/11320

tynes (approved):

PR #256, opened 2024-06-24
interop: Superchain WETH

smartcontracts:
**Description**
Specification for the Superchain WETH contract. Refer to [this PR](https://github.com/ethereum-optimism/design-docs/pull/25/files) in the design repo for context.

### DIFF
--- a/specs/interop/superchain-weth.md
+++ b/specs/interop/superchain-weth.md
@@ -1,0 +1,126 @@
+# Superchain WETH
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Motivation and Constraints](#motivation-and-constraints)
+  - [Handling native assets other than ETH](#handling-native-assets-other-than-eth)
+  - [Minimizing protocol complexity](#minimizing-protocol-complexity)
+- [Constants](#constants)
+- [SuperchainWETH](#superchainweth)
+  - [Invariants](#invariants)
+    - [`deposit`](#deposit)
+    - [`withdraw`](#withdraw)
+    - [`sendERC20`](#senderc20)
+    - [`sendERC20To`](#senderc20to)
+    - [`relayERC20`](#relayerc20)
+- [ETHLiquidity](#ethliquidity)
+  - [Invariants](#invariants-1)
+    - [Global Invariants](#global-invariants)
+    - [`burn`](#burn)
+    - [`mint`](#mint)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+Superchain WETH is a version of the standard WETH contract that allows ETH to be interoperable across the Superchain.
+Superchain WETH treats ETH as an ERC-20 token for interoperability and avoids native ETH support.
+Superchain WETH also introduces a liquidity contract used to provide liquidity for native ETH across chains.
+
+## Motivation and Constraints
+
+ETH is the native asset of Ethereum and has by extension also become the native asset of many Ethereum L2 blockchains.
+In its role as a native asset, ETH can be used to pay for transaction fees and can be transferred from account to
+account via calls with attached value. ETH plays a significant role in the economics of most L2s and any protocol that
+enables interoperability between chains must be able to account for ETH.
+
+### Handling native assets other than ETH
+
+Not all chains using the OP Stack use ETH as the native asset. We would like these chains to be able to interoperate
+with chains that *do* use ETH as a native asset. Certain solutions that might work when all chains use ETH as a native
+asset begin to break down when alternative native assets are introduced. For example, a protocol that burns the native
+asset on one chain and mints it on another will work if both chains use the same native asset but will obviously fail if
+either chain uses a different native asset.
+
+### Minimizing protocol complexity
+
+Support for native ETH opens the door to unnecessary complexity. Any solution to this problem should aim to minimize the
+amount of protocol code required to support native ETH. This generally points towards an app-layer solution if possible
+but does not preclude a protocol-layer solution as long as we minimize implementation size.
+
+## Constants
+
+| Name                     | Value                                        |
+| ------------------------ | -------------------------------------------- |
+| `SuperchainWETH` Address | `0x4200000000000000000000000000000000000024` |
+| `ETHLiquidity` Address   | `0x4200000000000000000000000000000000000025` |
+
+## SuperchainWETH
+
+### Invariants
+
+#### `deposit`
+
+- Reverts if triggered on a chain that does not use ETH as a native token.
+
+#### `withdraw`
+
+- Reverts if triggered on a chain that does not use ETH as a native token.
+
+#### `sendERC20`
+
+- Reverts if attempting to send more than the sender's available balance.
+- Reduce's the sender's balance by the sent amount.
+- Emits a transfer event from sender to null address for the sent amount.
+- Emits a `SendERC20` event.
+- Burns liquidity by sending the sent amount of ETH into the `ETHLiquidity` contract if native token is ETH.
+  - Must not revert.
+- Sends a message to `SuperchainWETH` on the recipient chain finalizing a send of WETH.
+  - Must not revert.
+
+#### `sendERC20To`
+
+- All invariants of `sendERC20`.
+- Message sent to `SuperchainWETH` on recipient chain includes `to` address as recipient.
+
+#### `relayERC20`
+
+- Reverts if called by any address other than the `L2ToL2CrossDomainMessenger`.
+- Reverts if `crossDomainMessageSender` is not the `SuperchainWETH` contract.
+- Mints liquidity from the `ETHLiquidity` contract if native token is ETH.
+  - Must not revert.
+- Increases the recipient's balance by the sent amount.
+- Emits a transfer event from null address to recipient for the sent amount.
+- Emits a `RelayERC0` event.
+
+## ETHLiquidity
+
+### Invariants
+
+#### Global Invariants
+
+- Initial balance must be set to `type(uint248).max` (wei). Purpose for using `type(uint248).max` is to guarantees that
+the balance will be sufficient to credit all use within the `SuperchainWETH` contract but will never overflow on calls
+to `burn` because there is not ETH in the total ETH supply to cause such an overflow. Invariant that avoids overflow is
+maintained by  `SuperchainWETH` but could theoretically be broken by some future contract that is allowed to integrate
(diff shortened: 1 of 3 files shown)


specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,107 @@
+# Superchain WETH
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
```
tynes, line 25: We will want a constants table that includes the addresses for the 2 new predeploys and an overview section similar to what you put together in the design doc, just framing the problem and the solution at a high level. We will want to mention the custom gas token chain compatibility

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,107 @@
...
+  function bridgeERC20(uint256 wad, uint256 chainId) external {
+    _bridgeERC20To(msg.sender, wad, chainId, true);
+  }
+
+  function bridgeERC20To(address to, uint256 wad, uint256 chainId) external {
+    _bridgeERC20To(to, wad, chainId, true);
+  }
+
+  function bridgeERC20Wrapped(uint256 wad, uint256 chainId) external {
```
tynes: We will want to be mindful of the ABI as we will want to implement superchain erc20, which the latest is here: https://github.com/ethereum-optimism/specs/pull/71#discussion_r1631562710

There may be additional functions that would be nice to have but we should be careful about expanding the ABI too much as it increases loc and therefore chances of bugs

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,107 @@
...
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+Superchain WETH is a version of the standard WETH contract that allows ETH to be interoperable across the Superchain.
+Superchain WETH treats ETH as an ERC-20 token for interoperability and avoids native ETH support.
+Superchain WETH also introduces a liquidity contract used to provide liquidity for native ETH across chains.
+
+## SuperchainWETH
+
```
tynes, line 59: We can leave out implementation details of superchain erc20 as that spec would be expected to cover those, i think defining the invariants around the various functions would be good in the context of weth

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,161 @@
...
+        // Great.
+      } catch {
+        // Just leave it wrapped.
+      }
+    }
+  }
+
+  function unwrap(address to, uint256 wad) public {
+    if (msg.sender != address(this)) revert Unauthorized();
+
```
tynes: Why do we need `unwrap` when we have `withdraw`?

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,161 @@
...
+      liquidity.mint(wad);
+    }
+
+    // Mint the WETH balance.
+    balanceOf[to] += wad;
+    emit Bridge(to, wad)
+
+    // If user wants to unwrap, try unwrapping.
+    if (unwrap) {
+      try this.unwrap(to, wad) {
```
tynes: I think it'll be less error prone to put this in an if/else rather than a try catch

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,161 @@
...
+  function sendERC20To(address to, uint256 wad, uint256 chainId) external {
+    _sendWETH(to, wad, chainId, true);
+  }
+
+  function finalizeSendWETH(address to, uint256 wad, bool unwrap) external {
+    if (msg.sender != address(messenger)) revert Unauthorized();
+    if (messenger.crossDomainMessageSender() != address(this)) revert Unauthorized();
+
+    // Mint the required liquidity.
+    if (l1Block.isCustomGasToken()) {
```
tynes: This needs to be `(!l1Block.isCustomGasToken())`

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,161 @@
...
+TODO
+
+### Implementation Sketch
+
+```solidity
+contract ETHLiquidity {
+  address internal weth = Predeploys.SUPERCHAIN_WETH;
+  L1Block internal l1Block = L1Block(Predeploys.L1_BLOCK);
+
+  function burn() external payable {
```
tynes: This function is not used, it existed to try to make `totalSupply` of weth more accurate

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,161 @@
...
+
+  function sendERC20(uint256 wad, uint256 chainId) external {
+    _sendWETH(msg.sender, wad, chainId, true);
+  }
+
+  function sendERC20To(address to, uint256 wad, uint256 chainId) external {
+    _sendWETH(to, wad, chainId, true);
+  }
+
+  function finalizeSendWETH(address to, uint256 wad, bool unwrap) external {
```
tynes: This breaks the superchain erc20 interface technically, we would want `finalizeSendERC20`. If you are strongly opinionated about having the unwrap ability, perhaps we could extend with `finalizeSendETH` which assumes unwrap is true?

specs/interop/superchain-weth.md
smartcontracts: Problem is if unwrap reverts for any reason then the whole thing fails and the tokens get lost.

specs/interop/superchain-weth.md
smartcontracts: `withdraw` is for `msg.sender` while `unwrap` allows specifying a `to` address so that we can basically trigger a withdrawal on behalf of another user.

specs/interop/superchain-weth.md
smartcontracts: Honestly at that point it might just be easier to give `SuperchainWETH` the massive ETH supply and call it a day. Is there a specific class of issues that we're avoiding with `ETHLiquidity`?

specs/interop/superchain-weth.md
tynes: ah ok this makes sense

specs/interop/superchain-weth.md
tynes: I definitely prefer to have 2 contracts here but am open to changing. My reasoning is as follows:
- The design [here](https://oplabs.notion.site/Public-Interoperable-ETH-Prototype-3fb07a8310304eeb83ba217f42932eac?pvs=4) already shows it works
- Preserve ~accurate `totalSupply` which is nice for data team
- Prevent unknown unknown bugs with very explicit conditions for being able to mint
- Generally things should do 1 thing well and be composed together
- Enable `ETHLiquidity` to be a primitive that can be used by the system in other ways in the future, this is about flexibility with future upgrades

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,168 @@
...
+    _mint(msg.sender, wad);
+  }
+}
+```
+
+## ETHLiquidity
+
+### Invariants
+
+TODO
```
tynes: Still need invariants for `ETHLiquidity`

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    // Mint to user's balance.
+    _mint(msg.sender, wad);
+  }
+}
+```
+
+## ETHLiquidity
+
```
tynes, line 97: We want to mention that its seeded with `type(uint248).max` ether

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,168 @@
...
+but does not preclude a protocol-layer solution as long as we minimize implementation size.
+
+### We want to maximize functionality
+
+Any solution should strive to maximize the functionality of the underlying message-passing protocol. Support for native
+ETH should not unnecessarily restrict the interop protocol.
+
+### Streamlining the user experience
+
+An ideal solution would make it possible to streamline the user experience such that ETH transfers between chains look
```
tynes: We probably want to delete this section now

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,168 @@
...
+Superchain WETH also introduces a liquidity contract used to provide liquidity for native ETH across chains.
+
+## Motivation and Constraints
+
+ETH is the native asset of Ethereum and has by extension also become the native asset of many Ethereum L2 blockchains.
+In its role as a native asset, ETH can be used to pay for transaction fees and can be transferred from account to
+account via calls with attached value. ETH plays a significant role in the economics of most L2s and any protocol that
+enables interoperability between chains must be able to account for ETH.
+
+### We want to handle native assets other than ETH
```
tynes: small nit: usually not best to say things like "we want" in specs, instead say something like "Enable Interop between chains with different gas paying assets"

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,168 @@
...
+asset on one chain and mints it on another will work if both chains use the same native asset but will obviously fail if
+either chain uses a different native asset.
+
+### We want to minimize protocol complexity
+
+Support for native ETH opens the door to unnecessary complexity. Any solution to this problem should aim to minimize the
+amount of protocol code required to support native ETH. This generally points towards an app-layer solution if possible
+but does not preclude a protocol-layer solution as long as we minimize implementation size.
+
+### We want to maximize functionality
```
tynes: We can delete this section now

tynes:
This needs a spec lint

specs/interop/superchain-weth.md
```diff
@@ -0,0 +1,199 @@
...
+#### `finalizeSendERC20`
+
+- Reverts if called by any address other than the `L2ToL2CrossDomainMessenger`.
+- Reverts if `crossDomainMessageSender` is not the `SuperchainWETH` contract.
+- Mints liquidity from the `ETHLiquidity` contract if native token is ETH.
+  - Must not revert.
+- Increases the recipient's balance by the sent amount.
+- Emits a transfer event from null address to recipient for the sent amount.
+
+### Implementation Sketch
```
tynes: I think that we can delete this implementation sketch now that we have the actual contract implemented. Some of the other specs still have source code in them for pseudo code purposes of specific functions

specs/interop/token-bridging.md
```diff
@@ -0,0 +1,97 @@
+# Token Bridging
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
```
tynes: Did you mean to have this file in this PR?

tynes:
Looks like a linting issue

tynes (approved):

PR #257, opened 2024-06-24
Interop: SuperchainERC20 Standard

0xParti:
**Description**
The `SuperchainERC20` standard allows to interact with ERC20s across the Superchain.

This is a follow-up PR to https://github.com/ethereum-optimism/specs/pull/71.

We would love to
- Hear your feedback on the design decisions. 
- Discuss function and variable naming.

We will include invariants and create a design docs to discuss Factories for `SuperchainERC20`s soon.

### DIFF
--- a/specs/interop/token-bridging.md
+++ b/specs/interop/token-bridging.md
@@ -28,7 +28,8 @@ Without a standardized security model, bridged assets may not be fungible with e
 The `SuperchainERC20` unifies ERC20 token bridging to make it fungible across the Superchain.
 It builds on top of the messaging protocol, as the most trust minimized bridging solution.
 
-Unlike other standards, such as [xERC20](https://www.xerc20.com/), where users interact with a bridge possessing mint and burn
+Unlike other standards, such as [xERC20](https://www.xerc20.com/),
+where users interact with a bridge possessing mint and burn
 privileges on cross-chain enabled tokens, this approach allows users to call methods directly on the token contract.
 
 ## Interface
@@ -52,7 +53,8 @@ sendERC20(address _to, uint256 _amount, uint256 _chainId, bytes _data)
 
 #### `relayERC20`
 
-Process incoming messages IF AND ONLY IF if they where initiated by the same contract (token) address on a different chain
+Process incoming messages IF AND ONLY IF if they where initiated
+by the same contract (token) address on a different chain
 and come from the `L2ToL2CrossChainMessenger` in the local chain.
 It will mint `_amount` to address `_to`, as defined in `sendERC20`.
 If the message contained `_data`,
@@ -137,7 +139,8 @@ function relayERC20(address _to, uint256 _amount, bytes memory _data) external {
 }
 ```
 
-Note: Some other naming options that were considered were `bridgeERC20`, `send()`, `xTransfer()`, `Transfer` (with a chainId parameter).
+Note: Some other naming options that were considered were
+`bridgeERC20`, `send()`, `xTransfer()`, `Transfer` (with a chainId parameter).
 
 ## Invariants
 
@@ -175,8 +178,10 @@ Besides the ERC20 invariants, the SuperchainERC20 will require the following int
 
 ### Factory
 
-A token factory predeploy can ensure that `SuperchainERC20` tokens can be permissionlessly and deterministically deployed.
-Without a deterministic deployment scheme, maintaining a mapping of the token addresses between all of the chains will not be scalable.
+A token factory predeploy can ensure that `SuperchainERC20`
+tokens can be permissionlessly and deterministically deployed.
+Without a deterministic deployment scheme, maintaining a mapping of the token
+addresses between all of the chains will not be scalable.
 
 ### Migration to the Standard
 
@@ -187,17 +192,21 @@ Without a deterministic deployment scheme, maintaining a mapping of the token ad
 - Tokens that are `OptimismMintableERC20Token` (corresponding to locked liquidity in L1)
   fall into the above category of already deployed and non updatable.
   They do have a special property though, which is burn/mint permissions granted to the `L2StandardBridge`.
-  This makes the convert method particularly appealing. See [Liquidity Migration](https://github.com/ethereum-optimism/design-docs/blob/098435155471ed3bcd60a50f049897495c901733/protocol/superc20/liquidity-migration.md)
-  for more context.
+  This makes the convert method particularly appealing.
 
 ## Future Considerations
 
 ### Cross Chain `transferFrom`
 
-In addition to standard locally initialized bridging, it is possible to allow contracts to be cross-chain interoperable.
-For example, a contract in chain A could send pre-approved funds from a user in chain B to a contract in chain C.
+In addition to standard locally initialized bridging,
+it is possible to allow contracts to be cross-chain interoperable.
+For example, a contract in chain A could send pre-approved funds
+from a user in chain B to a contract in chain C.
 
-For the moment, the standard will not include any specific functionality to facilitate such an action and relay on the usage of `permit2`.
-If, at some point in the future, these actions were to be included in the standard, a possible design could introduce a `remoteTransferFrom()` function.
+For the moment, the standard will not include any specific functionality
+to facilitate such an action and relay on the usage of `permit2`.
+If, at some point in the future, these actions were to be included in the standard,
+a possible design could introduce a `remoteTransferFrom()` function.
 
-Notice that this action is no longer atomic, so the contract would need a special implementation resume execution once the funds arrive in the target chain.
+Notice that this action is no longer atomic, so the contract would
+need a special implementation resume execution once the funds arrive in the target chain.
--- a/specs/interop/token-bridging.md
+++ b/specs/interop/token-bridging.md
@@ -5,12 +5,14 @@
 
 - [Overview](#overview)
 - [Implementation](#implementation)
+- [Invariants](#invariants)
 - [Deployment and migrations](#deployment-and-migrations)
   - [Factory](#factory)
   - [Migration to the Standard](#migration-to-the-standard)
-- [Future Considerations: Cross Chain `transferFrom`](#future-considerations-cross-chain-transferfrom)
-  - [Approvals](#approvals)
-  - [`remoteTransferFrom()`](#remotetransferfrom)
+- [Future Considerations](#future-considerations)
+  - [Cross Chain `transferFrom`](#cross-chain-transferfrom)
+    - [Approvals](#approvals)
+    - [`remoteTransferFrom()`](#remotetransferfrom)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -26,9 +28,10 @@ The standard will build on top of ERC20 and include the following three external
 
 - `sendERC20(uint256 amount, uint256 chainId)(bool)`
 - `sendERC20To(address to, uint256 amount, uint256 chainId)(bool)`
+- `sendERC20To(address to, uint256 amount, uint256 chainId, bytes data)(bool)`
 - `finalizeSendERC20(address to, uint256 amount)`
 
-`sendERC20(uint256 amount, uint256 chainId)` and `sendERC20To(address to, uint256 amount, uint256 chainId)` will burn `amount` tokens and initialize a message to the `L2ToL2CrossChainMessenger` to mint the `amount` in the target address at `chainId`.
+`sendERC20(uint256 amount, uint256 chainId)` and `sendERC20To(address to, uint256 amount, uint256 chainId)` will burn `amount` tokens and initialize a message to the `L2ToL2CrossChainMessenger` to mint the `amount` in the target address at `chainId`. `sendERC20To(address to, uint256 amount, uint256 chainId, bytes data)` will do the same, but also include a message.
 
 `finalizeSendERC20(address to, uint256 amount)` will process incoming messages from the `L2ToL2CrossChainMessenger` initiated from the same token on a different `chainId` and mint `amount` to the `to` address.
 
@@ -38,6 +41,8 @@ for both replay protection and domain binding.
 
 [l2-to-l2]: ./predeploys.md#l2tol2crossdomainmessenger
 
+[TODO] should we add a sendERC20To specific event?
+
 ```solidity
 function sendERC20(uint256 _amount, uint256 _chainId) external returns (bool) {
   return sendERC20To(msg.sender, _amount, _chainId);
@@ -45,15 +50,32 @@ function sendERC20(uint256 _amount, uint256 _chainId) external returns (bool) {
 
 function sendERC20To(address _to, uint256 _amount, uint256 _chainId) public returns (bool success) {
   _burn(msg.sender, _amount);
-  bytes memory _message = abi.encodeCall(this.finalizeSendERC20, (_to, _amount));
+  bytes memory _message = abi.encodeWithSignature("finalizeSendERC20(address,uint256)", _to, _amount);
+  _sendMessage(_chainId, _message);
+  return true
+}
+
+function sendERC20To(address _to, uint256 _amount, uint256 _chainId, bytes memory _data) public returns (bool success) {
+  _burn(msg.sender, _amount);
+  bytes memory _message = abi.encodeWithSignature("finalizeSendERC20(address,uint256,bytes)", _to, _amount, _data);
   _sendMessage(_chainId, _message);
   return true
 }
 
 function finalizeSendERC20(address _to, uint256 _amount) external {
+  // this will be a modifier, inline here for clarity
+  require(msg.sender == address(L2ToL2CrossChainMessenger));
+  require(L2ToL2CrossChainMessenger.crossDomainMessageSender() == address(this));
+  //
+  _mint(_to, _amount);
+}
+
+function finalizeSendERC20(address _to, uint256 _amount, bytes memory _data) external {
   require(msg.sender == address(L2ToL2CrossChainMessenger));
   require(L2ToL2CrossChainMessenger.crossDomainMessageSender() == address(this));
   _mint(_to, _amount);
+  (bool success, ) = _to.call(_data);
+  require(success, "External call failed");
 }
 
 /////// Internal functions ///////////
@@ -63,6 +85,34 @@ function _sendMessage(uint256 _destination, bytes memory _data) internal virtual
 ```
 
 Note: Some other naming options that were considered were `bridgeERC20`, `send()`, `xTransfer()`, `Transfer` (with a chainId parameter).
+
+## Invariants
+
+Besides the ERC20 invariants, the SuperchainERC20 will require the following interop specific properties:
+
+- Conservation of finalized `totalSupply`: The minted `amount` in `finalizeSendERC20()` should match the `amount` that was burnt in `sendERC20To()`, as long as target chain has the initiating chain in the dependency set. This implies that finalized cross-chain transactions will conserve the sum of `totalSupply` for each chain in the Superchain.
+  - Corollary 1: each in flight messages will decrease the `totalSupply` exactly by the burnt `amount`.
+  - Corollary 2: `SuperchainERC20s` should not charge a token fee or increase the balance when moving cross-chain.
+  - Question: what should we assume for chains not in the dependency set? a rollback method could modify this invariant to also consider that case.
+- Freedom of movement: Users should be able to send tokens into any target chain that has initiating chain in its dependency set.
+  - Question: should we add "deployed" to the invariant or assume its not going to be an issue?
+- [To discuss] Unique Messenger: The `sendERC20()` and `sendERC20To()` functions must exclusively use the `L2toL2CrossDomainMessenger` for messaging. Similarly, the `finalizeSendERC20()` function should only process messages originating from the L2toL2CrossDomainMessenger.
+  - Corollary: xERC20 and other standards from third party bridges should use different functions.
+- [To discuss] Locally initiated: The bridging action should be initialized from the chain where funds are located only.
+  - This is because same address might correspond to different users cross-chain. For example, two SAFEs with the same address in two chains might have different owners. It it possible to distiguish an EOA from a contract by checking the size of the code in the caller's address, but this will probably change with [EIP-7702](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md).
+  - A way to allow for remotely initiated bridging is to include remote approval, i.e. approve a certain address in a certain chainId to spend local funds.
+- [To discuss] Bridge Event: `sendERC20()` and `sendERC20To()` should emit a `Bridged` event.
+  - We already have a lot of events triggering from these actions. Should we include yet another one?
+
+**Desired properties**
+
+- Spatial symmetry: Same implementation and address across chains.
+
+**Open questions**
+
+- Should we define invariants for upgrades?
+- Does native minting introduce any invariant?
+  - If every factory is "equal", how do you choose which one creates the initial supply?
 
 ## Deployment and migrations
 
@@ -80,12 +130,15 @@ A token factory predeploy can ensure that `SuperchainERC20` tokens can be permis
   - Some options are wrapping, converting (if burn-mint rights can be modified) or Mirroring.
 - Tokens that are `OptimismMintableERC20Token` (corresponding to locked liquidity in L1) fall into the above category of already deployed and non updatable. They do have a special property though, which is burn/mint permissions granted to the `L2StandardBridge`. This makes the convert method particularly appealing. See [Liquidity Migration](https://github.com/ethereum-optimism/design-docs/blob/098435155471ed3bcd60a50f049897495c901733/protocol/superc20/liquidity-migration.md) for more context.
 
-## Future Considerations: Cross Chain `transferFrom`
+## Future Considerations
 
-In addition to standard bridging, it is possible to allow contracts to be cross-chain interoperable. For example, a contract in chain A
-could send pre-approved funds from a user in chain B to a contract in chain C.
+### Cross Chain `transferFrom`
 
-The standard should not include a remotely initialized `transfer` implementation as this should be handled directly on the chain where the funds live with the `sendERC20()` function. `transferFrom`, on the other hand, might get initialized from a call to a contract that is localized in a different chain from the funds. To be composable, the standard could introduce a `remoteTransferFrom()` function.
+In addition to standard locally initialized bridging, it is possible to allow contracts to be cross-chain interoperable. For example, a contract in chain A could send pre-approved funds from a user in chain B to a contract in chain C.
+
+The standard should not include a remotely initialized `transfer` implementation as this should be handled directly on the chain where the funds live with the `sendERC20()` function. `transferFrom`, on the other hand, might get initialized from a call to a contract that is localized in a different chain from the funds.
+
+For the moment, the standard will not include any specific functionality to facilitate such an action and relay on the usage of `permit2`. If, at some point in the future, these actions were to be included in the standard, a possible design could introduce a `remoteTransferFrom()` function.
 
 Notice that this action is no longer atomic, so the contract would need a special implementation resume execution once the funds arrive in the target chain.
 
@@ -122,7 +175,7 @@ function relayAndExecute(
 ) external;
 ```
 
-### Approvals
+#### Approvals
 
 To enable `remoteTransferFrom()`, it's necessary to include a separate `xAllowance` mapping that keeps track of the `chainId`. This allowance should be set locally using a new `xApprove()` function (the naming is open for discussion).
 
@@ -147,7 +200,7 @@ function allowance(address _owner, address _spender, uint256 _chainId) external 
 
 It is technically possible for the standard to also have remote approvals (initialize in chain A and approval for funds in chain B). This feature, however, was discarded as an address might have different owners in different chains (think of smart wallets for instance). Moreover, even if this was not an issue, switching `chainId` is not that relevant UX-wise to include a dedicated function in the token standard.
 
-### `remoteTransferFrom()`
+#### `remoteTransferFrom()`
 
 As already mentioned, the `remoteTransferFrom()` will allow using funds on a remote chain on behalf of another party. The function should initiate a call to the `L2ToL2CrossChainMessenger` without burning anything on the origin chain.
 
--- a/specs/interop/token-bridging.md
+++ b/specs/interop/token-bridging.md
@@ -14,9 +14,6 @@
 - [Diagram](#diagram)
 - [Implementation](#implementation)
 - [Invariants](#invariants)
-- [Deployment and migrations](#deployment-and-migrations)
-  - [Factory](#factory)
-  - [Migration to the Standard](#migration-to-the-standard)
 - [Future Considerations](#future-considerations)
   - [Cross Chain `transferFrom`](#cross-chain-transferfrom)
 
@@ -53,11 +50,11 @@ sendERC20(address _to, uint256 _amount, uint256 _chainId, bytes _data)
 
 #### `relayERC20`
 
-Process incoming messages IF AND ONLY IF if they where initiated
+Process incoming messages IF AND ONLY IF initiated
 by the same contract (token) address on a different chain
 and come from the `L2ToL2CrossChainMessenger` in the local chain.
 It will mint `_amount` to address `_to`, as defined in `sendERC20`.
-If the message contained `_data`,
+If the message containes `_data`,
 it will also include an external call to address `_to` with `_data`.
 
 ```solidity
@@ -139,9 +136,6 @@ function relayERC20(address _to, uint256 _amount, bytes memory _data) external {
 }
 ```
 
-Note: Some other naming options that were considered were
-`bridgeERC20`, `send()`, `xTransfer()`, `Transfer` (with a chainId parameter).
-
 ## Invariants
 
 Besides the ERC20 invariants, the SuperchainERC20 will require the following interop specific properties:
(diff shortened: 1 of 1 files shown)


specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+`finalizeBridgeERC20(address to, uint256 amount)` will process incoming messages from the `L2ToL2CrossChainMessenger` initiated from the same token on a different `chainId` and mint `amount` to the `to` address.
+
+A possible implementation would look like this:
+
+```solidity
+function bridgeERC20(uint256 _amount, uint256 _chainId) external returns (bool success) {
+    success = bridgeERC20To(msg.sender, _amount, _chainId);
+  }
+
+function bridgeERC20To(address _to, uint256 _amount, uint256 _chainId) public returns (bool) {
```
tynes: Thoughts on replacing the `bridgeXXX` with `sendXXX` as the goal is to make it feel like one chain, so we can make the API more like the messenger which doesn't use the bridge language, per https://github.com/ethereum-optimism/specs/pull/71#discussion_r1631562710

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+function finalizeBridgeERC20(address _to, uint256 _amount) external onlyMessenger {
+  _mint(_to, _amount);
+}
+
+/////// Internals and modifiers ///////////
+function _sendMessage(uint256 _destination, bytes memory _data) internal virtual {
+  L2ToL2CrossDomainMessenger.sendMessage(_destination, address(this), _data);
+}
+
+modifier onlyMessenger() virtual {
```
tynes: For the spec, its easier to read the invariants when they are inlined rather than in a modifier, but the modifier is def easier to maintain

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+
+The first two functions will burn `amount` tokens and initialize a message to the `L2ToL2CrossChainMessenger` to mint the `amount` in the destination (`to` address in `chainId`).
+
+`finalizeBridgeERC20(address to, uint256 amount)` will process incoming messages from the `L2ToL2CrossChainMessenger` initiated from the same token on a different `chainId` and mint `amount` to the `to` address.
+
+A possible implementation would look like this:
+
+```solidity
+function bridgeERC20(uint256 _amount, uint256 _chainId) external returns (bool success) {
+    success = bridgeERC20To(msg.sender, _amount, _chainId);
```
tynes: nit: inconsistent named return args vs not

specs/interop/token-bridging.md
```diff
@@ -2,42 +2,71 @@
...
+  function xApprove(address spender, uint256 chainId, uint256 amount) external returns (bool);
+  function remoteTransferFrom(
+    address owner,
+    address recipient,
+    uint256 spendChainId,
+    uint256 executeChainId,
+    uint256 amount,
+    bytes memory data
+  ) external returns (bool);
+	function relayTransferFrom(
```
tynes: nit: looks like a tab got in here

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+```solidity
+function bridgeERC20(uint256 _amount, uint256 _chainId) external returns (bool success) {
+    success = bridgeERC20To(msg.sender, _amount, _chainId);
+  }
+
+function bridgeERC20To(address _to, uint256 _amount, uint256 _chainId) public returns (bool) {
+  _burn(msg.sender, _amount);
+  bytes memory _message = abi.encodeCall(this.finalizeBridgeERC20, (_to, _amount));
+  _sendMessage(_chainId, _message);
+  return true;
```
tynes: Why return a `bool` here if its always true? So that `_sendMessage` can return `false` under a different implementation?

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+[Disclaimer]
+
+- Threat the following code as experimental. It will have errors and will change function names. The goal is to discuss functionalities.
+- Current function names are open to discussion.
+
+## Implementation
+
+### Cross Chain transfers
+
+The standard will include the following three external functions:
```
tynes: I think it will be easier to maintain this spec if we get rid of the solidity code block above that defines the interface and then just define the interface like you did below, but include possible return arguments like:

```
bridgeERC20(uint256 amount, uint256 chainId)(bool)
```

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+Notice that this action is no longer atomic, so the contract will need a special implementation to use it and resume execution once the funds arrive in the chain.
+
+**Approvals**
+
+To enable `remoteTransferFrom()`, we must include a separate `xAllowance` mapping that keeps track of the `chainId`. This allowance should be set locally using a new `xApprove()` function (the naming is open for discussion).
+
+Here is what a possible implementation could look like:
+
+```solidity
+function xApprove(address _spender, uint256 _chainId, uint256 _amount) external returns (bool) {
```
tynes: I like the idea of overloading `approve` directly but there were strong opinions to not to that for `transfer`. Maybe we do `remoteApprove` to make it explicit?

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+  // TODO [research] _chainId wildcard as type(uint256).max (this is very risky)
```
tynes: I think we should leave out the wildcard because that would conflict where the chainid is `type(uint256).max`. A future `permit3` contract could exist in the future and add that functionality

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+
+function allowance(address _owner, address _spender, uint256 _chainId) external view returns (uint256) {
+  if (_chainId == block.chainId) return _allowance[_owner][_spender];
+  return _xAllowance[_owner][_spender][_chainId];
+}
+```
+
+It is possible for the standard to also have remote approvals. This feature, however, was discarded as an address might have different owners in different chains (think of smart wallets for instance). Moreover, even if this was not an issue, switching `chainId` is not that relevant UX-wise to include a dedicated function in the token standard.
+
+**`remoteTransferFrom()`**
```
tynes: small nit: elsewhere in the specs we usually do subsections like `### remoteTransferFrom` so it shows up in the table of contents, i definitely like subsections for each function that explains the invariants

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+) external onlyMessenger {
+  uint256 allowed = _xAllowances[_owner][_spender][_originChainId];
+
+  if (allowed != type(uint256).max) {
+    _xAllowances[_owner][_spender][_originChainId] = allowed - _amount;
+  }
+
+  _burn(_owner, _amount);
+
+  if (block.chain == _executionChainId){
```
tynes: nit: `block.chainid` and no space after `)`

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+  require(L2ToL2CrossChainMessenger.crossDomainMessageSender() == address(this));
+  _;
+}
+```
+
+Note: `send()` or `xTransfer()` can also be good fits instead of `bridgeERC20()`.
+
+### Cross Chain `transferFrom`
+
+The standard will not include a remotely initialized `transfer` implementation as we think this should be handled directly on the chain where the funds live with the `bridgeERC20()` function. `transferFrom`, on the other hand, might get initialized from a call to a contract that is localized in a different chain from the funds. To be composable, the standard will introduce a `remoteTransferFrom()` function.
```
tynes: small nit: better to try to not use language like "we think" in a spec and instead state directly why it should be used this way

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+
+```solidity
+function xApprove(address _spender, uint256 _chainId, uint256 _amount) external returns (bool) {
+  // TODO [research] _chainId wildcard as type(uint256).max (this is very risky)
+  _xAllowances[msg.sender][_spender][_chainId] = _amount;
+  emit XApproval(msg.sender, _spender, _chainId, _amount);
+  return true;
+}
+// this will override the ERC20 view function
+function allowance(address _owner, address _spender) external view returns (uint256) {
```
tynes: Do we actually need this override?

tynes:
This doesn't yet mention the rollback functionality per https://github.com/ethereum-optimism/specs/pull/71#discussion_r1631563223 but that hasn't been implemented yet in the `L2ToL2CrossDomainMessenger`, we need to be mindful of scope and perhaps launch without the rollback feature to make sure we can be code frozen by end of July

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+- Current function names are open to discussion.
+
+## Implementation
+
+### Cross Chain transfers
+
+The standard will include the following three external functions:
+
+- `bridgeERC20(uint256 amount, uint256 chainId)`
+- `bridgeERC20To(address to, uint256 amount, uint256 chainId)`
```
tynes: Can we add a `sendERC20To(address,uint256,uint256,bytes)` overload per https://github.com/ethereum-optimism/specs/pull/71#issuecomment-1990504944 ? Could this replace the `relayAndExecute`?

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+  if (_chainId == block.chainId) return _allowance[_owner][_spender];
+  return _xAllowance[_owner][_spender][_chainId];
+}
+```
+
+It is possible for the standard to also have remote approvals. This feature, however, was discarded as an address might have different owners in different chains (think of smart wallets for instance). Moreover, even if this was not an issue, switching `chainId` is not that relevant UX-wise to include a dedicated function in the token standard.
+
+**`remoteTransferFrom()`**
+
+As already mentioned, the `remoteTransferFrom()` \*\*\*\*will allow using funds on a remote chain on behalf of another party. The function should initiate a call to the `L2ToL2CrossChainMessenger` without burning anything on the origin chain.
```
tynes: nit: extra `\*\*`

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+
+To process a `remoteTransferFrom()` on the destination, the standard should include a `relayTransferFrom()` function that checks local allowance and calls `burn()`. Then, it will check if the target lives on the same `chainId` to call `relayAndExecute()` to the corresponding contract.
+
+Finally, the standard should included a `relayAndExecute()` function that might be used to resume execution on the origin chain. This is how the atomicity lost in the `transferFrom()` gets addressed.
+
```
tynes, line 114: A mermaid diagram of the control flow would be super helpful here

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+It is possible for the standard to also have remote approvals. This feature, however, was discarded as an address might have different owners in different chains (think of smart wallets for instance). Moreover, even if this was not an issue, switching `chainId` is not that relevant UX-wise to include a dedicated function in the token standard.
+
+**`remoteTransferFrom()`**
+
+As already mentioned, the `remoteTransferFrom()` \*\*\*\*will allow using funds on a remote chain on behalf of another party. The function should initiate a call to the `L2ToL2CrossChainMessenger` without burning anything on the origin chain.
+
+To process a `remoteTransferFrom()` on the destination, the standard should include a `relayTransferFrom()` function that checks local allowance and calls `burn()`. Then, it will check if the target lives on the same `chainId` to call `relayAndExecute()` to the corresponding contract.
+
+Finally, the standard should included a `relayAndExecute()` function that might be used to resume execution on the origin chain. This is how the atomicity lost in the `transferFrom()` gets addressed.
```
tynes: There is a tradeoff between simplicity and an atomicity guarantee, do you feel strongly that the complexity is worth the atomicity guarantee here?

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+
+  if (allowed != type(uint256).max) {
+    _xAllowances[_owner][_spender][_originChainId] = allowed - _amount;
+  }
+
+  _burn(_owner, _amount);
+
+  if (block.chain == _executionChainId){
+    _relayAndExecute(_owner, _recipient, _spender, _originChainId, block.chainid, _amount, _data);
+  } else {
```
tynes: I am confused by this if else branch, can you explain it?

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+As already mentioned, the `remoteTransferFrom()` \*\*\*\*will allow using funds on a remote chain on behalf of another party. The function should initiate a call to the `L2ToL2CrossChainMessenger` without burning anything on the origin chain.
+
+To process a `remoteTransferFrom()` on the destination, the standard should include a `relayTransferFrom()` function that checks local allowance and calls `burn()`. Then, it will check if the target lives on the same `chainId` to call `relayAndExecute()` to the corresponding contract.
+
+Finally, the standard should included a `relayAndExecute()` function that might be used to resume execution on the origin chain. This is how the atomicity lost in the `transferFrom()` gets addressed.
+
+A possible implementation for these functionalities would look like this:
+
+```solidity
+function remoteTransferFrom(
```
tynes: After discussing with @smartcontracts, we think its a good idea to remove the remote transfer from logic from the first iteration of the superchain erc20 standard to simplify things as an MVP. It will reduce the amount of test coverage we need and in the short term a `permit2` like contract could be used to enable cross chain approvals. We can revisit this part of the interface again in the future if people are really asking for it

specs/interop/token-bridging.md
0xParti: We were trying to match the ERC20 standard, where transfer returns a bool. We can remove it if you think its not worth.

specs/interop/token-bridging.md
```diff
@@ -72,26 +101,200 @@ contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
...
+  }
+
+function bridgeERC20To(address _to, uint256 _amount, uint256 _chainId) public returns (bool) {
+  _burn(msg.sender, _amount);
+  bytes memory _message = abi.encodeCall(this.finalizeBridgeERC20, (_to, _amount));
+  _sendMessage(_chainId, _message);
+  return true;
+}
+
+function finalizeBridgeERC20(address _to, uint256 _amount) external onlyMessenger {
```
0xParti: if we add `bool` returns to the other functions to match ERC20, we should probably also add it here.

specs/interop/token-bridging.md
0xParti: Moving this to future work.

That said, agree we should make the name more explicit. I don't like `remoteApprove`, as the approval is local and would be a different convention from `remoteTransferFrom`, which is a non-local action.  Maybe `approveRemote`?

specs/interop/token-bridging.md
0xParti: We don't need it, but it would make the code cleaner. If we did not override here, the contract would have two allowances mappings, one with and another without chainId.

specs/interop/token-bridging.md
0xParti: good question. I think the `remoteTransferFrom` flow is essential in an interop nirvana. It might be possible to do it otherwise, but atomicity loss will remain an issue, as the initiating call has no way to guarantee execution on its own.

specs/interop/token-bridging.md
0xParti: Agree. Will add!

specs/interop/token-bridging.md
0xParti: The recipient might be in a chain which is different from where the funds are (where the `relayTransferFrom` call is happening). This if will check if its on the same chain (and perform an internal call) or not (and craft a message to execute the call cross-chain).

0xParti: to clarify, the first if corresponds to:
Contract in chain A calls remoteTransferFrom to chain B with a recipient in chain B 
The second corresponds to:
Contract in chain A calls remoteTransferFrom to chain B with a recipient in chain C

specs/interop/token-bridging.md
0xParti: Ok, this makes sense. The tradeoff with permit here is devex. We can upgrade later with more time to test, if we think its worth.

Do you think it makes sense to introduce a predeploy that implements permit2 and/or message relaying for these tokens? This would be a middle ground between having it in the standard and leaving all the work to the interacting contract. 
These predeploys could also be implemented afterwards

specs/interop/token-bridging.md
0xParti: Yes, we can add this, its a good idea. It should be able to replace `relayAndExecute()` if used with additional logic.

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,296 @@
...
+## Implementation
 
-```solidity
-interface ISuperchainERC20 {
-  function bridgeERC20(uint256 amount, uint256 chainId) external;
-  function bridgeERC20To(address to, uint256 amount, uint256 chainId) external;
-  function finalizeBridgeERC20(address to, uint256 amount) external;
-}
-```
+The standard will build on top of ERC20 and include the following three external functions:
```
mds1: Removing `three` since there are four :)
```suggestion
The standard will build on top of ERC20 and include the following external functions:
```

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,296 @@
...
 
-## Implementation
+- `sendERC20(uint256 amount, uint256 chainId)(bool)`
+- `sendERC20To(address to, uint256 amount, uint256 chainId)(bool)`
+- `sendERC20To(address to, uint256 amount, uint256 chainId, bytes data)(bool)`
+- `finalizeSendERC20(address to, uint256 amount)`
```
mds1: What is the rationale for the return values? Three return a bool but the last one doesn't. Should all they return nothing and just revert on failure? 

Also, do we need different function names and arg ordering?  What do you think about:
```solidity
sendERC20(uint256 amount, uint256 chainId)
sendERC20(uint256 amount, uint256 chainId, address to)
sendERC20(uint256 amount, uint256 chainId, address to, bytes data)
```

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,296 @@
...
+
+`sendERC20(uint256 amount, uint256 chainId)` and `sendERC20To(address to, uint256 amount, uint256 chainId)` will burn `amount` tokens and initialize a message to the `L2ToL2CrossChainMessenger` to mint the `amount` in the target address at `chainId`. `sendERC20To(address to, uint256 amount, uint256 chainId, bytes data)` will do the same, but also include a message.
```
mds1: What happens with `data`? e.g. what is the use case there, given that there is no corresponding `target` address for the data

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,296 @@
...
 for both replay protection and domain binding.
 
 [l2-to-l2]: ./predeploys.md#l2tol2crossdomainmessenger
 
+[TODO] should we add a sendERC20To specific event?
+
 ```solidity
-contract OptimismSuperchainERC20 is ERC20, ISuperchainERC20 {
-  constructor(string memory _name, string memory _symbol, uint256 _decimals) ERC20(_name, _symbol, _decimals) {}
+function sendERC20(uint256 _amount, uint256 _chainId) external returns (bool) {
```
mds1: I would suggest diagrams over code for design docs

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,296 @@
...
+}
+```
+
+Note: Some other naming options that were considered were `bridgeERC20`, `send()`, `xTransfer()`, `Transfer` (with a chainId parameter).
+
+## Invariants
+
+Besides the ERC20 invariants, the SuperchainERC20 will require the following interop specific properties:
+
+- Conservation of finalized `totalSupply`: The minted `amount` in `finalizeSendERC20()` should match the `amount` that was burnt in `sendERC20To()`, as long as target chain has the initiating chain in the dependency set. This implies that finalized cross-chain transactions will conserve the sum of `totalSupply` for each chain in the Superchain.
```
mds1: > as long as target chain has the initiating chain in the dependency set

What happens if the initiating chain is not in the dependency set? Do the send methods need to validate that the chain ID is in the dependency set, or does the executing message just never get included? Who is expected to call the finalize method?

specs/interop/token-bridging.md
0xParti: yes, it should all return bool as per the ERC20 standard or not. I was waiting for Mark's answer to change all together. Do you have any take here? Mark suggested to return nothing in a different comment.

 like your suggestion to have a single function name! Will change it :)

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,296 @@
...
+  - Corollary 1: each in flight messages will decrease the `totalSupply` exactly by the burnt `amount`.
```
mds1: Does "in flight" mean "initiated message is safe but executing message has not happened yet"? Is it possible the executing message never occurs?

I would redefine this invariant so there is no corollary to simplify it, e.g. the "total supply is conserved across all safe blocks" (or whatever the actual invariant is)

specs/interop/token-bridging.md
0xParti: This is illustrated below in the implementation, but should also be mentioned here. The target is the `to`, and this can be used to send funds and execute something in target. It is possible to add another `target` address different from `to` as well.

specs/interop/token-bridging.md
0xParti: Thats a very good open question, and its something we have to discuss further. We considered adding rollback functionalities for these cases, but we think its out of scope for the moment.
For that reason, I included the dependancy set clarification into the invariant. If there were a way to handle that scenario, the invariant could be extended.

specs/interop/token-bridging.md
0xParti: yes, in flight means exactly that, I can clarify. I'm not sure the two sentences are equivalent. One is saying that the totalSupply is conserved over safe blocks, the other that in between the state changed in a specific way.
Both are corolaries of the first sentence of the invariant now that I think about it more.

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,298 @@
...
-}
-```
+The standard will build on top of ERC20 and include the following five external functions:
 
-## Implementation
+- `sendERC20(uint256 amount, uint256 chainId)(bool)`
+- `sendERC20(address to, uint256 amount, uint256 chainId)(bool)`
+- `sendERC20(address to, uint256 amount, uint256 chainId, bytes data)(bool)`
+- `finalizeSendERC20(address to, uint256 amount)`
+- `finalizeSendERC20(address to, uint256 amount, bytes data)`
```
tynes: small nit: can we modify the finalize functions to be `relayERC20` to match the interface of the messenger? just to keep naming consistent

https://github.com/ethereum-optimism/specs/blob/main/specs/interop/predeploys.md#relaymessage-invariants

tynes: Also thoughts on just having one `relayERC20` function that optionally makes a `CALL` to `to` if `data.length > 0`? Seems like a simplification that we could make to make the interface smaller. We could either keep both `sendERC20` functions or only have the one with `data`, I lean towards keeping both so it feels like `transfer` but open to keeping both to keep the standard smaller

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,298 @@
...
+
+## Invariants
+
+Besides the ERC20 invariants, the SuperchainERC20 will require the following interop specific properties:
+
+- Conservation of finalized `totalSupply`: The minted `amount` in `finalizeSendERC20()` should match the `amount` that was burnt in `sendERC20()`, as long as target chain has the initiating chain in the dependency set.
+  - Corollary 1: Cross-chain transactions will conserve the sum of `totalSupply` for each chain in the Superchain across safe blocks.
+  - Corollary 2: Each initiated but not finalized message (included in initiating chain but not yet in target chain) will decrease the `totalSupply` exactly by the burnt `amount`.
+  - Corollary 2: `SuperchainERC20s` should not charge a token fee or increase the balance when moving cross-chain.
+  - Question: what should we assume for chains not in the dependency set? a rollback method could modify this invariant to also consider that case.
```
tynes: We could note above that an assumption is that the remote chain has the local chain in its dependency set, and if this is not held then the funds will be locked similarly to sending funds to the wrong address

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,298 @@
...
+- Freedom of movement: Users should be able to send tokens into any target chain that has initiating chain in its dependency set.
+  - Question: should we add "deployed" to the invariant or assume its not going to be an issue?
```
tynes: I think we could skip over the deployed idea here and have offchain tooling help to prevent footguns

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,298 @@
...
+- [To discuss] Unique Messenger: The `sendERC20()` functions must exclusively use the `L2toL2CrossDomainMessenger` for messaging. Similarly, the `finalizeSendERC20()` function should only process messages originating from the L2toL2CrossDomainMessenger.
+  - Corollary: xERC20 and other standards from third party bridges should use different functions.
+- [To discuss] Locally initiated: The bridging action should be initialized from the chain where funds are located only.
+  - This is because same address might correspond to different users cross-chain. For example, two SAFEs with the same address in two chains might have different owners. It it possible to distiguish an EOA from a contract by checking the size of the code in the caller's address, but this will probably change with [EIP-7702](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md).
+  - A way to allow for remotely initiated bridging is to include remote approval, i.e. approve a certain address in a certain chainId to spend local funds.
+- [To discuss] Bridge Event: `sendERC20()` should emit a `Bridged` event.
```
tynes: We could make a `SendERC20` event be emitted, we would need to define it as part of the spec, same for `RelayERC20`, so these should be added to a list near where the functions are defined above with a small explanation of when they should be emitted

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,298 @@
...
+  - We already have a lot of events triggering from these actions. Should we include yet another one?
+
+**Desired properties**
+
+- Spatial symmetry: Same implementation and address across chains.
```
tynes: We could say that an invariant exists that the addresses are the same between chains and remove this desired properties section

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,298 @@
...
+
+**Open questions**
+
+- Should we define invariants for upgrades?
```
tynes: We can define the upgrade in the future, i think its fine to remove open q's. This defines the interface and the convert approach is one usage of the interface that we will implement first

specs/interop/token-bridging.md
tynes: > Do you think it makes sense to introduce a predeploy that implements permit2 and/or message relaying for these tokens?

We can scope this in future work, for now I'd like to leave this out

specs/interop/token-bridging.md
```diff
@@ -2,96 +2,298 @@
...
+A possible implementation for these functionalities would look like this:
 
-A token factory predeploy can be used to ensure that `SuperchainERC20` tokens
-can be permissionlessly and deterministically deployed. If a deterministic deployment
-scheme is not used, maintaining a mapping of the token addresses between all of the
-chains will not be scalable.
+```solidity
+function remoteTransferFrom(
+  address _owner,
+  address _recipient,
```
tynes: Is it ok if we remove this implementation/definition of the interface for now and instead just talk about it as being a future addition? Just to keep the spec more lean as we may learn new things and it may end up being different and then we would need to modify this spec

specs/interop/token-bridging.md
mds1: As we talked about during the design my example suggestion of `"total supply is conserved across all safe blocks" (or whatever the actual invariant is)` is not practical, since there is no guarantee an executing message is ever sent.

From my new understanding, some invariants we have are:
- For a fixed supply token, the sum of totalSupply across chains should never increase. (It stays the same when every initiating message has its executing message, but decreases if some executing messages are never sent)
    - This expands similarly for a token that has some way to be minted:
        - In the absence of non-crosschain-mints (i.e. ones that are expected to increase total supply) we get the fixed supply token invariant
        - When there are non-crosschain-mints, the sum of total supply across chains increases by the exact non-crosschain-mint amount
- The sum of all user balances across all chains must be exactly equal to the sum of the total supply across all chains. 
    - We preserve the regular ERC20 invariant that the sum of all user balances on a single chain must be exactly equal to total supply on that chain

specs/interop/token-bridging.md
mds1: Yea I agree with @tynes—the `sendERC20` methods don't need to match ERC20, and "the call was successful unless it reverts" is a simpler and common pattern

specs/interop/token-bridging.md
0xParti: I like the fixed supply clarification. 
Also, the sum of user's balance vs total supply is a very strong invariant we had missed, good catch.

Adding on top of that one, the sum of each user's balance across chains will behave like the totalSupply invariant, i.e. it should not increase for tokens of fix supply.

I do feel like "sum of totalSupply across chains should never increase" is not precise enough. In any situation where this is not conserved (for fixed supply tokens) then its decrease exactly the bridged amount and it's because the transaction has not been completed.

0xParti: Regarding the cross-chain mint, do you feel like this is something that should be implemented inside the sendERC20/relayERC20 logic?

specs/interop/token-bridging.md
0xParti: I really like the idea of simplifying both `relayERC20` and `sendERC20` to a single function. 
- I don't feel the analogy with `transfer` is strong enough to keep both, as it's a whole different action. 
- UX wise, we should trust the integrations and keep the standard as lean as possible.
- Gas wise, there will be an advantage for keeping the functions separate, but its only a very small check.

0xParticle: We are thinking about changing the - `sendERC20(address to, uint256 amount, uint256 chainId, bytes data)` to include a `target` input. In this way, we could decouple the call from the `to`. That might be an argument for of keeping both implementations.

specs/interop/token-bridging.md
0xParti: Yes, will add this note, with the small caveat that, if at some point the target chain includes the initiating in the dependency set, funds could be recovered.

specs/interop/token-bridging.md
0xParti: absolutely, will add in the next commit

specs/interop/token-bridging.md
0xParti: agree

specs/interop/token-bridging.md
0xParti: perfect

specs/interop/token-bridging.md
0xParti: closing for now, as we are not including this flow for the moment.

specs/interop/token-bridging.md
0xParti: closing for now, as we are not including this flow for the moment.

specs/interop/token-bridging.md
0xParti: closing for now, as we are not including this flow for the moment.

specs/interop/token-bridging.md
0xParti: removed

specs/interop/token-bridging.md
0xParticle: added: https://github.com/ethereum-optimism/specs/blob/25c6d31850a1919ab4fe819c7634a09572ad8add/specs/interop/token-bridging.md#diagram

0xParti: resolving, as we are not implementing this for the moment

specs/interop/token-bridging.md
0xParti: resolving, as we are not doing this for the moment

specs/interop/token-bridging.md
0xParti: done

specs/interop/token-bridging.md
0xParti: added a mermaid flow: https://github.com/ethereum-optimism/specs/blob/25c6d31850a1919ab4fe819c7634a09572ad8add/specs/interop/token-bridging.md#diagram

specs/interop/token-bridging.md
0xParti: moved into single `sendERC20` and `relayERC20` in the latest commit. We can keep the discussion open just in case.

specs/interop/token-bridging.md
0xParti: added

tynes (approved):
Awesome!

PR #258, opened 2024-06-25
interop: Specify deposit handling

Inphi:
**Description**

Add early specification of deposits for interop. Introduces a "deposit context" to derivation as a useful model for thinking about deposits in the EVM.


specs/interop/predeploys.md
```diff
@@ -416,6 +426,30 @@ dependency set called `dependencySet()`. This function MUST return the array of
...
+Only the `CrossL2Inbox` is authorized to call `isDeposit`.
+
+#### `setDepositing`
+
+```solidity
+function setDepositing(bool _isDepositing)
+```
+A new method added to indicate whether the current execution is in a deposit context.
+
+Only the `DEPOSITOR_ACCOUNT` is authorized to call `setDepositing`.
```
clabby: When is this called? For each deposit (assuming `DEPOSITOR_ACCOUNT` is the only party allowed to call it)?

specs/interop/predeploys.md
```diff
@@ -416,6 +426,30 @@ dependency set called `dependencySet()`. This function MUST return the array of
...
 MUST return the length of the dependency set array.
 
+### Deposit Context
+
+The `L1Block` contract is updated with the following methods to support [deposit contexts](./derivation.md#deposit-context).
+
+#### `isDeposit`
+
+```solidity
+function isDeposit() (return bool)
```
clabby: Would fit very well as a transient storage variable.

clabby (commented):
I like the idea of the deposit context, offers a pretty clean separation.

Wondering how we can lift this outside of explicit actions, though - having each deposit (or the beginning/last deposit) hook into a call to `L1Block` seems expensive over time.

Maybe prior to executing the L1 info transaction, the block executor can set the transient `isDepositing` variable to `true` in `L1Block`, and then manually clear it after all deposits have been processed? This would ensure there's no gas impact on the downstream depositors, and it becomes a consensus rule.

specs/interop/predeploys.md
Inphi: It's called twice, at the start of deposits processing and at the end (to unset `_isDepositing`). Rather than for every deposit.

clabby: What enforces that the calls are made?

Inphi: Derivation is altered to inserts sandwich depositing transactions with `setDepositing` calls. That is, when signaling payload-attribute transactions to execution.
Verifiers would then check that the calls were made, in the correct order.

Inphi: Effectively, the rollup node inserts the calls as system transactions when constructing payload-attributes.

tynes: The way that I thought this would work is the `setXXX` would `sstore(isDeposit, true)` and then the deposits would be sandwiched with another system deposit that calls a public `onlyDepositor` function that `sstore(isDeposit, false)`

specs/interop/predeploys.md
```diff
@@ -416,6 +426,30 @@ dependency set called `dependencySet()`. This function MUST return the array of
...
+```
+
+A new method added to `L1Block` indicating whether the current execution is in a [deposit context](./derivation.md#deposit-context). The method has the following signature:
+
+
+Only the `CrossL2Inbox` is authorized to call `isDeposit`.
```
tynes, line 610: Could we add a follow up sentence with the rationale? ie if any contract could call this, then apps can build censorship easily

tynes:
We out of band decided on a sandwich approach where the L1 attributes tx sets `isDeposit = true` then derivation creates another deposit tx that is at the tail end of the deposits that sets `isDeposit = false`

specs/interop/derivation.md
```diff
@@ -0,0 +1,22 @@
...
+
+- [Overview](#overview)
+  - [Deposit Context](#deposit-context)
+- [Security Considerations](#security-considerations)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview 
+
+This is an experimental section and may be changed in the future. It is not required
```
tynes, line 16: We do want this for the initial release of interop, what I think we should cut from the initial mainnet release is the ability for deposits to trigger executing messages, we can follow up with that functionality

specs/interop/upgrade.md
```diff
@@ -29,3 +30,7 @@ The exact definitions for these upgrade transactions are still to be defined.
 ## Security Considerations
 
 TODO
+
+### Deposit Transactions During Upgrades
+
+TODO
```
tynes, line 75: small nit: usually security considerations goes at the very bottom

Inphi:
I'm realizing that we need stronger guarantees on the cumulative gas usage of the L1 attributes tx and user deposits. Otherwise, the `isDeposit = false` transaction sequenced after user deposits can fail. It'll be very bad if `isDeposit` remains set to `true` as that would mean all executing messages are treated as deposits and thus not handled until `isDeposit` is successfully reset. To illustrate:
1. L1 info deposit transaction succeeds: setting L1 block attributes and `isDeposit = true`
2. User deposits are processed.
3. L1 info deposit transaction to set `isDeposit = false` fails due to insufficient gas

The issue still occurs even if there aren't any user deposits as the first L1 info tx may be enough to fill the entire block (in theory, though unlikely in practice).

We could re-examine the `ResourceConfig.maxResourceLimit` setting to ensure there's ample gas available for both L1 attributes txs. Then the only thing left to do is to have a hard gas limit on reduce the gasLimit of the initial L1 attributes tx. Basically the invariant must be:
```
l1AttributeTx.gasLimit + resourceConfig.maxResourceLimit + isDepositFalseTx.gasLimit <= 30000000
```
Right now this isn't the case on mainnet. As the `l1AttributeTx.gasLimit` is the dominating term above at 150 M. Even though it typically uses much less gas than that, we should guarantee that there's always sufficient gas available.
However, these gas and resource limits seem appropriate to me and I don't think we should reduce them.

Other options as we've discussed earlier that don't have this problem:
- Update execution to honor `isSystemTx=true` for deposits. This eliminates gas concerns for executing `isDeposit = true`.
- Mutate deposits once again. Ensures that `isDeposit` is always reset at the expense of broken deposit gas estimation. Not great UX.
- `isDeposit` updates outside of the EVM via transient storage. Requires changes to the execution engine. 

We could also simply live with the possibility of interop being disabled temporarily. While this may be OK for a devnet, it  shouldn't happen on mainnet.

tynes:
This 100% is a valid concern, I think we should:
>We could re-examine the ResourceConfig.maxResourceLimit
and lower it if necessary.

We shouldn't have to think about this, its left over from old decisions and I think that we can solve this problem using a deposit queue which i made a draft spec for here: https://github.com/ethereum-optimism/specs/pull/82

specs/interop/derivation.md
```diff
@@ -0,0 +1,52 @@
...
+
+These two operations wrap user deposits, such that `isDeposit = true` occurrs during the first L1 attributes
+transaction and `isDeposit = false` occurs immediately either after either the last user deposit,
+if one exists, or after the first L1 attributes transaction if there are no user deposits.
+
+The order of deposit transactions occurs as follows:
+
+1. L1 attributes transaction calling [`setL1BlockValuesInterop()`](./l1_attributes.md).
+1. User deposits
+1. L1 attributes transaction calling [`resetIsDeposit()`](./l1_attributes.md)
```
skeletor-spaceman:
```suggestion
1. L1 attributes transaction calling [`depositsComplete()`](./l1_attributes.md)
```

specs/interop/predeploys.md
```diff
@@ -26,6 +27,9 @@
 - [L1Block](#l1block)
   - [Static Configuration](#static-configuration)
   - [Dependency Set](#dependency-set)
+  - [Deposit Context](#deposit-context)
+  - [`isDeposit()`](#isdeposit)
+    - [`resetIsDeposit()`](#resetisdeposit)
```
```suggestion
    - [`depositsComplete()`](#depositscomplete)
```

specs/interop/predeploys.md
```diff
@@ -416,6 +428,28 @@ dependency set called `dependencySet()`. This function MUST return the array of
...
+```
+
+### `isDeposit()`
+
+Returns true iff the current execution occurs in a [deposit context](./derivation.md#deposit-context).
+
+Only the `CrossL2Inbox` is authorized to call `isDeposit`.
+This is done to prevent apps from easily detecting and censoring deposits.
+
+#### `resetIsDeposit()`
```
```suggestion
#### `depositsComplete()`
```

specs/interop/predeploys.md
```diff
@@ -416,6 +428,28 @@ dependency set called `dependencySet()`. This function MUST return the array of
...
+
+Called after processing the first L1 Attributes transaction and user deposits to destroy the deposit context.
+
+Only the `DEPOSITOR_ACCOUNT` is authorized to call `resetIsDeposit()`.
```
```suggestion
Only the `DEPOSITOR_ACCOUNT` is authorized to call `depositsComplete()`.
```

skeletor-spaceman (commented):
renaming `resetIsDeposit` in favor of `depositsComplete`

tynes:
Looks like this needs a rebase now @skeletor-spaceman, happy to review again when its ready

skeletor-spaceman:
> happy to review again when its ready

@tynes merge ready :)

tynes:
@skeletor-spaceman The CI is failing

specs/interop/predeploys.md
```diff
@@ -150,6 +154,14 @@ function executeMessage(Identifier calldata _id, address _target, bytes calldata
 
 Note that the `executeMessage` function is `payable` to enable relayers to earn in the gas paying asset.
 
+### Deposit Handling
+
+A call to `executeMessage` reverts if the call is made in a [deposit context](./derivation.md#deposit-context).
```
tynes: We should be particular about the event `ExecutingMessage` rather than a call to the function

tynes: since there are multiple entrypoints now

specs/interop/predeploys.md
```diff
@@ -430,6 +442,29 @@ dependency set called `dependencySet()`. This function MUST return the array of
 `L1Block` MUST also provide a public getter to get the dependency set size called `dependencySetSize()`. This function
 MUST return the length of the dependency set array.
 
+### Deposit Context
+
+The `L1Block` contract is updated with new methods to interact with [deposit contexts](./derivation.md#deposit-context).
+
+```solidity
+function isDeposit() public view returns (bool);
+function resetIsDeposit() public;
```
tynes: The function name is `resetIsDeposit` here but `depositsComplete` elsewhere - what is the final name?

skeletor-spaceman: good catch! thanks :)

specs/interop/predeploys.md
skeletor-spaceman: true! fixed :)

specs/interop/derivation.md
```diff
@@ -0,0 +1,52 @@
...
+As such, there is exactly one deposit context per block.
+
+A deposit context is created by two operations:
+
+- An L1 attributes transaction that sets `isDeposit = true` on the `L1Block` contract.
+This instantiates a deposit context for the current block.
+- An L1 attributes transaction that sets `isDeposit = false`. This destroys the existing deposit context.
+
+These two operations wrap user deposits, such that `isDeposit = true` occurrs during the first L1 attributes
+transaction and `isDeposit = false` occurs immediately either after either the last user deposit,
```
0xDiscotech:
```suggestion
transaction and `isDeposit = false` occurs immediately after the last user deposit,
```

specs/interop/derivation.md
```diff
@@ -0,0 +1,52 @@
...
+if one exists, or after the first L1 attributes transaction if there are no user deposits.
```
0xDiscotech:
```suggestion
if any exists, or after the first L1 attributes transaction if there are no user deposits.
```

specs/interop/derivation.md
```diff
@@ -0,0 +1,52 @@
...
+
+The order of deposit transactions occurs as follows:
+
+1. L1 attributes transaction calling [`setL1BlockValuesInterop()`](./l1_attributes.md).
+1. User deposits
+1. L1 attributes transaction calling [`depositsComplete()`](./l1_attributes.md)
```
0xDiscotech:
```suggestion
1. L1 attributes transaction calling [`setL1BlockValuesInterop()`](./l1_attributes.md).
2. User deposits
3. L1 attributes transaction calling [`depositsComplete()`](./l1_attributes.md)
```

specs/interop/derivation.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Gas Considerations
+
+There must be sufficient gas available in the block to destroy deposit context.
+There's no guarantee on the minimum gas available for the second L1 attributes transaction as the block
+may be filled by the other deposit transactions. As a consequence, a deposit context may spill onto multiple blocks.
```
0xDiscotech:
```suggestion
may be filled by the other deposit transactions. As a consequence, a deposit context may spill into multiple blocks.
```

specs/interop/predeploys.md
```diff
@@ -430,6 +443,28 @@ dependency set called `dependencySet()`. This function MUST return the array of
 `L1Block` MUST also provide a public getter to get the dependency set size called `dependencySetSize()`. This function
 MUST return the length of the dependency set array.
 
+### Deposit Context
+
+The `L1Block` contract is updated with new methods to interact with [deposit contexts](./derivation.md#deposit-context).
```
0xDiscotech:
```suggestion
The `L1BlockInterop` contract will extend from `L1Block` to add new methods to interact with [deposit contexts](./derivation.md#deposit-context).
```

specs/interop/predeploys.md
```diff
@@ -430,6 +443,28 @@ dependency set called `dependencySet()`. This function MUST return the array of
...
+
+```solidity
+function isDeposit() public view returns (bool);
+function depositsComplete() public;
+```
+
+### `isDeposit()`
+
+Returns true iff the current execution occurs in a [deposit context](./derivation.md#deposit-context).
```
0xDiscotech:
```suggestion
Returns true if the current execution occurs in a [deposit context](./derivation.md#deposit-context).
```

specs/interop/upgrade.md
```diff
@@ -26,6 +29,43 @@ The execution payload MUST set `noTxPool` to `true` for this block.
 
 The exact definitions for these upgrade transactions are still to be defined.
 
+## L1 Attributes
```
0xDiscotech, line 36: Here did you confuse `setL1BlockValuesInterop` with `setL1BlockIsthmus`?

0xDiscotech (commented):
Solid spec! Left few suggestions and a question

specs/interop/derivation.md
```diff
@@ -0,0 +1,52 @@
...
+and concluding immediately after the execution of the final deposited transaction within that block.
+As such, there is exactly one deposit context per block.
+
+A deposit context is created by two operations:
+
+- An L1 attributes transaction that sets `isDeposit = true` on the `L1Block` contract.
+This instantiates a deposit context for the current block.
+- An L1 attributes transaction that sets `isDeposit = false`. This destroys the existing deposit context.
+
+These two operations wrap user deposits, such that `isDeposit = true` occurrs during the first L1 attributes
```
0xDiscotech:
```suggestion
These two operations wrap user deposits, such that `isDeposit = true` occurs during the first L1 attributes
```

specs/interop/derivation.md
skeletor-spaceman: oh, this works just fine, and we don't have to worry about re-ordering the numbers if we introduce an intermediary step

specs/interop/predeploys.md
skeletor-spaceman: great catch

specs/interop/upgrade.md
skeletor-spaceman: yes, this is outdated

specs/interop/derivation.md
0xDiscotech: ah ok, don't you think using this is better then?
```
* L1 attributes transaction calling [`setL1BlockValuesInterop()`](./l1_attributes.md).
* User deposits
* L1 attributes transaction calling [`depositsComplete()`](./l1_attributes.md)
```

specs/interop/predeploys.md
```diff
@@ -581,6 +594,28 @@ dependency set called `dependencySet()`. This function MUST return the array of
 `L1Block` MUST also provide a public getter to get the dependency set size called `dependencySetSize()`. This function
 MUST return the length of the dependency set array.
 
+### Deposit Context
+
+The `L1BlockInterop` contract will extend from `L1Block` to add new methods to interact with [deposit contexts](./derivation.md#deposit-context).
```
tynes: I would recommend we don't mention `L1BlockInterop` here since that leaks our development flow into the specs, ultimately we want to merge all features directly into `L1Block` but we don't want development features in the genesis of new chains or to block other releases like holocene

specs/interop/upgrade.md
```diff
@@ -30,6 +33,43 @@ The execution payload MUST set `noTxPool` to `true` for this block.
...
+
+On the Interop activation block, and if Interop is not activated at Genesis,
+the L1 Attributes Transaction includes a call to `setL1BlockValuesEcotone`.
+
+Every subsequent L1 Attributes transaction should include a call to the new `setL1BlockValuesInterop` function.
```
tynes: nit: we renamed this to `setL1BlockValuesIsthmus` right?

specs/interop/derivation.md
```diff
@@ -0,0 +1,52 @@
...
+
+- An L1 attributes transaction that sets `isDeposit = true` on the `L1Block` contract.
+This instantiates a deposit context for the current block.
+- An L1 attributes transaction that sets `isDeposit = false`. This destroys the existing deposit context.
+
+These two operations wrap user deposits, such that `isDeposit = true` occurs during the first L1 attributes
+transaction and `isDeposit = false` occurs immediately after the last user deposit,
+if any exists, or after the first L1 attributes transaction if there are no user deposits.
+
+The order of deposit transactions occurs as follows:
```
tynes, line 36: If we wanted to optimize this, per the current spec (before deposit queue), deposits only exist when the sequence number is 0, so we could alter between calling ecotone vs isthmus, but i think its fine for now to not optimize this

specs/interop/upgrade.md
skeletor-spaceman: oops yes! nice catch.

specs/interop/derivation.md
skeletor-spaceman: agreed to not over-optimize for now :)

specs/interop/predeploys.md
skeletor-spaceman: pls re-check how it reads now

tynes (approved):

PR #267, opened 2024-07-08
Fix SWS typo in glossary.md

miltonjonat:
**Description**

Glossary was stating that the default Sequencer Window Size was measured in "epochs" instead of blocks

**Tests**

N/A

**Additional context**

N/A

**Metadata**

N/A

### DIFF
--- a/specs/glossary.md
+++ b/specs/glossary.md
@@ -233,7 +233,7 @@ A sequencing window is a range of L1 blocks from which a [sequencing epoch][sequ
 A sequencing window whose first L1 block has number `N` contains [batcher transactions][batcher-transaction] for epoch
 `N`. The window contains blocks `[N, N + SWS)` where `SWS` is the sequencer window size.
 
-The current default `sws` is 3600 epochs.
+The current default `sws` is 3600 blocks.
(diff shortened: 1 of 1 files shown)


specs/glossary.md
```diff
@@ -233,7 +233,7 @@ A sequencing window is a range of L1 blocks from which a [sequencing epoch][sequ
 A sequencing window whose first L1 block has number `N` contains [batcher transactions][batcher-transaction] for epoch
 `N`. The window contains blocks `[N, N + SWS)` where `SWS` is the sequencer window size.
 
-The current default `sws` is 3600 epochs.
+The current default `sws` is 3600 blocks.
```
ajsutton, line 236: We probably should be explicit that these are L1 blocks. The original epochs, while being a rather circular reference, is more accurate than just "blocks" because each L1 block is an epoch (but an epoch may contain multiple L2 blocks).   It's already strongly implied by the context but people often get confused about whether "blocks" means L1 or L2 blocks so best to be clear.

emhane:
need any help here addressing review @miltonjonat ?

PR #268, opened 2024-07-09
interop: token standard without data

0xParti:
**Description**
This PR is a rewrite of [71](https://github.com/ethereum-optimism/specs/pull/71) removing the data parameter from the `sendERC()` and `relayERC20()` functions and corresponding events. The text has been modified accordingly and I added a new section in Future considerations showcasing the cross-chain concatenation flow without using the call in the standard.

The goal of this PR is to open the discussion on whether or not we should include the call optionality as part of the standard.

### DIFF
--- a/specs/interop/token-bridging.md
+++ b/specs/interop/token-bridging.md
@@ -2,7 +2,6 @@
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
-
 **Table of Contents**
 
 - [Overview](#overview)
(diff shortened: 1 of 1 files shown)


0xParti:
The main reason we are considering to remove the call is simplicity. It is possible to achieve the same result without adding this feature to the standard, with two separate calls to the `L2toL2CrossDomainMessenger`. We have been leaning towards simplicity in the standard and we think it makes sense to keep that intuition here as well.

The main issue we can see for the user without the integrated call is the possibility of having message reordering at the sequencer level. This would be a serious issue if and only if the call does not depend on the funds arriving, as otherwise the message would revert but can be replayed later. Notice the integrated call flow also had a large risk when funds where burnt and the call reverted (this can be replayed as well, but if the call was badly formatted for example, funds would be locked).

It would be amazing if there was a way to ensure at the messaging layer that concatenated messages (maybe a new type of message?) get included in the right ordering.

tynes (approved):

tynes:
Fire!

tynes:
Just needs a linting fix then good to go

PR #270, opened 2024-07-10
Alt-DA: Callout No Challenge Mechanism for Generic Commitments

axelKingsley:
This callout was missed in spec previously, and it is an important callout - DA Layers using Generic Commitment types don't have a challenge resolution mechanism.


specs/experimental/alt-da.md
```diff
@@ -149,6 +149,11 @@ Data availability is guaranteed via a permissionless challenge contract on the L
 Users have a set number of L1 blocks (`challengeWindow`) during which they are able to call
 the `challenge` method of the contract with the following inputs:
 
+**Note:** Resolving Input Data through the Data Availability Challenge Contract is implemented
+only for `type=0 (keccak)` commitments. This is because `type=1 (da-service)` commitments are designed to be
+handled by a DA Server which is repsonsible the mapping between commitment and input datas.
```
ajsutton:
```suggestion
handled by a DA Server which is responsible for the mapping between commitment and input data.
```

`data` is already plural. :)

ajsutton (approved):

PR #271, opened 2024-07-11
Proposed: Simplify L1Fee Fjord

MSilb7:
**Description**

Proposing changes:
- Rename `L1Cost` to `L1Fee` to align with transaction receipts
- Add `l1` in front of scalar names to align with transaction receipts
- Delete the text saying that l1GasUsed is updated, since it's deprecated

### DIFF
--- a/specs/fjord/exec-engine.md
+++ b/specs/fjord/exec-engine.md
@@ -23,7 +23,7 @@ Fjord updates the L1 cost calculation function to use a FastLZ-based compression
 The L1 cost is computed as:
 
 ```pseudocode
-l1FeeScaled = baseFeeScalar*l1BaseFee*16 + blobFeeScalar*l1BlobBaseFee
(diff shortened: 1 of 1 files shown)


ajsutton (approved):
LGTM but I tend to think that the `L1GasUsed` field should still be mentioned. Its deprecated but I think that behaviour change still exists and is worth documenting until we actually remove `L1GasUsed`.

tynes:
> LGTM but I tend to think that the `L1GasUsed` field should still be mentioned. Its deprecated but I think that behaviour change still exists and is worth documenting until we actually remove `L1GasUsed`.

I agree with this but now i believe the concept of L1 gas used is in the `GasPriceOracle` predeploy ? Which it is still documented here: https://github.com/ethereum-optimism/specs/blob/cf8bf2c8c56d693e2489719760944da905fd4618/specs/fjord/predeploys.md?plain=1#L57

PR #283, opened 2024-07-23
interop: Add initial indexing api spec

ajsutton:
**Description**

Adds a spec for the first method in the "indexing backend" RPC API. This is designed to specify a standard API for tools that indexing logs across chains to optimise lookups for sequencing and verifying nodes. Written in a way that tries to be independent of any particular design for the indexer.

Still todo:

* [x] `GetCrossUnsafe` method
* [x] `NextDeriveTask` method
* [x] `OnDerived` method
* [x] `CrossSafe` method
* [x] `TryFinalize` method


specs/interop/indexing_api.md
```diff
@@ -0,0 +1,92 @@
+# Indexing API
```
tynes: Thoughts on calling this the "Superchain API"?

ajsutton: I'm slightly against that because I think this is an entirely optional thing that could be used as part of implementing interop. Calling it Superchain API makes it sound like a critical component. While our implementation will depend on it, it's certainly possible to implement interop without it or with a different API if you make different architectural decisions.

ajsutton:
@protolambda I think we moved away from this approach again and should just close this right?

protolambda:
@ajsutton I am not entirely sure yet. We'll likely run into something like this as part of devnet 2 reorg support planning, although maybe we can simplify it. I'm fine with keeping this open for now.

tynes:
We likely should revisit this PR to make sure that the supervisor API is specified

@protolambda @tyler-smith

PR #285, opened 2024-07-23
Initial draft of Alligator Interop spec

corydickson:
This adds the additional specification to allow interop functionality to the `Alligator` contract: https://github.com/ethereum-optimism/specs/pull/247

### DIFF
--- a/specs/governance/alligator-interop.md
+++ b/specs/governance/alligator-interop.md
@@ -97,7 +97,7 @@ The contract may support a function to get the chainId or another modifier to pr
 ## Backwards Compatibility
 
 Previous instances of `Alligator` will not be able to interface with the OP mainnet contract given that they do not support the cross-chain message passing. However, the existing state
-of voting power on the superchain shall remained unchanged.
+of voting power on the non-mainnet chains shall remained unchanged.
 
 ## Security Considerations
 
--- a/specs/governance/alligator-interop.md
+++ b/specs/governance/alligator-interop.md
@@ -56,11 +56,7 @@ sequenceDiagram
 ```solidity
 
 function getChainId() external view returns (uint256) {
-    uint256 chainId;
-    assembly {
-         chainId := chainid()
-    }
-
+    uint256 chainId = block.chainid;
     return chainId;
 }
 
--- a/specs/governance/alligator-interop.md
+++ b/specs/governance/alligator-interop.md
@@ -14,7 +14,7 @@
 
 # Overview
 
-The goal of this specification is to allow interoperability between the base and superchains for that use the [`Alligator`](alligator.md) predeploy. This allows the support of the
+The goal of this specification is to allow interoperability between the OP mainnet and other chains that use the [`Alligator`](alligator.md) predeploy. This allows the support of the
 subdelegation flow with for the [`GovernanceToken`](gov-token.md) by making a few changes to the existing interface for cross-chain message passing as it pertains to the checkpoint state.
 
 ## Interface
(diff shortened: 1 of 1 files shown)


specs/governance/alligator-interop.md
```diff
@@ -0,0 +1,136 @@
...
+  - [Diagram](#diagram)
+  - [Implementation](#implementation)
+  - [Backwards Compatibility](#backwards-compatibility)
+  - [Security Considerations](#security-considerations)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Overview
+
+The goal of this specification is to allow interoperability between the base and superchains for that use the [`Alligator`](alligator.md) predeploy. This allows the support of the
```
tynes: One common misconception is that each chain is called a superchain, but really the superchain is made up of individual chains. What do you mean the base here?

specs/governance/alligator-interop.md
```diff
@@ -0,0 +1,136 @@
...
+    address dst,
+    uint256 amount
+) external
+```
+
+In addition to performing it's standard functionality this method will now use the `L2ToL2CrossDomainMessenger` to send a message to a new method on the `Alligator` contract, `afterTokenTransferInterop`. It will ensure prior to doing so that the `chainId` is not OP Mainnet i.e. 10(0xa). Moreover, it will still handle a check that both address have been migrated using
+the new storage field and update the voting power of the delegate.
+
+```solidity
+function afterTokenTransferInterop(
```
tynes: Can we just add this to `afterTokenTransfer` instead of having 2 functions? ie update the post hook?

specs/governance/alligator-interop.md
```diff
@@ -0,0 +1,136 @@
...
+            (uint256 oldWeight, uint256 newWeight) = _writeCheckpoint(_checkpoints[dst], _add, amount);
+            emit DelegateVotesChanged(dst, oldWeight, newWeight);
+        }
+
+        uint256 nativeChainId;
+        assembly {
+             nativeChainId := chainid()
+        }
+
+        if (nativeChainId != uint256(10)) {
```
tynes: This is not going to be scalable as every token transfer on a remote chain will require a relayer to send a transaction on op mainnet. This means gov will need to front the costs here because nobody else has the incentive to do so. Wondering if there is a way that we can:
- send batched diffs
- not need to be aware of the chainid

Going to brainstorm a bit on this

specs/governance/alligator-interop.md
```diff
@@ -0,0 +1,136 @@
...
+    address src,
+    address dst,
+    uint256 amount
+) external {
+    require(msg.sender == address(L2ToL2CrossChainMessenger));
+    require(L2ToL2CrossChainMessenger.crossDomainMessageSender() == address(this));
+
+    uint256 nativeChainId;
+    assembly {
+         nativeChainId := chainid()
```
tynes: You can do `block.chainid` in modern solidity

specs/governance/alligator-interop.md
corydickson: Explicated the use of other OP chains that are not mainnet and their interaction with the `Alligator` contract

specs/governance/alligator-interop.md
corydickson: Done, updates now only happens if the contract is deployed on OP mainnet

tynes:
I believe there has been a lot of progress on this that has not yet been specified

PR #286, opened 2024-07-23
feat: add new legacy section, move l2OO

smartcontracts:
Adds a new legacy section and moves the L2 Output Oracle spec into the new section. Replaces current Proposals doc with a temporary pointer to the fault proofs doc.


smartcontracts:
Pending #287

specs/protocol/proposals.md
```diff
@@ -1,181 +1,14 @@
...
-    if rollup_status.finalized_l2.number >= next_checkpoint_block:
-        output = op_node_client.output_at_block(next_checkpoint_block)
-        tx = send_transaction(output)
-    time.sleep(poll_interval)
-```
-
-A `CHALLENGER` account can delete multiple output roots by calling the `deleteL2Outputs()` function
-and specifying the index of the first output to delete, this will also delete all subsequent outputs.
-
-## L2 Output Commitment Construction
```
tynes, line 78: I think we can leave this section in here, the definition of the output since it didn't change with fault proofs

PR #294, opened 2024-07-24
interop: liquidity migration specs

0xParti:
**Description**

The Liquidity Migration Specs will cover the required modifications to allowed converting between legacy and `SuperchainERC20` tokens using the `L2StandardBridge`.

**Additional context**

This specs formalizes the [Convert Unified Design Docs](https://github.com/ethereum-optimism/design-docs/pull/46#discussion_r1678444097)

### DIFF
--- a/specs/interop/liquidity_migration.md
+++ b/specs/interop/liquidity_migration.md
@@ -236,7 +236,7 @@ The conversion flow will need to conserve the following invariants:
 
 ### Access Control
 
-Each `SuperchainERC20` will correspond to a a remote token and can be connected
+Each `SuperchainERC20` will correspond to a remote token and can be connected
 to multiple existing legacy representations.
 Furthermore, each legacy representation can be connected to various
 `SuperchainERC20` (one per metadata).
--- a/specs/interop/liquidity_migration.md
+++ b/specs/interop/liquidity_migration.md
@@ -2,7 +2,6 @@
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
-
 - [Overview](#overview)
 - [Implementation changes](#implementation-changes)
   - [L2StandardBridge](#l2standardbridge)
--- a/specs/interop/liquidity_migration.md
+++ b/specs/interop/liquidity_migration.md
@@ -2,6 +2,7 @@
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
 - [Overview](#overview)
 - [Implementation changes](#implementation-changes)
   - [External functions](#external-functions)
@@ -59,13 +60,13 @@ access control function looks like this:
 
 ```solidity
 function convertToSuper(address _legacyAddr, address _superAddr, uint256 _amount) public {
-	require(checkPair(_legacyAddr, _superAddr), "Invalid address pair");
-	require(IERC20(_legacyAddr).decimals() == IERC20(_superAddr).decimals(), "Decimals do not match")
-	
-	IERC20(_legacyAddr).burn(msg.sender, _amount);
-	IERC20(_superAddr).mint(msg.sender, _amount);
-	
-	emit ConvertedToSuper(_legacyAddr, _superAddr, msg.sender, _amount);
+  require(checkPair(_legacyAddr, _superAddr), "Invalid address pair");
+  require(IERC20(_legacyAddr).decimals() == IERC20(_superAddr).decimals(), "Decimals do not match")
+
+  IERC20(_legacyAddr).burn(msg.sender, _amount);
+  IERC20(_superAddr).mint(msg.sender, _amount);
+
+  emit ConvertedToSuper(_legacyAddr, _superAddr, msg.sender, _amount);
 }
 ```
 
@@ -319,7 +320,7 @@ and adjust the amounts and events accordingly:
 ```solidity
 function convertFromSuper(address _legacyAddr, address _superAddr, uint256 _amount) public {
   require(checkPair(_legacyAddr, _superAddr), "Invalid address pair");
-	uint256 _adjustedAmount = computeAmountOut(_legacyAddr, _superAddr, _amount);
+  uint256 _adjustedAmount = computeAmountOut(_legacyAddr, _superAddr, _amount);
 
   IERC20(_superAddr).burn(msg.sender, _amount);
   IERC20(_legacyAddr).mint(msg.sender, _adjustedAmount);
--- a/specs/interop/liquidity_migration.md
+++ b/specs/interop/liquidity_migration.md
@@ -5,19 +5,21 @@
 
 - [Overview](#overview)
 - [Implementation changes](#implementation-changes)
-  - [External functions](#external-functions)
+  - [L2StandardBridge](#l2standardbridge)
     - [`convertToSuper`](#converttosuper)
     - [`convertFromSuper`](#convertfromsuper)
-  - [Events](#events)
     - [`ConvertedToSuper`](#convertedtosuper)
     - [`ConvertedFromSuper`](#convertedfromsuper)
-  - [Internal Functions](#internal-functions)
     - [`checkPair`](#checkpair)
+  - [OptimismMintableERC20Factory](#optimismmintableerc20factory)
+    - [`deployments`](#deployments)
+    - [`createOptimismMintableERC20`](#createoptimismmintableerc20)
 - [Diagram](#diagram)
 - [Invariants](#invariants)
 - [Considerations](#considerations)
 - [Appendix](#appendix)
   - [Access Control](#access-control)
+  - [Backwards compatibility](#backwards-compatibility)
   - [Decimal management](#decimal-management)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
@@ -35,13 +37,15 @@ corresponding `SuperchainERC20`.
 
 ## Implementation changes
 
-The `L2StandardBridge` will be extended to include the following functions and events:
+Bothe the `L2StandardBridge` and the `OptimismMintableERC20TokenFactory` will be extended
+to allow for liquidity migration:
 
-### External functions
+### L2StandardBridge
 
 #### `convertToSuper`
 
-Converts `_amount` of legacy tokens with address `_legacyAddr`
+The `L2StandardBridge` SHOULD add a `convertToSuper` external function that
+converts `_amount` of legacy tokens with address `_legacyAddr`
 to the same `_amount` of `SuperchainERC20` tokens in address `_superAddr`.
 
 ```solidity
@@ -71,7 +75,8 @@ function convertToSuper(address _legacyAddr, address _superAddr, uint256 _amount
 
 #### `convertFromSuper`
 
-Converts `_amount` of `SuperchainERC20` tokens in address `_superAddr`
+The `L2StandardBridge` SHOULD add a `convertFromSuper` external function that
+converts `_amount` of `SuperchainERC20` tokens in address `_superAddr`
 to the same `_amount` of legacy tokens with address `_legacyAddr`.
 
 ```solidity
@@ -98,11 +103,10 @@ function convertFromSuper(address _legacyAddr, address _superAddr, uint256 _amou
 }
 ```
 
-### Events
-
 #### `ConvertedToSuper`
 
-It MUST trigger when anyone converts legacy tokens
+The `L2StandardBridge` SHOULD add a `ConvertedToSuper` event
+that MUST trigger when anyone converts legacy tokens
 to `SuperchainERC20` with `convertToSuper`.
 
 ```solidity
@@ -111,18 +115,18 @@ event ConvertedToSuper(address indexed legacyAddr, address indexed superAddr, ad
 
 #### `ConvertedFromSuper`
 
-It MUST trigger when anyone converts `SuperchainERC20`
+`ConvertedFromSuper` event
+that MUST trigger when anyone converts `SuperchainERC20`
 into legacy tokens with `convertFromSuper`.
 
 ```solidity
 event ConvertedFromSuper(address indexed legacyAddr, address indexed superAddr, address indexed caller, uint256 amount);
 ```
 
-### Internal Functions
-
 #### `checkPair`
 
-Checks the validity of the `_legacyAddr` and `_superAddr` pair.
+The `L2StandardBridge` SHOULD add a `checkPair` internal function that
+checks the validity of the `_legacyAddr` and `_superAddr` pair.
 To do so, it will verify:
 
 1. Valid legacy address:
@@ -170,6 +174,38 @@ function checkPair(address _legacyAddr, address _superAddr) internal view return
 Note: Code can be simplified by using || in
 the requires or conditionals.
 The tradeoff would be losing error precision.
+
+### OptimismMintableERC20Factory
+
+#### `deployments`
+
+The `OptimismMintableERC20Factory` SHOULD include a new mapping that stores the remote
+token address for each deployed `OptimismMintableERC20`.
+
+```solidity
+mapping(address => address) public deployments;
+```
+
+The mapping will be updated on new updates and only once
+manually via storage manipulation to include the list of allowed tokens
+(see the [Appendix](#backwards-compatibility)).
+
+#### `createOptimismMintableERC20`
+
+The function should be modified to update the `deployments` mapping.
+
+```solidity
+function createOptimismMintableERC20(address _remoteToken, string memory _name, string memory _symbol) public returns (address) {
+  require(_remoteToken != address(0), "OptimismMintableERC20Factory: must provide remote token address");
+  bytes32 salt = keccak256(abi.encode(_remoteToken, _name, _symbol));
+  address localToken = address(new OptimismMintableERC20{salt: salt}(BRIDGE, _remoteToken, _name, _symbol));
+  ...
+  // Add the following line
+  deployments[localToken] = _remoteToken;
+  //
+  return localToken;
+}
+```
 
 ## Diagram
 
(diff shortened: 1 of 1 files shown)


specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,358 @@
...
+to the same `_amount` of `SuperchainERC20` tokens in address `_superAddr`.
+
+```solidity
+convertToSuper(address _legacyAddr, address _superAddr, uint256 _amount)
+```
+
+The function will
+
+1. Check that `_legacyAddr` and `_superAddr` are valid and paired.
+2. Check both tokens have the same amount of decimals.
```
smartcontracts: I think this is probably the best option. Figuring out how to handle decimal changes is a big can of worms.

specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,358 @@
...
+#### `checkPair`
+
+Checks the validity of the `_legacyAddr` and `_superAddr` pair.
+To do so, it will verify:
+
+1. Valid `OptimismMintableERC20` address:
+
+   1. For representations minted before
+      the `OptimismMintableERC20Factory` upgrade:
+      Check an allowed token list that connects L2 representations with the
```
smartcontracts: How will this list be populated?

specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,358 @@
...
+      Verify the `OptimismMintableERC20Factories` deployment lists.
+      This can be the same mapping used for 1a.
+
+   This approach requires updating
+   the `OptimismMintableERC20Factory` to store
+   deployments as a mapping from `OptimismMintableERC20`
+   to its `REMOTE_TOKEN`
+   (both allowed list and new deployments).
+
+2. Valid `SuperchainERC20` address:
```
smartcontracts: Sanity check: do we assume that the remote token address is always an L1 address? Is there any case where the remote token address may not be an L1 address?

specs/interop/liquidity_migration.md
smartcontracts: It would be relatively straightforward to do this trustlessly but would require that we have an external function that can be called to populate the mapping. I would prefer that over a trusted solution.

0xng: @smartcontracts the idea was to do it in a trusted manner, here's a relevant comment from @tynes: https://github.com/ethereum-optimism/design-docs/pull/46#discussion_r1678444097, proto had suggested a slightly different approach using an onion-hash, which is the R&D discord

specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,358 @@
...
+
+You can read more details about access control in the [Appendix](#decimal-management).
+
+A reference implementation would look like follows
+
+```solidity
+IOptimismMintableERC20Factory public factory = IOptimismMintableERC20Factory(Predeploys.OptimismMintableERC20Factory);
+ISuperchainERC20Factory public superFactory = ISuperchainERC20Factory(Predeploys.SuperchainERC20Factory);
+
+function isValidPair(address _legacyAddr, address _superAddr) internal view returns (bool) {
```
skeletor-spaceman: should this be `checkPair`? or should that be renamed to `isValidPair`?

specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,358 @@
...
+}
+```
+
+#### `convertFromSuper`
+
+Converts `_amount` of `SuperchainERC20` tokens in address `_superAddr`
+to the same `_amount` of legacy tokens with address `_legacyAddr`.
+
+```solidity
+convertToSuper(address _legacyAddr, address _superAddr, uint256 _amount)
```
tynes: Should this be `convertFromSuper`?

tynes:
@mds1 It would be really good to have your review on this if you have time

specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,358 @@
...
+The function will
+
+1. Check that `_legacyAddr` and `_superAddr` are valid and paired.
+2. Check both tokens have the same amount of decimals.
+3. Burn `_amount` of `_legacyAddr` from `msg.sender`.
+4. Mint `_amount` of `_superAddr` to `msg.sender`.
+
+An example implementation that depends on the `checkPair()`
+access control function looks like this:
+
```
tynes, line 62: Wondering how you feel about the following ABI:

```solidity
function convert(address _from, address _to, uint256 _amount) public {
  // perhaps modularize the internals of these branches a bit, to dedupe code
  if (_isOptimismMintableERC20(_from)) {
    require(_isOptimismSuperchainERC20(_to));
    require(checkPair(_from, _to));
    _mintAndBurn(_from, _to, _amount);
  } else if (isOptimismMintableERC20(_to)) {
    require(_isOptimismSuperchainERC20(_from));
    require(checkPair(_to, _from));
   _mintAndBurn(_to, _from, _amount);
  } else {
    revert();
  }
}

function _mintAndBurn(address _from, address _to, _amount) internal {
    IERC20(_from).burn(msg.sender, _amount);
    IERC20(_to).mint(msg.sender, _amount);
}
```

I think this simplifies the devex as you only have a single function that can be used to convert either direction. The [isOptimismMintableERC20Token](https://github.com/ethereum-optimism/optimism/blob/3c1d3b87573bc3c50eb88a64fff7cd633270d4f4/packages/contracts-bedrock/src/universal/StandardBridge.sol#L398) implementation already exists, we would just need to add one for the superchain erc20

I also think that the decimals check could go inside of `checkPair`

tynes: The safety of this is in the burn, see [here](https://github.com/Vectorized/solady/blob/1f43cc8005cc3b3c8361dd7dbdd2cdeaf0f99e66/src/tokens/ERC20.sol#L442)

tynes: Just walking through a user flow here when interacting with a smart contract, you could `approve` a contract, the contract could `transferFrom` and then `convert` and send cross chain

specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,358 @@
...
+Legacy token owners must convert into a `SuperchainERC20` representation, as defined in the [standard](token-bridging.md),
+to move across the Superchain.
+
+The conversion method will use the `L2StandardBridge` mint/burn rights
+over the legacy tokens to allow easy migration to and from the
+corresponding `SuperchainERC20`.
+
+## Implementation changes
+
+The `L2StandardBridge` will be extended to include the following functions and events:
```
tynes: Perhaps you could make a `### L2StandardBridge` section and a `### OptimismMintableERC20TokenFactory` section so that we have headers for the specific contracts getting changes

tynes: This is good that it lists the specific changes for the `L2StandardBridge`, it would be good to note the specific changes to the factory, ie the new mapping for deployed contracts

specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,358 @@
...
+### Internal Functions
+
+#### `checkPair`
+
+Checks the validity of the `_legacyAddr` and `_superAddr` pair.
+To do so, it will verify:
+
+1. Valid `OptimismMintableERC20` address:
+
+   1. For representations minted before
```
tynes: I think that we can leave this bullet point for a backwards compatibility section and just say that the validity involves checking the canonical token mapping in the factory + a decimals check. Do we need to do checks for other erc20 metadata equality?

specs/interop/liquidity_migration.md
tynes: We can assume that the mapping will be populated correctly as part of this spec, we can work on a design doc for how to populate the mapping separately

specs/interop/liquidity_migration.md
tynes: I see the following pseudocode below:

```solidity
  address _legacyRemoteToken = factory.deployments(_legacyAddr);
```

specs/interop/liquidity_migration.md
tynes: I think this function should also be responsible for checking erc20 metadata such as decimals

specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,358 @@
...
+and `IOptimismMintableERC20`.
+The Token List will be a one-time update conducted by Optimism and will
+live in the `OptimismMintableERC20Factory` contract.
+If the legacy token is not included in the token list
+(for new tokens, for instance),
+the `checkPair` will verify validity against the factory.
+
+### Decimal management
+
+It is possible to remove the revert on decimal difference
```
tynes: I lean towards just asserting that the decimals are the same and not worrying about scaling them

specs/interop/liquidity_migration.md
0xng: @tynes The approach you mention is great IMO. 

This is a timely comment as well, I was debating whether the SuperchainERC20 implementation for convert should inherit from `OptimismMintableERC20` as they share most of the functionality (gives permissions to the bridge, and has a remoteToken it wraps, so very similar to the non-legacy interface). The only thing that annoyed me a bit was inheriting the legacy getters. Do you have any preference on this? As in whether to:

1. Inherit from OptimismMintableERC20 and override `supportsInterface` with `type(IOptimismSuperchainERC20).interfaceId`

2. Don't inherit but copy the non-legacy logic. 

Something to note is that if we want to allow for direct deposits, we should add this new interface to the `StandardBridge` as well

specs/interop/liquidity_migration.md
0xParti: > should this be `checkPair`? or should that be renamed to `isValidPair`?

Yes, that was a typo (I changed the function name last min). Will fix in the next commit.

> I think this function should also be responsible for checking erc20 metadata such as decimals

Absolutely. We have been keeping functions as specific as possible, but checking decimal compatibility can be considered part of "checking the Pair validity". Will move it.

specs/interop/liquidity_migration.md
0xParti: For the scope of this spec, we can assume the remote token is an L1 address. 
We had an internal discussion on whether to name it `l1Address` or something similar for this spec. Still, we went for the remote naming because we think a similar logic will eventually be used for L2 native tokens, where the remote address will not correspond to L1 but to the native L2.

specs/interop/liquidity_migration.md
0xParti: Yes, that seems to be the consensus. 

For context, L1 deposits do not check decimals at all, and it's perfectly the same to do the same here (remove the check altogether). In that case, the market should adjust the pricing for each representation.
So, both approaches are safe:
- enforce the same decimals
- don't enforce and let the market price the assets
I think it's mostly a UX decision.

specs/interop/liquidity_migration.md
0xParti: That's a great idea, will do it

specs/interop/liquidity_migration.md
0xParti: We introduced two functions instead of one to keep functions as specific as possible and reduce gas while leaving "ABI" complexities to the integrations. 

That said, the simplification looks elegant, and the gas increase is negligible, so it is probably a good idea.

What do you think @tynes about moving the interface checker inside `checkPair` and removing the conditionals from the main function?

specs/interop/liquidity_migration.md
0xParti: I agree it creates some confusion for first time readers. I could move the details to an Appendix. 
No need to check other metadata equality I think.

specs/interop/liquidity_migration.md
0xParti: Yes, that seems to be the consensus. Would you remove the whole section?

specs/interop/liquidity_migration.md
0xParti: Also:
- `_mintAndBurn()` should only be called as `_mintAndBurn(_from, _to, _amount)`, never as `_mintAndBurn(_to, _from, _amount)`, right?
- The `require(_isOptimismSuperchainERC20(...))` is redundant. This is already checked in `checkPair` by verifying the token was deployed by the `SuperchainERC20Factory`.

0xParti: What do you think about something like this?

```solidity
function convert(address _from, address _to, uint256 _amount) public {
  _validatePair(_from, _to);
  IERC20(_from).burn(msg.sender, _amount);
  IERC20(_to).mint(msg.sender, _amount);
  emit Converted(_from, _to, msg.sender, _amount);
}

function _validatePair(address _from, address _to) internal {
  require(IERC20Metadata(_from).decimals() == IERC20Metadata(_to).decimals(), "Decimals do not match")
  if (_isOptimismMintableERC20(_from)) {
    _validateFactories(_from, _to)
  } else {
    _validateFactories(_to, _from)
  }
}

function _validateFactories(address _legacyAddr, address _superAddr) internal {
  address _legacyRemoteToken = factory.deployments(_legacyAddr);
  require(_legacyRemoteToken != address(0), "Invalid Legacy address");

  address _superRemoteToken = superFactory.deployments(_superAddr);
  require(_superRemoteToken != address(0), "Invalid SuperchainERC20 address");
  
  require(_legacyRemoteToken == _superRemoteToken, "Invalid token pair");
}
```

We feel like we can merge the `_isOptimismMintableERC20(...)` check with the `factory.deployments(_from)!=address(0)`, but we should check.

specs/interop/liquidity_migration.md
0xParti: Moved to the Appendix

specs/interop/liquidity_migration.md
0xParti: Done

specs/interop/liquidity_migration.md
0xParti: We can also remove the decimal check and not adjust, wdyt?

specs/interop/liquidity_migration.md
tynes: This definitely looks good to me. I think its sufficient - no need for the interface checks this way

tynes: > This is a timely comment as well, I was debating whether the SuperchainERC20 implementation for convert should inherit from OptimismMintableERC20

@0xng I lean towards not inheriting from it given its based on oz rather than solady, we could update the version of it to be based on solady but i think it makes sense to just implement a net new contract. open to ideas tho

specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,389 @@
...
+Legacy token owners must convert into a `SuperchainERC20` representation, as defined in the [standard](token-bridging.md),
+to move across the Superchain.
+
+The conversion method will use the `L2StandardBridge` mint/burn rights
+over the legacy tokens to allow easy migration to and from the
+corresponding `SuperchainERC20`.
+
+## Implementation changes
+
+Bothe the `L2StandardBridge` and the `OptimismMintableERC20TokenFactory` will be extended
```
tynes: typo: `both`

specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,389 @@
...
+(see the [Appendix](#backwards-compatibility)).
+
+#### `createOptimismMintableERC20`
+
+The function should be modified to update the `deployments` mapping.
+
+```solidity
+function createOptimismMintableERC20(address _remoteToken, string memory _name, string memory _symbol) public returns (address) {
+  require(_remoteToken != address(0), "OptimismMintableERC20Factory: must provide remote token address");
+  bytes32 salt = keccak256(abi.encode(_remoteToken, _name, _symbol));
```
tynes: I think that we can leave out most of this pseudocode and just define the salt and the new `deployments` mapping

specs/interop/liquidity_migration.md
0xParti: Updated to a single `convert` in the last commit.

specs/interop/liquidity_migration.md
0xParti: Done

specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,266 @@
...
+Note: Code can be simplified by using || in
+the requires or conditionals.
+The tradeoff would be losing error precision.
+
+### OptimismMintableERC20Factory
+
+#### Version
+
+Ensure the `OptimismMintableERC20Factory` implementation uses CREATE2 instead of CREATE and decimals as part of the salt.
+This is in line with the latest stable implementation in the
```
tynes: This comment is going to go stale, meaning that at some point it will no longer be the latest implementation

specs/interop/liquidity_migration.md
```diff
@@ -0,0 +1,266 @@
...
+
+Note: Code can be simplified by using || in
+the requires or conditionals.
+The tradeoff would be losing error precision.
+
+### OptimismMintableERC20Factory
+
+#### Version
+
+Ensure the `OptimismMintableERC20Factory` implementation uses CREATE2 instead of CREATE and decimals as part of the salt.
```
tynes: I think that we should add the spec for how the salt is computed. This is useful so that we can reason about invariants around how the `deployments` mapping is populated.

Right now it is:

```solidity
bytes32 salt = keccak256(abi.encode(_remoteToken, _name, _symbol, _decimals));
```

Since this isn't done with create3, a change in the solidity compiler version will result in different addresses. If it is easy to migrate to a create3 deployer, that would be preferred, but certainly not required. For the new factory we would definitely want to use create3

specs/interop/liquidity_migration.md
tynes: I recommend just removing this paragraph and adding a paragraph that says that the mapping may also be filled with addresses computed from alternative mechanisms for some chains

tynes (approved):
This is amazing! Great work :D

specs/interop/liquidity_migration.md
0xParti: I changed the paragraph to be as open as possible, only discouraging the use of CREATE, discussing the issue with CREATE2 and suggesting CREATE3 eventually. 

Would you like us to work on a dedicated spec for the salt of the `OptimismMintableERC20Factory` then?

tynes: What is the additional work in your mind besides what i linked above?

0xParti: Nevermind, lol. It would not be worth.
I modified the paragraph in the latest commit. Let me know what do you think of it.

tynes (approved):

PR #297, opened 2024-07-26
interop: Initial honest node spec

ajsutton:
**Description**

Adds a very very rough honest node spec to work through how a node should use the supervisor API.

### DIFF
--- a/specs/interop/honest-nodes.md
+++ b/specs/interop/honest-nodes.md
@@ -1,0 +1,183 @@
+# Honest Nodes
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Overview](#overview)
+- [Sequencing](#sequencing)
+- [Unsafe](#unsafe)
+- [Cross-Unsafe](#cross-unsafe)
+- [Local Safe](#local-safe)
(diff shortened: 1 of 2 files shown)


specs/interop/honest-nodes.md
```diff
@@ -0,0 +1,183 @@
...
+efficient verification of executing message dependencies. Implementations MAY differ from this approach providing they
+still adhere to the protocol rules.
+
+## Sequencing
+
+Sequencing nodes continue building new `unsafe` blocks on top of the existing `unsafe` chain head. For
+each [ExecutingMessage event][event] emitted by a transaction being added to the new block, the sequencer MUST
+call [`interop_checkMessage`] with the identifier and payload hash from the event.
+
+Sequencers MUST remove any transactions with any `conflicts` or `unknown` events from the block.
```
tynes, line 34: It could be good to use language that implies "drop transactions while building the block" to make it clear its not removing any transactions from sealed blocks

ajsutton:
@protolambda I think we moved away from this approach again and should just close this right?

protolambda:
@ajsutton (as with other spec doc) I am not entirely sure yet. We'll likely run into something like this as part of devnet 2 reorg support planning, although maybe we can simplify it. I'm fine with keeping this open for now.

ajsutton:
Given how old this is now, I'm going to close it.  It will still be around for reference if we need it but so much has changed it's unlikely to be useful as-is anyway.

PR #299, opened 2024-07-26
linting: add filenames linting task

tynes:
**Description**

Adds a just recipe for linting filenames. This blocks any filename
that uses an underscore in its name. Dashes should be preferred
over underscores.


protolambda (approved):
It appears that the assets that were renamed were not referenced in the spec to begin with, so no broken links to fix there.

PR #300, opened 2024-07-28
Added details about the ERC20 Withdrawal Back to Domain function?

nidz-the-fact:
**Description**

Added details about the `ERC20 Withdrawal Back to Domain` function? Please consider this.
cc. https://specs.optimism.io/experimental/custom-gas-token.html#user-flow

- For When ETH is the Native Asset
should call the function `L2StandardBridge`.**withdraw**`(address,uint256,uint32,bytes)`
cc. https://github.com/ethereum-optimism/optimism/blob/op-contracts/v2.0.0-beta.2/packages/contracts-bedrock/src/L2/L2StandardBridge.sol#L93-L106

- For When an ERC20 Token is the Native Asset
should call the function `L2StandardBridge`.**bridgeERC20**`(address,address,uint256,uint32,bytes)`
cc. https://github.com/ethereum-optimism/optimism/blob/op-contracts/v2.0.0-beta.2/packages/contracts-bedrock/src/L2/L2StandardBridge.sol#L86-L88

**_Please consider this._**

### DIFF
--- a/specs/experimental/custom-gas-token.md
+++ b/specs/experimental/custom-gas-token.md
@@ -335,6 +335,7 @@ meaning that "sending to the other domain" indicates it can be used for both dep
 | Native Asset Deposit | `OptimismPortal.depositTransaction(address,uint256,uint64,bool,bytes) payable` | None |
 | ERC20 Send to Other Domain | `L1StandardBridge.bridgeERC20(address,address,uint256,uint32,bytes)` | Approve `L1StandardBridge` for ERC20 |
 | Native Asset Withdrawal | `L2ToL1MessagePasser.initiateWithdrawal(address,uint256,bytes) payable` | None |
+| ERC20 Withdrawal Back to Domain | `L2StandardBridge.withdraw(address,uint256,uint32,bytes)` | Approve `L2StandardBridge` for ERC20 |
 
 There are multiple APIs for users to deposit or withdraw `ether`. Depending on the usecase, different APIs should be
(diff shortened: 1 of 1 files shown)


tynes:
`L2StandardBridge.withdraw(address,uint256,uint32,bytes)` is a legacy method and we do not recommend users to use it. The modern methods can be found on the [StandardBridge](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v2.0.0-beta.2/packages/contracts-bedrock/src/universal/StandardBridge.sol) and are symmetric on both L1 and L2

tynes:
I don't understand what you mean by recommending `L2StandardBridge.bridgeERC20(address,address,uint256,uint32,bytes)` because it will not deposit native ERC20 asset

nidz-the-fact:
> I don't understand what you mean by recommending `L2StandardBridge.bridgeERC20(address,address,uint256,uint32,bytes)` because it will not deposit native ERC20 asset

I mean the detailed introduction of **ERC20** withdrawal from L2 back to L1. 
To call `0x420000000000000000000000000000000010` Use the `bridgeERC20` function.

Because the L2 withdrawal of ERC20 is quite different from Native and calls the function
Is it necessary to add this detail and information?
erc20-w

tynes:
The table is meant to illustrate how to interact with native asset and is not meant to show how to send erc20 tokens

nidz-the-fact:
Ok, thanks for replying to the comment. I will close this PR because it is not yet suitable for improvement.

PR #303, opened 2024-07-31
Refactor the Security Council Safe specs into Admin and Auth model specs

maurelian:
Renames and updates  the former Security Council Safe spec to a more comprehensive Superchain Administration and Authorization Model. 

This includes detailed descriptions of the various safes, their purposes, their thresholds, and ownership models.

Further PRs will be stacked on this one to keep the diff as readable as possible.

### DIFF
--- a/specs/experimental/admin-and-auth.md
+++ b/specs/experimental/admin-and-auth.md
@@ -86,7 +86,9 @@ extensions.
 
 ## Ownership model diagram
 
-The following diagram outlines the control relationships between the contracts in the
+The following diagram outlines the control relationships between the contracts in the system.
+
+
 
 ```mermaid
 flowchart LR
--- a/specs/experimental/admin-and-auth.md
+++ b/specs/experimental/admin-and-auth.md
@@ -129,9 +129,7 @@ flowchart LR
 
    %% Declare a class to make the outer boxes somewhat transparent to provide some contrast
    classDef outer fill:#ffffff44
-   class SuperchainSystem outer
-   class GuardianSystem outer
-   class UpgradeSystem outer
+   class SuperchainSystem,GuardianSystem,UpgradeSystem outer
 ```
 
 The remainder of this document outlines each safe, including its key configuration parameters, and
--- a/specs/experimental/admin-and-auth.md
+++ b/specs/experimental/admin-and-auth.md
@@ -1,9 +1,13 @@
-# Safe contract configuration
+# Superchain Administration and Authorization Model
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 **Table of Contents**
 
+- [Overview](#overview)
+- [List of Safes](#list-of-safes)
+- [Ownership model diagram](#ownership-model-diagram)
+- [Guardian Safe](#guardian-safe)
 - [Deputy guardian module](#deputy-guardian-module)
   - [Deputy Guardian Module Security Properties](#deputy-guardian-module-security-properties)
 - [Liveness checking mechanism](#liveness-checking-mechanism)
@@ -27,9 +31,13 @@
 
 ## Overview
 
-Administrative actions in the Superchain are managed by a system of multisignature Safe contracts,
- as well as custom extensions to those Safe contracts which provide additional functionality. This
-document describes the system of Safe's and their purposes.
+Authorization of administrative actions in the Superchain are managed by a system of multisignature
+ Safe contracts, as well as custom extensions to those Safe contracts which provide additional
+functionality. This document describes the system of Safe's and their purposes.
+
+## List of Safes
+
+This list outlines the various Safes, including thresholds, ownership and extensions.
 
 1. **The ProxyAdminOwner Safe:** The name of this Safe is slightly misleading. While it does control
    the `ProxyAdmin` contract, can therefore upgrade contracts in the system, more generally it is
@@ -71,6 +79,8 @@ document describes the system of Safe's and their purposes.
    This Safe has a threshold of 5 and has the same 7 owners as the Foundation Upgrade Safe.
 
 ## Ownership model diagram
+
+The following diagram outlines the control relationships between the contracts in the
 
 ```mermaid
 flowchart LR
--- a/specs/experimental/admin-and-auth.md
+++ b/specs/experimental/admin-and-auth.md
@@ -7,7 +7,6 @@
 - [Overview](#overview)
 - [List of Safes](#list-of-safes)
 - [Ownership model diagram](#ownership-model-diagram)
-- [Guardian Safe](#guardian-safe)
 - [Deputy guardian module](#deputy-guardian-module)
   - [Deputy Guardian Module Security Properties](#deputy-guardian-module-security-properties)
 - [Liveness checking mechanism](#liveness-checking-mechanism)
--- a/specs/experimental/admin-and-auth.md
+++ b/specs/experimental/admin-and-auth.md
@@ -138,13 +138,6 @@ Note: in the diagram above, the [`ProtocolVersions`
    Protocol", because the decision to follow the version signals in the contract is optional. It is
    included here for completeness, but is not considered as either Safety or Liveness affecting.
 
-## Guardian Safe
-
-The Security Council (at
-[eth:0xc2819DC788505Aac350142A7A707BF9D03E3Bd03](https://etherscan.io/address/0xc2819DC788505Aac350142A7A707BF9D03E3Bd03))
-uses a specially extended Safe multisig contract to provide additional security guarantees on top of
-those provided by the Safe contract.
-
 ## Deputy guardian module
 
 The Security Council acts as the Guardian, which is authorized to activate the [Superchain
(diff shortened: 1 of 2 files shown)


maurelian:
* **#303** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/specs/303?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @maurelian and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

specs/experimental/admin-and-auth.md
```diff
@@ -25,21 +28,131 @@
...
+   classDef outer fill:#ffffff44
+   class SuperchainSystem,GuardianSystem,UpgradeSystem outer
+```
+
+Note: in the diagram above, the [`ProtocolVersions`
+   contract](../protocol/superchain-upgrades.md#protocolversions-l1-contract) is listed as "Out of
+   Protocol", because the decision to follow the version signals in the contract is optional. It is
+   included here for completeness, but is not considered as either Safety or Liveness affecting.
 
 ## Deputy guardian module
```
maurelian, line 137: Note to reviewers regarding the diff below this line: 

1. I ran an in IDE 'rewrap' which resulted in some whitespace changes below, but they should be superficial only. 
2. Some of the content below is outdated, I plan to address this in a follow up PR.

specs/experimental/admin-and-auth.md
```diff
@@ -1,9 +1,12 @@
-# Security Council Safe
+# Superchain Administration and Authorization Model
```
mds1, line 1: Maybe a nit, but is there a difference you are trying to convey between "Administration" and "Authorization"? Wondering if we should shorten this to 
```suggestion
# Superchain Administration Model
```

specs/experimental/admin-and-auth.md
mds1: Got it, so I will not review below this line

specs/experimental/admin-and-auth.md
```diff
@@ -25,21 +28,131 @@
...
-those provided by the Safe contract.
+## Overview
+
+Authorization of administrative actions in the Superchain are managed by a system of multisignature
+ Safe contracts, as well as custom extensions to those Safe contracts which provide additional
+functionality. This document describes the system of Safe's and their purposes.
+
+## List of Safes
+
+This list outlines the various Safes, including mainnet addresses, thresholds, ownership, and any
```
mds1, line 39: Should specs include addresses? They aren't technically part of the spec and I don't think we list them anywhere else in the spec (excluding e.g. predeploys). Maybe we talk with:
-  @tessr about adding them to the superchain registry? I think they would fit here: https://github.com/ethereum-optimism/superchain-registry/blob/4810f97b7ee9407e383761c00ff89856c33d6156/superchain/configs/mainnet/superchain.toml#L2-L3
- @sbvegan about adding them to the docs?

specs/experimental/admin-and-auth.md
```diff
@@ -25,21 +28,131 @@
...
+extensions.
+
+1. **[The ProxyAdminOwner
```
mds1: We should document the safe versions we are using for each of these safes and note that this spec and the extensions are only compatible and tested with those versions

specs/experimental/admin-and-auth.md
```diff
@@ -25,21 +28,131 @@
...
+   Safe](https://etherscan.io/address/0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A):** The name of
+    this Safe is slightly misleading. While it does control the `ProxyAdmin` contract, can therefore
+   upgrade contracts in the system, more generally it is in charge of _safety_, meaning it should
+   control any action which has an impact on the determination of a valid L2 state, or the custody
+   of bridged assets. This includes but is not limited to upgrading L1 contracts, and modifying the
+   implementation of the dispute game.
```
mds1: Hmm, I think we should avoid the specs becoming the source of truth for the current configuration and roles in the system (current config should be in docs, and roles should be a separate specs page, similar to [this docs page](https://docs.optimism.io/chain/security/privileged-roles)), and instead have them focus on the extensions and how they're intended to be used with Safes. The "List of Safes" content seems better suited for docs, but the rest of the spec seems like the right content

specs/experimental/admin-and-auth.md
```diff
@@ -25,21 +28,131 @@
...
+1. **[The Foundation Operations
+   Safe](https://etherscan.io/address/0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A):** This Safe acts
+   as the Deputy Guardian, meaning that (via the Guardian Safes's `DeputyGuardianModule`) can call
+   any functions in the system which impact liveness.
+
+   This Safe has a threshold of 5 and has the same 7 owners as the Foundation Upgrade Safe.
+
+## Ownership model diagram
+
+The following diagram outlines the control relationships between the contracts in the
```
mds1: sentence is cut off

specs/experimental/admin-and-auth.md
```diff
@@ -25,21 +28,131 @@
...
+
+```mermaid
```
mds1, line 83: I love this diagram, so glad it's going to be in the specs now

specs/experimental/admin-and-auth.md
```diff
@@ -25,21 +28,131 @@
...
+flowchart LR
+   subgraph SuperchainSystem[Superchain System]
+      subgraph Liveness
```
mds1, line 86: I see the right side of this diagram changed a bit from the original version to be more generic/abstract, whereas the older one was more concrete. Personally I prefer the original one because it's more explicit. I would be curious to hear more about the motivation for this change. (Original version below for reference)

I think one argument in favor of your version is that it's easier to maintain by being more generic, but that also makes it harder (maybe impossible?) to build the contracts just from the spec. We should ideally have enough detail that an implementer can re-implement our architecture properly even if they can't see the OP Mainnet contracts and code

image

specs/experimental/admin-and-auth.md
sbvegan: They should be in the docs, ideally the privileged roles doc gets refactored to be privileged roles in the Superchain. All members of the superchain should be included by pulling that information from the superchain registry

specs/experimental/admin-and-auth.md
maurelian: My thinking is that Administration refers to what actions can be taken, and Authorization refers to how permissions are granted and verifed. 

If I had to choose, I think I'd prefer to keep the latter.

specs/experimental/admin-and-auth.md
maurelian: Adding them to the registry and docs seem like a good idea to me.

maurelian: Adding the addresses was a last minute thing I'm fine to remove them.

specs/experimental/admin-and-auth.md
maurelian: > roles should be a separate specs page

I'm not sure about that, it seems to me that roles are tightly linked with the Safes/modules/guards that help to define those roles.

Some clarifying questions:
- would you also move the diagram to that page? 
- how should the fact that the modules are specifically associated with roles be handled? 

I imagine there would be a lot of hyperlinking between the two docs. 

> The "List of Safes" content seems better suited for docs

I understand the point regarding the config content (threshold/owners etc.), but the other information about the purpose of a given Safe and what it controls feels very relevant (and is basically a prose version of the diagram).

specs/experimental/admin-and-auth.md
maurelian: IMO if the audience understands the difference between Liveness and Safety, they should be able to re-implement the system. 

The diagram should serve as a guide rather than a spec, otherwise it will get quite messy (the screenshot is missing the POA's auth to call the DGF's `setImplementation` and `setInitBond` for example). 

To address this comment, I propose updating the 'List of Safes' content to specify more precisely which contracts/methods each account can call.

specs/experimental/admin-and-auth.md
maurelian: 11a9254

specs/experimental/admin-and-auth.md
maurelian: Actually, for Safess owned by other Safes (the POA and Guardian Safe) it still seems necessary to keep the threshold and ownership info. WDYT?

specs/experimental/admin-and-auth.md
maurelian: ToDo: check my list of items to document. There may be something relevant to changes introduced in v1.4.0.

specs/experimental/admin-and-auth.md
maurelian: > Hmm, I think we should avoid the specs becoming the source of truth for the current configuration and roles in the system (current config should be in docs, and roles should be a separate specs page, similar to [this docs page](https://docs.optimism.io/chain/security/privileged-roles)), and instead have them focus on the extensions and how they're intended to be used with Safes. The "List of Safes" content seems better suited for docs, but the rest of the spec seems like the right content

We've agreed to: 
1. assume stage 1 is a requirement for implementing the protocol
2. create a new specs doc dedicated to that requirement, it will include the list of safes and diagram
3. have this document focus on the Safe extensions.

maurelian:
Replaced by the stack on top of #326.

PR #304, opened 2024-08-01
feat: change tense of events to match other contracts

0xng:
**Description**

Changes the tense of the events to match the tenses of the events emitted in other contracts like [CrossDomainMessenger](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/universal/CrossDomainMessenger.sol#L194) and [L2ToL2CrossDomainMessenger](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L2/L2ToL2CrossDomainMessenger.sol#L183)


tynes:
This is a good catch, better to keep things standard even tho i prefer it not that way

tynes:
Curious what you think @mds1 ?

mds1:
We do tend to use past tense so I'm supportive of this. We should also update the style guide accordingly

tynes (approved):

tynes:
I think that past tense is a bit of an anti pattern, for example [Swap](https://github.com/Uniswap/v3-core/blob/d8b1c635c275d2a9450bd6a78f3fa2484fef73eb/contracts/UniswapV3Pool.sol#L786) rather than `Swapped` but we do it around the codebase already

tynes (approved):

PR #306, opened 2024-08-01
specs: standardize folder structure

tynes:
**Description**

Move fjord to the same folder structure as the other
hardfork directories. Previously, fjord was special
as it was in a subdirectory. It should be siblings
to the other hardforks.


protolambda (approved):

tynes:
looks like some links are failiing, will fix

ajsutton:
Could we add redirects to `book.toml` to avoid breaking external links please?

tynes:
@ajsutton I added the redirect links, good call

PR #309, opened 2024-08-02
feat: update `OptimismMintableERC20Factory` to use CREATE3

0xParti:
**Description**

Updated the `OptimismMintableERC20Factory` specs in `predeploys.md` to use `CREATE3` and store deployments.

This update will allow to implement the liquidity migration process as described in the [Liquidity Migration specs](https://github.com/ethereum-optimism/specs/pull/294).

The implementation link was temporarily removed from the document.

### DIFF
--- a/specs/protocol/predeploys.md
+++ b/specs/protocol/predeploys.md
@@ -298,7 +298,9 @@ and implement the interface required by the `StandardBridge` to just work with d
 Each `OptimismMintableERC20` contract that is created by the `OptimismMintableERC20Factory` allows for the `L2StandardBridge` to mint
 and burn tokens, depending on if the user is depositing from L1 to L2 or withdrawing from L2 to L1.
 
-The factory will deploy using `CREATE3`, where the salt will depend on the user inputs: `remoteToken`, `name`, `symbol`, and `decimals`. The use of `CREATE3` over `CREATE2` removes any dependence on the compiler.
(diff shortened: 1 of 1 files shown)


tynes:
Ah apologies @0xParticle - i meant `specs/interop/predeploys.md` for the file to update, in `predeploys.md` it describes the bedrock release and then we have hardfork specific modifications in the hardfork folders. Generally the idea is to try to keep similar filenames as in the root and in hardfork specific folders

specs/protocol/predeploys.md
```diff
@@ -289,16 +289,33 @@ upgraded by changing its proxy's implementation key.
...
 and implement the interface required by the `StandardBridge` to just work with deposits and withdrawals.
 
-Each ERC20 contract that is created by the `OptimismMintableERC20Factory` allows for the `L2StandardBridge` to mint
-and burn tokens, depending on if the user is depositing from L1 to L2 or withdrawing from L2 to L1.
+Each `OptimismMintableERC20` contract that is created by the `OptimismMintableERC20Factory`
+allows for the `L2StandardBridge` to mint
+and burn tokens, depending on if the user is
+depositing from L1 to L2 or withdrawing from L2 to L1.
+
+The factory will deploy using `CREATE3`, where the salt will depend on the user inputs:
```
tynes, line 303: It would be really useful to have a high level overview of what create3 is someplace in the specs, not sure where it makes the most sense. It could be a `create3.md` file

PR #310, opened 2024-08-02
chore: Remove hint routes from the fault proof specification

clabby:
## Overview

Removes hint routes from the fault proof specification.

### Rationale

Hint routes are an implementation-specific construct. How a client/host program choose to hint data is not of the concern of the specification to define; Rather the specification should define the types of preimages that may be received over the preimage oracle ABI (as it does with preimage key types)

Hints on-chain are a no-op; They play no critical role except for in the off-chain component, where the host may choose its own methods of populating the preimage key-value store based off of whatever hints it receives.

### DIFF
--- a/specs/fault-proof/index.md
+++ b/specs/fault-proof/index.md
@@ -25,15 +25,6 @@
   - [Main content](#main-content)
   - [Epilogue](#epilogue)
   - [Pre-image hinting routes](#pre-image-hinting-routes)
-    - [`l1-block-header <blockhash>`](#l1-block-header-blockhash)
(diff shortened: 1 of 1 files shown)


tynes:
I believe that we decided to not merge this

tynes:
Can you confirm @clabby ?

clabby:
> Can you confirm @clabby ?

That decision hasn't been made. I still think we should remove this.

Inphi:
It's useful to have the hint routes documented somewhere. I'd move these routes to an "implementation notes" section in the spec that simply says these routes are currently used by the op-program.

tynes:
Going to close this since its the source of truth for the hint docs

PR #314, opened 2024-08-05
interop: update predeploys.md

0xParti:
**Description**
Updated the `predeploys.md` spec to include details on the new `OptimismMintableERC20Factory`. 
I only focused on the updates to keep it similar to the L1Block section (also an update), but did not make any comments about the actual public and external functions, or invariants. I can include all of this and turn it into a more explicit update or a full spec.


specs/interop/predeploys.md
```diff
@@ -417,6 +421,46 @@ dependency set called `dependencySet()`. This function MUST return the array of
...
+## OptimismMintableERC20Factory
+
+| Constant | Value                                        |
+| -------- | -------------------------------------------- |
+| Address  | `0x4200000000000000000000000000000000000012` |
+
+### OptimismMintableERC20
+
+The `OptimismMintableERC20Factory` creates ERC20 contracts on L2 that can be used to deposit
+native L1 tokens into (`OptimismMintableERC20`). Anyone can deploy `OptimismMintableERC20`s contracts.
```
tynes: nit: trailing `s` reads strange in "`OptimismMintableERC20`s"

specs/interop/predeploys.md
```diff
@@ -417,6 +421,46 @@ dependency set called `dependencySet()`. This function MUST return the array of
...
+allows for the `L2StandardBridge` to mint
+and burn tokens, depending on whether the user is
+depositing from L1 to L2 or withdrawing from L2 to L1.
+
+### Creation method
+
+The `OptimismMintableERC20Factory` is updated to use `CREATE3` instead of `CREATE2`.
+This will remove the compiler's dependency on address computation.
+
+The salt SHOULD depend on the inputs:
```
tynes: A quick sentence for rationale here would be good, like these are used to ensure that there is a single unique token address per set of ERC20 metadata

tynes (approved):

tynes:
Looks good to me, one could say that the `SHOULD`s could be changed to `MUST` since its not optional, see https://datatracker.ietf.org/doc/html/rfc2119

0xParti:
@tynes Do you think its worth mentioning something about the external functions and invariants here? I also haven't mentioned anything about the `deployments` one time update with the token list, which will be part of the first update.

tynes:
> external functions and invariants here

I do think that would be useful here and keep a high bar for the specs

tynes:
I don't think we want to go deep into the one time update for the tokenlist here, that is something we should do a design doc for and not add to the specs

0xParti:
> > external functions and invariants here
> 
> I do think that would be useful here and keep a high bar for the specs

Added both. Let me know if you agree.

agusduha:
We are adding a new contract inheriting form `OptimismMintableERC20Factory` to make these updates.

Should we create a new function `createWithCreate3` to be called from `createOptimismMintableERC20WithDecimals` so that we can add it to the new contract? Or do u prefer to directly change the code in this function?

0xParti:
> We are adding a new contract inheriting form `OptimismMintableERC20Factory` to make these updates.
> 
> Should we create a new function `createWithCreate3` to be called from `createOptimismMintableERC20WithDecimals` so that we can add it to the new contract? Or do u prefer to directly change the code in this function?

@tynes this would affect the specs. We could simply modify the existing `createOptimismMintableERC20WithDecimals` or adda new function that `createOptimismMintableERC20WithDecimals` will call (similar to how previous upgrades where extended). wdyt?

0xParticle:
I noticed you pushed some changes to main within the same file. I will create a new branch with those changes included.

tynes:
I thought we decided to use `CREATEX` and also not update the `OptimismMintableERC20Factory` to use it?

specs/interop/predeploys.md
```diff
@@ -420,6 +432,187 @@ dependency set called `dependencySet()`. This function MUST return the array of
...
+The `OptimismMintableERC20Factory` is updated to
+
+- use `CREATE3` instead of `CREATE2` to remove the compiler's dependency on address computation.
+- include a `deployments` mapping
+  to store the `remoteToken` address for each deployed `OptimismMintableERC20`.
+  - This is essential for the liquidity migration process defined in the liquidity migration spec.
+
+### Functions
+
+#### `createWithCreate3`
```
tynes: This leaks an implementation detail into the name, not a fan of this style of naming

tynes (approved):

specs/interop/predeploys.md
```diff
@@ -420,6 +430,151 @@ dependency set called `dependencySet()`. This function MUST return the array of
...
+It MUST use `CREATE2` to deploy new contracts. The salt MUST depend on the following inputs:
+`_remoteToken`, `_name`, `_symbol`, and `_decimals`.
+This will ensure a unique `OptimismMintableERC20` for each set of ERC20 metadata.
+
+It MUST store the `_remoteToken` address for each deployed `OptimismMintableERC20` in a `deployments` mapping.
+
+A reference implementation looks like the following:
+
+```solidity
+function createOptimismMintableERC20WithDecimals(
```
tynes: I think we could get away without implementing the whole function here and instead precisely define the salt in natural language and the invariant that the mapping must be updated

0xParti: perfect. Will also remove the reference implementation for the others

tynes (approved):

PR #322, opened 2024-08-08
feat: add Cannon compiler assumptions

smartcontracts:
Adds a new section to the Cannon page that describes some security model things related to the assumptions that Cannon makes about compilers that users are using.

### DIFF
--- a/specs/fault-proof/cannon-fault-proof-vm.md
+++ b/specs/fault-proof/cannon-fault-proof-vm.md
@@ -17,6 +17,9 @@
   - [Pre-image Communication](#pre-image-communication)
     - [Pre-image I/O Alignment](#pre-image-io-alignment)
 - [Exceptions](#exceptions)
+- [Security Model](#security-model)
+  - [Compiler Correctness](#compiler-correctness)
+  - [Compiler Assumptions](#compiler-assumptions)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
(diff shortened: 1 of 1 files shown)


specs/fault-proof/cannon-fault-proof-vm.md
```diff
@@ -218,3 +221,49 @@ transition. Nominally, the FPVM must raise an exception in at least the followin
...
+Many compilers, including the Golang compiler, will not generate code that would trigger these conditions under bug-free
+operation. Given the inherent reliance on [Compiler Correctness](#compiler-correctness) in applications using Cannon, the
+tests and defense mechanisms that must necessarily be employed by Cannon users to protect their particular programs
+against compiler bugs should also suffice to surface bugs that would break these compiler assumptions. Stated simply, Cannon
+can rely on specific compiler behaviors because users inherently must employ safety nets to guard against compiler bugs.
+
+Cannon users should be careful to utilize compilers that satisfy these assumptions. The following compilers satisfy these
+assumptions:
+
+- Standard Golang MIPS compiler
```
ajsutton: I don't think we can state this without also specifying the exact versions involved given that compatibility could be broken in later versions.  I would suggest removing the list entirely - there should be tests/static analysis in place to verify the compiled code meets cannons expectations.

specs/fault-proof/cannon-fault-proof-vm.md
```diff
@@ -218,3 +221,49 @@ transition. Nominally, the FPVM must raise an exception in at least the followin
...
+Cannon is designed to prove the correctness of a particular state transition that emulates a MIPS32 machine.
+Cannon does not guarantee that the MIPS32 instructions correctly implement the program that the user intends to prove.
+As a result, Cannon's use as a Fault Proof system inherently depends on some extent on the correctness of the compiler
+used to generate the MIPS32 instructions over which Cannon operates.
+
+To illustrate this concept, suppose that a user intends to prove simple program `input + 1 = output`.
+Suppose then that the user's compiler for this program contains a bug and errantly generates the MIPS instructions for a
+slightly different program `input + 2 = output`. Although Cannon would correctly prove the operation of this compiled program,
+the result proven would differ from the user's intent. Cannon proves the MIPS state transition but makes no assertion about
+the correctness of the translation between the user's high-level code and the resulting MIPS program.
```
ajsutton, line 238: It feels a bit weird to me to call out bugs in compilers specifically for cannon when the exact same risk exists for native op-node and op-geth but isn't called out in the spec.

specs/fault-proof/cannon-fault-proof-vm.md
smartcontracts: Fair enough, will remove

specs/fault-proof/cannon-fault-proof-vm.md
smartcontracts: Realistically I would prefer that we call out every risk as explicitly as possible, whenever possible. I do think the risk is somewhat unique here as people may not already have built up a mental model that makes the compiler risk obvious. I wouldn't be surprised if the compiler risk is non-obvious to a lot of people.

clabby: For this specific case, I think it's worth while. We don't call out compiler bugs in our standard native software, yes, but the MIPS target is the least-actively-maintained target in the Go compiler I'd bet. The risk is substantially higher than average native Go code.

Inphi (approved):

PR #324, opened 2024-08-12
add interop start timestamp explainer to spec docs

AmadiMichael:
**Description**

This pr adds the concept of Interop Start Timestamp to the Specs.

**Metadata**

- Fixes a part of [this issue](https://github.com/ethereum-optimism/optimism/issues/11367)


specs/interop/predeploys.md
```diff
@@ -117,6 +129,7 @@ A simple implementation of the `executeMessage` function is included below.
 function executeMessage(Identifier calldata _id, address _target, bytes calldata _msg) public payable {
     require(_id.timestamp <= block.timestamp);
     require(L1Block.isInDependencySet(_id.chainid));
+    require(_id.timestamp > L1Block.interopStart());
```
tynes: We could choose to put this value directly in the CrossL2Inbox to save a call, what do you think?

tynes:
This doesn't match the implemention exactly in the pseudocode

https://github.com/ethereum-optimism/optimism/pull/11398

We should migrate more pseudocode to natural language when it makes sense to but no need to do that now for this

I think the implementation it the right way to do it

specs/interop/predeploys.md
AmadiMichael: Yeah, I think its worth the gas savings

AmadiMichael: Interestingly, this already exists in Crossl2Inbox and not L1Block, it's my mistake in the reference implementation

tynes (approved):

PR #325, opened 2024-08-12
feat: add cspell spellchecking

smartcontracts:
Adds the cspell spellchecker to the linting process. Fixes a bunch of spelling errors that cspell caught.

New words can be automatically added to the wordlist by running `just lint-specs-spelling-fix`. This will add all unknown words to the wordlist. Since this means typos can be added by accident, this command is not added to the main `just lint-specs-fix` command to prevent people from abusing it and creating more burden on reviewers.

NOTE: This PR also bumps pnpm to version 9 since that was the version used in the monorepo prior to its removal and is the version we're using everywhere else.


words.txt
```diff
@@ -0,0 +1,192 @@
...
+Depositer
+devnet
+dialable
+Discv
+discv
+disincentivized
+DOCTOC
+doctoc
+EBADF
+Ectone
```
mbaxter: Looks like this is a typo (should be Ecotone)

words.txt
```diff
@@ -0,0 +1,192 @@
...
+fastlz
+fcntl
+Filecoin
+FINDNODES
+flippin
+Forkchoice
+forkchoice
+FPAC
+FPVM
+froms
```
mbaxter: Looks like a typo (from)

words.txt
```diff
@@ -0,0 +1,192 @@
...
+frontran
```
mbaxter: Should be frontrun?

words.txt
```diff
@@ -0,0 +1,192 @@
+addi
```
mds1, line 1: What's `addi`?

words.txt
```diff
@@ -0,0 +1,192 @@
...
+GETFL
+gindex
+griefed
+Hardfork
+hardfork
+hardforks
+hardfork’s
+idxs
+IERC
+Impls
```
mbaxter, line 76: Looks like this is being pulled from [a code block](https://github.com/ethereum-optimism/specs/blob/992df6bf62f4ab1f6181bd07cae2b55e3ba0b577/specs/fault-proof/stage-one/dispute-game-interface.md?plain=1#L166)?  Does the `ignoreRegExpList` regex need to be updated?

words.txt
```diff
@@ -0,0 +1,192 @@
...
+newrank
+nodehash
+offchain
+opstack
+outputroot
+parseable
+Pausability
+pausability
+peerstore
+permisionlessly
```
mbaxter: Looks like a typo

words.txt
smartcontracts: `addi` is a MIPS instruction

Justfile
```diff
@@ -37,6 +37,14 @@ lint-links:
     		--exclude twitter.com --exclude explorer.optimism.io --exclude linux-mips.org --exclude vitalik.eth.limo \
     		--exclude-mail /input/README.md "/input/specs/**/*.md"
 
+# Validates spelling using cspell
+lint-specs-spelling:
+    npx cspell "./**/*.md"
+
+# Updates cspell words file with new words
+lint-specs-spelling-fix:
+    npx cspell --words-only --unique "./**/*.md" | sort --ignore-case | uniq > words.txt
```
mbaxter, line 46: Should we lowercase the entries so we can avoid all the casing variations in words.txt?

mbaxter (commented):
Looks good!  Just a few questions on some of the words.txt entries.  Also looks like we may be spell-checking some code blocks, but looks like we're trying to avoid that with the `ignoreRegExpList` config.

Justfile
smartcontracts: cspell is case sensitive which probably makes sense for some proper nouns. I think it's likely best to leave it as-is.

smartcontracts:
@mbaxter I decided to nix the code that ignores stuff inside code blocks. Spell checker found enough interesting things inside code blocks and it's easy enough to add words to the dictionary if needed.

mbaxter (approved):

.circleci/config.yml
```diff
@@ -27,6 +27,9 @@ jobs:
       - image: <<pipeline.parameters.ci_builder_image>>
     steps:
       - checkout
+      - run:
+          name: Install pnpm
```
tynes, line 31: This isn't ideal, we don't have a docker image that is great for this, ci-builder is too heavy, but this can work for now

tynes (approved):

tynes:
Ideally mdbook has spell check baked in to reduce the number of tools that we rely on but looks like no support given this open issue: https://github.com/rust-lang/mdBook/issues/141

PR #328, opened 2024-08-12
feat: add protocol diagrams to overview

smartcontracts:
Ports diagrams from contracts-bedrock to specs repository so they can be maintained in one place.


smartcontracts:
I had to update Mermaid to the latest version to be able to support the syntax we used

tynes (approved):

PR #331, opened 2024-08-13
maint: clean up the introductory docs

smartcontracts:
Introductory docs were quite outdated and were using old diagrams. Cleans up these two docs heavily and reworks the diagrams into Mermaid so they can be maintained more easily going forward.


tynes (approved):
Nice cleanup

PR #327, opened 2024-08-12
Rename to Safe Extensions

maurelian:
Mostly just some minor clean up.

- Rename to Safe Extensions
- Add new stage one document
- Clean up headings and document outline

PR #326, opened 2024-08-12
Rewrap safe document

maurelian:
Just cleans up white space in the Security Council safe doc. This is intended to keep the diff small with further changes incoming.

PR #333, opened 2024-08-14
Add new Stage 1 document

maurelian:
- Renames to `stage-1.md` to be more consistent with usage elsewhere.
- Adds 'overview' content to the document, which sets the tone for what will be included (how to configure an OP Stack for Stage 1) and what is left out (the generic requirements for Stage 1).

### DIFF
--- a/specs/protocol/stage-1.md
+++ b/specs/protocol/stage-1.md
@@ -2,5 +2,42 @@
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Overview](#overview)
+- [Permissionless Fault Proofs](#permissionless-fault-proofs)
+- [Roles for Stage 1](#roles-for-stage-1)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+This document outlines configuration requirements (including roles and other parameters)
+for an implementation of the OP Stack to satsify Stage One decentralization requirements as defined
+by
+L2 Beat [[1](https://medium.com/l2beat/introducing-stages-a-framework-to-evaluate-rollups-maturity-d290bb22befe), [2](https://medium.com/l2beat/stages-update-security-council-requirements-4c79cea8ef52)].
+
+## Permissionless Fault Proofs
+
+Stage 1 requires a chain to be operating with Permissionless Fault Proofs.
+
+## Roles for Stage 1
+
+Within the context of an OP Stack, the following roles are required for Stage 1:
+
+1. **Upgrade Controller:** ALthough named for its ability to perform an upgarde, more generally this
+   account is in charge of _safety_, meaning it MUST control any action which has an impact
+   on the determination of a valid L2 state, or the custody of bridged assets.
+
+   This includes upgrading L1 contracts, and modifying the implementation of the dispute game, and
+   any other safety-critical function.
+
+2. **Guardian:** This account is in charge of _liveness_, meaning it MUST control any action which
+   may cause a delay in the finalization of L2 states and the resulting settlement on L1.
+
+   This includes but is not limited to pausing code paths related to withdrawals.
+
+There may be additional [roles](./configurability.md#admin-roles) in the system, however they MUST
+not be able to perform any actions which have an impact on either the safety or liveness of the
+system. Any example of such a role is the `SystemConfig` owner (AKA Chain Operator), which can
+modify fees and other protocol parameters.
--- a/specs/protocol/stage-1.md
+++ b/specs/protocol/stage-1.md
@@ -42,4 +42,3 @@ not be able to perform any actions which have an impact on either the safety or 
 system. An example of such a role is the `SystemConfig` owner (AKA Chain Operator), which can
 modify fees and other protocol parameters. Constraints MUST be placed on the allowable range of
 values for these parameters so that they do not result in a loss of liveness or safety.
-
--- a/specs/protocol/stage-1.md
+++ b/specs/protocol/stage-1.md
@@ -25,20 +25,18 @@ Stage 1 requires a chain to be operating with Permissionless Fault Proofs.
 
 Within the context of an OP Stack, the following roles are required for Stage 1:
 
-1. **Upgrade Controller:** Although named for its ability to perform an upgarde, more generally this
-   account is in charge of _safety_, meaning it MUST control any action which has an impact
-   on the determination of a valid L2 state, or the custody of bridged assets.
+1. **Upgrade Controller:** Although named for its ability to perform an upgrade, more generally this
+   account MUST control any action which has an impact on the determination of a valid L2 state,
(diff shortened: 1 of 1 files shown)


maurelian:
* **#334** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/specs/334?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#333** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/specs/333?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#327** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/specs/327?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#326** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/specs/326?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @maurelian and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

specs/protocol/stage-1.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Permissionless Fault Proofs
+
+Stage 1 requires a chain to be operating with Permissionless Fault Proofs.
+
+## Roles for Stage 1
+
+Within the context of an OP Stack, the following roles are required for Stage 1:
+
+1. **Upgrade Controller:** ALthough named for its ability to perform an upgarde, more generally this
```
mds1:
```suggestion
1. **Upgrade Controller:** Although named for its ability to perform an upgarde, more generally this
```

mds1: Is "Upgrade Controller" a new term we want to standardize around? If so, we will need to update this in a lot of places for consistency since it's "L1 ProxyAdmin Owner" everywhere (standard config, docs, superchain registry, code, etc.)

Or is this intended to be an "abstract" role that exists in the context of L2 Beat requirements even though we don't have a role with this name in the system?

specs/protocol/stage-1.md
```diff
@@ -0,0 +1,44 @@
...
+   account is in charge of _safety_, meaning it MUST control any action which has an impact
+   on the determination of a valid L2 state, or the custody of bridged assets.
+
+   This includes upgrading L1 contracts, and modifying the implementation of the dispute game, and
+   any other safety-critical function.
```
mds1: nit
```suggestion
   This includes upgrading L1 contracts, modifying the implementation of the dispute game, and
   any other safety-critical functions.
```

specs/protocol/stage-1.md
```diff
@@ -0,0 +1,44 @@
...
+
+2. **Guardian:** This account is in charge of _liveness_, meaning it MUST control any action which
+   may cause a delay in the finalization of L2 states and the resulting settlement on L1.
```
mds1: Hmm, we may need to loosen the clear safety vs. liveness phrasing that defines these two roles—the guardian can also change the respected game type, which is more of a safety role than a liveness one. Quoting from the contract:
```solidity
/// @notice Sets the respected game type. Changing this value can alter the security properties of the system,
///         depending on the new game's behavior.
/// @param _gameType The game type to consult for output proposals.
function setRespectedGameType(GameType _gameType) external {
```

specs/protocol/stage-1.md
```diff
@@ -0,0 +1,44 @@
...
+
+   This includes but is not limited to pausing code paths related to withdrawals.
+
+There may be additional [roles](./configurability.md#admin-roles) in the system, however they MUST
+not be able to perform any actions which have an impact on either the safety or liveness of the
+system. An example of such a role is the `SystemConfig` owner (AKA Chain Operator), which can
```
mds1: The system config owner can set the gas fees extremely high, which then affects liveness because no one can afford transactions? Perhaps the right framing is to say these additional roles don't impact conformance to the L2 Beat stage 1 requirements, as opposed to safety or liveness

specs/protocol/stage-1.md
maurelian: For now I would like to just stealthily introduce this term into the lexicon without attempting to replace the current L1 Proxy Admin owner, though I do think we should try to do that over time (I'll add that to the roles design review I have in draft).

As you'll see in the next PR, we still refer to the 2 of 2 safe with the existing language.

specs/protocol/stage-1.md
maurelian: IMO the phrasing here is right, and if the implementation does't do this then we need to fix it. :)

Fortunately, I did look into this earlier, and determined that although the Guardian can change the game type by calling `setRespectedGameType()`, it is only the Upgrade Controller which can determine the available options by calling `DGF.setImplementation()`. https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/dispute/DisputeGameFactory.sol#L199-L201 

I don't think that the natspec is wrong, but it's somewhat incomplete.

mds1: I guess it's a bit of semantics depending on how you think of it. The UC determines the available options, but the guardian changing from the PDG to FDG does change safety properties. It's just that both were approved by the UC

maurelian: After some offline discussion, I think that there is some room for this semantic discussion, but that the current phrasing reflects our assessment of the stage 1 requirements. 

Are you OK to let it stand as such for now?

maurelian: Actually, I do have an idea for a small rephrasing, standby.

specs/protocol/stage-1.md
maurelian: You're right. However I've realized that he issue is actually not about liveness, but about allowing withdrawals. I will update this text to clarify that.

On that basis, we can double check the requirements. So, in order for withdrawals to continue, we need:
1. force inclusion (deposit tx) to work. This can be blocked by either:
  - resource config params being outside of a reasonable range, which is why all [these checks exist](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L1/SystemConfig.sol#L446-L461).
  - gas limit params being outside of a reasonable range (prevented by [these checks](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L1/SystemConfig.sol#L409-L410)).
2. permissionless output proving
3. withdrawals being unpaused

Since 1 is enforced in the contract, and 2 and 3 are guardian actions that's all good. 

As you note, there is also the [`setGasConfig`](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L1/SystemConfig.sol#L363-L398) logic which the chain operator has access to, however these values do not impact on the deposit path, only sequenced tx, so they cannot censor withdrawals.

maurelian: This document is now written with more precise language, so that instead of 'safety/liveness', it focuses on the state transition validity vs. withdrawals.

Accordingly, I think we can resolve this issue.

specs/protocol/stage-1.md
maurelian: You make a good point, however I think it's best to keep this document as is for now.

mds1 (approved):

PR #339, opened 2024-08-15
Fix definition of predeploy namespace

geoknee:
Drive-by question:

Has it always been the case that predeploys live in a **1 byte** namespace? 

I think this is a counterexample https://optimistic.etherscan.io/address/0x4200000000000000000000000000000000000800

### DIFF
--- a/specs/protocol/predeploys.md
+++ b/specs/protocol/predeploys.md
@@ -39,7 +39,7 @@ network forking.
 
 Predeploy addresses exist in 1 byte namespace `0x42000000000000000000000000000000000000xx`.
 Proxies are set at each possible predeploy address except for the
-`GovernanceToken` and the `ProxyAdmin`.
+`GovernanceToken`.
 
 The `LegacyERC20ETH` predeploy lives at a special address `0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000`
 and there is no proxy deployed at that account.
--- a/specs/protocol/predeploys.md
+++ b/specs/protocol/predeploys.md
@@ -39,7 +39,7 @@ network forking.
 
 Predeploy addresses exist in 1 byte namespace `0x42000000000000000000000000000000000000xx`.
 Proxies are set at each possible predeploy address except for the
-`GovernanceToken`.
+`GovernanceToken` and `WETH`.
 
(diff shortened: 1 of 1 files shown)


tynes:
This is definitely a bug, there are a total of 2048 predeploys, that is bigger than a 1 byte namespace

specs/protocol/predeploys.md
```diff
@@ -39,7 +39,7 @@ network forking.
 
 Predeploy addresses exist in 1 byte namespace `0x42000000000000000000000000000000000000xx`.
 Proxies are set at each possible predeploy address except for the
-`GovernanceToken` and the `ProxyAdmin`.
+`GovernanceToken`.
```
tynes: I believe this should be updated, i believe weth has no proxy

geoknee: Yes, double checked here 
https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/libraries/Predeploys.sol#L141-L145

and updated here fce5a999ca87c33d8606a3721c5ae6a96827f8cd.

geoknee:
> This is definitely a bug, there are a total of 2048 predeploys, that is bigger than a 1 byte namespace

Updated here [51b1497](https://github.com/ethereum-optimism/specs/pull/339/commits/51b14977ede24ae05360500f6831e847a5233ef0)

specs/protocol/predeploys.md
```diff
@@ -37,9 +37,9 @@ Predeploys are used instead of precompiles to make it easier for multiclient
 implementations as well as allowing for more integration with hardhat/foundry
 network forking.
 
-Predeploy addresses exist in 1 byte namespace `0x42000000000000000000000000000000000000xx`.
-Proxies are set at each possible predeploy address except for the
-`GovernanceToken` and the `ProxyAdmin`.
+Predeploy addresses exist in a prefixed namespace `0x420xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx`.
```
tynes: Don't you think that we should only use `x` in the least significant 3 nibbles to represent the fact that they are the only things that can change?

geoknee: [4f2b9a3](https://github.com/ethereum-optimism/specs/pull/339/commits/4f2b9a332350d79b4f25e643fc247a0078439bd3)

tynes (approved):

PR #342, opened 2024-08-16
governance: update migration logic in `GovernanceDelegation`

cairoeth:
**Description & Context**

For interop, the governor will have to be updated use the L1 block number instead of the OP L2 block number as networks in their superchain will have their own cadence of blocks.

Proposed by @smartcontracts, the interop update introduces breaking changes so it's better to have the delegation state separated: the legacy state in the `GovernanceToken` (with L2 block number), and the new state in the `GovernanceDelegation` contract. This means that the `GovernanceToken` doens't call the `GovernanceDelegation` in its getter functions, and vice versa.

### DIFF
--- a/specs/experimental/gov-delegation.md
+++ b/specs/experimental/gov-delegation.md
@@ -12,6 +12,7 @@
     - [`delegateFromToken`](#delegatefromtoken)
     - [`delegateBatched`](#delegatebatched)
     - [`afterTokenTransfer`](#aftertokentransfer)
+    - [`migrateAccounts`](#migrateaccounts)
     - [`_delegate`](#_delegate)
   - [Getters](#getters)
     - [`checkpoints`](#checkpoints)
@@ -20,6 +21,7 @@
     - [`getVotes`](#getvotes)
     - [`getPastVotes`](#getpastvotes)
     - [`getPastTotalSupply`](#getpasttotalsupply)
+    - [`migrated`](#migrated)
     - [`delegations`](#delegations)
   - [Events](#events)
     - [`DelegationCreated`](#delegationcreated)
@@ -28,12 +30,15 @@
 - [Types](#types)
   - [`Delegation`](#delegation)
   - [`AllowanceType`](#allowancetype)
+- [Migration](#migration)
+- [Backwards Compatibility](#backwards-compatibility)
 - [User Flow](#user-flow)
   - [Partial Delegations](#partial-delegations)
   - [Absolute & Relative Delegations](#absolute--relative-delegations)
   - [Differences](#differences)
 - [Delegation Validation](#delegation-validation)
 - [Security Considerations](#security-considerations)
+  - [Dependence on GovernanceDelegation](#dependence-on-governancedelegation)
   - [Connection with GovernanceToken](#connection-with-governancetoken)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
@@ -71,7 +76,9 @@ any existing delegations of the delegator.
 function delegate(Delegation _delegation) external
 ```
 
-The function MUST call the [`_delegate`](#_delegate) function with the `msg.sender` and `_newDelegations` parameters.
+This function MUST enforce the migration logic, as specified in the [Migration](#migration) section, for `msg.sender`
+and the `_delegatee` address in the delegation. Afterwards, the function MUST call the [`_delegate`](#_delegate) function
+with the `msg.sender` and `_newDelegations` parameters.
 
 At the end, the `delegate` function MUST emit a `DelegationCreated` event with the given function parameters.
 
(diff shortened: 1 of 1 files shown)


tynes:
I'd like a review from @smartcontracts given he has the most context here

specs/experimental/gov-delegation.md
```diff
@@ -309,17 +305,20 @@ enum AllowanceType {
...
-`GovernanceToken`.
-
 ## Migration
 
 All write functions in the `GovernanceDelegation` MUST check if the users interacting with it have been migrated by
 checking the `migrated` mapping from its [storage](#storage). If a user has not been migrated, the `GovernanceDelegation`
 MUST copy the delegation and checkpoint data from the token contract to its own state. After copying the data, the
-`GovernanceDelegation` MUST update the `migrated` mapping to reflect that the address has been migrated.
+`GovernanceDelegation` MUST update the `migrated` mapping to reflect that the address has been migrated, and remove the
+state from the `GovernanceToken` contract via a function that can only be called by the `GovernanceDelegation` contract.
```
smartcontracts: Function can be callable by any address. Basically just have a function like:

```solidity
function migrate(address _addr) public {
    address tmp = delegates[_addr];
    if (tmp != address(0)) {
        _delegate(_addr, address(0);
        governanceDelegation.delegateFromToken(address, tmp);
    }
}
```

specs/experimental/gov-delegation.md
```diff
@@ -309,17 +305,20 @@ enum AllowanceType {
...
-The `GovernanceDelegation` contract ensures backwards compatibility by allowing the migration of delegation state from the
-`GovernanceToken`.
-
 ## Migration
 
 All write functions in the `GovernanceDelegation` MUST check if the users interacting with it have been migrated by
 checking the `migrated` mapping from its [storage](#storage). If a user has not been migrated, the `GovernanceDelegation`
 MUST copy the delegation and checkpoint data from the token contract to its own state. After copying the data, the
-`GovernanceDelegation` MUST update the `migrated` mapping to reflect that the address has been migrated.
+`GovernanceDelegation` MUST update the `migrated` mapping to reflect that the address has been migrated, and remove the
```
smartcontracts: I actually don't think we need the `migrated` mapping anymore? Since we're not forwarding stuff to the original contract.

specs/experimental/gov-delegation.md
smartcontracts: Must call this function whenever the user tries to create a delegation.

smartcontracts: Please also include an invariant in this spec that says that the user MUST NOT have an active delegation in both contracts at the same time. It's one or the other. We can say that a delegation to `address(0)` is an inactive delegation.

specs/experimental/gov-delegation.md
cairoeth: yeah the `GovernanceDelegation` contract doesn't really need it, but it could be a helper for external contracts like governor: if `migrated` is true, get data from `GovernanceDelegation`, otherwise from `GovernanceToken`. 

if these contracts check `delegates[_addr]` only in `GovernanceToken`, they won't be able to know if there's a delegation in `GovernanceDelegation`, so they will always have to call both. wdyt?

specs/experimental/gov-delegation.md
cairoeth: agreed. added in https://github.com/ethereum-optimism/specs/pull/342/commits/fd3c52df61b48f751f3ba72636162cf27f501ec8

specs/experimental/gov-delegation.md
smartcontracts: Ah since we maintain the invariant that you can't have a delegation in both contracts then what you can do is just check if `delegates[_addr]` inside `GovernanceToken` is `address(0)`

specs/experimental/gov-delegation.md
```diff
@@ -309,17 +305,28 @@ enum AllowanceType {
 | `Absolute`  | `0`    | The amount of votes delegated is fixed and denominated in the `GovernanceToken`'s decimals  |
 | `Relative`  | `1`    | The amount of votes delegated is relative and denominated in percentages.                   |
 
-## Backwards Compatibility
+## Migration
 
-The `GovernanceDelegation` contract ensures backwards compatibility by allowing the migration of delegation state from the
-`GovernanceToken`.
+All write functions in the `GovernanceDelegation` MUST check if the users interacting with it (such as creating or updating
```
smartcontracts: Technically we don't need this in *all* write functions, only in functions that would create/update a delegation

specs/experimental/gov-delegation.md
```diff
@@ -309,17 +305,28 @@ enum AllowanceType {
...
+a delegation) have been migrated by checking the `migrated` mapping from its [storage](#storage). If a user has not been
+migrated, the `GovernanceDelegation` MUST copy the delegation and checkpoint data from the token contract to its own state.
+After copying the data, the `GovernanceDelegation` MUST update the `migrated` mapping to reflect that the address has been
+migrated, and clear the delegation in the `GovernanceToken` contract.
 
-## Migration
+The `GovernanceDelegation` MUST enforce the following invariants for the migration logic:
+
+1. A user MUST NOT have an active delegation in both `GovernanceToken` and `GovernanceDelegation` at the same time.
```
smartcontracts, line 305: Technically some of these statements are implied by the other statements but I think it's good enough to just be explicit and list all of the statements

specs/experimental/gov-delegation.md
cairoeth: agreed, updated in https://github.com/ethereum-optimism/specs/pull/342/commits/eee936c9321706d7939f83103eb6246aa4b6e4f0

specs/experimental/gov-delegation.md
cairoeth: yeah i think it helps to have some redundancy in this specific case

smartcontracts (approved):

tynes (approved):

PR #354, opened 2024-08-26
Typo fix: replace `type-3` with `type-2` in reference to 1559 tx

NiloCK:
**Additional context**

See class Transaction1559Envelope in https://eips.ethereum.org/EIPS/eip-1559#specification

Possibly this is a 0-indexing vs 1-indexing confusion, but I don't think I've encountered anyone referring to legacy tx types as `type-1`. More likely this is a simple typo.


protolambda (approved):
Good catch, yes this was a 0-indexing mistake. It's legacy (X, sometimes considered 0), then access-list (1), then 1559 (2).

PR #357, opened 2024-08-27
Holocene: Add Holocene Derivation Spec

### LINKED ISSUE #11313: Holocene-D: Write Steady Batch Derivation Specs
Write specs, based on [design doc](https://github.com/ethereum-optimism/design-docs/pull/20).

### LINKED ISSUE #11513: Holocene-D: Write Partial Span Batch Validity specs

sebastianst:
**Description**

Holocene Derivation Specification

**Metadata**

- Fixes https://github.com/ethereum-optimism/optimism/issues/11513
- Fixes https://github.com/ethereum-optimism/optimism/issues/11313


specs/protocol/holocene/derivation.md
```diff
@@ -0,0 +1,186 @@
...
+## Steady Batch Derivation
+
+Steady Batch Derivation changes the derivation rules for invalid batches, and instead of giving them
+a second chance for a worst-case full sequencing window, they are immediately derived as
+deposit-only blocks. There are two stages at which this new rule takes effect, in the batch queue
+and in the engine queue.
+
+### Batch Queue
+
+With Holocene, whereever a singular batch would be dropped before, it is now immediately replaced
```
refcell: Pedantic

```suggestion
With Holocene, wherever a singular batch would be dropped before, it is now immediately replaced
```

specs/protocol/holocene/derivation.md
```diff
@@ -0,0 +1,186 @@
...
+by a deposit-only batch.
+
+TODO: details
+
+### Engine Queue
+
+Holocene changes the way in which the Engine Queue responds to payload attributes errors: On a
+payload processing error, the payload is immediately replaced by the respecive empty deposit-only
```
tynes: What sorts of processing errors can result in this?

Possible things?
- invalid block (tx with wrong nonce, tx with not enough balance for fees)
- invalid parent hash
- error responses here: https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#response-1

specs/protocol/holocene/derivation.md
```diff
@@ -0,0 +1,186 @@
...
+validity was treated atomically, so earlier pending valid singular batches could be invalidated by a
+later invalid singular batch pulled from a span batch.
+To accomplish this new behavior, the [Delta span batch batch queue rules](../delta/span-batches.md#batch-queue)
+are changed in the following ways.
+
+The max sequencer time drift check is applied to each singluar batch derived from a span batch
+individually. Once a singular batch is found to be invalid, only the remaining span batch is
+dropped. Note that the sequencing window check still applies to the span batch in full, because the
+oldest blocks would be the first to be invalidated and if dropped would also invalidate newer
+blocks.
```
refcell: Does this mean that a singular batch that was marked valid or turned into a deposit-only batch could still be invalidated due to the sequencing window check applying to the span batch in full? I don't really understand the "span batch in full" wording here.

specs/protocol/holocene/derivation.md
```diff
@@ -0,0 +1,186 @@
...
+
+TODO: other changes to validation checks.
+
+## Strict Batch Ordering
+
+The new Strict Batch Ordering rule requires batches within and across channels to be strictly
+ordered. Note that this is already guaranteed for singular batcher derived from span batches by design.
```
refcell: pedantic nit
```suggestion
ordered. Note that this is already guaranteed for singular batches derived from span batches by design.
```

specs/protocol/holocene/derivation.md
```diff
@@ -0,0 +1,186 @@
...
+### Frame Queue
+
+The frame queue retains its function and holds all frames of the last batcher transaction. Holocene
+still allows multiple frames per batcher transaction. As before, this allows for optionally filling
+up remaining space of a batcher transactions with a starting frame of the next channel.
+
+### Channel Bank
+
+The Channel Bank becomes much simpler and only builds up a single channel.
+Channel frames have to arrive in order, and only one channel can be open at a time.
```
refcell: Is it worth noting here that the current implementation shouldn't _need to_ change and supporting multiple channels is required for backwards-compatibility?

specs/protocol/holocene/derivation.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### Timeout
+
+The timeout is applied as before, just only to the single staging channel.
+
+#### Reading & Frame Loading
+
+Since the channel bank now only holds a single staging channel, and frames have to arrive in
+order, frames are read until the channel is complete, and then the complete channel can be pulled
+form the channel bank.
```
refcell: nit
```suggestion
from the channel bank.
```

specs/protocol/holocene/derivation.md
```diff
@@ -0,0 +1,186 @@
...
+- A first frame for a new channel starts a new channel as the staging channel.
+  - If there already is an open staging channel, it is dropped and replaced by this new channel.
+- An out of order frame is immediately dropped.
+  - If it belongs to the currently open channel, that channel is also dropped as the staging channel.
+  - This rule also covers dropping duplicate frames, as before Holocene.
+
+Otherwise, the pre-Holocene rules are applied:
+- If the current channel is timed-out, but not yet pruned, and the incoming frame would be the next
+  correct frame for this channel, it is dropped.
+- If a channel got closed by a _last_ frame, new frames for that channel are dropped.
```
refcell: nit
```suggestion
- If a channel was closed by a _last_ frame, new frames for that channel are dropped.
```

specs/protocol/holocene/derivation.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Batch Queue
+
+The batch queue is also simplified in that batches are required to arrive strictly ordered, and gaps
```
refcell: What is considered a gap?

refcell (commented):
Reads very well

specs/protocol/holocene/derivation.md
sebastianst: Spec got reworked. New rules drop invalid batches, only replacing invalid payloads with deposit-only blocks.

specs/protocol/holocene/derivation.md
sebastianst: The pre-Holocene derivation still has to be supported yes! I suggest consensus clients actually implement alternative stages pre and post holocene and then switch over at the L1 holocene activation block, including clearing the derivation pipeline state. This is also described in the new activation section.

So the legacy channel bank that supports buffering of multiple channels should be retained. However, after the FP challenge window expired, we could switch FP program implementations to a simpler DP implementation that only supports Holocene derivation, since FPs don't need to support historical derivation after a fork is active for long enough.

specs/protocol/holocene/derivation.md
sebastianst: Specs got rewritten, should be clearer now.

specs/protocol/holocene/derivation.md
sebastianst: We do not list the types of errors that can cause this in the original spec either. There, we just say
> - On payload processing errors the attributes must be dropped, and the forkchoice state must be left unchanged.
>   - Eventually the derivation pipeline will produce alternative payload attributes, with or without batches.
>   - If the payload attributes only contained deposits, then it is a critical derivation error if these are invalid.

Do you think we should start listing the types of failures that can cause this?

specs/protocol/holocene/derivation.md
```diff
@@ -1,13 +1,332 @@
...
+Note that when the L1 origin of the batch queue moves forward, it is guaranteed that it is empty,
+because future batches aren't buffered any more.
+
+### Fast Channel Invalidation
+
+Furthermore, upon finding an invalid batch, the remaining channel it got derived from is also discarded.
+
+_TODO: I believe that the batch queue, similarly to the channel bank, now actually only holds at most
+one single staging batch, because we eagerly derive payloads from any valid singular batch. And the
+span batch stage before it would similarly only hold at most one staging span batch._
```
sebastianst: We can update this after getting more familiar with the implementation of Holocene derivation

specs/protocol/holocene/derivation.md
```diff
@@ -1,13 +1,332 @@
...
+even just the following L2 block with the current safe head as parent?), we can
+safely continue derivation from this channel because no previous derivation pipeline state could
+have influenced the L2 safe head.
+- If the first frame we find is a non-first frame, then we need to go back a full channel
+timeout window to see if we find the start of that channel.
+  - If we find the starting frame, we can continue derivation from it.
+  - If we don't find the starting frame, we need to go back a full sequencing window. (TODO: verify)
+
+_TODO: above needs to be verified, we don't introspect batcher transactions and frame metadata
+during sync start yet._
```
sebastianst: I touched on the TODOs in this section in discussions with @protolambda and @clabby, curious what you make of it.

specs/protocol/holocene/derivation.md
```diff
@@ -1,13 +1,332 @@
...
+The timeout is applied as before, just only to the single staging channel.
+
+### Reading & Frame Loading
+
+The frame queue is guaranteed to hold ordered and contiguous frames, per channel. So reading and
+frame loading becomes simpler in the channel bank:
+
+- A first frame for a new channel starts a new channel as the staging channel.
+  - If there already is an open, non-completed staging channel, it is dropped and replaced by this
+  new channel. This is consistent with how the frame queue drops all frames of an non-closed channel
```
ajsutton:
```suggestion
  new channel. This is consistent with how the frame queue drops all frames of a non-closed channel
```

specs/protocol/holocene/derivation.md
ajsutton: Seems like an excellent invariant to assert in tests.

specs/protocol/holocene/derivation.md
```diff
@@ -1,13 +1,332 @@
...
+
+When the L1 traversal stage of the derivation pipeline moves its origin to the L1 block whose
+timestamp matches the Holocene activation timestamp, the derivation pipeline's state is mostly reset
+by **discarding**
+
+- all frames in the frame queue,
+- channels in the channel bank, and
+- all batches in the batch queue.
+
+The three stages are then replaces by the new Holocene frame queue, channel bank and batch queue
```
ajsutton:
```suggestion
The three stages are then replaced by the new Holocene frame queue, channel bank and batch queue
```

specs/protocol/holocene/derivation.md
```diff
@@ -1,13 +1,332 @@
...
+start deriving from this L1 origin.
+
+- If the first frame we find is a _first frame_ for a channel that includes the safe head (TODO: or
+even just the following L2 block with the current safe head as parent?), we can
+safely continue derivation from this channel because no previous derivation pipeline state could
+have influenced the L2 safe head.
+- If the first frame we find is a non-first frame, then we need to go back a full channel
+timeout window to see if we find the start of that channel.
+  - If we find the starting frame, we can continue derivation from it.
+  - If we don't find the starting frame, we need to go back a full sequencing window. (TODO: verify)
```
ajsutton: The batch for an L2 block is only valid to be included on L1 _after_ the L2 block's L1 origin. So if we don't find a starting frame, the channel we did find a frame from must be timed out and would be discarded.  The block we're looking for can't be in any channel that timed out before its L1 origin so we shouldn't need to search any further back.

Even pre-holocene we don't search back a full sequencing window if we hit the finalized L2 block first - worst case is one ChannelTimeout prior to the L1 origin of the L2 finalized head.

specs/protocol/holocene/derivation.md
ajsutton: I'm actually not sure this needs to be in the spec to be honest.  It's perfectly valid to always start a full ChannelTimeout earlier and the optimisation to optimistically start from the L1 origin and only go back if a mid-channel frame is found is an implementation detail.  Though our specs already blur spec and implementation a fair bit...

specs/protocol/holocene/derivation.md
```diff
@@ -1,13 +1,332 @@
...
+block derived from a valid batch will be determined by the engine stage before it pulls new payload
+attributes from the previous stage. This avoids larger derivation pipeline resets.
+
+## Less Defensive Protocol
+
+The stricter derivation rules lead to a less defensive protocol. The old protocol rules allowed for
+second chances for invalid payloads and submitting frames and batches within channels out of order.
+Experiences from running OP Stack chains for over one and a half years have shown that these relaxed
+derivation rules are (almost) never needed, so stricter rules that improve worst-case scenarios for
+Fault Proofs and Interop are favorable.
```
ajsutton, line 252: I'd also add that the more relaxed rules created a lot more corner cases and complex interactions, which made it harder to reason about and test the protocol, increasing the risk of chain splits between different implementations.

specs/protocol/holocene/derivation.md
```diff
@@ -1,13 +1,332 @@
...
+
+The only conceivable scenarios in which a _valid_ batch leads to an _invalid_ payload are
+
+- a buggy or malicious sequencer+batcher
+- in the future, that an previously valid Interop dependency referenced in that payload is later
+invalidated, while the block that contained the Interop dependency got already batched.
+
+It is this latter case that inspired the Steady Block Derivation rule. It guarantees that the
+secondary effects of an invalid Interop dependency are contained to a single block only, which
+avoids a cascade of cross-L2 Interop reorgs.
```
ajsutton: It can still cause a cascade of cross-L2 interop reorgs because the deposit-only block potentially has fewer events emitted which may invalidate blocks on other chains that depend on them. However it makes resolving those invalidation linear instead of NP complete (or possibly just completely unsolvable)  because the invalid block can't be replaced by one from a later batch which includes new transactions (which then may make previously invalidated blocks on other chains valid again).

ajsutton (approved):
LGTM.

refcell (approved):

specs/protocol/holocene/derivation.md
sebastianst: That's true, it certainly doesn't need to be in the spec. It's sufficient to only contain a _working_ version of the sync start, that doesn't necessarily need to be optimal. However, I still think it can be beneficial to include some high level implementation pointers in the spec. We could just move it into the "Security and Implementation Considerations" section.

specs/protocol/holocene/derivation.md
```diff
@@ -1,13 +1,332 @@
...
+before the batch queue, so that the batch queue pulls singular batches from this previous Span Batch
+stage. When encountering an invalid singular batch, it is dropped, as is the remaining span batch
+for consistency reasons. We call this _forwards-invalidation_. However, we don't
+_backwards-invalidate_ previous valid batches that came from the same span batch, as pre-Holocene.
+
+An implementation may choose to put the current span batch, from which singular batches are derived
+from, into a separate stage, or process it within the batch queue under the new rules.
+
+When a batch derived from the current staging channel is a singular batch, it is directly forwarded
+to the batch queue. Otherwise, it is set as the current span batch in the span batch stage. The
```
geoknee: I could be missing something, but this reads to me like you are offering two implementation hints but then proceeding to describe just one of those implementations in more detail? If so, I think it could be made clearer.

specs/protocol/holocene/derivation.md
```diff
@@ -1,13 +1,332 @@
...
+- Note that we still allow span batches to overlap with the safe chain (`span_start.timestamp < next_timestamp`).
+
+If any of the above checks invalidate the span batch, it is dropped and the remaining channel from
+which the span batch was derived, is also immediately dropped (see also
+[Fast Channel Invalidation](#fast-channel-invalidation)).
+
+## Batch Queue
+
+The batch queue is also simplified in that batches are required to arrive strictly ordered, and any
+batches that violate the ordering requirements are immediately dropped, instead of buffered.
```
geoknee, line 145: If a batch arrives out of order, say due to a bug in the batcher, would this mean the safe chain would halt and not start again until the gap was filled?

specs/protocol/holocene/derivation.md
```diff
@@ -1,13 +1,332 @@
...
+have influenced the L2 safe head.
+- If the first frame we find is a non-first frame, then we need to go back a full channel
+timeout window to see if we find the start of that channel.
+  - If we find the starting frame, we can continue derivation from it.
+  - If we don't find the starting frame, we need to go back a full sequencing window. (TODO: verify)
+
+_TODO: above needs to be verified, we don't introspect batcher transactions and frame metadata
+during sync start yet._
+
+# Rationale
```
geoknee, line 205: Wondering if this section should eventually move back to the design doc? Do we typically commit the rational into the specs themselves?

specs/protocol/holocene/derivation.md
```diff
@@ -1,13 +1,332 @@
...
+
+## Less Defensive Protocol
+
+The stricter derivation rules lead to a less defensive protocol. The old protocol rules allowed for
+second chances for invalid payloads and submitting frames and batches within channels out of order.
+Experiences from running OP Stack chains for over one and a half years have shown that these relaxed
+derivation rules are (almost) never needed, so stricter rules that improve worst-case scenarios for
+Fault Proofs and Interop are favorable.
+
+# Security and Implementation Considerations
```
geoknee, line 258: Similarly, this material could end up in the FMA. Do we also want it in the specs?

specs/protocol/holocene/derivation.md
geoknee: Picked up elsewhere, but also wondering if implementation related material belongs here? Seems very useful though.

specs/protocol/holocene/derivation.md
sebastianst: Exactly. This is intended. So the batcher should continuously monitor the sync status of the sequencer, and if its L1 inclusion block is past the last batcher transaction, but the safe chain hasn't advanced as expected, it should batch the non-safe unsafe section of the chain again. This is described in the batcher hardening section.

specs/protocol/holocene/derivation.md
sebastianst: We intentionally started doing this, yes, to give better context for implementers. I also kept the Rationale section here more streamlined than the design discussion, which talks about different options etc.
We also started adding security considerations.

specs/protocol/holocene/derivation.md
sebastianst: Yes, we want to have a brief security discussion here, but not of the extend as the FMA. E.g. if you look at `fjord/derivation.md`, each feature has its own "Rationale" and "Security Considertaions" section.

specs/protocol/holocene/derivation.md
sebastianst: On second thought, I think this is such a low-level implementation detail that I will just remove this sentence and not even move it below to the "Implementation Considerations" section.

specs/protocol/holocene/derivation.md
sebastianst: added verbatim, thanks

specs/protocol/holocene/derivation.md
sebastianst: Definitely. We may also enforce it in the implementation. I'll leave the comment in and update later during implementation.

specs/protocol/holocene/derivation.md
sebastianst: moved to implementation section

specs/protocol/holocene/derivation.md
sebastianst: That makes sense, fixed and added with explanation, thanks!

specs/protocol/holocene/derivation.md
protolambda: Yep, it avoids deeper reorg issues, but not with more than 1 step per chain, which seems like the best solution thus far. It's manageable though, especially if we prepare the deposit-only block (ideally on warm preimage memory, thus faster), so that it doesn't require a later sub-program step.

@sebastianst for the text to be technically correct: you could change the text to state "avoids a cascade of cross-L2 interop reorgs that revisit L2 chains more than once".

sebastianst: Thanks both, changed as you suggested proto.

Issue #340, opened 2024-08-16
Holocene: Make Elasticity configurable via SystemConfig


roberto-bayardo:
In EIP-1559, the gas target of a block is (gas limit / elasticity). And since elasticity is a fixed constant of 6 as of the Fjord upgrade, one cannot adjust the gas target without increasing the gas limit. This has led Base to implement a hack whereby a different limit (a "soft" gas limit) is enforced by the sequencer during block building in order to allow us to increase the gas target without also increasing the (effective) limit. This unfortunately breaks things like priority fee suggestion, and is also simply unintuitive to users who see a gaslimit in the SystemConfig contract that isn't the one actually being enforced.

I would like to propose we make Elasticity configurable via the SystemConfig L1 contract, just as the Gas Limit is configurable, allowing gas target and gas limit to be independently configured.

From @tynes: We might also consider adding the denominator to be dynamic as well. When the basefee rises super quickly when the target is exceeded, it creates this permanent pegging to the target throughput. It also This is a complex system and not sure what the best params ought to be for a high throughput chain. My guess is that the "ideal" params differ at different gas throughputs or usage patterns.

The proposal to improve l1 attribute handling by separating out static values should be a prerequisite for this work:
https://github.com/ethereum-optimism/specs/issues/122

tynes:
Per https://github.com/ethereum-optimism/specs/issues/122#issuecomment-2308584680, this makes a lot of sense to implement as a `ConfigType`

roberto-bayardo:
noting that we have decided not to implement #122 just yet and instead handle the new l1 attributes in the "normal" way to avoid making too many diffs to the derivation pipeline.

BlocksOnAChain:
@roberto-bayardo: +1 to not implementing https://github.com/ethereum-optimism/specs/issues/122 at the moment.
Since we kinda made things simpler with this approach + already merged [358](https://github.com/ethereum-optimism/specs/pull/358), do you have an estimate on when we close this line of work, having in mind that Holocene "code freeze" date is October 16th, 2024?

BlocksOnAChain:
@roberto-bayardo do we know what else is needed from your side to close the work on this issue?

PR #358, opened 2024-08-28
holocene: smart contract changes

### LINKED ISSUE #350: L2 Genesis Simplification (L2 MCP)
**Is your feature request related to a problem? Please describe.**

Right now there is not a great way to do L2 contracts releases because there is network specific configuration in the L2 genesis allocs in the form of storage and bytecode (immutables). This makes it difficult to release an `allocs.json` artifact as a release of the L2 genesis, as well as ensuring that no malicious code has been set in the genesis state of a chain. Ideally there should be a single deterministic L2 genesis state per hardfork and all network specific config should be sourced from L1.

This project involves moving all of the network specific configuration in the L2 predeploys to be sourced from L1. We want a system similar to interop's `setConfig` as described [here](https://github.com/ethereum-optimism/specs/issues/122#issuecomment-2272191134) that sets the network specific values into storage via deposit transactions when the `SystemConfig` is initialized.

Examples of this network specific config:
- [L2StandardBridge.otherBridge](https://github.com/ethereum-optimism/optimism/blob/1dbe9d829361b65320c64a8512b0e82898bd6b05/packages/contracts-bedrock/src/L2/L2StandardBridge.sol#L65)
- [L2CrossDomainMessenger.otherMessenger](https://github.com/ethereum-optimism/optimism/blob/1dbe9d829361b65320c64a8512b0e82898bd6b05/packages/contracts-bedrock/src/L2/L2CrossDomainMessenger.sol#L31)
- [FeeVault config](https://github.com/ethereum-optimism/optimism/blob/1dbe9d829361b65320c64a8512b0e82898bd6b05/packages/contracts-bedrock/src/L2/L1FeeVault.sol#L21), for each of the 3 fee vaults
- L2 ProxyAdmin owner

Instead of reading these values from storage on L2, they could be exposed via getters on the `L1Block` contract and then read from a single source of truth. The L1 `setConfig` path would be able to issue deposit transactions that can modify these values. This would decouple the L2 genesis state from network specific config and simplify the process of spinning up a new chain because all chains can use the same `allocs.json`

tynes:
**Description**

This PR is the spec for:
- https://github.com/ethereum-optimism/specs/issues/350
- https://github.com/ethereum-optimism/specs/issues/340

We are explicitly not scoping https://github.com/ethereum-optimism/specs/issues/122 because it will result in a large diff to the proofs. In the future we can work on proof optimizations.


specs/protocol/holocene/configurability.md
```diff
@@ -0,0 +1,89 @@
...
+`ConfigType`.
+
+- `setGasLimit(uint64)`
+- `setFeeScalars(uint32,uint32)`
+- `setEIP1559Params(uint64,uint64)`
+- `setBatcherHash(bytes32)`
+- `setUnsafeBlockSigner(address)`
+- `setGasPayingToken(address)`
+- `setBatchInbox(address)`
+
```
tynes, line 129: @roberto-bayardo @blockchaindevsh  What was the API that we wanted to standardize on?

specs/protocol/holocene/configurability.md
```diff
@@ -0,0 +1,89 @@
...
+| Name | Value | Description |
+| ---- | ----- | --- |
+| `SET_GAS_LIMIT` | `uint8(0)` | Sets the L2 gas limit |
+| `SET_FEE_SCALARS` | `uint8(1)` | Sets the L2 fee configuration |
+| `SET_EIP_1559_PARAMS` | `uint8(2)` | Sets the EIP-1559 elasticity and denominator |
+| `SET_BATCHER_HASH` | `uint8(3)` | Sets the batcher role |
+| `SET_UNSAFE_BLOCK_SIGNER` | `uint8(4)` | Sets the p2p sequencer key |
+| `SET_GAS_PAYING_TOKEN` | `uint8(5)` | Modifies the gas paying token for the chain |
+| `SET_BATCH_INBOX` | `uint8(6)` | Sets the canonical batch inbox address |
```
tynes: We need an enum for `REMOTE_CHAIN_ID` - this would deposit the L1 chainid into L2 on `SystemConfig` initialization, enabling the L1 chainid to be stored in the L2 `L1Block` predeploy. We depend on the L1 chain id in an L2 predeploy, so to have a deterministic L2 genesis we need to not embed it in there.

specs/protocol/holocene/configurability.md
```diff
@@ -0,0 +1,89 @@
...
+transaction to not post duplicate data between transactions which reduces database
+size growth over time.
+
+## Constants
+
+### `ConfigType`
+
+The `ConfigType` enum represents configuration that can be modified.
+
+| Name | Value | Description |
```
tynes, line 38: Now that I look at this, this is defining the enum value, that is the key in `setConfig`, we should also define the value in terms of the `bytes` in `setConfig(ConfigType,bytes)`. This is useful because the serialization is different for each type

specs/protocol/holocene/configurability.md
```diff
@@ -0,0 +1,89 @@
...
+| ---- | ----- | --- |
+| `SET_GAS_LIMIT` | `uint8(0)` | Sets the L2 gas limit |
+| `SET_FEE_SCALARS` | `uint8(1)` | Sets the L2 fee configuration |
+| `SET_EIP_1559_PARAMS` | `uint8(2)` | Sets the EIP-1559 elasticity and denominator |
```
tynes: @roberto-bayardo just confirming with you here, this will set the EIP 1559 params, it can `abi.encode(uint64,uint64)`

specs/protocol/holocene/configurability.md
```diff
@@ -0,0 +1,89 @@
...
+| `SET_BATCHER_HASH` | `uint8(3)` | Sets the batcher role |
+| `SET_UNSAFE_BLOCK_SIGNER` | `uint8(4)` | Sets the p2p sequencer key |
+| `SET_GAS_PAYING_TOKEN` | `uint8(5)` | Modifies the gas paying token for the chain |
```
blockchaindevsh: The `SET_GAS_PAYING_TOKEN` functionality is currently implemented in the [Interop contract](https://github.com/ethereum-optimism/optimism/blob/aebf669c108240298e14bd0e29068a18e0b0a52e/packages/contracts-bedrock/src/L2/L1BlockInterop.sol#L63) . Should we consider moving this implementation into the Holocene contract?

specs/protocol/holocene/configurability.md
```diff
@@ -0,0 +1,89 @@
...
+| `SET_BATCH_INBOX` | `uint8(6)` | Sets the canonical batch inbox address |
+
+TODO: make sure that `SET_BATCH_INBOX` doesn't break the fault proof
+
+## `SystemConfig`
+
+### Interface
+
+Each listed function calls `OptimismPortal.setConfig(ConfigType)` with its corresponding
+`ConfigType`.
```
blockchaindevsh: Should be : Each listed function calls `OptimismPortal.setConfig(ConfigType)` with its corresponding
value ?

specs/protocol/holocene/configurability.md
blockchaindevsh: It was `setFeeScalars`, more specifically how we want to implement `StaticConfig.encodeSetFeeScalars`. I think for the Holocene upgrade we still want to maintain the [scalar](https://github.com/ethereum-optimism/optimism/blob/d7e5418745ef5da2df999a98da11cde228a4fbaa/packages/contracts-bedrock/src/L1/SystemConfigHolocene.sol#L47) for backwards compatibility, that's why I implemented `StaticConfig.encodeSetFeeScalars` by encoding the calculated scalar: [source](https://github.com/blockchaindevsh/optimism/blob/e876fe4029ddac58df548661fdc892267b975004/packages/contracts-bedrock/src/L1/SystemConfigHolocene.sol#L50). Do we want to implement it as:

```solidity
StaticConfig.encodeSetFeeScalars(uint32 _basefeeScalar, uint32 _blobBasefeeScalar) internal pure returns (bytes memory)  {
    return abi.encode(_basefeeScalar, _blobBasefeeScalar);
}
```

 instead ?

specs/protocol/holocene/configurability.md
blockchaindevsh: Currently, all the `StaticConfig.encodeXXX` functions are implemented as wrappers around `abi.encode`. Should we consider establishing this as a standard practice?

tynes: That library is specific for encoding and decoding the values passed through the `setConfig` API. We don't necessarily need to ABI encode them, we could use `calldataload` directly, but its most simple to just `abi.encode`

specs/protocol/holocene/configurability.md
tynes: Yes we will want to move this to the holocene contract

specs/protocol/holocene/configurability.md
tynes: Yes this was a typo, thank you

specs/protocol/holocene/configurability.md
```diff
@@ -0,0 +1,136 @@
...
+The following `ConfigUpdate` enum is defined where the `CONFIG_VERSION` is `uint256(0)`:
+
+| Name | Value | Definition | Usage |
+| ---- | ----- | --- | -- |
+| `BATCHER` | `uint8(0)` | `abi.encode(address)` | Modifies the account that is authorized to progress the safe chain |
+| `GAS_CONFIG` | `uint8(1)` | `(uint256(0x01) << 248) \| (uint256(_blobbasefeeScalar) << 32) \| _basefeeScalar` | Modifies the fee scalars |
+| `GAS_LIMIT` | `uint8(2)` | `abi.encode(uint64 _gasLimit)` | Modifies the L2 gas limit |
+| `UNSAFE_BLOCK_SIGNER` | `uint8(3)` | `abi.encode(address)` | Modifies the account that is authorized to progress the unsafe chain |
+| `EIP_1559_PARAMS` | `uint8(4)` | `abi.encode(uint64 _denominator, uint64 _elasticity)` | Modifies the EIP-1559 denominator and elasticity |
+
```
tynes, line 62: We could add a section on `Initialization` that clearly defines which calls should be made during an initialize call

specs/protocol/holocene/configurability.md
tynes: I would prefer to make no diff to this and just leave the most significant byte as a version (0 or 1) as to reduce need for diff in the derivation pipeline.

specs/protocol/holocene/predeploys.md
```diff
@@ -0,0 +1,252 @@
...
+      - [`l1ERC721Bridge`](#l1erc721bridge)
+      - [`remoteChainId`](#remotechainid)
+  - [FeeVault](#feevault)
+    - [Interface](#interface-1)
+      - [`config`](#config)
+  - [L2CrossDomainMessenger](#l2crossdomainmessenger)
+    - [Interface](#interface-2)
+  - [L2ERC721Bridge](#l2erc721bridge)
+    - [Interface](#interface-3)
+  - [L2StandardBridge](#l2standardbridge)
```
tynes, line 31: This is missing sections for `OptimismMintableERC721Factory` and `L2ERC721Bridge`

specs/protocol/holocene/configurability.md
```diff
@@ -0,0 +1,136 @@
...
+The following `ConfigUpdate` enum is defined where the `CONFIG_VERSION` is `uint256(0)`:
+
+| Name | Value | Definition | Usage |
+| ---- | ----- | --- | -- |
+| `BATCHER` | `uint8(0)` | `abi.encode(address)` | Modifies the account that is authorized to progress the safe chain |
+| `GAS_CONFIG` | `uint8(1)` | `(uint256(0x01) << 248) \| (uint256(_blobbasefeeScalar) << 32) \| _basefeeScalar` | Modifies the fee scalars |
+| `GAS_LIMIT` | `uint8(2)` | `abi.encode(uint64 _gasLimit)` | Modifies the L2 gas limit |
+| `UNSAFE_BLOCK_SIGNER` | `uint8(3)` | `abi.encode(address)` | Modifies the account that is authorized to progress the unsafe chain |
+| `EIP_1559_PARAMS` | `uint8(4)` | `abi.encode(uint64 _denominator, uint64 _elasticity)` | Modifies the EIP-1559 denominator and elasticity |
```
tynes: This should be a `bytes32` rather than 2 words, no reason to `abi.encode`

specs/protocol/holocene/configurability.md
```diff
@@ -0,0 +1,136 @@
...
+## `SystemConfig`
+
+### `ConfigUpdate`
+
+The following `ConfigUpdate` enum is defined where the `CONFIG_VERSION` is `uint256(0)`:
+
+| Name | Value | Definition | Usage |
+| ---- | ----- | --- | -- |
+| `BATCHER` | `uint8(0)` | `abi.encode(address)` | Modifies the account that is authorized to progress the safe chain |
+| `GAS_CONFIG` | `uint8(1)` | `(uint256(0x01) << 248) \| (uint256(_blobbasefeeScalar) << 32) \| _basefeeScalar` | Modifies the fee scalars |
```
roberto-bayardo: why not just call this FEE_SCALARS

tynes: I prefer this

tynes:
The `permit2` predeploy has the L2 chainid embedded in the bytecode. We need a way to handle this, I think we should just wipe the chainid to by `0` so it just uses `block.chainid` rather than a cached version

specs/protocol/holocene/predeploys.md
```diff
@@ -0,0 +1,252 @@
...
+  BaseFeeVault -- "baseFeeVaultConfig()(address,uint256,uint8)" --> L1Block
+  SequencerFeeVault -- "sequencerFeeVaultConfig()(address,uint256,uint8)" --> L1Block
+  L1FeeVault -- "l1FeeVaultConfig()(address,uint256,uint8)" --> L1Block
+  L2CrossDomainMessenger -- "l1CrossDomainMessenger()(address)" --> L1Block
+  L2StandardBridge -- "l1StandardBridge()(address)" --> L1Block
+  L2ERC721Bridge -- "l1ERC721Bridge()(address)" --> L1Block
+  OptimismMintableERC721Factory -- "remoteChainId()(uint256)" --> L1Block
+  end
+  OptimismPortal -- "setConfig(uint8,bytes)" --> L1Block
+```
```
tynes, line 83: With all of these calls, we need to think about the upgrade path. To not introduce a tight synchrony assumption between L1 and L2 upgrades, we may want to have the calls to the `L1Block` contract fall back to local storage if they fail or return the zero value. This would allow a progressive migration

tynes:
We need to think about the semantics of the upgrade.

For the L1Block storage part of the upgrade, we either need to:
- Include a "migration contract" that deterministically migrates the state
- Have fallbacks in each implementation contract that try to read from L1Block then fall back to state

Both have tradeoffs

The migration contract approach would look like upgrading the L1Block contract to a contract that is able to make a bunch of calls to get the data it needs and then storing it in its storage at the correct locations. So this would be fetching the storage values, for example calling `L2CrossDomainMessenger.otherMessenger`. We would need to make sure that all of the values can be accessed for this to work. The problem with this approach is that it will use a lot of gas and we need to be mindful of not using too much gas during the upgrade, otherwise it will conflict with user deposits. This is a risk we will always face until we have deposit queue. We could also make the semantics such that the deposit transactions are done over multiple blocks, but this is not flexible with certain configurations of the OP Stack.

The fallback approach will not force the storage migration to ever happen and we will likely be stuck with the fallback forever or at least a good amount of time until we can be sure that a `SystemConfig.initialize` does send all of the correct information to L2

specs/protocol/holocene/l1-attributes.md
```diff
@@ -0,0 +1,28 @@
...
+| blobBaseFeeScalar | uint32  | 8-11           |         |
+| sequenceNumber    | uint64  | 12-19          |         |
+| l1BlockTimestamp  | uint64  | 20-27          |         |
+| l1BlockNumber     | uint64  | 28-35          |         |
+| basefee           | uint256 | 36-67          | 2       |
+| blobBaseFee       | uint256 | 68-99          | 3       |
+| l1BlockHash       | bytes32 | 100-131        | 4       |
+| batcherHash       | bytes32 | 132-163        | 5       |
+| 1559Denominator   | uint64  | 164-171        | 6       |
+| 1559Elasticity    | uint64  | 172-179        |         |
```
roberto-bayardo:
```suggestion
| eip1559Denominator   | uint64  | 164-171        | 6       |
| eip1559Elasticity    | uint64  | 172-179        |         |
```

you seem to use the "eip" prefix elsewhere.

roberto-bayardo:
overall this LGTM other than the unresolved comments.

BlocksOnAChain:
Async reminder:
 @protolambda feel free to add your thoughts on any potential risks we might have with this approach.

protolambda:
@BlocksOnAChain no immediate concerns.

From a purist view, I like if the post-genesis, but pre-user, state changes are immediately after genesis, rather than at any point (there may be gaps, because of the fixed block time). Should not be a concern if deposits are only enabled after the post-genesis system deposits complete. 

And the 1559 fee parameters have fun lower/upper bound cases, but won't brick the system as long as they can be modified through L1-initiated deposits.

tynes:
I will rebase this before merge

specs/protocol/holocene/l1-attributes.md
tynes: resolved this

tynes:
> We need to think about the semantics of the upgrade.
> 
> For the L1Block storage part of the upgrade, we either need to:
> 
> * Include a "migration contract" that deterministically migrates the state
> * Have fallbacks in each implementation contract that try to read from L1Block then fall back to state
> 
> Both have tradeoffs
> 
> The migration contract approach would look like upgrading the L1Block contract to a contract that is able to make a bunch of calls to get the data it needs and then storing it in its storage at the correct locations. So this would be fetching the storage values, for example calling `L2CrossDomainMessenger.otherMessenger`. We would need to make sure that all of the values can be accessed for this to work. The problem with this approach is that it will use a lot of gas and we need to be mindful of not using too much gas during the upgrade, otherwise it will conflict with user deposits. This is a risk we will always face until we have deposit queue. We could also make the semantics such that the deposit transactions are done over multiple blocks, but this is not flexible with certain configurations of the OP Stack.
> 
> The fallback approach will not force the storage migration to ever happen and we will likely be stuck with the fallback forever or at least a good amount of time until we can be sure that a `SystemConfig.initialize` does send all of the correct information to L2

I solved this with a `setHolocene` function on the `L1Block` contract that does the storage migration

specs/protocol/holocene/predeploys.md
tynes: solved in https://github.com/ethereum-optimism/specs/pull/358#issuecomment-2358797255

PR #364, opened 2024-09-09
Fix broken links to security-council-safe.md

geoknee:
Following #327


maurelian (approved):
This LGTM to so am approving, but the CI error needs to be fixed.

tynes (approved):

PR #365, opened 2024-09-09
Some tweaks to the batch derivation spec

geoknee:


specs/protocol/derivation.md
```diff
@@ -396,6 +390,11 @@ The Fjord upgrade introduces an additional [versioned channel encoding
 format](./fjord/derivation.md#brotli-channel-compression) to support alternate compression
 algorithms.
 
+In the future this channel identification feature also allows the [batcher][g-batcher] to employ multiple signers
+(private keys) to submit one or multiple channels in parallel.
+This may help to alleviate issues where, because of transaction nonce values affecting the L1 tx-pool and thus inclusion:
+multiple transactions made by the same signer are stuck waiting on the inclusion of a previous transaction.
+
```
sebastianst: Honestly, let's just delete this. Speculation about future features IMO shouldn't be part of the specs and are confusing at worst. Plus, With the strict ordering requirements that we introduce in Holocene, we need to rethink parallel batcher keys anyways.

sebastianst (approved):

specs/protocol/derivation.md
geoknee: That works for me, [ffbb882](https://github.com/ethereum-optimism/specs/pull/365/commits/ffbb8820ba586c1272e0dbf098f5a2aeb5db7161)

PR #368, opened 2024-09-10
Add google analytics scripts to head.hbs

geoknee:
https://rust-lang.github.io/mdBook/format/configuration/renderers.html?highlight=analytics#html-renderer-options

### DIFF
--- a/theme/head.hbs
+++ b/theme/head.hbs
@@ -1,0 +1,8 @@
+<!-- Google tag (gtag.js) -->
+<script async src="https://www.googletagmanager.com/gtag/js?id=G-YNLKSPKGWN"></script>
+<script>
+    window.dataLayer = window.dataLayer || [];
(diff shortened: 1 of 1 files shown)


ajsutton (approved):

baokiddy:
Thank you for setting this up, @geoknee!

To give context on this edit, there is a current data push to set up more formalised tracking on both the developer documentation and the stack specifications with the goal of understanding developer and collaborator activity. The starting point is to work on creating a baseline to better understand how users find these resources and how they typically use them. Additionally, these tools also allow the team to better track  errors and remain on top of any issues with the sites during usage. 

All this data is being collected with the key point of better informing improvements and updates to address documentation gaps or pain points. This data will also used for the team to understand the rating of our pages in terms of accessibility and usability, particularly for mobile sites. This is all part of my larger technical content strategy being led by , I'm trying to get this set up for all of my like the main touch points.

PR #370, opened 2024-09-10
interop: specify deposits-complete handling

protolambda:
**Description**

Specify the deposits-complete system transaction functionality, the exact transaction contents, and the source-hash computation.

Also expand the tx-pool section with the effective-gas-limit change, to accommodate the system transaction.

Question: previously it was effectively a 70k gas reservation, so expanding it to 100k means the gas limit effectively shrinks by 30k. Setting the total gas limit from 30M to 30,030k seems a little silly though. But some product feedback may be useful here, if anyone strongly relies on having that 30k gas available.


tynes (approved):
Confirmed that this matches the the implementation in https://github.com/ethereum-optimism/optimism/pull/11712

tynes (approved):

tynes (approved):

PR #375, opened 2024-09-17
specs: remove diff based L1 attributes tx

tynes:
**Description**

This spec should generally be considered dead, it is much better
as a github issue since its gone stale and there are way higher priority
things. I introduced this early on to push forward the specs process and
it is no longer necessary.


ajsutton (approved):

Java7981:
Gooooood 🎉 🎉

PR #379, opened 2024-09-18
configurability: Add additional fault dispute game settings

ajsutton:
**Description**

Add missing fault dispute game settings to the configurability list.


Inphi (approved):

specs/protocol/configurability.md
```diff
@@ -41,23 +41,33 @@ The recommended way to deploy L1 contracts for an OP chain that meet the standar
...
+| Bond Withdrawal Delay                                                                                                                                                       | The length of time that must pass before dispute game bonds can be withdrawn.                                                                                                                                                                                | Static                              | 7 days                                                                                                                                                                                                        |                                                                                                                                                                                                                                                                                              |
```
mds1: Does Bond Withdrawal Delay map to `DelayedWETH.DELAY_SECONDS`? If we can note in each entry to where in the code it's configured, that would be really helpful for chains (and us with OPSM) to make sure standard chain deployments are configured correctly. I'm indifferent to whether it's a link in the first column, or just in the notes column something like "Corresponds to `DelayedWETH.DELAY_SECONDS`"

specs/protocol/configurability.md
```diff
@@ -41,23 +41,33 @@ The recommended way to deploy L1 contracts for an OP chain that meet the standar
...
+| Fault Game Max Depth                                                                                                                                                        | The maximum depth of fault dispute games.                                                                                                                                                                                                                    | Static                              | 73                                                                                                                                                                                                            |                                                                                                                                                                                                                                                                                              |
```
mds1: Can we add the rationale for each value in the notes column? It feels like a lot of magic numbers currently. I assume there's some complexity behind the rationale for these values, in which case the notes column can link somewhere else in the specs instead if that's easier

ajsutton: Added some notes but in a lot of cases they are actually pretty arbitrary, or at least there's a very broad range of values that would work fine.

specs/protocol/configurability.md
ajsutton: It maps to `faultGameWithdrawalDelay` in the deploy config JSON but that's not part of the specs and from there it does wind up  as `DelayedWETH.DELAY_SECONDS`.

I've added links to relevant sections of the specs - not sure how much that helps find the respective part of the implementation but should give a bit more detail.

ajsutton:
Once this is merged, I'm going to follow up by taking these out of a table format. They're utterly unreadable on the site, requiring a heap of weird side scrolling in a narrow little fixed size part of the window and even more unmaintainable in the actual markdown.

specs/protocol/configurability.md
mds1: Thank you!

mds1 (approved):

specs/protocol/configurability.md
mds1: Thanks, we will take a pass at transcribing these to solidity in the OPSM code, and likely will ask @clabby or @Inphi to review afterwards

maurelian:
> Once this is merged, I'm going to follow up by taking these out of a table format. They're utterly unreadable on the site, requiring a heap of weird side scrolling in a narrow little fixed size part of the window and even more unmaintainable in the actual markdown.

Apprecate this, as it makes the diff unreadable as well.

mslipper (approved):

PR #381, opened 2024-09-20
configurability: Remove tables

ajsutton:
**Description**

Switch from using tables to layout configurability requirements to labels. The tables required side-scrolling both in edit and read mode and created huge unreadable diffs when changes were made.

Also fixed invalid syntax in interop/tx-pool.md which was causing the book build to fail.


pauldowman (approved):
This will be a lot easier to maintain, the linting rules make it hard to edit large tables (every line changes when the length of the widest value in a column changes).

clabby (approved):

mslipper (approved):

PR #382, opened 2024-09-20
Isthmus: operator fee 

yuwen01:
## Overview

We propose adding additional fee scalars to the fee formula, which allow for more flexibility for chains that leverage alt-DA, ZK proving, or custom gas tokens.

This spec goes with [this design doc](https://github.com/ethereum-optimism/design-docs/pull/81/files).


specs/protocol/isthmus/configurability.md
```diff
@@ -0,0 +1,24 @@
...
+## `SystemConfig`
+
+### `ConfigUpdate`
+
+The following `ConfigUpdate` enum is defined where the `CONFIG_VERSION` is `uint256(1)`:
+
+| Name | Value | Definition | Usage |
+| ---- | ----- | --- | -- |
+| `BATCHER` | `uint8(0)` | `abi.encode(address)` | Modifies the account that is authorized to progress the safe chain |
+| `FEE_SCALARS` | `uint8(1)` | `(uint256(0x01) << 248) \| (uint256(_configurablefeescalar) << 128 \| (uint256(_configurablefeeconstant) << 64 \| (uint256(_blobbasefeeScalar) << 32) \| _basefeeScalar` | Modifies the fee scalars |
```
sebastianst: To keep it consistent with the (blob) base fee scalar variable names, we should use cameCase.
```suggestion
| `FEE_SCALARS` | `uint8(1)` | `(uint256(0x01) << 248) \| (uint256(_configurableFeeScalar) << 128 \| (uint256(_configurableFeeConstant) << 64 \| (uint256(_blobBaseFeeScalar) << 32) \| _baseFeeScalar` | Modifies the fee scalars |
```

specs/protocol/isthmus/exec-engine.md
```diff
@@ -0,0 +1,45 @@
...
+    - [Configuring scalars](#configuring-scalars)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Fees
+
+New OP stack variants have different resource consumption patterns, and thus require a more flexible
+pricing model. To enable more customizable fee structures, Isthmus adds a new component to the fee
+calculation: the `ConfigurableFee`, which is parameterized by two scalars: the `configurableFeeScalar`
+and the `configurableFeeConstant`.
```
sebastianst: I find the use of the prefix "configurable" a bit meaningless for this feature. Other fee parameters, like the `(blob)BaseFeeScalar`s are also "configurable". Maybe we use a prefix that better describes the reason for their introduction, like
* `OperatorFee`
* `operatorFeeScalar`
* `operatorFeeConstant`
or something similar that attaches more meaning to them? `fixedFee...` could also work.

specs/protocol/isthmus/l1-attributes.md
```diff
@@ -0,0 +1,40 @@
...
+| configurableFeeScalar  | uint64  | 180-187        |         |
+| configurableFeeConstant    | uint64  | 188-195        |         |
+
+In the first L2 block after the Isthmus activation block, the Isthmus L1 attributes are first used.
+
+The pre-Isthmus values are migrated over 1:1.
+Blocks after the Isthmus activation block contain all pre-Isthmus values 1:1,
+and also set the following new attributes:
+
+- The `configurableFeeScalar` is set to `0`.
```
sebastianst: Don't we want to set it to `1`? Otherwise there's no fees any more.

specs/protocol/isthmus/exec-engine.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Configurable fees
+
+The configurable fee is set as follows:
+
+`configurableFee = gas_used * configurableFeeScalar + configurableFeeConstant`
```
sebastianst: So we don't need any fractional scaling, like we introduced with Fjord for the model parameters? I mean something like
```suggestion
`configurableFee = (gas_used * configurableFeeScalar + configurableFeeConstant) / 1e6`
```
to allow for a decimal precision of 6.

specs/protocol/isthmus/l1-attributes.md
```diff
@@ -0,0 +1,40 @@
...
+| l1BlockTimestamp  | uint64  | 20-27          |         |
+| l1BlockNumber     | uint64  | 28-35          |         |
+| basefee           | uint256 | 36-67          | 2       |
+| blobBaseFee       | uint256 | 68-99          | 3       |
+| l1BlockHash       | bytes32 | 100-131        | 4       |
+| batcherHash       | bytes32 | 132-163        | 5       |
+| eip1559Denominator   | uint64  | 164-171        | 6       |
+| eip1559Elasticity    | uint64  | 172-179        |         |
+| configurableFeeScalar  | uint64  | 180-187        |         |
+| configurableFeeConstant    | uint64  | 188-195        |         |
```
sebastianst: Do we really want or need 64 bits instead of 32 bits size for the new parameters? E.g. the `(blob)baseFeeScalar`s also worked with 32 bits (and also a decimal scaling factor, see other comment).

sebastianst (commented):
Also curious to hear from @tynes and @roberto-bayardo, who's implemented changes to the fee function in Fjord.

I propose to use a prefix for this feature that conveys more meaning, like `OperatorFee` or `FixedFee`.

specs/protocol/isthmus/l1-attributes.md
yuwen01: The configurableFeeScalar is only scaled by the gas used -- it doesn't scale any of the existing fees. The goal is to add a separate component to the fee calculation, like base fee and priority fee.

specs/protocol/isthmus/exec-engine.md
yuwen01: Good point -- it makes sense for users to be able to have fractional scalars. However, I don't know why a user would want to have a fractional constant. The only reason I can think would be to save bits -- see my other comment.

specs/protocol/isthmus/l1-attributes.md
yuwen01: Thanks for the feedback! I agree with your point about renaming to `operatorFee` and allowing for 6 decimal points of precision, but I was a little unsure about reducing the bit width of the `operatorFeeConstant` and `operatorFeeScalar`. 

I think it should be fine to decrease the `Scalar` to 32 bits, but I'm concerned that 32 bits won't be enough to represent the constant factor. For example, in this transaction [https://optimistic.etherscan.io/tx/0xa6dfc18c35bf39fa60823e9280bde18496e27e9016040f7ad9ded6797c374f05](transaction), the total transaction fee in wei requires 43 bits to represent.

If we scale the constant term by a fixed factor we could fit it in 32 bits. But I don't know how much control a user might want over this constant.

specs/protocol/holocene/predeploys.md
```diff
@@ -255,6 +258,21 @@ The following functions are updated to read from the `L1Block` contract by calli
...
+#### Interface
+
+##### `getOperatorFee`
+
+This function calculates the operator fee based on the expected amount of gas used for a certain transaction.
+
+```function
+function getOperatorFee(uint256 gasUsed)(uint256)
+```
+
```
tynes: We will also need an `isHolocene` bool as well as branching logic in the fee estimation functions

specs/protocol/holocene/system-config.md
```diff
@@ -0,0 +1,39 @@
+# System Config
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [System config contents (version 0)](#system-config-contents-version-0)
+  - [Scalars](#scalars)
+    - [Holocene `scalar`, `overhead` (`uint256,uint256`) change](#holocene-scalar-overhead-uint256uint256-change)
+
```
tynes: We will also want to include the specific ABI changes to the `SystemConfig`

specs/protocol/holocene/system-config.md
```diff
@@ -0,0 +1,39 @@
...
+After Holocene activation:
+
+- A new version to the `scalar` encoding is added. It is described as follows.
+
+\*Byte ranges are indicated with `[` (inclusive) and `)` (exclusive).
+
+- `0`: scalar-version byte
+- `[1, 32)`: depending scalar-version:
+  - Scalar-version `2`:
+    - `[1, 8)`: padding, must be zero.
```
refcell: I believe this should be padded to the start of the `operatorFeeScalar`.

i.e.
```suggestion
    - `[1, 12)`: padding, must be zero.
```

refcell (commented):
Nice work.

specs/protocol/isthmus/predeploys.md
```diff
@@ -243,6 +269,40 @@ The following functions are updated to read from the `L1Block` contract by calli
...
+### GasPriceOracle
+
+In order to maintain accurate offchain fee estimation, the `GasPriceOracle` must be updated to allow users
+to estimate the operator fee. We also add a new boolean `isIsthmus` to help with evaluating the operator fee.
+
+#### Interface
+
+##### `setIsthmus`
+
+This function is meant to be called once on the activation block of the isthmus network upgrade.
```
yuwen01: on second thought maybe this isn't necessary -- if we read the scalars from the l1block before isthmus is set, we'll just read from uninitialized storage, which defaults to 0. So the derived operator fee will just be 0.

specs/protocol/isthmus/configurability.md
```diff
@@ -10,22 +10,34 @@
 - [`SystemConfig`](#systemconfig)
   - [`ConfigUpdate`](#configupdate)
   - [Initialization](#initialization)
+  - [Modifying Operator Fee Parameters](#modifying-operator-fee-parameters)
   - [Interface](#interface)
+    - [Operator fee parameters](#operator-fee-parameters)
+      - [`operatorFeeScalar`](#operatorfeescalar)
+      - [`operatorFeeConstant`](#operatorfeeconstant)
+      - [`setOperatorFeeScalars`](#setoperatorfeescalars)
+      - [`setOperatorFeeManager`](#setoperatorfeemanager)
```
tynes: Can we create a role in the `SuperchainConfig` called "operator fee manager" and source the value from there? https://github.com/ethereum-optimism/specs/blob/d6b979ab67bd98bc3a15bf4988df08a066788047/specs/protocol/isthmus/superchain-config.md

This will make standard definition more simple, use the same superchain config that op gov recognizes, chain can also use their own superchain config if they are their own L2

specs/protocol/isthmus/configurability.md
```diff
@@ -67,19 +82,73 @@ The following actions should happen during the initialization of the `SystemConf
 - `emit ConfigUpdate.GAS_LIMIT`
 - `emit ConfigUpdate.UNSAFE_BLOCK_SIGNER`
 - `emit ConfigUpdate.EIP_1559_PARAMS`
+- `emit ConfigUpdate.OPERATOR_FEE_PARAMS`
+- `emit ConfigUpdate.OPERATOR_FEE_MANAGER`
```
tynes: We don't need to have the manager be a `ConfigUpdate`, you can remove this

specs/protocol/isthmus/configurability.md
```diff
@@ -67,19 +82,73 @@ The following actions should happen during the initialization of the `SystemConf
...
+
+This function sets the `operatorFeeScalar` and `operatorFeeConstant`.
+
+This function MUST only be callable by the [`OperatorFeeManager`](#operator-fee-manager).
+
+```solidity
+function setOperatorFeeScalar(uint32 _operatorFeeScalar, uint64 _operatorFeeConstant)()
+```
+
+##### `setOperatorFeeManager`
```
tynes: I would like to delete this setter and instead source the fee manager from the superchain config as mentioned in a previous comment

specs/protocol/isthmus/predeploys.md
tynes: This is useful for the `GasOracle` predeploy that attempts to do L2 fee estimation

tynes: You can see [here](https://github.com/ethereum-optimism/optimism/blob/1f106f7868e15a80184145322f165fb8db0bc83c/packages/contracts-bedrock/src/L2/GasPriceOracle.sol#L57) how we have handled this in the past. I believe we will need to include a calculation of the operator fee in the `GasPriceOracle` now

tynes:
This spec is looking great so far! Implementation will want to build on top of https://github.com/ethereum-optimism/optimism/pull/12057 which is targeting audit Nov 4th, so it should be merged in over the next 2 weeks

specs/protocol/isthmus/configurability.md
tynes: Follow up: trying to align ongoing work that is changing with this so that things are coherent, [this](https://github.com/ethereum-optimism/design-docs/blob/4980faff965e6caa1fc35ea015654c17c5931d10/protocol/standard-l2-genesis.md#systemconfig) is the design that we are landing on.

The `SystemConfig.initialize` function would accept a `Roles` struct, it would be good to add the `operatorFeeManager` to that `Roles` struct and have no setter function meaning the operator can only change with another call to `initialize`. This helps keep the ABI slim and allows for each chain operator to have its own individual account that is used.

It is still possible to have each chain have its own individual accounts in the world where its sourced from the `SuperchainConfig`, it just means that chains that aren't governed by Optimism would deploy their own `SuperchainConfig`. Apologies for the back and forth, was trying to be consistent with designs in the Standard L2 Genesis project and we were flip flopping internally on how we were going to do it

yuwen01: This makes more sense to me than the superchain solution, happy to proceed in this direction. Once the spec lands for the `Roles` struct then I'll update this spec accordingly -- I've been yet unable to find reference to the design choices you linked.

refcell:
Any updates here?

yuwen01:
Hi! Sorry for the delay, but I've updated the spec setting the operator fee manager in the Roles struct. See `system-config.md` in the isthmus folder.

specs/protocol/isthmus/l1-attributes.md
```diff
@@ -0,0 +1,37 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+The L1 block attributes transaction is updated to include the operator fee parameters.
+
+| Input arg         | Type    | Calldata bytes | Segment |
+| ----------------- | ------- | -------------- | ------- |
+| {0x098999be}      |         | 0-3            | n/a     |
```
maurelian, line 17: what is this?

yuwen01: That's the call signature for the function. Those 4 bytes are the first four bytes of `keccak("setL1BlockValuesIsthmus()")`. I'll clarify this in the spec.

specs/protocol/isthmus/configurability.md
```diff
@@ -67,19 +80,59 @@ The following actions should happen during the initialization of the `SystemConf
...
+```
+
+##### `setOperatorFeeScalars`
+
+This function sets the `operatorFeeScalar` and `operatorFeeConstant`.
+
+This function MUST only be callable by the [`OperatorFeeManager`](#operator-fee-manager).
+
+```solidity
+function setOperatorFeeScalar(uint32 _operatorFeeScalar, uint64 _operatorFeeConstant)()
```
tynes: small nit: I think its fine to omit the trailing `()` when there is no return arg

specs/protocol/isthmus/configurability.md
```diff
@@ -157,3 +216,28 @@ The following fields are included:
...
+
+## Consensus Parameters
+
+The operator fee scalar and constant are new consensus parameters, so we define standard values for them.
+
+### [Operator Fee Scalar](exec-engine.md#operator-fees)
+
+**Description:** Operator fee scalar -- used to calculate the operator fee<br/>
+**Administrator:** [Operator Fee Manager](#operator-fee-manager)<br/>
+**Requirement:** Between 0 and 0.5 * (baseFee + priorityFee) <br/>
```
tynes: This isn't noted anywhere, apologies about that, but its probably best to put these configurability requirements in [this](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/configurability.md) file so they are all in a single place

Also we would like to say that the requirement for these values is 0 for now for the superchain, with op-succinct you can use them however you would like, then when we do add zk as part of the superchain's proof system we can define the ranges of values. We don't have bandwidth to really think deeply about what the standard values should be from a product perspective and while what you have now could make sense, I don't want to ratify something thru gov that ultimately isnt right

tynes:
This generally looks good to me. Some work started in https://github.com/ethereum-optimism/optimism/pull/12847 to define the config for isthmus that you can build on. cc @vdamle for visibility

tynes:
cc @trianglesphere come leave a review please

specs/protocol/configurability.md
```diff
@@ -441,4 +459,11 @@ configuration of the permissioned dispute game.<br/>
 **Requirement:** No requirement<br/>
 **Notes:** <br/>
 
+### Operator Fee Manager
+
+**Description:** Account authorized to modify the operator fee scalar. <br/>
+**Administrator:** [System Config Owner](#admin-roles)<br/>
```
tynes: The standard value should be `address(0)` so that they cannot be turned on for standard config chains, otherwise its going to require a lot more work with governance and getting the fee split set up for this for superchain production chains

tynes (approved):

specs/protocol/configurability.md
tynes: Good with merge once this is updated

yuwen01: Fixed!

specs/protocol/isthmus/exec-engine.md
```diff
@@ -141,3 +145,42 @@ Withdrawals list in the block body is encoded as an empty RLP list.
...
+
+Where:
+
+- `gasUsed` is amount of gas used by the transaction.
+- `operatorFeeScalar` is a `uint32` scalar set by the chain operator, scaled by `1e6`.
+- `operatorFeeConstant` is a `uint64` scalar set by the chain operator.
+
+#### Configuring Parameters
+
+`operatorFeeScalar` and `operatorFeeConstant` are loaded in a similar way to the `baseFeeScalar` and
```
trianglesphere, line 190: I saw comment somewhere, maybe gone by now, that there would be restrictions on how large the operator fee scalar and fee constant should be set. If this feature becomes a standard chain features I think that this restriction is very important to protect user funds.

specs/protocol/isthmus/l1-attributes.md
refcell: Bump on adding this to the spec to make it clear what this is

refcell (approved):
This looks good, just add a clarifier for the l1 block attributes function signature so it's clear what that value is at the top of the table in `l1-attributes.md`.

yuwen01:
> This looks good, just add a clarifier for the l1 block attributes function signature so it's clear what that value is at the top of the table in `l1-attributes.md`.

Thanks! Just fixed.

specs/protocol/isthmus/exec-engine.md
yuwen01: I think that the operator fee will probably not become part of the superchain config since it adds complexity. Its purpose is more intended for non-standard chains, like op-succinct chains, for example. 

My previous proposal wa the at the constant was between 0 and 600 Gwei and that the scalar was between 0 and 0.5x (basefee + priorityfee).

yuwen01:
@sebastianst If you have a moment could you please review this spec?

tynes:
cc @emhane @vdamle - Would be good to have your eyes on this as its isthmus related work

specs/protocol/isthmus/predeploys.md
```diff
@@ -192,6 +209,46 @@ The following functions are updated to read from the `L1Block` contract by calli
...
+- `isIsthmus`
+
+`operatorFeeScalar` and `operatorFeeConstant` are read from the `L1Block` contract, and `isIsthmus`
+is read directly from storage. If `isIsthmus` is false, then this function MUST return `0`.
+
+```function
+function getOperatorFee(uint256 gasUsed)(uint256)
+```
+
+##### `getL1Fee`
```
K-Ho: If a user wanted to calculate the L1 data fee and the Operator fee separately as this is specified, then...

`l1DataFee=getL1Fee-getOperatorFee`
and `operatorFee=getOperatorFee`
which is strange. I can totally imagine a wallet accidentally thinking that the way to calculate the total non-L2 fee would be to do `getOperatorFee + getL1Fee`
While I also would like to keep as much backwards compatibility as possible, upon thinking about this more, I could imagine this being confusing, since the operatorFee doesn't actually have to have anything to do with L1 costs - for op-succinct it accounts for off-chain proving costs, which are completely separate from L1.

I think it makes more sense to keep getL1Fee as is (just L1 data fees) and have the separate getOperatorFee.

Then total fee is just getL1Fee + getOperatorFee

specs/protocol/isthmus/system-config.md
```diff
@@ -0,0 +1,29 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+The `SystemConfig` is updated to set a new role: the `OPERATOR_FEE_MANAGER`. This role is set once upon
+initialization and can only be set again with another call to `initialize`.
+
+### `Roles`
+
+The `Roles` struct is updated to include the new `OPERATOR_FEE_MANAGER` role.
```
leruaa: @tynes About the possible conflicts in `SystemConfig`, my concern is mainly about that part with the `Roles` struct that don't exist yet, but added in https://github.com/ethereum-optimism/optimism/pull/12932

hashcashier (changes requested):
While the simplicity here is appealing, this method of using gas to meter proving costs won't really work. The underlying assumption here is that gas correlates with proving cost, which doesn't really hold at all, and would lead often to a higher fee than what should be paid if the constant & scalar billing parameters are set to assume the worst-case ratio.

For example, SSTORE costs differ based on the delta from zero to non-zero values, and have gas refunds for clearing values. Cycle counts for a trie update in the stateless zk clients don't really follow these dynamics.

Do you think we can generalize this further into a more customizeable or precise metering approach?

tynes:
@leruaa Do you mind sharing an update of the status of this?

leruaa:
@tynes I think it's ready to be merged, all the feedbacks have been handled.

specs/protocol/isthmus/derivation.md
```diff
@@ -0,0 +1,25 @@
...
+**Table of Contents**
+
+- [Network upgrade automation transactions](#network-upgrade-automation-transactions)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Network upgrade automation transactions
+
+The Isthmus hardfork activation block contains the following transactions, in this order:
+
```
tynes: This section is underspecified. What are the definitions of the network upgrade transactions? The ecotone specs contain examples of this. This is important because the derivation pipeline has to deterministically create these transactions, ie who is the `from`, what is the `to`, what is the `data`? For the contract deployments, we need finalized bytecode, which will need to come from a release

specs/protocol/isthmus/exec-engine.md
```diff
@@ -141,3 +145,42 @@ Withdrawals list in the block body is encoded as an empty RLP list.
...
+New OP stack variants have different resource consumption patterns, and thus require a more flexible
+pricing model. To enable more customizable fee structures, Isthmus adds a new component to the fee
+calculation: the `operatorFee`, which is parameterized by two scalars: the `operatorFeeScalar`
+and the `operatorFeeConstant`.
+
+### Operator Fee
+
+The operator fee, is set as follows:
+
+`operatorFee = (gasUsed * operatorFeeScalar / 1e6) + operatorFeeConstant`
```
hashcashier, line 180:
```suggestion
`operatorFee = (gasLimit * operatorFeeScalar / 1e6) + operatorFeeConstant`
```

specs/protocol/isthmus/exec-engine.md
```diff
@@ -141,3 +145,42 @@ Withdrawals list in the block body is encoded as an empty RLP list.
...
+
+Where:
+
+- `gasUsed` is amount of gas used by the transaction.
```
hashcashier, line 184:
```suggestion
- `gasLimit` is the maximum amount of gas that can be used by the transaction.
```

hashcashier (changes requested):
Suggesting we use the transaction gas limit instead of metered usage to minimize the potential abuse of gas refunds on proving costs.

specs/protocol/isthmus/derivation.md
ratankaliani: Can you send a link to one of the ecotone specs that contains an example of the network upgrade transaction definitions? 

I don't see a clear format in the derivation spec: https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/ecotone/derivation.md#ecotone-blob-retrieval

tynes:
Another question - now that standard L2 genesis is being decoupled from this, what is the state of the implementation of the system config + operator fee vault? how is the `RECIPIENT` set?

specs/protocol/isthmus/derivation.md
tynes: Looks like they are in the wrong place: https://github.com/ethereum-optimism/specs/blob/87e13cba8f34a68858a334fc2587041b712facfa/specs/protocol/derivation.md#network-upgrade-automation-transactions

I can move them to the correct place

tynes:
We need to handle both the upgrade case and the from genesis case. How do we handle the new predeploy in the upgrade case? We have a requirement that the derivation pipeline upgrade transactions are deterministic, its not straight forward to set the `RECIPIENT` due to this constraint. I recommend the network upgrade tx deploys a `FeeVault` with the `RECIPIENT` set to `address(0)` and then chains that upgrade will need to set a new implementation with the `RECIPIENT` that they want. Without standard L2 genesis, there isn't much of a better way to do this

tynes:
> Suggesting we use the transaction gas limit instead of metered usage to minimize the potential abuse of gas refunds on proving costs.

cc @yuwen01 @ratankaliani @leruaa thoughts on this?

tynes:
Opened https://github.com/ethereum-optimism/specs/pull/562 so that you can see what a spec for network upgrade txs look like

leruaa:
@tynes about @hashcashier suggestion, we think that gas used is more representative of proving costs. Furthermore, we are in favor of an incremental approach, rather than updating the current implementation, at the risk of delaying delivery. Especially considering it's possible that Isthmus could be advanced to cycle 35, and that an audit needs to be done.

tynes:
@leruaa I am more concerned about delivery being delayed due to underspecification of key parts that i have commented on without it being addressed in a timely manner

> we think that gas used is more representative of proving costs.

Please explain why

puma314:
The `gas_used` represents the actual amount of gas/computation that is used for processing the transaction. The way that proving works is that every opcode that is processed incurs a certain amount of "cycles" in the zkVM, and so gas_used is a direct measure of the opcodes that are processed. The `gas_limit` is NOT correlated with this, so it is quite obvious to me that `gas_used` is significantly more correlated with proving costs vs. `gas_limit`?

I think @hashcashier should specify what potential "abuse" there might be on gas refunds, as that is not clear to me.

K-Ho:
> The `gas_used` represents the actual amount of gas/computation that is used for processing the transaction. The way that proving works is that every opcode that is processed incurs a certain amount of "cycles" in the zkVM, and so gas_used is a direct measure of the opcodes that are processed. The `gas_limit` is NOT correlated with this, so it is quite obvious to me that `gas_used` is significantly more correlated with proving costs vs. `gas_limit`?
> 
> I think @hashcashier should specify what potential "abuse" there might be on gas refunds, as that is not clear to me.

This makes sense to me.
1. We have agreed this is a v0 of how to accurately charge for zk proving costs, and that we will iterate on it in the future.
2. This feature is growing stale and as long as there is no security risk (which should be reviewed in the FMA review), we should not optimize towards getting this out into the real world and testing it over making the perfect formula on our first try. 

We've iterated many times on the fee formula for accurately charging for L1 data fees, and I guarantee we will continue to make improvements to it

tynes:
> The `gas_used` represents the actual amount of gas/computation that is used for processing the transaction. The way that proving works is that every opcode that is processed incurs a certain amount of "cycles" in the zkVM, and so gas_used is a direct measure of the opcodes that are processed. The `gas_limit` is NOT correlated with this, so it is quite obvious to me that `gas_used` is significantly more correlated with proving costs vs. `gas_limit`?
> 
> I think @hashcashier should specify what potential "abuse" there might be on gas refunds, as that is not clear to me.

See the following EIP for an explanation: https://eips.ethereum.org/EIPS/eip-7778

It is a real thing, given you do not get your electricity and gpu time back in the same way that you get gas back onchain

specs/protocol/isthmus/configurability.md
```diff
@@ -39,10 +40,11 @@ The `ConfigType` enum represents configuration that can be modified.
 | `BASE_FEE_VAULT_CONFIG` | `uint8(1)` | Sets the Fee Vault Config for the `BaseFeeVault` |
 | `L1_FEE_VAULT_CONFIG` | `uint8(2)` | Sets the Fee Vault Config for the `L1FeeVault` |
 | `SEQUENCER_FEE_VAULT_CONFIG` | `uint8(3)` | Sets the Fee Vault Config for the `SequencerFeeVault` |
-| `L1_CROSS_DOMAIN_MESSENGER_ADDRESS` | `uint8(4)` | Sets the `L1CrossDomainMessenger` address |
-| `L1_ERC_721_BRIDGE_ADDRESS` | `uint8(5)` | Sets the `L1ERC721Bridge` address |
-| `L1_STANDARD_BRIDGE_ADDRESS` | `uint8(6)` | Sets the `L1StandardBridge` address |
-| `REMOTE_CHAIN_ID` | `uint8(7)` | Sets the chain id of the base chain |
+| `OPERATOR_FEE_VAULT_CONFIG` | `uint8(4)` | Sets the Fee Vault Config for the `OperatorFeeVault` |
```
tynes: Given we are not combining this with L2 standard genesis, we are going to need to rethink how we go about configuring the system. How does the existing solidity implementation work?

specs/protocol/isthmus/derivation.md
```diff
@@ -16,11 +23,261 @@ The Isthumus hardfork activation block contains the following transactions, in t
...
+```bash
+cast keccak $(cast concat-hex 0x0000000000000000000000000000000000000000000000000000000000000002 $(cast keccak "Isthmus: L1 Block Deployment"))
+# 0x3b2d0821ca2411ad5cd3595804d1213d15737188ae4cbd58aa19c821a6c211bf
+```
+
+Verify `data`:
+
+```bash
+git checkout 905311d64d3df67649837ec57f0b0a4b6214ce31
```
tynes: This is a copypasta - we don't use pnpm anymore. Please prove to me that the bytecode in the txt file is legit

tynes: We are going to want to use a commit of a release. This means we will want to remove any rc suffix on the semver of the specific contracts being upgraded

specs/protocol/isthmus/configurability.md
leruaa: Seems like there was a bad merge in this file, I will have a look.

Currently I added a new method `setOperatorFeeScalars(uint32, uint64)` to `SystemConfig`. But reading the same file in the main branch, I see a section about [Fee Vault Config](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/isthmus/configurability.md#fee-vault-config), so I guess I should conform with that and add a `function setOperatoreFeeVaultConfig(address,uint256,WithdrawalNetwork)` method. Is it what you have in mind?

leruaa: Ah ok, I saw the Fee Vault Config section is part of L2 standard genesis, so I won't need to conform with that.

specs/protocol/isthmus/exec-engine.md
tynes: We are going to go with gas used for now and optimize for avg case rather than worst case, see https://github.com/ethereum-optimism/specs/pull/382#issuecomment-2625644968

specs/protocol/isthmus/predeploys.md
```diff
@@ -67,6 +73,7 @@ graph LR
   BaseFeeVault -- "getConfig(ConfigType.GAS_PAYING_TOKEN)(address,uint256,uint8)" --> L1Block
   SequencerFeeVault -- "getConfig(ConfigType.SEQUENCER_FEE_VAULT_CONFIG)(address,uint256,uint8)" --> L1Block
   L1FeeVault -- "getConfig(ConfigType.L1_FEE_VAULT_CONFIG)(address,uint256,uint8)" --> L1Block
+  OperatorFeeVault -- "getConfig(ConfigType.OPERATOR_FEE_VAULT_CONFIG)(address,uint256,uint8)" --> L1Block
```
tynes: i am confused about this integration into std L2 genesis, will need to look at what the implementation does

specs/protocol/isthmus/derivation.md
leruaa: I updated to `make build-contracts`

specs/protocol/isthmus/predeploys.md
leruaa: I detangled the specs, now there are 2 separate sections in configurability.md:
- Fee vault config
- Operator fee parameters

tynes:
@leruaa lets get https://github.com/ethereum-optimism/specs/pull/567 merged then rebase this on top. will unblock this getting merged

specs/protocol/isthmus/configurability.md
```diff
@@ -39,9 +48,70 @@ The following actions should happen during the initialization of the `SystemConf
...
-TODO: operator fee
+#### Fee Vault Config
+
+For each `FeeVault`, there is a setter for its config. The arguments to the setter include
+the `RECIPIENT`, the `MIN_WITHDRAWAL_AMOUNT` and the `WithdrawalNetwork`.
+Each of these functions should be `public` and only callable by the chain governor.
+
+Each function calls `OptimismPortal.setConfig(ConfigType,bytes)` with its corresponding `ConfigType`.
+
+##### `setBaseFeeVaultConfig`
```
tynes: This still follows the standard L2 genesis style. We should not have `setConfig`. There should be no diff to the `FeeVault`

specs/protocol/isthmus/configurability.md
```diff
@@ -39,9 +48,70 @@ The following actions should happen during the initialization of the `SystemConf
...
+
+```solidity
+function setBaseFeeVaultConfig(address,uint256,WithdrawalNetwork)
+```
+
+##### `setL1FeeVaultConfig`
```
tynes: We should remove this one

specs/protocol/isthmus/configurability.md
```diff
@@ -39,9 +48,70 @@ The following actions should happen during the initialization of the `SystemConf
...
+
+```solidity
+function setL1FeeVaultConfig(address,uint256,WithdrawalNetwork)
```
tynes: We should remove this one

specs/protocol/isthmus/configurability.md
```diff
@@ -39,9 +48,70 @@ The following actions should happen during the initialization of the `SystemConf
...
+```
+
+##### `setSequencerFeeVaultConfig`
```
tynes: We should remove this one

specs/protocol/isthmus/derivation.md
tynes: We should leave the exact commit and bytecode empty for now, we can merge the contracts in their own PR and then cut a release of them to know the exact bytecode

specs/protocol/isthmus/predeploys.md
```diff
@@ -31,4 +30,11 @@ call each getter for the network specific config and set the returndata into sto
```
tynes: Can you delete this fee vault section? There are no changes to that contract

specs/static/bytecode/isthmus-gas-price-oracle-deployment.txt
```diff
@@ -0,0 +1 @@
...006106d5565b81600181146119cb57600281146119d5576119f1565b60019150506106d5565b60ff8411156119e6576119e661184d565b50506001821b6106d5565b5060208310610133831016604e8410600b8410161715611a14575081810a6106d5565b611a1e8383611938565b807fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff04821115611a5057611a5061184d565b029392505050565b6000610d6e8383611999565b6000808212827f7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff03841381151615611a9e57611a9e61184d565b827f8000000000000000000000000000000000000000000000000000000000000000038412811615611ad257611ad261184d565b50500190565b60007f7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff600084136000841385830485118282161615611b1957611b1961184d565b7f80000000000000000000000000000000000000000000000000000000000000006000871286820588128184161615611b5457611b5461184d565b60008712925087820587128484161615611b7057611b7061184d565b87850587128184161615611b8657611b8661184d565b505050929093029392505050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052603260045260246000fd5b60007fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff8203611bf457611bf461184d565b506001019056fea164736f6c634300080f000a
```
tynes: Lets wait until we have a real release to pull this from, meaning we merge the contract code and cut a tag and build from that

specs/protocol/isthmus/configurability.md
```diff
@@ -39,9 +48,70 @@ The following actions should happen during the initialization of the `SystemConf
 - `emit ConfigUpdate.GAS_LIMIT`
```
tynes: We need to add the operator fee params config update type to the table above here

specs/protocol/configurability.md
```diff
@@ -441,4 +459,11 @@ configuration of the permissioned dispute game.<br/>
...
+**Requirement:** `address(0)`<br/>
```
tynes: Looking at this again, the requirement of `address(0)` makes no sense. We cannot require the system config owner to be null. There currently is no such thing as the operator fee manager in the code

```solidity
    function setOperatorFeeScalars(uint32 _operatorFeeScalar, uint64 _operatorFeeConstant) external onlyOwner {
        _setOperatorFeeScalars(_operatorFeeScalar, _operatorFeeConstant);
    }
```

We should delete this section and just make the above 2 be managed by the System Config Owner

specs/protocol/configurability.md
```diff
@@ -283,6 +286,21 @@ contracts deployed on layer 1.<br/>
 **Requirement:** Disabled<br/>
 **Notes:** Simple clear restriction.<br/>
 
+### [Operator Fee Scalar](exec-engine.md#operator-fees)
+
+**Description:** Operator fee scalar -- used to calculate the operator fee<br/>
+**Administrator:** [Operator Fee Manager](#operator-fee-manager)<br/>
```
tynes: See comment [below](https://github.com/ethereum-optimism/specs/pull/382/files#r1938081685), this should be System Config Owner

specs/protocol/isthmus/configurability.md
```diff
@@ -39,9 +45,44 @@ The following actions should happen during the initialization of the `SystemConf
...
 - `emit ConfigUpdate.UNSAFE_BLOCK_SIGNER`
 - `emit ConfigUpdate.EIP_1559_PARAMS`
+- `emit ConfigUpdate.OPERATOR_FEE_PARAMS`
 
 These actions MAY only be triggered if there is a diff to the value.
 
+`ConfigUpdate.OPERATOR_FEE_PARAMS` MAY be emitted. If it is not emitted, the `operatorFeeScalar` and
```
tynes: I recommend removing this from being called in `initialize` see https://github.com/ethereum-optimism/optimism/pull/12166/files#r1938083725

specs/protocol/configurability.md
yuwen01: The idea here is that "standard" chains cannot have an operator fee manager. Maybe the "Requirement" description is a bit misleading here.

But I agree that the operator fee manager isn't super necessary as a role. I can't imagine many situations where you'd need it to be distinct from the system config owner

specs/protocol/isthmus/configurability.md
```diff
@@ -39,9 +45,44 @@ The following actions should happen during the initialization of the `SystemConf
...
+`operatorFeeConstant` are set to 0 by default.
+
+### Modifying Operator Fee Parameters
+
+A new `SystemConfig` `UpdateType` is introduced that enables the modification of
+the `operatorFeeScalar` and `operatorFeeConstant` by the [`OperatorFeeManager`](../configurability.md#operator-fee-manager).
```
tynes: There is no operator fee manager in practice, the code shows the system config owner being the one with permissions. This is fine, lets just move forward with that

specs/protocol/isthmus/configurability.md
```diff
@@ -39,9 +45,44 @@ The following actions should happen during the initialization of the `SystemConf
...
+
+```solidity
+function operatorFeeConstant()(uint64)
+```
+
+##### `setOperatorFeeScalars`
+
+This function sets the `operatorFeeScalar` and `operatorFeeConstant`.
+
+This function MUST only be callable by the [`OperatorFeeManager`](../configurability.md#operator-fee-manager).
```
tynes: https://github.com/ethereum-optimism/specs/pull/382/files#r1938084307

specs/protocol/isthmus/derivation.md
```diff
@@ -17,11 +24,261 @@ The Isthumus hardfork activation block contains the following transactions, in t
...
+`TODO`.
+
+## Operator fee vault deployment
+
+A new `OperatorFeeVault` contract has been created to recieve the operator fees. The contract is created with
+the following arguments:
+
+- Recipient address: The base fee vault
+- Min withdrawal amount: 0
+- Withdrawal network: L2
```
tynes, line 126: Perfect, this is exactly what we want

specs/protocol/isthmus/derivation.md
```diff
@@ -17,11 +24,261 @@ The Isthumus hardfork activation block contains the following transactions, in t
...
+
+## OperatorFeeVault Proxy Update
+
+This transaction updates the GasPriceOracle Proxy ERC-1967 implementation slot to point to the new GasPriceOracle
+deployment.
+
+A deposit transaction is derived with the following attributes:
+
+- `from`: `0x0000000000000000000000000000000000000000`
+- `to`: `0x420000000000000000000000000000000000001B` (Gas Price Oracle Proxy)
```
tynes: typo: gas price oracle proxy? this is operator fee vault proxy

specs/protocol/isthmus/predeploys.md
```diff
@@ -27,8 +25,11 @@ This function is meant to be called once on the activation block of the holocene
...
 call each getter for the network specific config and set the returndata into storage.
 
-### FeeVault
+### OperatorFeeVault
+
+This vault implements `FeeVault`, like `BaseFeeVault`, `SequencerFeeVault`, and `L1FeeVault`.
+No special logic is needed in order to insert or withdraw funds.
 
-TODO
+Its address will be `0x420000000000000000000000000000000000001b`.
```
tynes, line 34: Confirmed in the code that this matches

specs/protocol/isthmus/derivation.md
```diff
@@ -17,11 +24,261 @@ The Isthumus hardfork activation block contains the following transactions, in t
...
+
+```bash
+cast keccak $(cast concat-hex 0x0000000000000000000000000000000000000000000000000000000000000002 $(cast keccak "Isthmus: L1 Block Deployment"))
+# 0x3b2d0821ca2411ad5cd3595804d1213d15737188ae4cbd58aa19c821a6c211bf
+```
+
+Verify `data`:
+
+```bash
+git checkout TODO
```
refcell, line 71: Need to update this `TODO`

specs/protocol/isthmus/derivation.md
```diff
@@ -17,11 +24,261 @@ The Isthumus hardfork activation block contains the following transactions, in t
...
+make build-contracts
+jq -r ".bytecode.object" packages/contracts-bedrock/forge-artifacts/L1Block.sol/L1Block.json
+```
+
+This transaction MUST deploy a contract with the following code hash
+`TODO`.
```
refcell, line 77: Also missing `TODO` here

specs/protocol/isthmus/derivation.md
```diff
@@ -17,11 +24,261 @@ The Isthumus hardfork activation block contains the following transactions, in t
...
+
+```bash
+cast keccak $(cast concat-hex 0x0000000000000000000000000000000000000000000000000000000000000002 $(cast keccak "Isthmus: Gas Price Oracle Deployment"))
+# 0xfc70b48424763fa3fab9844253b4f8d508f91eb1f7cb11a247c9baec0afb8035
+```
+
+Verify `data`:
+
+```bash
+git checkout TODO
```
refcell, line 111: `TODO`

specs/protocol/isthmus/derivation.md
```diff
@@ -17,11 +24,261 @@ The Isthumus hardfork activation block contains the following transactions, in t
...
+make build-contracts
+jq -r ".bytecode.object" packages/contracts-bedrock/forge-artifacts/GasPriceOracle.sol/GasPriceOracle.json
+```
+
+This transaction MUST deploy a contract with the following code hash
+`TODO`.
```
refcell, line 117: `TODO`

specs/protocol/isthmus/derivation.md
```diff
@@ -17,11 +24,261 @@ The Isthumus hardfork activation block contains the following transactions, in t
...
+
+```bash
+cast keccak $(cast concat-hex 0x0000000000000000000000000000000000000000000000000000000000000002 $(cast keccak "Isthmus: Operator Fee Vault Deployment"))
+# 0x107a570d3db75e6110817eb024f09f3172657e920634111ce9875d08a16daa96
+```
+
+Verify `data`:
+
+```bash
+git checkout TODO
```
refcell, line 156: `TODO`

specs/protocol/isthmus/derivation.md
```diff
@@ -17,11 +24,261 @@ The Isthumus hardfork activation block contains the following transactions, in t
...
+make build-contracts
+jq -r ".bytecode.object" packages/contracts-bedrock/forge-artifacts/OperatorFeeVault.sol/OperatorFeeVault.json
+```
+
+This transaction MUST deploy a contract with the following code hash
+`TODO`.
```
refcell, line 162: `TODO`

specs/static/bytecode/isthmus-gas-price-oracle-deployment.txt
```diff
@@ -0,0 +1 @@
+TODO
```
refcell, line 1: ?

specs/static/bytecode/isthmus-l1-block-deployment.txt
```diff
@@ -0,0 +1 @@
+TODO
```
refcell, line 1: ?

specs/static/bytecode/isthmus-operator-fee-deployment.txt
```diff
@@ -0,0 +1 @@
+TODO
```
refcell, line 1: ?

refcell (changes requested):
There's a few `TODO` items here that need to be filled out ASAP

specs/static/bytecode/isthmus-gas-price-oracle-deployment.txt
leruaa: It's because we need a release version

specs/static/bytecode/isthmus-l1-block-deployment.txt
leruaa: Same

specs/static/bytecode/isthmus-operator-fee-deployment.txt
leruaa: Same

specs/protocol/isthmus/derivation.md
leruaa: The code hash need to be from a contract with a release version

specs/protocol/isthmus/derivation.md
leruaa: It need to be a commit with a release tag

specs/protocol/isthmus/derivation.md
leruaa: Fixed

specs/protocol/isthmus/configurability.md
leruaa: Removed

specs/protocol/isthmus/configurability.md
leruaa: Removed

specs/protocol/isthmus/configurability.md
leruaa: Done

specs/protocol/configurability.md
leruaa: Done

specs/static/bytecode/isthmus-gas-price-oracle-deployment.txt
leruaa: Done

specs/protocol/isthmus/derivation.md
leruaa: It need to be a commit with a release tag

specs/protocol/isthmus/derivation.md
leruaa: The code hash need to be from a contract with a release version

specs/protocol/isthmus/derivation.md
leruaa: Same

specs/protocol/isthmus/derivation.md
leruaa: Same

tynes (approved):

tynes:
@leruaa Will merge after the lint error is fixed

PR #384, opened 2024-09-20
SuperchainERC20 redesign

0xParti:
**Description**

This PR introduces changes to `token-bridging.md` and `predeploys.md`. 
The main idea is migrating the `sendERC20` and `relayERC20` functions from the `SuperchainERC20` standard to a new `InteropStandardBridge`. The reasons behind this change are in [this PR to the Design Doc repo](https://github.com/ethereum-optimism/design-docs/pull/83).

### DIFF
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -60,7 +60,7 @@
     - [`Converted`](#converted)
   - [Invariants](#invariants)
   - [Conversion Flow](#conversion-flow)
-- [InteropStandardBridge](#interopstandardbridge)
+- [SuperchainERC20Bridge](#superchainerc20bridge)
   - [Overview](#overview-2)
   - [Functions](#functions-2)
     - [`sendERC20`](#senderc20)
@@ -796,7 +796,7 @@ sequenceDiagram
   L2StandardBridge-->L2StandardBridge: emit Converted(from, to, Alice, amount)
 ```
 
-## InteropStandardBridge
+## SuperchainERC20Bridge
 
 | Constant | Value                                        |
 | -------- | -------------------------------------------- |
@@ -804,7 +804,7 @@ sequenceDiagram
 
 ### Overview
 
-The `InteropStandardBridge` is an abstraction on top of the `L2toL2CrossDomainMessenger`
+The `SuperchainERC20Bridge` is an abstraction on top of the `L2toL2CrossDomainMessenger`
 that facilitates token bridging using interop.
 It has mint and burn rights over `SuperchainERC20` tokens
 as described in the [token bridging spec](./token-bridging.md).
@@ -861,12 +861,12 @@ The following diagram depicts a cross-chain transfer.
 ```mermaid
 sequenceDiagram
   participant from
-  participant L2SBA as InteropStandardBridge (Chain A)
+  participant L2SBA as SuperchainERC20Bridge (Chain A)
   participant SuperERC20_A as SuperchainERC20 (Chain A)
   participant Messenger_A as L2ToL2CrossDomainMessenger (Chain A)
   participant Inbox as CrossL2Inbox
   participant Messenger_B as L2ToL2CrossDomainMessenger (Chain B)
-  participant L2SBB as InteropStandardBridge (Chain B)
+  participant L2SBB as SuperchainERC20Bridge (Chain B)
   participant SuperERC20_B as SuperchainERC20 (Chain B)
 
   from->>L2SBA: sendERC20To(tokenAddr, to, amount, chainID)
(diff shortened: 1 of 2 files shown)


0xParti:
With the change of structure, there is room for adding additional functionalities to the `InteropStandardBridge`. I could not think of any crucial one, but happy to hear from your end.

specs/interop/predeploys.md
```diff
@@ -785,6 +796,124 @@ sequenceDiagram
...
+    If the target chain includes it later, these could be unlocked.
+- Freedom of movement: Users should be able to send and receive tokens in any target
+  chain with the initiating chain in its dependency set
+  using `sendERC20()` and `relayERC20()`, respectively.
+- Unique Messenger: The `sendERC20()` function must exclusively use the `L2toL2CrossDomainMessenger` for messaging.
+  Similarly, the `relayERC20()` function should only process messages originating from the `L2toL2CrossDomainMessenger`.
+- Unique Address: The `sendERC20()` function must exclusively send a message
+  to the same address on the target chain.
+  Similarly, the `relayERC20()` function should only process messages originating from the same address.
+  - Note: The `Create2Deployer` preinstall and the custom Factory will ensure same address deployment.
```
skeletor-spaceman: > The `Create2Deployer` preinstall 

could we link to this pre-install?

specs/interop/token-bridging.md
```diff
@@ -86,87 +75,52 @@ The following diagram depicts a cross-chain transfer.
...
+  L2SBB-->L2SBB: emit RelayedERC20(tokenAddr, from, to, amount, source)
 ```
 
 ## Implementation
 
-An example implementation that depends on deterministic deployments across chains
-for security is provided.
-This construction builds on top of the [L2ToL2CrossDomainMessenger][l2-to-l2]
+An example implementation for the `sendERC20` and `relayERC20` functions is provided.
+This construction builds on top of the
```
skeletor-spaceman: missing text?

tynes:
I kind of think that we should remove the interface portion of superchain erc20 and just make the requirements be that the bridge contract has mint/burn rights and the token address is the same across all chains. We don't even really need the deterministic addresses, it just makes accounting way easier

0xParti:
> I kind of think that we should remove the interface portion of superchain erc20 and just make the requirements be that the bridge contract has mint/burn rights and the token address is the same across all chains. We don't even really need the deterministic addresses, it just makes accounting way easier

@tynes Yes, thats what I did on this last PR. Let me know if I wasn't clear enough.
For same address, agree. Its not a hard requirement, but it simplifies things a lot, as we wont need to create or maintain an access control list cross-chain, aka a registry that knows what token address to mint in chain B based on the burnt token in chain A.

tynes (approved):

skeletor-spaceman (approved):

specs/interop/token-bridging.md
```diff
@@ -24,60 +18,55 @@
...
-Unlike other standards, such as [xERC20](https://www.xerc20.com/),
-where users interact with a bridge possessing mint and burn
-privileges on cross-chain enabled tokens, this approach allows users to call methods directly on the token contract.
+## `SuperchainERC20` standard
 
-## Interface
+The standard will build on top of ERC20 and include the following properties:
 
-### Functions
+1. Give `mint` and `burn` rights to the `SuperchainERC20Bridge`.
```
simplyoptimistic, line 29: Is it possible to make the mint / burn requirements wrt rate limits more explicit (for clarity around xerc20-superchain implementations)? 

Specifically, I think it would help if this states whether infinite mint/burn permissions are necessary or if rate limited as per the vanilla xerc20 implementation is fine. If it doesn't matter then I think this should still be stated for clarity.

tynes: I don't think that we should enshrine any particular logic in the bridge about rate limiting. There isn't a universal way of doing rate limits that will solve every usecase, so any rate limits should be done in the token contracts themselves. It is a good callout that the spec could mention rate limiting implemented in the token or another contract

simplyoptimistic: Yeah that's fine. Just wanted to make it crystal clear whether a rate limited erc20 was compatible with superchainerc20 or not. And it sounds like the answer is that it is compatible.

skeletor-spaceman: it should be, yes, and let's say the rate limit is exceeded, you'd have 2 options.
1) allow for replaying the message when the rate limit enables it.
2) expire the message and trigger a "rollback" interop message, which would re-mint the funds back on origin.

tynes (approved):

tynes:
Amazing work @0xParticle !

PR #391, opened 2024-09-25
cannon: Add spec for new multithreaded Cannon VM

### LINKED ISSUE #11664: Finalize MTCannon Specification
Depends on: https://github.com/ethereum-optimism/optimism/issues/11644

Finalize the Multithreaded Cannon (MTCannon) [draft spec](https://www.notion.so/oplabs/Multithreaded-Cannon-8a2e8079764a46f094804436b123332f) and update the specs repo.  While we are supporting both MT and original Cannon implementations, we should keep the existing [Cannon spec](https://specs.optimism.io/fault-proof/cannon-fault-proof-vm.html) and add an additional spec for MTCannon.

mbaxter:
**Description**

Add a spec for the new multithreaded Cannon VM.  This is currently filed under "Experimental" because it is not approved / released yet.  

**For Reviewers**

The first commit copies the existing Cannon VM spec.  Exclude this commit to view the diff. 

**Metadata**

Fixes https://github.com/ethereum-optimism/optimism/issues/11664


specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -0,0 +1,511 @@
...
+
+#### Robustness
+
+To ensure no unexpected behaviors are triggered, MTCannon will now raise an exception if unrecognized syscalls are
+encountered during program execution.
+
+## Multithreading
+
+The MTCannon FPVM rotates between threads to provide
+[concurrent execution](https://en.wikipedia.org/wiki/Concurrent_computing) rather than
```
Inphi: Maybe refer to this as [_multitasking_](https://en.wikipedia.org/wiki/Computer_multitasking) which is similar to concurrent execution, but is a more appropriate concept at this level of abstraction.

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -0,0 +1,511 @@
...
+The VM must read/write the maximum amount of bytes possible without crossing the input address alignment boundary.
+For example, the effect of a write request for a 3-byte aligned buffer must be exactly 3 bytes.
+If the buffer is misaligned, then the VM may write less than 3 bytes depending on the size of the misalignment.
+
+## Exceptions
+
+The FPVM may raise an exception rather than output a post-state to signal an invalid state
+transition. Nominally, the FPVM must raise an exception in at least the following cases:
+
+- Invalid instruction (either via an invalid opcode or an instruction referencing registers
```
Inphi, line 506: Let's expand this list to include the invalid thread states.

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -0,0 +1,511 @@
...
+the left stack is "active", and the right is "inactive".
+
+Representing the set of threads as two stacks allows for a succinct commitment to the contents of all threads.
+For details, see [“Thread Stack Hashing”](#thread-stack-hashing).
+
+### Thread Traversal Mechanics
+
+Threads are traversed deterministically by moving from the first thread to the last thread,
+then back to the first thread repeatedly.  For example, given the set of threads: {0,1,2,3},
+the FPVM would traverse to each as follows: 0, 1, 2, 3, 2, 1, 0, 1, 2, 3, 2, ….
```
Inphi: each traversal visits the last item in the stack twice. So this should be:
```suggestion
the FPVM would traverse to each as follows: 0, 1, 2, 3, 3, 2, 1, 0, 0, 1, 2, 3, 3, 2, ….
```

Worth clarifying why this happens.

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -0,0 +1,511 @@
...
+value:
+
+- An empty stack is represented by the value:
+  - `c0 = hash(bytes32(0) ++ bytes32(0))`
+- To push a thread to the stack, hash the concatenation of the current stack commitment with the thread hash:
+  - `c1 = hash(c0 ++ hash(el0))`.
+- To push another thread:
+  - `c2 = hash(c1 ++ hash(el1))`.
+- To pop an element from the stack, peel back the last hash operation:
+  - `c3 = c1`
```
Inphi: it's not clear that we're popping `c2` in the `c3` example. Consider using a function notation for clarity:
```
push(c1, el1) => c2 = hash(c1 ++ hash(el1))
pop(c2) => c3 = c1
```

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -0,0 +1,511 @@
...
+1. `heap` - 32-bit base address of the most recent memory allocation via mmap.
+1. `llReservationActive` - 8-bit boolean indicator of whether a memory reservation,
+   which is reserved via a Load Linked Word (`ll`) instruction, is active.
+1. `llAddress` - 32-bit address of the currently active memory reservation if one exists.
+1. `llOwnerThread` - 32-bit id of the thread that initiated the current memory reservation if one exists.
+1. `exitCode` - 8-bit exit code.
+1. `exited` - 8-bit boolean valuel indicating whether the VM has exited.
+1. `step` - 64-bit step counter.
+1. `stepsSinceLastContextSwitch` - 64-bit step counter that tracks the number of steps executed on the current
+   thread since the last [context switch](https://en.wikipedia.org/wiki/Context_switch).
```
Inphi: Thinking we should rename this to `stepsSinceLastPreemption` since preemption is identical to context switch here. Or at least we should use "preemption" to describe this field.

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -0,0 +1,511 @@
...
+| 4005 | open          | 🚫              | 🚫               | 🚫           | 🚫               | Sets errno to `0x9`.                                                                                                                                                                                                                                       |
+| 4166 | nanosleep     | 🚫              | 🚫               | 🚫           | 🚫               | Preempts the active thread and returns 0.                                                                                                                                                                                                                  |
+| 4263 | clock_gettime | uint32 clock_id | uint32 addr      | 🚫           | 🚫               | Supports `clock_id`'s `REALTIME`(0) and `MONOTONIC`(1). For other `clock_id`'s, sets errno to `0x16`.  Calculates a deterministic time value based on the state's `step` field.  Seconds are set at memory address `addr` and nsecs are set at `addr + 4`. |
```
Inphi: This needs to explain how the steps are converted to seconds via `HZ`.

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -0,0 +1,511 @@
...
+  - [Compiler Assumptions](#compiler-assumptions)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+This is a description of the second iteration of the Cannon Fault Proof Virtual Machine (FPVM).
+When necessary to distinguish this version from the initial implementation,
+it can be referred to as Multithreaded Cannon (MTCannon).  Like the original Cannon implementation,
+the MTCannon FPVM emulates a minimal Linux-based system running on big-endian 32-bit MIPS32 architecture.
```
Inphi: Emphasize that it's a _uniprocessor_  Linux-based system. The most frequent question I get regarding MT-Cannon is whether it executes threads in parallel.

Inphi (commented):
Layout and wording looks good overall.

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -0,0 +1,511 @@
...
+Thread management and scheduling are typically handled by the
+[operating system (OS) kernel](https://en.wikipedia.org/wiki/Kernel_%28operating_system%29):
+programs make thread-related requests to the OS kernel via [syscalls](https://en.wikipedia.org/wiki/System_call).
+As such, this implementation includes a few new Linux-specific thread-related [syscalls](#syscalls).
+Additionally, the [FPVM state](#fpvm-state) has been modified in order to track the set of active threads
+and thread-related global state.
+
+#### Robustness
+
+To ensure no unexpected behaviors are triggered, MTCannon will now raise an exception if unrecognized syscalls are
```
tynes: This hints that cannon did not do this, without context it may not make sense, could be good to be explicit about the cannon behavior

tynes:
This is amazing, good work @mbaxter!

specs/experimental/cannon-fault-proof-vm-mt.md
mbaxter: good catch!

specs/experimental/cannon-fault-proof-vm-mt.md
mbaxter: Yeah - good point.  We should clean this up.

mbaxter: Added a ticket [here](https://github.com/ethereum-optimism/optimism/issues/12147) just to make sure we don't forget.  But I can probably push that change up real quick.

specs/experimental/cannon-fault-proof-vm-mt.md
mbaxter: Updated by adding:
```
- Invalid thread state:
  - There are no threads - both thread stacks are empty.
  - The active thread stack is empty.
```

The first bullet is kind of redundant given the second bullet, but we do explicitly check for it in MIPS2.sol.

Also wonder if we should make that [MIPS2.sol check](https://github.com/ethereum-optimism/optimism/blob/3210a8c6bb7f98639213fb49718919d84aa8774a/packages/contracts-bedrock/src/cannon/MIPS2.sol#L165-L167) stricter by only verifying that the active thread stack is not empty?

Inphi: Yup, we should add that check. Created an [issue](https://github.com/ethereum-optimism/optimism/issues/12160) for it.

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -0,0 +1,557 @@
...
+| 4166 | nanosleep     | 🚫              | 🚫               | 🚫           | 🚫               | Preempts the active thread and returns 0.                                                                                                                                                                                                                                                                                                                                                                                                                                                           |
+| 4263 | clock_gettime | uint32 clock_id | uint32 addr      | 🚫           | 🚫               | Supports `clock_id`'s `REALTIME`(0) and `MONOTONIC`(1). For other `clock_id`'s, sets errno to `0x16`.  Calculates a deterministic time value based on the state's `step` field and a constant `HZ` (10,000,000) where `HZ` represents the approximate clock rate (steps / second) of the FPVM: $$\text{seconds} = \text{step} / \text{HZ}$$ $$ \text{nsecs} = \frac{(\text{step} \% \text{HZ}) \times 10^9}{\text{HZ}} $$ Seconds are set at memory address `addr` and nsecs are set at `addr + 4`. |
```
Inphi: The LaTeX math expression isn't displayed correctly here. Try using `$...$` instead.

mbaxter: How are you previewing it?  I think it looks ok on my machine (running `just serve`):
Screenshot 2024-09-30 at 4 51 34 PM

Inphi: Ah. Maybe it only works on the webapp. I'm previewing it on github - https://github.com/ethereum-optimism/specs/blob/cd6ef8e7d96cfba604daab621bbddddb72220e77/specs/experimental/cannon-fault-proof-vm-mt.md#supported-syscalls

Inphi: We can merge it if it only works on specs.optimism.io. But worth fixing it on gh too.

mbaxter: I couldn't get alternative LaTeX expressions rendering correctly.  Switched to useing a `code` style - wdyt?:

Screenshot 2024-10-01 at 3 00 39 PM

Inphi: LGTM. Thanks!

Inphi (approved):

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -0,0 +1,558 @@
...
+  - If the currently active thread is not a match:
+    - The active thread is preempted.
+    - If the right thread stack is now empty:
+      - This means all threads have been visited (the traversal begins by moving
+        left, then right so this is the end of the traversal).
+      - The wakeup traversal completes <sup>*</sup>.
+
+> **<sup>*</sup>** Wakeup traversal is completed by setting the FPVM state's `wakeup` field to `0xFFFFFFFF` (-1),
+> causing the FPVM to resume normal execution.
+
```
ajsutton: Nit: We could use the footnote support tin markdown for this.

```suggestion
      - The wakeup traversal completes[^traversal-completion].

[^traversal-completion]: Wakeup traversal is completed by setting the FPVM state's `wakeup` field to `0xFFFFFFFF` (-1),
causing the FPVM to resume normal execution.

```

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -0,0 +1,558 @@
...
+### Exited Threads
+
+When the VM encounters an active thread that has exited, it is popped from the active thread stack, removing it from
+the VM state.
+
+### Waiting Threads
+
+Threads enter a waiting state when a futex wait syscall is successfully executed, setting the thread's
+`futexAddr`, `futexVal`, and `futexTimeout` fields according to the futex syscall arguments.
```
ajsutton: Should this be:
```suggestion
`futexAddr`, `futexVal`, and `futexTimeoutStep` fields according to the futex syscall arguments.
```

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -0,0 +1,558 @@
...
+For example, an on-chain FPVM that relies on pre-supplied merkle proofs for memory access may
+raise an exception if the supplied merkle proof does not match the pre-state `memRoot`.
+
+## Security Model
+
+### Compiler Correctness
+
+Cannon is designed to prove the correctness of a particular state transition that emulates a MIPS32 machine.
+Cannon does not guarantee that the MIPS32 instructions correctly implement the program that the user intends to prove.
+As a result, Cannon's use as a Fault Proof system inherently depends on some extent on the correctness of the compiler
```
ajsutton:
```suggestion
As a result, Cannon's use as a Fault Proof system inherently depends to some extent on the correctness of the compiler
```

ajsutton (approved):
LGTM.

specs/experimental/cannon-fault-proof-vm-mt.md
mbaxter: good call - wasn't aware of this syntax

PR #392, opened 2024-09-26
fix: correct p2p block signature data

ncitron:
**Description**

Fixes an ambiguity in how the V3 p2p payloads are signed. This change makes it match the current implementation of op-node.


protolambda (approved):

protolambda:
Line length is too long. Lint is failing. Will try and push a fix commit

PR #393, opened 2024-09-26
docs: adding opcm mermaid diagram

blmalone:
This is a draft implementation of the OPCM mermaid diagram for specs. The feature is still currently under development so it is subject to change. 

Note: could split this into three diagrams.


blmalone:
Closing this as we have new docs about OPCM.

PR #395, opened 2024-09-27
Fix link to OptimismPortal

p-offtermatt:
**Description**

This is just a minor PR to fix a broken link in the spec that I stumbled into while going through the spec.

I didn't see anything encouraging or discouraging tiny PRs like this, so thought I might as well fix after I noticed the dead link :)


tynes:
Great fix! Can definitely merge this given CI passes, looks like some sort of linting change, likely wrong version of something

p-offtermatt:
Thanks, I fixed the failure :)

tynes (approved):

PR #396, opened 2024-09-27
Isthmus: L2 Withdrawals root spec updates

### LINKED ISSUE #394: Isthumus: Spec updates for L2 withdrawals-root in block header
Peeling off the work associated with updating specs mentioned in: https://github.com/ethereum-optimism/optimism/issues/12044

General notes: keep the specs categorized in the `isthumus ` directory, and file names consistent across upgrades:
e.g. Execution layer changes go `exec-engine.md`.

- [x] Specs note that the block-body withdrawals list is encoded as an empty RLP list.
- [x] Specs note that any block-after-transaction simulation should include an empty withdrawals-root;
  the block is sealed at the end of a block.
- [x] Specs note that the withdrawals-root in the genesis block,
  if Isthumus is active, is always the empty withdrawals root, regardless of L2 state.
- [x] Specs note that the withdrawals-root starting at shanghai, pre-Isthumus,
  is set to an empty withdrawals list, this equals the same root as an empty storage root.
  The withdrawals are there in L2 state, just not elevated into the withdrawals-root.
  Pre-Isthumus output-root construction should be careful not to use the header withdrawals-root.
- [x] Specs note that at the time of state-processing, with an in-progress header,
  the withdrawals-root should not be available to the EVM / application layer.
- [x] Specs Engine-API modification: `ExecutableData` gets a `withdrawalsRoot` attribute. 
- [x] Specs note that for an accurate storage-root to be known,
  the block state has to be committed first, before the withdrawals storage root can be set accurately.
  This attribute should thus be set right after setting the state-root attribute of the header.
- [x] Specs EL P2P note: block-bodies and block-headerts responses are handled the same in isolation:
  we compute the body-withdrawal-hashes from the received withdrawals list in the block body.
  Then, when we compare the body-withdrawal-hashes to the header withdrawal-hashes,
  and if optimism Isthumus is active for the given header, we check if the body-withdrawal-hash matches the empty list root,
  while allowing the header-withdrawal-hash to be any non-null value.
  By verifying at this later stage, we have the header timestamp, and can apply the fork logic conditionally.
- [x] Specs update to describe CL `ExecutionPayload` SSZ encoding with new withdrawals-root included.
    - ~[ ] We remove the `withdrawals` list from the encoding, diverging from the L1 beacon-chain `ExecutionPayload` encoding~.
- [x] Spec that we continue to use `engine_newPayloadV3` engine-API,
  but now with the additional `ExecutionPayload` attribute, that is omitted pre-Isthumus.
- [x] Specs update to describe new CL P2P gossip topic, with updated `ExecutionPayload` SSZ encoding,
  and a gossip-validation rule that checks if the withdrawals root is non-nil, and withdrawals list is empty.
  This spec change goes into the main `rollup-node-p2p.md`, to be consistent with the other gossip topics.
- [x] Update fault-proof-program spec that describes the output-claim verification,
  and link to the Isthumus specs change of the block header.

vdamle:
**Description**

Spec Updates for Isthumus: L2 Withdrawals root spec updates

**Tests**

None

**Metadata**

- Fixes #394


specs/protocol/holocene/exec-engine.md
```diff
@@ -21,6 +29,10 @@ After the Holocene hardfork's activation, the L2 block header's `withdrawalsRoot
 insertions and deletions have been applied to the trie. In other words, the storage root should be the same root
 that is returned by `eth_getProof` at the given block number.
```
tynes: A good way to make this very clear is saying that its the storage root corresponding to the post state, exactly how the post global state root is included in the block header

specs/protocol/holocene/exec-engine.md
```diff
@@ -59,6 +71,29 @@ places a burden on users of the system in a post-fault-proofs world, where:
...
+at such a block's header, we expect that the computed withdrawal hash must match the hash of an empty list of withdrawals.
+The header `withdrawalsRoot` hash can be any non-null value in this case. Verification is performed after a header is available
+so that the header's timestamp can be checked to see if Holocene is active.
+
+#### Backwards Compatibility Considerations
+
+Beginning at Shanghai and prior to Holocene activation, the `withdrawalsRoot` field is set to an empty withdrawals list.
+This is the same root as an empty storage root. The withdrawals are captured in the L2 state, however they are not
+reflected in the `withdrawalsRoot`. Any implementation that caclulates output root should be careful not to use the
+header `withdrawalsRoot`.
```
tynes: Should be careful to only use the withdrawals root after holocene is active? which can be determined if its not empty

tynes:
Nice work on this so far!

specs/protocol/holocene/exec-engine.md
```diff
@@ -59,6 +71,29 @@ places a burden on users of the system in a post-fault-proofs world, where:
 Placing the [`L2ToL1MessagePasser`][l2-to-l1-mp] account storage root in the `withdrawalsRoot` field alleviates this burden
 for users and protocol participants alike, allowing them to propose and verify other proposals with lower operating costs.
 
+#### Genesis Block
+
+If Holocene is active at genesis block, the withdrawals root is the empty withdrawals root, regardless of L2 state.
```
protolambda: at the genesis block

missing "the"

specs/protocol/holocene/exec-engine.md
```diff
@@ -59,6 +71,29 @@ places a burden on users of the system in a post-fault-proofs world, where:
...
+#### P2P
+
+During sync, the block body withdrawal hash is computed from the withdrawals list in the block body. If Holocene is active
+at such a block's header, we expect that the computed withdrawal hash must match the hash of an empty list of withdrawals.
+The header `withdrawalsRoot` hash can be any non-null value in this case. Verification is performed after a header is available
+so that the header's timestamp can be checked to see if Holocene is active.
+
+#### Backwards Compatibility Considerations
+
+Beginning at Shanghai and prior to Holocene activation, the `withdrawalsRoot` field is set to an empty withdrawals list.
```
protolambda: not set to an empty list, but set to the MPT root of an empty list

specs/protocol/holocene/exec-engine.md
protolambda: If the smart contract storage is empty, it matches the empty withdrawals list, even though it represents something different pre-holocene.

And pre-holocene, even though the withdrawals root is present, and a valid MPT root in the block header, it shouldn't be used.

And post-holocene, if there was never any withdrawal contract storage, a MPT root equal to one of an empty list, can be valid to use.

Not sure how to best word a warning for these possible easy things to mistake, but I think it's worth mentioning something to this extent.

protolambda (commented):
Great start

specs/protocol/holocene/exec-engine.md
vdamle: added this on at the end of the current description.

specs/protocol/rollup-node-p2p.md
```diff
@@ -277,6 +282,11 @@ A block is structured as the concatenation of:
   - `signature`: A `secp256k1` signature, always 65 bytes, `r (uint256), s (uint256), y_parity (uint8)`
   - `parentBeaconBlockRoot`: L1 origin parent beacon block root, always 32 bytes
   - `payload`: A SSZ-encoded `ExecutionPayload`, always the remaining bytes.
+- V4 topic
+  - `signature`: A `secp256k1` signature, always 65 bytes, `r (uint256), s (uint256), y_parity (uint8)`
+  - `parentBeaconBlockRoot`: L1 origin parent beacon block root, always 32 bytes
```
vdamle, line 287: Question: Is my assumption correct that this is an additive list?

specs/protocol/rollup-node-p2p.md
```diff
@@ -403,6 +416,8 @@ Implementations may opt for a different limit, since this sync method is optiona
   matching the `ExecutionPayload` SSZ definition of the L1 Merge, L2 Bedrock and L2 Regolith, L2 Canyon versions.
 - `1`: SSZ-encoded `ExecutionPayloadEnvelope` with Snappy framing compression,
   matching the `ExecutionPayloadEnvelope` SSZ definition of the L2 Ecotone version.
+- `2`: SSZ-encoded `ExecutionPayload` with Snappy framing compression,
```
vdamle, line 419: From the proposed changes [here](https://github.com/ethereum-optimism/op-geth/pull/383/files#diff-50e0dca0bd0ac667ca6a4a3b762c9346030f346fdd9b48f3f83be675a43423eaR91-R94) and as further elaborated [here](https://github.com/ethereum-optimism/optimism/issues/12044#issuecomment-2369618003), the `ExecutionPayload` (and not `ExecutionPayloadEnvelope`) will be extended with `withdrawalsRoot`

specs/protocol/rollup-node-p2p.md
protolambda: the block header RLP list? Kind of, attributes have been reused / changed meaning in the past, but it's meant to be backward compatible.
The SSZ version that is used for Gossip, with the `parentBeaconBlockRoot` added to it, is already non-standard, and not strictly speaking SSZ anymore (due to a missing offset value).

specs/protocol/isthmus/exec-engine.md
```diff
@@ -24,22 +33,29 @@ The storage root of the `L2ToL1MessagePasser` is included in the block header's
 
 Isthmus, like other network upgrades, is activated at a timestamp.
 Changes to the L2 Block execution rules are applied when the `L2 Timestamp >= activation time`.
+Changes to the L2 block header are applied when it is considering data from a L1 Block whose timestamp
+is greater than or equal to the activation timestamp.
```
ajsutton: I'm not sure what this actually means.  Is this when the L2 block's origin is an L1 block with timestamp > activation or if the L2 block came from a batch included on L1 after activation time (and is that when the channel starts or ends?).

Also, why would different parts of the L2 block activate Isthmus at different times?

specs/protocol/isthmus/exec-engine.md
```diff
@@ -24,22 +33,29 @@ The storage root of the `L2ToL1MessagePasser` is included in the block header's
...
 
 ## `L2ToL1MessagePasser` Storage Root in Header
 
-After Holocene's activation, the L2 block header's `withdrawalsRoot` field will consist of the 32-byte
+After Isthumus hardfork's activation, the L2 block header's `withdrawalsRoot` field will consist of the 32-byte
 [`L2ToL1MessagePasser`][l2-to-l1-mp] account storage root _after_ the block has been executed, and _after_ the
 insertions and deletions have been applied to the trie. In other words, the storage root should be the same root
-that is returned by `eth_getProof` at the given block number.
+that is returned by `eth_getProof` at the given block number -- it is the storage root of the post state,
+similar to how the state root of the post state is stored in the block header.
```
ajsutton: Would it be simpler to say this should be the account store root from the world state identified by the stateRoot field in the block header?

specs/protocol/isthmus/exec-engine.md
```diff
@@ -24,22 +33,29 @@ The storage root of the `L2ToL1MessagePasser` is included in the block header's
...
+
+For an accurate storage root to be determined, the block state has to be committed first and only then, the
+`withdrawalsRoot` can be set accurately. Hence the `withdrawalsRoot` attribute should be set right after setting
+the state root attribute in the header.
```
ajsutton: This seems like an implementation detail which isn't necessarily true. The withdrawals root could easily be set once its final state is known even if hashing some other part of the world state trie is ongoing to calculate the overall state root.

specs/protocol/isthmus/exec-engine.md
```diff
@@ -63,6 +79,33 @@ places a burden on users of the system in a post-fault-proofs world, where:
 Placing the [`L2ToL1MessagePasser`][l2-to-l1-mp] account storage root in the `withdrawalsRoot` field alleviates this burden
 for users and protocol participants alike, allowing them to propose and verify other proposals with lower operating costs.
 
+#### Genesis Block
+
+If isthumus is active at the genesis block, the withdrawals root is the empty withdrawals root, regardless of L2 state.
```
ajsutton: Why wouldn't this be set to match the actual L2 state? Given genesis is the initial safe head having this be inconsistent could lead to creating invalid proposals.

specs/protocol/isthmus/exec-engine.md
```diff
@@ -63,6 +79,33 @@ places a burden on users of the system in a post-fault-proofs world, where:
...
+
+At the time of state processing, the header for which transactions are being validated should not make it's `withdrawalsRoot`
+available to the EVM/application layer.
+
+#### P2P
+
+During sync, the block body withdrawal hash is computed from the withdrawals list in the block body. If Isthumus is
+active at such a block's header, we expect that the computed withdrawal hash must match the hash of an empty list of
+withdrawals. The header `withdrawalsRoot` MPT hash can be any non-null value in this case. Verification is performed
+after a header is available so that the header's timestamp can be checked to see if Isthumus is active.
```
ajsutton: Do we need to be clear about what this validation is and exactly when it is performed? ie does a node have to go back and recheck the header of every block after sync completes or just performs that check as each header is received?

Actually, how does a node do that verification if it's snap syncing given it won't have the world state? I _think_ just verifying the final block is enough since it's parent hash commits to the previous world states and the withdrawalsRoot header and snap sync assumes that the chain it downloads has valid execution of transactions (since it skips executing them).

specs/protocol/isthmus/exec-engine.md
```diff
@@ -76,3 +119,24 @@ an outbound withdrawal for a long period of time, the node may not have access t
 [`L2ToL1MessagePasser`][l2-to-l1-mp]. In this case, the client would be unable to keep consensus. However, most modern
 clients are able to at the very least reconstruct the account storage root at a given block on the fly if it does not
 directly store this information.
+[l2-to-l1-mp]: ../../protocol/predeploys.md#L2ToL1MessagePasser
+[output-root]: ../../glossary.md#l2-output-root
+
+##### Transaction Simulation
+
+An empty withdrawals root should be included at the end of all applied simulated transactions, when the block is sealed.
```
ajsutton: Why is this empty and not the actual root?  Is the state root field not updated for the simulated transactions?

Also, what does simulated transactions mean in this context? I'm assuming for some RPC calls but we probably should be clear about the context here.

specs/protocol/isthmus/exec-engine.md
```diff
@@ -76,3 +119,24 @@ an outbound withdrawal for a long period of time, the node may not have access t
...
+
+## Block Body Withdrawals List
+
+withdrawals list in the block body is encoded as an empty RLP list.
```
ajsutton:
```suggestion
Withdrawals list in the block body is encoded as an empty RLP list.
```

ajsutton (commented):
Looks good generally. I've left a bunch of comments but primarily aimed at ensuring we're being really clear in the spec rather than disagreeing with the design (apart from the withdrawalRoot for genesis which seems weird to me but maybe I'm missing something).

specs/protocol/isthmus/exec-engine.md
tynes: This is true, if the world state is being built concurrently, so its an implementation detail. The most important thing is that its the post state

specs/protocol/isthmus/exec-engine.md
tynes: This saves some effort when building the L2 genesis. If its going to break the proposer, its definitely something that we can support. There is not technically much utility about making a claim about the genesis state to facilitate a withdrawal since there are none in the genesis state of a standard config chain

specs/protocol/isthmus/exec-engine.md
tynes: Verifying the final block is indeed enough, but this is definitely a weird edge case around reorging that I am not exactly sure how snapsync handles

specs/protocol/isthmus/exec-engine.md
```diff
@@ -63,6 +84,33 @@ places a burden on users of the system in a post-fault-proofs world, where:
...
+#### P2P
+
+During sync, the block body withdrawal hash is computed from the withdrawals list in the block body. If Isthmus is
+active at such a block's header, we expect that the computed withdrawal hash must match the hash of an empty list of
+withdrawals. The header `withdrawalsRoot` MPT hash can be any non-null value in this case. Verification is performed
+after a header is available so that the header's timestamp can be checked to see if Isthmus is active.
+
+#### Backwards Compatibility Considerations
+
+Beginning at Shanghai and prior to Isthmus activation, the `withdrawalsRoot` field is set to the MPT root
```
tynes: Which OP Stack is shanghai in? Could be better to use the OP Stack fork name

specs/protocol/isthmus/exec-engine.md
tynes: The withdrawals root isn't in EVM context? Transaction simulation via `eth_call` doesn't build a block?

specs/protocol/isthmus/exec-engine.md
vdamle: I believe it is the former -- the L2 block's origin is an L1 block with timestamp > activation.

> Also, why would different parts of the L2 block activate Isthmus at different times?

Definitely didn't mean to imply the above. Also not sure where I've implied the above! Please let me know and I'll fix it.

specs/protocol/isthmus/exec-engine.md
vdamle: Thanks for the suggestion, that definitely sounds more straight forward. Will make the change

specs/protocol/isthmus/exec-engine.md
vdamle: Ah, makes sense! I've removed this.

specs/protocol/isthmus/exec-engine.md
ajsutton: It's not so much that there's utility in proposing for the genesis block, but that this breaks two invariants:
1. That starting from Isthmus the withdrawalsRoot header is the `L2ToL1MessagePasser` storage root
2. That the safe head can always be used to produce a valid proposal

Everywhere that reads the withdrawalRoot would need to have special case handling for genesis. So I'd be very inclined to have the extra state read in L2 genesis generation to make all other cases simple.

ajsutton: As an example of the bugs that come from this special case - this PR adds a paragraph to `proposal.md` that says the withdrawalsRoot can be used post-Isthmus but forgets the genesis special case where it can't.

specs/protocol/isthmus/exec-engine.md
vdamle: Good point - Per [this issue](https://github.com/ethereum-optimism/optimism/issues/6728) and [blog post](https://blog.oplabs.co/canyon-hardfork/) , it is in `Canyon` hard fork. Modified to specify Canyon.

specs/protocol/isthmus/exec-engine.md
protolambda: The withdrawal-contract genesis storage-root will keep changing, and having many special genesis values to put in the block-header makes things complicated. It couples the state content and block-header generation to tightly. We should just default the withdrawals-root in the genesis block to an empty withdrawals MPT root. I think not being allowed to withdraw from the genesis block is fine, as there aren't supposed to be withdrawals anyway.

specs/protocol/isthmus/exec-engine.md
vdamle: Agree the wording makes it unclear, will clarify that the validation is performed on the final synced block.

specs/protocol/isthmus/exec-engine.md
vdamle: > As an example of the bugs that come from this special case - this PR adds a paragraph to proposal.md that says the withdrawalsRoot can be used post-Isthmus but forgets the genesis special case where it can't.

🤦🏼‍♂️ Indeed, I will add that.

specs/protocol/isthmus/exec-engine.md
vdamle: I see that the wording doesn't make clear the situations in which this is applicable. There is a [`eth_simulate`](https://github.com/ethereum/execution-apis/pull/484) proposed addition where transactions can be simulated spread across multiple blocks. The intent here is to specify that in such cases, an empty withdrawal root is included in the block header. Will clarify this.

specs/protocol/isthmus/exec-engine.md
```diff
@@ -63,6 +80,36 @@ places a burden on users of the system in a post-fault-proofs world, where:
 Placing the [`L2ToL1MessagePasser`][l2-to-l1-mp] account storage root in the `withdrawalsRoot` field alleviates this burden
 for users and protocol participants alike, allowing them to propose and verify other proposals with lower operating costs.
 
+#### Genesis Block
+
+If isthmus is active at the genesis block, the withdrawals root is the empty withdrawals root, regardless of L2 state.
```
tynes: nit: Isthmus is capitalized elsewhere, we should be standard

specs/protocol/rollup-node-p2p.md
```diff
@@ -266,6 +267,10 @@ Canyon/Delta blocks are broadcast on `/optimism/<chainId>/1/blocks`.
 
 Ecotone blocks are broadcast on `/optimism/<chainId>/2/blocks`.
 
+### `blocksv4`
+
```
tynes, line 271: I do not think we need to add a new p2p topic for holocene anymore, so this should be good, waiting on a response: https://github.com/ethereum-optimism/specs/pull/428#issuecomment-2418099412

tynes (approved):
This looks good to me, would like approval from @ajsutton or @protolambda before merge

specs/protocol/isthmus/exec-engine.md
ajsutton: Typically to create genesis block you populate the world state from the genesis config and then insert that into the stateRoot field of the block header. That already includes the withdrawal contract storage root so there doesn't seem to be any additional dependencies being added here.

My concern is less about whether you can withdraw but more that it's really easy to forget genesis is a special case and do things like make invalid proposals or depend on the withdrawalsRoot as mentioned above when it's not actually accurate. Given we already have to know the genesis world state to create the block header it doesn't seem worth creating that foot gun.

specs/protocol/rollup-node-p2p.md
vdamle: Thanks for flagging that! Correct, @roberto-bayardo responded to your clarification - https://github.com/ethereum-optimism/specs/pull/428#issuecomment-2418259114

specs/protocol/isthmus/exec-engine.md
ajsutton: I think what I'm struggling with is this paragraph both says changes to the L2 block execution rules activate when L2 timestamp >= activation time (existed prior to this PR) _and_ the block header changes apply when considering data from a L1 block with timestamp >= activation time (added in this PR). So the L2 block changes are activating at different times which doesn't make sense to me.

For L2 block changes, I'd expect the first condition to be right - the changes activate when the L2 block timestamp >= activation time and it doesn't matter when that batch data is actually included on L1 or what the L1 origin timestamp is.

specs/protocol/isthmus/exec-engine.md
```diff
@@ -24,22 +38,25 @@ The storage root of the `L2ToL1MessagePasser` is included in the block header's
...
 
 ## `L2ToL1MessagePasser` Storage Root in Header
 
-After Holocene's activation, the L2 block header's `withdrawalsRoot` field will consist of the 32-byte
+After Isthmus hardfork's activation, the L2 block header's `withdrawalsRoot` field will consist of the 32-byte
 [`L2ToL1MessagePasser`][l2-to-l1-mp] account storage root _after_ the block has been executed, and _after_ the
 insertions and deletions have been applied to the trie. In other words, the storage root should be the same root
-that is returned by `eth_getProof` at the given block number.
+that is returned by `eth_getProof` at the given block number -- it is the account store root from the world state
+identified by the stateRoot field in the block header.
```
ajsutton: I'd just say:

```
After Isthmus hardfork's activation, the L2 block header's `withdrawalsRoot` field will consist of the 32-byte
 [`L2ToL1MessagePasser`][l2-to-l1-mp] account storage root from the world state identified but the stateRoot field in the block header.
```

specs/protocol/isthmus/exec-engine.md
```diff
@@ -63,6 +80,36 @@ places a burden on users of the system in a post-fault-proofs world, where:
...
+available to the EVM/application layer.
+
+#### P2P
+
+During sync, the block body withdrawal hash is computed from the withdrawals list in the block body.
+When verifying the final block that is synced, if Isthmus is active at such a block's header,
+we expect that the computed withdrawal hash must match the hash of an empty list of withdrawals.
+The header `withdrawalsRoot` MPT hash can be any non-null value in this case. Verification of the
+final block is performed after a header is available so that the header's timestamp can be checked
+to see if Isthmus is active.
```
ajsutton: I've got to admit, I still don't understand this. There are no withdrawals in any block for OP Stack so wouldn't the computed withdrawal root always be the empty hash? And when the final header is received if Isthmus is active shouldn't it have the message passer account storage root?

I know headers and bodies in devp2p are matched up by comparing roots (ie tx root, receipts root and withdrawals root) so during sync if you request a batch of block bodies, you'd expect that they all have an empty withdrawals list since OP Stack doesn't use withdrawals.  To match them up with the headers you'd have to essentially ignore the withdrawalsRoot (or treat the header as if it had an empty withdrawal root). I think this is trying to explain that.

specs/protocol/isthmus/exec-engine.md
```diff
@@ -63,6 +80,36 @@ places a burden on users of the system in a post-fault-proofs world, where:
...
+
+Beginning at Canyon (which includes Shanghai hardfork support) and prior to Isthmus activation,
+the `withdrawalsRoot` field is set to the MPT root of an empty withdrawals list. This is the
+same root as an empty storage root. The withdrawals are captured in the L2 state, however
+they are not reflected in the `withdrawalsRoot`. Hence, prior to Isthmus activation,
+even if a `withdrawalsRoot` is present and a MPT root is present in the header, it should not be used.
+Any implementation that calculates output root should be careful not to use the header `withdrawalsRoot`.
+
+After Isthmus activation, if there was never any withdrawal contract storage, a MPT root of an empty list
+can be set as the `withdrawalsRoot`
```
ajsutton, line 108: unless the block is genesis (though I think we should fix that).

specs/protocol/proposals.md
```diff
@@ -112,6 +113,12 @@ where:
    then the withdrawal against that storage root), we can prove against the L2toL1MessagePasser's storage root directly,
    thus reducing the verification cost of withdrawals on L1.
 
+   After the Isthmus hard fork, the `withdrawal_storage_root` is present in the
+   [block header as `withdrawalsRoot`][header-withdrawals-root] and can be used directly, instead of computing
+   the storage root of the L2toL1MessagePasser contract.
+
+   If the Isthmus hard fork is active at the genesis block, the `withdrawalsRoot` is the empty withdrawals root.
```
ajsutton: This leaves some ambiguity about what the correct output root should be.  Is the withdrawals root taken from the genesis header or from the genesis world state when constructing the output root for the genesis block?

specs/protocol/rollup-node-p2p.md
ajsutton: Note that at the top of the Gossip Topics section it says "There are three topics" but now there are four (we should rephrase that to not specify a number).

ajsutton (commented):
Looks good generally, just a couple of bits that I think could use more clarity.  For the P2P stuff I don't have all the context so may just be misunderstanding things.

I am very strongly of the opinion that it's a mistake to have the genesis root hard code an empty hash for withdrawalsRoot instead of using the actual value from the account storage. The `stateRoot` field will be changing with any changes to the actual storage root anyway.

specs/protocol/isthmus/exec-engine.md
vdamle: Removed the confusing new statement. The original description of timestamp activation conveys the intended behavior.

specs/protocol/isthmus/exec-engine.md
vdamle: I met with @protolambda / @ajsutton and we came to the agreement that even for the genesis block, we will set the `withdrawalsRoot` field in the header to the `L2ToL1MessagePasser` storage root, if Isthmus is active -- this is a change from the original spec proposal that stated that it will be set to the empty withdrawals root.

specs/protocol/isthmus/exec-engine.md
vdamle: no longer relevant as genesis header withdrawalsRoot shall also be an MPT root of an empty list (we decided not to special case the genesis withdrawalsRoot).

specs/protocol/proposals.md
vdamle: Updated.

specs/protocol/isthmus/exec-engine.md
vdamle: I've updated the description based on the conversation yesterday.

specs/protocol/isthmus/exec-engine.md
vdamle: Updated based on discussion yesterday.

specs/protocol/holocene/derivation.md
```diff
@@ -135,7 +135,7 @@ invalid.
 due to the new strict batch ordering rules.
 - If `span_end.timestamp < next_timestamp`, the span batch is set to have `past` validity, as it
 doesn't contain any new batches (this would also happen if applying timestamp checks to each derived
-singular batch individually). See below in the [Batch Queue][#batch-queue] section about the new
+singular batch individually). See below in the [Batch Queue](#batch-queue) section about the new
```
vdamle, line 138: fixed an incorrect link - this was not meant to reference a url declaration but just another section in the doc.

ajsutton (approved):
LGTM.

protolambda (approved):

PR #401, opened 2024-09-30
feat(holocene): Simplify engine queue modifications

clabby:
## Overview

Simplifies the holocene engine queue modifications.


tynes:
Seems good, will wait for @sebastianst to review

sebastianst (approved):
Thanks for fixing this confusion.

PR #404, opened 2024-10-01
feat(holocene): Simplify `eip1559Params`

clabby:
## Overview

Simplifies the dynamic EIP-1559 params feature of Holocene by including the parameters within the `PayloadAttributesV3` and the block `Header`.

### Rationale

Currently, the EIP-1559 parameters are included within the `L1Block` contract. However, unlike other variables within the L1 info transaction:
1. The EIP-1559 parameters are unlikely to change frequently enough to warrant being in the L1 info transaction. Most other fields change every L1 block, but even on the most ambitious OP Stack chains (Base), they only expect these values to change once per week.
2. In the execution engine, all OP Stack EVM implementations currently [load the L1 block info from the state database](https://github.com/ethereum-optimism/op-geth/blob/optimism/core/types/rollup_cost.go#L120-L182). However, since the EIP-1559 parameters are required at the beginning of the block, we would also need to lift this out of the EVM and into the block executor.
3. The EIP-1559 params in L1 info makes the function to compute the next block's basefee stateful. That becomes a fairly invasive change, where the pre-block environment fillers now have to have access to the database, L1 block info formation from storage slots logic, etc. In `reth`, for example, the base fee computation function is [within a lower-level config type, without access to the state or transactions](https://github.com/paradigmxyz/reth/blob/main/crates/ethereum/evm/src/lib.rs#L150-L152). Ultimately we lose database access several layers of abstraction higher up, warranting a refactor to make this change work. [In `geth`](https://github.com/ethereum-optimism/op-geth/blob/optimism/consensus/misc/eip1559/eip1559.go#L58-L98), it's similar, but the change is a bit less invasive higher up with only 2 layers to pass the state handle through.

There's precedent for `SystemConfig` variables that are needed at the beginning of a block, or that don't update frequently, to belong within the `PayloadAttributes` (The `gas_limit`). The block builder additionally having access to control the EIP-1559 parameters is constrained by consensus rule, and similarly to the `gasLimit`, is specified to be referenced from the `SystemConfig` in the derivation pipeline's payload builder. This change would simplify this feature's implementation within execution clients and save us from needlessly posting the parameters in the L1 info transaction every 2 seconds as such.

### Why in the block header?

https://github.com/ethereum-optimism/specs/pull/404/files#diff-b93100978f12975e6b4ee660c7362c71d4f2a416ef7c9f0e76e17ca0062842efR144-R152


clabby:
* **#404** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/specs/404?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

clabby:
dead-ing this proposal.

@roberto-bayardo made a good point in discord that, in order to move this value into the `PayloadAttributes`, we need to still commit to the EIP-1559 params in history as to not need to consult L1 in order to do historical execution of L2. Back to drawing board.

specs/protocol/holocene/exec-engine.md
```diff
@@ -73,5 +81,73 @@ an outbound withdrawal for a long period of time, the node may not have access t
...
+    eip1559Params: QUANTITY or null
+}
+```
+
+### `eip1559Params` encoding
+
+| Name          | Type  | Byte Offset |
+| ------------- | ----- | ----------- |
+| `denominator` | `u32` | `[0, 4)`    |
+| `elasticity`  | `u32` | `[4, 8)`    |
```
refcell: Previously these were encoded as `uint64` types. Why are we reducing to `u32`?

refcell: I see this is to fit in the u64 nonce field. Maybe we can add this as context?

roberto-bayardo: maybe since the field is limited to 32 bytes we should be frugal in case we want to put more things here in the future?  We don't need a lot of precision for either of these.

refcell (approved):

refcell:
Although I think this is sound, I'm a bit concerned of using the nonce in case we need it down the road for something. Placing these in the L1 info feels more scalable to me and is a more consistent way of introducing modifications, though there is the aforementioned tradeoff that the basefee computation becomes invasive.

specs/protocol/holocene/exec-engine.md
roberto-bayardo: You probably want to update the configurability.md to also specify u32 encoding in the configupdate event type?  Also while you're there can you change "enum" to "event"?

https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/holocene/configurability.md#configupdate

and here:
[
function setEIP1559Params(uint64 _denominator, uint64 _elasticity)](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/holocene/configurability.md#seteip1559params)

specs/protocol/holocene/exec-engine.md
```diff
@@ -73,5 +81,73 @@ an outbound withdrawal for a long period of time, the node may not have access t
...
+PayloadAttributesV3: {
+    timestamp: QUANTITY
+    random: DATA (32 bytes)
+    suggestedFeeRecipient: DATA (20 bytes)
+    withdrawals: array of WithdrawalV1
+    parentBeaconBlockRoot: DATA (32 bytes)
+    transactions: array of DATA
+    noTxPool: bool
+    gasLimit: QUANTITY or null
+    eip1559Params: QUANTITY or null
```
roberto-bayardo: Instead of QUANTITY shouldn't this just be DATA (8 bytes)?

roberto-bayardo: I also don't really like the "or null" here, or for gaslimit above, where it actually should never be null. It is an optional parameter and can be omitted by the L1, but should always be specified by an L2.

I see you are using an explicit "null" value to indicate "use the chain config" but "null" vs "unspecified because it's optional" gets a little ambiguous and could probably use clarification.  Ideally we just wouldn't allow null if the parameter is specified, and we can mark it mandatory for all Holocene  blocks (including the first one). Perhaps just use the all-0 8-byte array as the initializer?

clabby: I'm open to using the "all-0 8 byte array" as the initializer instead!

roberto-bayardo (approved):

tynes:
We should make sure that there is a getter for EIP1559 params on the `SystemConfig` so that its legible on chain what the config is. This will help to make standardness checks possible as part of the superchain-registry. This might also be in the spec, I just see the getters removed from the `L1Block` contract, would be good to double check

specs/protocol/holocene/configurability.md
```diff
@@ -50,15 +50,15 @@ The `ConfigType` enum represents configuration that can be modified.
...
+The following `ConfigUpdate` event is defined where the `CONFIG_VERSION` is `uint256(0)`:
 
 | Name | Value | Definition | Usage |
 | ---- | ----- | --- | -- |
 | `BATCHER` | `uint8(0)` | `abi.encode(address)` | Modifies the account that is authorized to progress the safe chain |
 | `FEE_SCALARS` | `uint8(1)` | `(uint256(0x01) << 248) \| (uint256(_blobbasefeeScalar) << 32) \| _basefeeScalar` | Modifies the fee scalars |
 | `GAS_LIMIT` | `uint8(2)` | `abi.encode(uint64 _gasLimit)` | Modifies the L2 gas limit |
 | `UNSAFE_BLOCK_SIGNER` | `uint8(3)` | `abi.encode(address)` | Modifies the account that is authorized to progress the unsafe chain |
-| `EIP_1559_PARAMS` | `uint8(4)` | `uint256(uint64(_denominator)) << 32 \| uint64(_elasticity)` | Modifies the EIP-1559 denominator and elasticity |
+| `EIP_1559_PARAMS` | `uint8(4)` | `uint256(uint64(uint32(_denominator))) << 32 \| uint64(uint32(_elasticity))` | Modifies the EIP-1559 denominator and elasticity |
```
sebastianst, line 63: Why are there `uint64` conversions around the `uint32` ones?

specs/protocol/holocene/exec-engine.md
```diff
@@ -4,30 +4,38 @@
...
+  - [Rationale](#rationale-2)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
+## Timestamp Activation
+
+Holocene, like other network upgrades, is activated at a timestamp.
+Changes to the L2 Block execution rules are applied when the `L2 Timestamp >= activation time`.
+Changes to the execution engine are applied when it is considering data from a L1 Block whose timestamp
+is greater than or equal to the activation timestamp.
```
sebastianst: Do we want to be more specific here?

specs/protocol/holocene/exec-engine.md
```diff
@@ -73,5 +81,73 @@ an outbound withdrawal for a long period of time, the node may not have access t
...
+  - if `eip1559Params` is zero, the [canyon base fee parameter constants](../exec-engine.md#1559-parameters) are
+    used.
+  - if `eip1559Params` are non-null, the values from the attributes are used.
+
+### Rationale
+
+This type is made available in the payload attributes to allow the block builder to dynamically control the EIP-1559
+parameters of the chain. As described in the [derivation - AttributesBuilder](./derivation.md#attributes-builder)
+section, the derivation pipeline must populate this field from the `SystemConfig` during payload building, similar to
+how it must reference the `SystemConfig` for `gasLimit` field.
```
sebastianst:
```suggestion
how it must reference the `SystemConfig` for the `gasLimit` field.
```
?

specs/protocol/holocene/exec-engine.md
```diff
@@ -73,5 +81,73 @@ an outbound withdrawal for a long period of time, the node may not have access t
...
+    withdrawals: array of WithdrawalV1
+    parentBeaconBlockRoot: DATA (32 bytes)
+    transactions: array of DATA
+    noTxPool: bool
+    gasLimit: QUANTITY or null
+    eip1559Params: DATA (8 bytes)
+}
+```
+
+### `eip1559Params` encoding
```
sebastianst, line 101: do we want to be specific about endianess?

specs/protocol/holocene/exec-engine.md
```diff
@@ -73,5 +81,73 @@ an outbound withdrawal for a long period of time, the node may not have access t
...
+
+## `eip1559Params` in Header
+
+Upon Holocene activation, the L2 block header's `nonce` field will consist of the 8-byte `eip1559Params` value.
+
+### Header Validity Rules
+
+Prior to Holocene activation, the L2 block header's `nonce` field is valid iff it is equal to `u64(0)`.
+
+After Holocene activation, The L2 block header's `nonce` field is valid iff it is non-zero.
```
sebastianst, line 135: Although the first Holocene block will still have the 0 value?

sebastianst (approved):

specs/protocol/holocene/configurability.md
clabby: Mainly to note that we do want to allow for an overflow in the shift. i.e. `uint32(0xFF_00_00_00) << 8 == 0`

specs/protocol/holocene/exec-engine.md
clabby: The first holocene block will not have the `0` value - it will insert the canyon base fee params into the `nonce` field.

sebastianst: oh gotcha!

tynes (approved):

PR #407, opened 2024-10-02
specs: migrate holocene dropped features to isthmus

tynes:
**Description**
Migrates the dropped features from holocene to isthmus


sebastianst (approved):
lgtm, but there seems to be new content besides just moved content, e.g. some new content in file `specs/protocol/holocene/system-config.md`. Is that intended?

tynes:
> lgtm, but there seems to be new content besides just moved content, e.g. some new content in file `specs/protocol/holocene/system-config.md`. Is that intended?

I renamed the file since now its only system config related things

clabby (approved):

PR #424, opened 2024-10-10
holocene: Add `past` batch validity handling

sebastianst:
**Description**

The spec also instructed past batches to cause the remaining span batch and channel to be dropped. We've decided that we still want to allow overlapping channels and span batches with Holocene.

**Tests**

### DIFF
--- a/specs/protocol/holocene/derivation.md
+++ b/specs/protocol/holocene/derivation.md
@@ -154,7 +154,7 @@ So the following changes are made to the [Bedrock Batch Queue](../derivation.md#
 future are now directly `drop`-ped. This effectively disallows gaps, instead of buffering future
 batches.
 - A new batch validity `past` is introduced. A batch has `past` validity if its timestamp is before
-or equal to the safe head's timestamp. This also applies to span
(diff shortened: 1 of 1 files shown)


refcell (approved):
Reads well to me 👍

specs/protocol/holocene/derivation.md
```diff
@@ -148,24 +151,34 @@ So the following changes are made to the [Bedrock Batch Queue](../derivation.md#
 
 - The reordering step is removed, so that later checks will drop batches that are not sequential.
 - The `future` batch validity status is removed, and batches that were determined to be in the
-future are now directly `drop`-ped. This effectively disallows gaps, instead of buffering future batches.
+future are now directly `drop`-ped. This effectively disallows gaps, instead of buffering future
+batches.
+- A new batch validity `past` is introduced. A batch has `past` validity if its timestamp is before
+or equal to the safe head's timestamp. This also applies to span
```
ajsutton:
```suggestion
or equal to the safe head's timestamp. This also applies to span batches.
```

?

ajsutton (approved):
LGTM.

PR #428, opened 2024-10-12
Holocene: eip-1559 parameters in extraData

roberto-bayardo:
**Description**

EIP-1559 parameters need to be in the ExecutionPayload, not just the PayloadAttributes, to allow geth to properly include the encoded parameters in the block header.  So instead of using nonce, we use the extraData field which is already in the ExecutionPayload. This will also reserve nonce to be used for sub-second block times.

### DIFF
--- a/specs/protocol/holocene/exec-engine.md
+++ b/specs/protocol/holocene/exec-engine.md
@@ -7,34 +7,54 @@
 - [Overview](#overview)
 - [Timestamp Activation](#timestamp-activation)
 - [Dynamic EIP-1559 Parameters](#dynamic-eip-1559-parameters)
-  - [Extended `PayloadAttributesV3`](#extended-payloadattributesv3)
-    - [PayloadID computation](#payloadid-computation)
-    - [`eip1559Params` encoding](#eip1559params-encoding)
-  - [Execution](#execution)
-  - [Rationale](#rationale)
-  - [`eip1559Params` in Header](#eip1559params-in-header)
-    - [Header Validity Rules](#header-validity-rules)
+  - [EIP-1559 Parameters in Block Header](#eip-1559-parameters-in-block-header)
+  - [EIP-1559 Parameters in `PayloadAttributesV3`](#eip-1559-parameters-in-payloadattributesv3)
     - [Encoding](#encoding)
-    - [Rationale](#rationale-1)
+    - [PayloadID computation](#payloadid-computation)
+  - [Execution](#execution)
+    - [Payload Attributes Processing](#payload-attributes-processing)
+    - [Base Fee Computation](#base-fee-computation)
+  - [Rationale](#rationale)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 ## Overview
 
-The EIP-1559 parameters are encoded in the block header's `nonce` field and can be
-configured dynamically through the `SystemConfig`.
+The EIP-1559 parameters are encoded in the block header's `extraData` field and can be configured dynamically through
+the `SystemConfig`.
 
 ## Timestamp Activation
 
-Holocene, like other network upgrades, is activated at a timestamp.
-Changes to the L2 Block execution rules are applied when the `L2 Timestamp >= activation time`.
+Holocene, like other network upgrades, is activated at a timestamp.  Changes to the L2 Block execution rules are applied
+when the `L2 Timestamp >= activation time`.
 
 ## Dynamic EIP-1559 Parameters
 
-### Extended `PayloadAttributesV3`
+### EIP-1559 Parameters in Block Header
+
+With the Holocene upgrade, the `extraData` header field of each block must have the following format:
+
+| Name          | Type               | Byte Offset |
+| ------------- | ------------------ | ----------- |
+| `version`     | `u8`               | `[0, 1)`    |
+| `denominator` | `u32 (big-endian)` | `[1, 5)`    |
+| `elasticity`  | `u32 (big-endian)` | `[5, 9)`    |
+
+Additionally,
+- `version` must be 0
+- `denominator` must be non-zero
+- there is no additional data beyond these 9 bytes
+
+Note that `extraData` has a maximum capacity of 32 bytes (to fit in the L1 beacon-chain `extraData` data-type) and its
+format may be modified/extended by future upgrades.
+
+Note also that if the chain had Holocene genesis, the genesis block must have an above-formated `extraData` representing
+the initial parameters to be used by the chain.
+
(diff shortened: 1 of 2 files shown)


specs/protocol/holocene/exec-engine.md
```diff
@@ -62,6 +63,36 @@ field.
 | `denominator` | `u32 (big-endian)` | `[0, 4)`    |
 | `elasticity`  | `u32 (big-endian)` | `[4, 8)`    |
 
+### Extended `ExecutionPayloadV3`
+
```
tynes, line 100: Does this mean there needs to be a new gossipsub topic for the modified execution payload? For unsafe blocks. Each version gets its own topic, see https://github.com/ethereum-optimism/specs/pull/396 which modifies the execution payload and a new topic is created

roberto-bayardo: Hmm, since it is just adding a new optional field and the block itself is not changed, I assumed no.  But I am not so familiar with how the p2p layer does validation and may have to dig deeper to confirm.

roberto-bayardo:
As discussed in discord, updated spec to use extraData instead of nonce.

specs/protocol/holocene/exec-engine.md
```diff
@@ -64,19 +63,19 @@ field.
...
+  - if `eip1559Params` in the `PayloadAttributesV3` is null, the attributes are to be considered invalid by the engine.
   - if `eip1559Params`' `denominator` is `0`, the attributes are to be considered invalid by the engine.
-  - if `parent_header.nonce` is zero, the [canyon base fee parameter constants](../exec-engine.md#1559-parameters) are
-    used for the block's base fee parameters.
-  - if `parent_header.nonce` is non-zero, the EIP-1559 parameters encoded within the parent header's `nonce` field are
-    used for the block's base fee parameters.
+  - if `parent_header.extraData` is empty, the [canyon base fee parameter constants](../exec-engine.md#1559-parameters)
+    are used for the block's base fee parameters.
+  - if `parent_header.extraData` is non-empty with the version byte (first byte) value 0, the remaining 64
+    bytes are used for the block's eip-1559 parameters.
```
protolambda:
```suggestion
  - if Holocene is not active in `parent_header.timestamp`, then `parent_header.extraData` may be set to anything valid in pre-Holocene context. The  [canyon base fee parameter constants](../exec-engine.md#1559-parameters), or prior EIP-1559 constants are used for the block's base fee parameters.
    - While not set in regular pre-Holocene blocks, the `extraData` was set in the genesis block of some pre-Holocene chains, and other blocks with pre-Bedrock legacy content. It is thus important to check the hardfork before parsing the `extraData`.
  - if Holocene is active in `parent_header.timestamp`, then `parent_header.extraData` is non-empty, with the following encoding:
    - the first byte is set to `0`, used as version byte for the content.
    - the next 8 bytes are used to encode the 1559 params following the [`eip1559Params`
encoding](#eip1559params-encoding).
    - the `extraData` ends there, at 9 bytes length. The `extraData` has a maximum capacity of 32 bytes (to fit in the L1 beacon-chain `extraData` data-type), and may be modified/extended in future versions.
```

A hardfork should also not retro-actively apply constraints on prior blocks, like the parent-block.
The spec needs to call out that the transition block does not read the 1559 params from the more arbitrary pre-holocene block. 
The previous spec rule would create weird behavior for the block right after genesis, if Holocene wasn't active then yet.

specs/protocol/holocene/system-config.md
```diff
@@ -27,10 +27,10 @@ The following `ConfigUpdate` event is defined where the `CONFIG_VERSION` is `uin
 | Name | Value | Definition | Usage |
 | ---- | ----- | --- | -- |
 | `BATCHER` | `uint8(0)` | `abi.encode(address)` | Modifies the account that is authorized to progress the safe chain |
-| `FEE_SCALARS` | `uint8(1)` | `(uint256(0x01) << 248) \| (uint256(_blobbasefeeScalar) << 32) \| _basefeeScalar` | Modifies the fee scalars |
+| `FEE_SCALARS` | `uint8(1)` | `abi.encode(uint256(0), (uint256(0x01) << 248) \| (uint256(_blobbasefeeScalar) << 32) \| _basefeeScalar)` | Modifies the fee scalars |
 | `GAS_LIMIT` | `uint8(2)` | `abi.encode(uint64 _gasLimit)` | Modifies the L2 gas limit |
 | `UNSAFE_BLOCK_SIGNER` | `uint8(3)` | `abi.encode(address)` | Modifies the account that is authorized to progress the unsafe chain |
-| `EIP_1559_PARAMS` | `uint8(4)` | `(uint256(_denominator) << 32) \| _elasticity` | Modifies the EIP-1559 denominator and elasticity |
+| `EIP_1559_PARAMS` | `uint8(4)` | `abi.encode((uint256(_denominator) << 32) \| _elasticity)` | Modifies the EIP-1559 denominator and elasticity |
```
protolambda, line 33: It may be worth clarifying that this is the format emitted by the `SystemConfig`. Not the same as the format in the block-header `extraData`.

roberto-bayardo: done

specs/protocol/holocene/exec-engine.md
roberto-bayardo: thanks, did not realize it could be non-empty in previous cases. updated with some tweaking to your language.

clabby (approved):
lgtm. If we want, we can also take this opportunity to address your nitpick in the encoding: `denominator, elasticity` -> `elasticity, denominator`.

roberto-bayardo:
> lgtm. If we want, we can also take this opportunity to address your nitpick in the encoding: `denominator, elasticity` -> `elasticity, denominator`.

Thanks, given the geth impl is all using this order now, may as well leave it be given how minor this is.

tynes (approved):
Looks good to me, would should let @protolambda have a final say before merge

tynes:
Just to clarify, no need to add a new p2p topic correct?

roberto-bayardo:
> Just to clarify, no need to add a new p2p topic correct?

Correct, we are no longer changing ExecutionPayload since it already contains the ExtraData field.

roberto-bayardo:
> Just to clarify, no need to add a new p2p topic correct?

Correct, we are no longer changing ExecutionPayload since it already contains the ExtraData field.

roberto-bayardo:
Oh shoot I think I borked this PR trying to rebase.

protolambda (approved):

protolambda (approved):

PR #435, opened 2024-10-17
interop: changed event tense to present for `ICrosschainERC20`

0xParti:
**Description**
Changes the event time from past to present: `CrosschainMinted` to `CrosschainMint` and `CrosschainBurnt` to `CrosschainBurn`.
The rationale is that the `ICrosschainERC20` should be a generic interface, and other standard use present tense (ERC20 as an example).


tynes (approved):

tynes:
Very in favor of this, I don't love the past tense standard that we accidentally originally adopted

PR #437, opened 2024-10-21
interop: add storage upgrade spec

agusduha:
**Description**

Adding spec for the `OptimismMintableERC20Factory` storage update to fully implement liquidity migration of `OptimismMintableERC20` tokens to `OptimismSuperchainERC20`, the `OptimismMintableERC20Factory` needs to update the deployment history through a hash onion strategy, as outlined in this [design doc](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/superchain-erc20/storage-upgrade.md).

### DIFF
--- a/specs/interop/storage-upgrade.md
+++ b/specs/interop/storage-upgrade.md
@@ -1,0 +1,114 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Migrated Liquidity storage update](#migrated-liquidity-storage-update)
+  - [Overview](#overview)
+  - [`OptimismMintableERC20Factory` predeploy](#optimismmintableerc20factory-predeploy)
+  - [Generate Hash Onion](#generate-hash-onion)
+  - [`superchain-ops` task](#superchain-ops-task)
+  - [Additional Notes](#additional-notes)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Migrated Liquidity storage update
+
+**Disclaimer:** This content should be placed in the protocol upgrade folder when it becomes available.
+
+## Overview
+
+To fully migrate liquidity from `OptimismMintableERC20` tokens to `OptimismSuperchainERC20`,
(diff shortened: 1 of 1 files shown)


blmalone (approved):
I can't think of a significant negative consequence of being constrained to setting storage in a strict order with the hash onion approach.
With this in mind, I think this approach makes sense 👍🏻

specs/interop/storage-upgrade.md
```diff
@@ -0,0 +1,121 @@
...
+    require(_localTokens.length == _remoteTokens.length, "TokensLengthMismatch");
+
+    // Unpeel the hash onion and store the deployments
+    bytes32 innerLayer = _startingInnerLayer;
+    for (uint256 i; i < _localTokens.length; i++) {
+        innerLayer = keccak256(abi.encodePacked(innerLayer, abi.encodePacked(_localTokens[i], _remoteTokens[i])));
+
+        deployments[_localTokens[i]] = _remoteTokens[i];
+
+        emit DeploymentStored(_localTokens[i], _remoteTokens[i]);
```
blmalone, line 68: I'm thinking it might help while testing, to also log the inner layer here.

specs/interop/storage-upgrade.md
```diff
@@ -0,0 +1,121 @@
...
+    require((Predeploys.L2_CROSS_DOMAIN_MESSENGER).xDomainMessageSender() ==
+        AddressAliasHelper.undoL1ToL2Alias((Predeploys.PROXY_ADMIN).owner()), "Unauthorized");
+
+    require(hashOnion == 0, "Already initialized");
+
+    hashOnion = _hashOnion;
+}
+
+function verifyAndStore(
+    address[] calldata _localTokens,
```
blmalone, line 52: If these token lists are large, what're the chances we see a stack too deep error? If so, we will need to consider how the upgrade task needs to change to account for this possibility.

agusduha: This `verifyAndStore` function can be called in batches, it is not needed to call the hole list at once.

Also this function is meant to be called directly on L2, and it is not part of the upgrade task to set the initial hash onion, so it is simpler to handle it.

agusduha:
Closing this spec due to being deprioritized, creating this issue https://github.com/ethereum-optimism/optimism/issues/12750 to keep track, will be revisited next year

PR #438, opened 2024-10-23
Add execution witness and account proof hint types

meyer9:
**Description**

Adds `l2-execution-witness` and `l2-account-proof` which are op-program hints that will replace the more granular `l2-code` and `l2-state-node` hints.

**Additional context**

optimism PR: https://github.com/ethereum-optimism/optimism/pull/12559

### DIFF
--- a/specs/fault-proof/index.md
+++ b/specs/fault-proof/index.md
@@ -34,6 +34,8 @@
     - [`l2-code <codehash>`](#l2-code-codehash)
     - [`l2-state-node <nodehash>`](#l2-state-node-nodehash)
     - [`l2-output <outputroot>`](#l2-output-outputroot)
+    - [`l2-execution-witness <blocknumber>`](#l2-execution-witness-blocknumber)
+    - [`l2-account-proof <blocknumber_and_address>`](#l2-account-proof-blocknumber_and_address)
   - [Precompile Accelerators](#precompile-accelerators)
 - [Fault Proof VM](#fault-proof-vm)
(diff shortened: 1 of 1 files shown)


clabby (approved):

specs/fault-proof/index.md
```diff
@@ -438,6 +440,18 @@ Requests the host to prepare the L2 Output at the l2 output root `<outputroot>`.
 The L2 Output is the preimage of a
 [computed output root](../protocol/proposals.md#l2-output-commitment-construction).
 
+#### `l2-execution-witness <blocknumber>`
+
+Requests the host to prepare all preimages used in the execution of the block at `<blocknumber>` height.
+
+#### `l2-account-proof <blocknumber_and_address>`
```
ajsutton, line 447: I think these should be supplying the block hash, not the block number. We need to know that the data comes from the exact same block even if there are later reorgs. It's unlikely there will be a reorg that affects this, but everything else is block hash for that reason.

ajsutton: We need request the ExecutionWitness as part of building the block so we don' know the target block hash yet, only the parent.  So ExecutionWitness will have to stay as block number which is unfortunate but seems unavoidable.

But we should switch l2-account-proof to pass a block hash.

meyer9: Sounds good, will open a new PR to update this. This will diverge from the Kona hints, so I'll open an issue on that repo to update this: https://github.com/anton-rs/kona/blob/main/bin/client/src/l2/chain_provider.rs#L191-L192

PR #444, opened 2024-10-25
Improve definition of channel encoding

geoknee:
I think this section was written before span batches were invented, and it doesn't use "rlp encode" accurately any more. 

Drive-by observation: I think the final `rlp_encode` (the only one, if using span batches) is superfluous and could in principle be removed from the protocol -- it adds a few bytes which don't contain any useful information, since the input is just a byte string.

https://github.com/ethereum-optimism/optimism/blob/02d5832349a1f469adb00191459333e2032b09b4/op-node/rollup/derive/batch.go#L82

### DIFF
--- a/specs/protocol/derivation.md
+++ b/specs/protocol/derivation.md
@@ -376,7 +376,8 @@ rlp_batches = rlp_encode(encoded_batches)
 
 where:
 
-- `batches` is the input, a sequence of batches each with a byte-encoder function `.encode()` as per the next section ("Batch Encoding")
+- `batches` is the input, a sequence of batches each with a byte-encoder
+function `.encode()` as per the next section ("Batch Encoding")
 - `encoded_batches` is a byte array: the concatenation of the encoded batches
 - `rlp_batches` is the rlp encoding of the concatenated encoded batches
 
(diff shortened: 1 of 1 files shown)


specs/protocol/derivation.md
```diff
@@ -368,15 +368,17 @@ where:
...
+rlp_batches = rlp_encode(encoded_batches)
 ```
 
 where:
 
-- `batches` is the input, a sequence of batches byte-encoded as per the next section ("Batch Encoding")
-- `rlp_batches` is the concatenation of the RLP-encoded batches
+- `batches` is the input, a sequence of batches each with a byte-encoder function `.encode()` as per the next section ("Batch Encoding")
+- `encoded_batches` is a byte array: the concatenation of the encoded batches
+- `rlp_batches` is the rlp encoding of the concatenated encoded batches
```
geoknee, line 382: I kept this name of this variable so that existing references don't get broken.

specs/protocol/derivation.md
```diff
@@ -368,15 +368,17 @@ where:
 A channel is encoded by applying a streaming compression algorithm to a list of batches:
 
 ```text
-rlp_batches = []
+encoded_batches = []
 for batch in batches:
-    rlp_batches.append(batch)
+    encoded_batches ++ batch.encode()
```
geoknee, line 373: Using `++` here to emphasize that there is no structure to `encoded_batches`, i.e. it is _not_ a `[][]byte`.

sebastianst (approved):

PR #447, opened 2024-11-01
cannon: Migrate MTCannon spec to 64-bit

### LINKED ISSUE #12253: 64-bit MT-FPVM Specification
Create a specification for the 64-bit FPVM. This replaces the 32-bit MT-FPVM spec.

mbaxter:
**Description**

Migrate the multi-threaded FPVM spec to 64-bit:
- Update various constants, syscall numbers, etc
- Add a section to "new features" briefly discussing 64-bit architecture
- Update stateful instructions section and document the new `llReservationStatus` field

**Metadata**

Fixes https://github.com/ethereum-optimism/optimism/issues/12253


specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -68,16 +71,18 @@ Thus, the trace of a program executed by the FPVM is an ordered set of VM states
...
 - `Hash` - A 256-bit fixed-size value produced by the Keccak-256 cryptographic hash function.
+- `UInt8` - An 8-bit unsigned integer value.
 - `UInt64` - A 64-bit unsigned integer value.
-- `Word` - A 32-bit value.
+- `Word` - A 64-bit value.
 
 #### Constants
 
+- `EBADF` - A Linux error number indicating a bad file descriptor: `0x9`.
 - `MaxWord` - A `Word` with all bits set to 1: `0xFFFFFFFF`.
```
Inphi:
```suggestion
- `MaxWord` - A `Word` with all bits set to 1: `0xFFFFFFFFFFFFFFFF`.
```

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -205,11 +215,13 @@ the FPVM preempts the active thread.
...
+
 The Load Linked Word (`ll`) and Store Conditional Word (`sc`) instructions provide the low-level
 primitives used to implement atomic read-modify-write (RMW) operations.  A typical RMW sequence might play out as
 follows:
 
-- `ll` place a "reservation" on a particular memory address.
+- `ll` places a "reservation" on a particular memory address.
 - Subsequent instructions take the value at this address and perform some operation on it:
   - For example, maybe a counter variable is reserved and incremented.
 - `sc` is called and the modified value is stored at the reserved address only if it has not been modified since the
```
Inphi: I forget what the behavior was pre-64, but this isn't exactly correct now.
`sc` succeeds only if the word at the reserved _effective_ address has been modified. This is stricter than what's currently in the spec as a single byte modified at that word but located at a different reserved address will invalidate the reservation. So I suggest rewording this.

It probably helps to use be explicit on the type of address referenced here (virtual/effective).

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -224,24 +236,36 @@ FPVM state to track memory reservations initiated by `ll` operations.
...
 Only a single memory reservation can be active at a given time - a new reservation will clear any previous reservation.
 
 When the VM writes any data to memory, these `ll`-related fields are checked and the memory reservation is cleared if
-a memory write touches a reserved `llAddress`.
+a memory write touches a `Word` that contains `llAddress`.
 
 When an `sc` instruction is executed, the operation will only succeed if:
 
-- There exists an active reservation (`llReservationActive == true`).
+- There exists an active 4-byte memory reservation (`llReservationStatus == 1`).
```
Inphi: Could you clarify this better.

This line can be parsed incorrectly in several ways. I originally read the 4-byte memory reservation as the width of the reserved memory location. [Which isn't the case](https://github.com/ethereum-optimism/optimism/blob/develop/cannon/mipsevm/multithreaded/mips.go#L331C1-L334C3).
My second read was that it referred to the 4-byte memory reservation as the width of `llReservationStatus`, which isn't correct either.

Upon my third read, I finally inferred that the line describes the bit length of the operation used to create the reservation (i.e. `ld`). Can we be more explicit what this means. The phrase "4-byte memory reservation" hasn't been defined prior to this so it'll be worthwhile to clarify what that means. You could also simply replace this line with:
> When `llReservationStatus` is `1`

to avoid any ambiguity on what "4-byte memory" means in this context.

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -205,11 +215,13 @@ the FPVM preempts the active thread.
...
 ## Stateful Instructions
 
+### Load Linked / Store Conditional Word
+
 The Load Linked Word (`ll`) and Store Conditional Word (`sc`) instructions provide the low-level
 primitives used to implement atomic read-modify-write (RMW) operations.  A typical RMW sequence might play out as
 follows:
 
-- `ll` place a "reservation" on a particular memory address.
+- `ll` places a "reservation" on a particular memory address.
```
Inphi: Maybe worth being explicit here:
```suggestion
- `ll` places a "reservation" on a particular virtual address.
```

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -224,24 +236,36 @@ FPVM state to track memory reservations initiated by `ll` operations.
...
-On success, `sc` stores a value at the target memory address, clears the memory reservation and returns `1`.
+On success, `sc` stores a value at the target memory address, clears the memory reservation by zeroing out
+`llReservationStatus`, `llOwnerThread`, and `llAddress` and returns `1`.
 On failure, `sc` returns `0`.
 
+### Load Linked / Store Conditional Doubleword
+
+With the transition to MIPS64, Load Linked Doubleword (`lld`), and Store Conditional Doubleword (`scd`) instructions
+are also now supported.
+These instructions are similar to `ll` and `sc`, but target 8-byte rather than 4-byte memory segments.
```
Inphi: Note that in Cannon, both `ll` and `lld` "target 8-byte memory segments" since the reservation is invalidated when any byte located at its effective address (which spans an 8-byte word on MIPS64) is written to.

specs/experimental/cannon-fault-proof-vm-mt.md
```diff
@@ -252,11 +276,12 @@ The FPVM is a state transition function that operates on a state object consisti
 1. `preimageKey` - [`Hash`] The value of the last requested pre-image key.
 1. `preimageOffset` - [`Word`] The value of the last requested pre-image offset.
 1. `heap` - [`Word`] The base address of the most recent memory allocation via mmap.
-1. `llReservationActive` - [`Boolean`] Indicates whether a memory reservation,
-   which is reserved via a Load Linked Word (`ll`) instruction, is active.
+1. `llReservationStatus` - [`UInt8`] The current memory reservation status where: `0` means there is no
+   reservation, `1` means a 4-byte memory segment is reserved, and `2` means an 8-byte memory segment is reserved.
```
Inphi: The meaning of "4-byte memory segment" isn't clear from context. I suggest just using `ld/sc` and `lld/scd` as both semantics have already been defined on the previous section.

Inphi (commented):
We should be more explicit when using "memory address" in the spec. On MIPS64, it becomes ambiguous since an address either refers to the natural alignment of memory access by a 32-bit instruction, or an effective address.

specs/experimental/cannon-fault-proof-vm-mt.md
mbaxter: good catch :D - fixed

specs/experimental/cannon-fault-proof-vm-mt.md
mbaxter: tweaked this a bit - let me know if you think it makes sense now.

specs/experimental/cannon-fault-proof-vm-mt.md
mbaxter: 👍 simplified this

specs/experimental/cannon-fault-proof-vm-mt.md
mbaxter: done

specs/experimental/cannon-fault-proof-vm-mt.md
mbaxter: ~Just cut this line altogether because it's confusing and doesn't really add much.~ Edit: Added this back in, slightly modified 😂

mbaxter:
> We should be more explicit when using "memory address" in the spec

I'll do another pass and see if I can better clarify

Inphi (approved):
Looks good and easy to read. Thanks!

ajsutton (approved):

PR #459, opened 2024-11-18
specs: remove message expiry

tynes:
**Description**

The message expiry design has changed and is no longer going through the
messenger. This commit removes the information to simplify the specs
and implement more closely what exists today.


0xParti (approved):

0xParti:
do you think it would be worth to create a design doc with more details on expired messages? the entrypoint design doc has some comments on it, but might feel insufficient.

tynes:
I would like to explore another type of design for message expiry. Will follow up with a GitHub issue

PR #461, opened 2024-11-19
Update hint docs to include payload witness hint

meyer9:
**Description**

Replaces `l2-execution-witness` with `l2-payload-witness`. Currently the encoding is a bit weird for this hint as it's much larger than other hints.

`l2-execution-witness` does not cover the case of transactions that are invalid, but still posted on L1. We need to execute these transactions up until the point of failure.

I would propose we just switch all hint encoding to use JSON instead if this encoding is too weird.

### DIFF
--- a/specs/fault-proof/index.md
+++ b/specs/fault-proof/index.md
@@ -34,7 +34,7 @@
     - [`l2-code <codehash>`](#l2-code-codehash)
     - [`l2-state-node <nodehash>`](#l2-state-node-nodehash)
     - [`l2-output <outputroot>`](#l2-output-outputroot)
-    - [`l2-execution-witness <blocknumber>`](#l2-execution-witness-blocknumber)
+    - [`l2-payload-witness <payload_attributes>`](#l2-payload-witness-payload_attributes)
     - [`l2-account-proof <blockhash_and_address>`](#l2-account-proof-blockhash_and_address)
   - [Precompile Accelerators](#precompile-accelerators)
 - [Fault Proof VM](#fault-proof-vm)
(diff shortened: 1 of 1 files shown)


specs/fault-proof/index.md
```diff
@@ -440,16 +440,17 @@ Requests the host to prepare the L2 Output at the l2 output root `<outputroot>`.
 The L2 Output is the preimage of a
 [computed output root](../protocol/proposals.md#l2-output-commitment-construction).
 
-#### `l2-execution-witness <blocknumber>`
+#### `l2-payload-witness <payload_attributes>`
 
-Requests the host to prepare all preimages used in the execution of the block at `<blocknumber>` height.
+Requests the host to prepare all preimages used in the building of the payload specified by `<payload_attributes>`.
+`<payload_attributes>` is a hex-encoded JSON with the fields `parentBlockHash` and `payloadAttributes`.
```
ajsutton: I don't think there's any reason we need to hex-encode this.  The hint is just a length prefixed string, so we can include the JSON directly for this hint and others can keep using a simple hex hash. So we'd still have literally `l2-payload-witness ` as the start of the string, followed by the JSON. The payload parsing already differs for different hint types so I don't see a problem using JSON here.

ajsutton: Ah, the prefetcher in op-program does a hexdecode on the payload as part of generic parsing, but I think it's probably better to just move that and return the original string and then we can hex decode or json decode based on the hint type.

meyer9: sounds good! will update this

ajsutton (approved):
LGTM.

PR #463, opened 2024-11-21
feat: add DeputyPauseModule spec

smartcontracts:
Adds the spec for the DeputyPauseModule.

### DIFF
--- a/specs/protocol/safe-extensions.md
+++ b/specs/protocol/safe-extensions.md
@@ -119,15 +119,20 @@ The following security properties must be upheld by the `DeputyGuardianModule`:
 The Deputy Guardian Safe (currently the Optimism Foundation Safe) utilizes the Deputy Pause Module
 to remove the need for brittle pre-signed pause transactions and to speed up the reaction speed of
 the Deputy Guardian Safe for the specific purpose of triggering the Superchain-wide pause action.
+The Deputy Pause Module is explicitly designed to be used with an Externally Owned Account and is
+not designed to function with a smart contract deputy.
 
 #### Invariants
 
-1. Must correctly enforce access control so that only the hard-coded Deputy account can act.
-1. Must always allow the hard-coded Deputy account to act, even if the private key for the Deputy
-   account has been leaked.
-1. Must only allow the hard-coded Deputy account to carry out the singular action of causing the
-   Guardian account to trigger the Superchain-wide pause function on the `SuperchainConfig`
-   contract (and must not allow any other actions).
+1. Must enforce that the Deputy account is an Externally Owned Account under the assumption that
+   the account will not attempt to circumvent the available mechanisms for doing so.
+1. Must correctly enforce access control so that only the Deputy account can act.
+1. Must always allow the Deputy account to act even if the private key for this account is leaked.
+1. Must not allow the Deputy to create authentication signatures that are indefinitely valid.
+1. Must only allow the Deputy account to carry out the singular action of causing the Guardian to
+   trigger the Superchain-wide pause function on the `SuperchainConfig` contract via the Deputy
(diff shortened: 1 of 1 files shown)


mds1 (approved):

ajsutton (approved):
Makes sense to me.

specs/protocol/safe-extensions.md
```diff
@@ -112,6 +116,46 @@ The following security properties must be upheld by the `DeputyGuardianModule`:
...
+
+The Deputy Guardian Safe (currently the Optimism Foundation Safe) utilizes the Deputy Pause Module
+to remove the need for brittle pre-signed pause transactions and to speed up the reaction speed of
+the Deputy Guardian Safe for the specific purpose of triggering the Superchain-wide pause action.
+The Deputy Pause Module is explicitly designed to be used with an Externally Owned Account and is
+not designed to function with a smart contract deputy.
+
+#### Invariants
+
+1. Must enforce that the Deputy account is an EOA.
```
maurelian, line 131: are we actually going to do a `tx.origin` style check here? I'd prefer not to just for the sake of flexibility.

If we really want to, I'd be OK with doing it at the time the account is added, but not when the pause is being called.

maurelian: OK, I see below that it's based on signature validation. That seems reasonable.

specs/protocol/safe-extensions.md
```diff
@@ -112,6 +116,46 @@ The following security properties must be upheld by the `DeputyGuardianModule`:
...
+   to these values must be implemented by re-deploying the contract, removing the old module, and
+   adding the new module.
+1. Deputy Pause Module verifies a signature from the Deputy over a fixed string in the constructor
+   to confirm that the Deputy is an EOA that can generate valid signatures and that the creator at
+   least indirectly has access to the key.
+1. Pause action is gated and must come with a valid signature from the Deputy account.
+1. Signed pause messages must contain an expiry timestamp so that the message can only be used to
+   trigger the pause when the block timestamp is less than the signed expiry. Signed expiry
+   timestamp cannot be more than a fixed number of seconds in the future as defined by a
+   constructor parameter of the module itself.
```
maurelian, line 152: What's the thinking here? 

It seems like we're trying to protect against a key leak which signs an indefinite pause sig, but why would any attacker with the key bother to do that and then just sit on the sig? 

OTOH from an operational perspective this feels confusing:

> can only be used to trigger the pause when the block timestamp is less than the signed expiry. 

So, the check is `require(block.timestamp < signedExpiry)`?

> Signed expiry timestamp cannot be more than a fixed number of seconds in the future

or is it `require(block.timestamp + FIXED_NUMBER_OF_SECONDS_IN_FUTURE < signedExpiry)`?

smartcontracts: Ah so the thinking here is that we can lock down the key material but we can distribute signatures with expiry dates to a wider audience. Signatures will expire after some amount of time so people who have access to those signatures at some point in the past do not have perpetual access to the ability to trigger the pause.

We would have two checks. We would check that the signed expiry is still in the future:

```solidity
require(expiry > block.timestamp, "signature is expired");
```

And then we would check that the signed expiry is not further than some maximum amount of time in the future:

```solidity
require(expiry - block.timestamp < MAX_FUTURE_EXPIRY_SECONDS, "expiry is too far into the future");
```

This second condition is mainly to prevent the key from accidentally signing off on something that is valid indefinitely but I think we can probably get rid of that condition if you wanted to.

maurelian: Thanks this is helpful. Agree that signatures should expire. 

Honestly I'm not super bothered by the second check now, and putting the max expiry into the source might actually be the best way of avoiding confusion about what the upper bound should be. 

Feel free to resolve this.

maurelian (approved):

PR #465, opened 2024-11-22
feat: add shared lockbox spec 

agusduha:
**Description**

Introducing the `SharedLockbox` spec and its corresponding upgrade process. This enhancement improves the Superchain’s interoperable ETH withdrawal user experience and prevents withdrawal failures caused by insufficient ETH in the `OptimismPortal`. 

**Additional context**

Design doc: https://github.com/ethereum-optimism/design-docs/pull/206


specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,127 @@
...
+  - [Interface and properties](#interface-and-properties)
+  - [Events](#events)
+- [Reference implementation](#reference-implementation)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+With interoperable ETH, withdrawals will fail if the referenced `OptimismPortal` lacks sufficient ETH.
+This is due to having the possibility to move ETH liquidity accross the different chains and it could happen
```
tynes: typo: `across`

specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,127 @@
...
+when a chain is added.
+
+### Interface and properties
+
+**`lockETH`**
+
+Deposits and locks ETH into the lockbox's liquidity pool.
+
+- The function MUST accept ETH.
+- Only authorized `OptimismPortal` addresses SHOULD be allowed to interact.
```
maurelian: I think it's worth adding "MUST not revert" when called by an allowed address.

specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,127 @@
...
+from any OP chain in the Superchain, regardless of where the ETH was initially deposited.
+
+## Design
+
+The `SharedLockbox` contract is designed to manage the unified ETH liquidity for the Superchain.
+It implements two main functions: `lockETH` for depositing ETH into the lockbox,
+and `unlockETH` for withdrawing ETH from the lockbox.
+These functions are called by the `OptimismPortal` contracts to manage the shared ETH liquidity
+when making deposits or finalizing withdrawals.
+These `OptimismPortal`s will be allowlisted by the `SuperchanConfig` using the `authorizePortal` function
```
maurelian:
```suggestion
These `OptimismPortal`s will be allowlisted by the `SuperchainConfig` using the `authorizePortal` function
```

specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Interface and properties
+
+**`lockETH`**
+
+Deposits and locks ETH into the lockbox's liquidity pool.
+
+- The function MUST accept ETH.
+- Only authorized `OptimismPortal` addresses SHOULD be allowed to interact.
+- The function MUST emit the `ETHLocked` event with the `portal` that called it and the `amount`.
```
maurelian: This is a bit of a product decision, but it might be helfpul to also emit an l2 chain id.

specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,127 @@
...
+
+```solidity
+function lockETH() external payable;
+```
+
+**`unlockETH`**
+
+Withdraws a specified amount of ETH from the lockbox's liquidity pool.
+
+- Only authorized `OptimismPortal` addresses SHOULD be allowed to interact.
```
maurelian: Nit: why isn't this a MUST?

specs/interop/shared-lockbox-upgrade.md
```diff
@@ -0,0 +1,128 @@
...
+- [Batch transaction process](#batch-transaction-process)
+  - [Diagram](#diagram)
+- [Future Considerations / Additional Notes](#future-considerations--additional-notes)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+Based on the assumption that a chain joining the dependency set is an irreversible process,
+the on-chain chains list is simplified by assuming that joining the Shared Lockbox is
```
maurelian: I lack some context here, what/where is the on-chain chains list? 
It seems like it's becoming equivalent to the dependency set, was it previously different?

specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,127 @@
...
+## Overview
+
+With interoperable ETH, withdrawals will fail if the referenced `OptimismPortal` lacks sufficient ETH.
+This is due to having the possibility to move ETH liquidity accross the different chains and it could happen
+that a chain ends up with more liquidity than its `OptimismPortal`.
+The `SharedLockbox` improves the Superchain's interoperable ETH withdrawal user experience and avoids this issue.
+To do so, it unifies ETH L1 liquidity in a single contract (`SharedLockbox`), enabling seamless withdrawals of ETH
+from any OP chain in the Superchain, regardless of where the ETH was initially deposited.
+
+## Design
```
maurelian: Is this contract proxied/upgradeable?

specs/interop/shared-lockbox-upgrade.md
```diff
@@ -0,0 +1,128 @@
...
+
+- Add the chain to the op-governed dependency set
+- Move ETH liquidity from `OptimismPortal` to `SharedLockbox`
+- Upgrade the code of `OptimismPortal` to include the `SharedLockbox` integration
+
+This process also requires that:
+
+- `SharedLockbox` is deployed
+- `SuperchainConfig` is upgraded to manage the dependency set
+- `SystemConfig` is upgraded to the interop contract version
```
maurelian: Comment: This process should be managed by an OPCM (there is a different OCPM for each l1 contracts release).
Thoughts: Chains in the initial interop set will upgrade together to this shared lockbox. But other chains will upgrade seperately to their own shared lockbox.
  Future versions of the OPCM will therefore need to enable some distinct upgrade path which enables moving ETH into the interop set's lockbox.

specs/interop/shared-lockbox.md
maurelian: I see in the impl that it is. Should be added to the spec, as otherwise the options for moving assets out of the lockbox again are much trickier.

specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,127 @@
...
+```solidity
+event ETHUnlocked(address indexed portal, uint256 amount);
+```
+
+**`PortalAuthorized`**
+
+MUST be triggered when `authorizePortal` is called
+
+```solidity
+event PortalAuthorized(address indexed portal);
```
maurelian: to be clear, there is no intention to enable deletion from the list right?

specs/interop/optimism-portal-interop.md
```diff
@@ -0,0 +1,49 @@
...
+The `OptimismPortal` contract is upgraded to integrate the `SharedLockbox` and start using the shared ETH liquidity.
+
+### Interface and properties
+
+The `OptimismPortal` contract will add the following storage layout and modified functions:
+
+**`SHARED_LOCKBOX`**
+
+- An immutable address pointing to the `SharedLockbox` contract.
+- This address MUST be immutable because all `OptimismPortals` will point to the same `SharedLockbox`,
```
smartcontracts: Immutable but not an `immutable` variable (just a note for implementation)

specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,127 @@
...
+
+- [Overview](#overview)
+- [Design](#design)
+  - [Interface and properties](#interface-and-properties)
+  - [Events](#events)
+- [Reference implementation](#reference-implementation)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
```
smartcontracts: General feedback, would like to see more of an invariant-driven approach to this specification. We need to understand the top-level properties that the contract must satisfy, then we can define the implementation spec that satisfies these properties.

specs/interop/shared-lockbox.md
tynes: We don't have the L2 chainid legible on L1 right now to be able to do this but I do agree this is useful

specs/interop/optimism-portal-interop.md
tynes: The value should never change but if we make it an `immutable` in practice, it means that chains that upgrade but don't join the cluster instantly will need to deploy their own portal contract, which just adds cost. Another way to do it would be to be the address of the lockbox on the superchain config and fetch it from there, it adds an extra call which isn't ideal but it does reduce the amount of config footguns

specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,127 @@
...
+    require(_authorizedPortals[msg.sender], "Unauthorized");
+
+    emit ETHLocked(msg.sender, msg.value);
+}
+
+function unlockETH(uint256 _value) external {
+    require(_authorizedPortals[msg.sender], "Unauthorized");
+
+    // Using `donateETH` to not trigger a deposit
+    IOptimismPortal(msg.sender).donateETH{ value: _value }();
```
tynes: What do you think about using the selfdestruct trick to transfer ether rather than using `donateETH`? we have a `SafeSend` library that triggers a balance transfer without any execution in the contract

specs/protocol/superchain-configuration.md
```diff
@@ -64,3 +67,55 @@ When the Pause is activated, the following methods are disabled:
...
+
+The `SuperchainConfig` contract will manage and keep track of the dependency graph.
+It will be queried as the source of truth to get which chains are part of the Superchain.
+It will also allow to add a chain to the op-governed cluster and update each chain’s dependency set.
+
+### Interface and properties
+
+The contract will add the following storage layout and function:
+
+**`SHARED_LOCKBOX`**
```
tynes: Small nit: `####` rather than `**` so there are hyperliinks directly to the section

specs/protocol/superchain-configuration.md
```diff
@@ -64,3 +67,55 @@ When the Pause is activated, the following methods are disabled:
...
+- This address MUST be immutable because there's only one `SharedLockbox` for each cluster.
+
+**`systemConfigs`**
+
+- A mapping that associates chain IDs with their respective SystemConfig addresses.
+- It will be used when updating dependencies along each chain.
+
+**`dependencySet`**
+
+- An `EnumerableSet` that stores the current list of chain IDs in the dependency set.
```
tynes: There isn't context on what an `EnumerableSet` is. Its an implementation detail, we should describe the properties of this data structure that we depend on

specs/interop/optimism-portal-interop.md
```diff
@@ -0,0 +1,49 @@
...
+
+- [Overview](#overview)
+  - [Interface and properties](#interface-and-properties)
+  - [Integrating `SharedLockbox`](#integrating-sharedlockbox)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+The `OptimismPortal` contract is upgraded to integrate the `SharedLockbox` and start using the shared ETH liquidity.
```
ControlCplusControlV: "using the shared ETH liquidity" is a bit ambiguous imo, as each `OptimismPortal` is practically storing its Ether in the shared lock box, so we should note that funds are being moved with this change

specs/interop/optimism-portal-interop.md
```diff
@@ -0,0 +1,49 @@
...
+
+- An immutable address pointing to the `SharedLockbox` contract.
+- This address MUST be immutable because all `OptimismPortals` will point to the same `SharedLockbox`,
+  and this address SHOULD not change.
+
+### Integrating `SharedLockbox`
+
+The integration with the `SharedLockbox` involves adding extra steps when executing deposit transactions
+or finalizing withdrawal transactions.
+These steps include locking and unlocking ETH without altering other aspects of the current `OptimismPortal` implementation.
```
ControlCplusControlV: I would also encourage more direct language here instead of "adding extra steps" and "include," are these definitively all the extra steps being added?

specs/interop/optimism-portal-interop.md
```diff
@@ -0,0 +1,49 @@
...
+To implement this solution, the following changes are needed:
+
+**`depositTransaction`**
+
+Calls `lockETH` on the `SharedLockbox` with the `msg.value`.
+
+- The function MUST call `lockETH` on the `SharedLockbox` if:
```
ControlCplusControlV: I would specify the value it must call with here as well

specs/interop/shared-lockbox-upgrade.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Overview
+
+Based on the assumption that a chain joining the dependency set is an irreversible process,
+the on-chain chains list is simplified by assuming that joining the Shared Lockbox is
+equivalent to joining the op-governed dependency set.
+
+The upgrade process consists of three main points:
+
+- Add the chain to the op-governed dependency set
```
ControlCplusControlV: How exactly is this defined? Which contract is involved here?

specs/interop/shared-lockbox-upgrade.md
```diff
@@ -0,0 +1,128 @@
...
+- `SuperchainConfig` is upgraded to manage the dependency set
+- `SystemConfig` is upgraded to the interop contract version
+
+### Add the chain to the op-governed dependency set
+
+The `SuperchainConfig` contract will be responsible for storing and managing the dependency set.
+Its `addChain` function will be used to add the chain to the dependency set and call the `SystemConfig` of each chain
+to keep them in sync.
+It will also allowlist the corresponding `OptimismPortal`, enabling it to lock and unlock ETH from the `SharedLockbox`.
+Once this process is complete, the system will be ready to process deposits and withdrawals.
```
ControlCplusControlV: I would replace "this process" for "addChain is called"

specs/interop/shared-lockbox-upgrade.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Migrate ETH liquidity from `OptimismPortal` to `SharedLockbox`
+
+The ETH will be transferred from the `OptimismPortal` to the `SharedLockbox` using an intermediate contract.
```
ControlCplusControlV: I think we should name this intermediate contract specifically, and just mention defined below

specs/interop/shared-lockbox-upgrade.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+### `OptimismPortal` code upgrade
+
+The `OptimismPortal` will start locking and unlocking ETH through the `SharedLockbox`.
+It will continue to handle deposits and withdrawals but won't directly hold the ETH liquidity.
+To set this up, the upgrade function will be called via `ProxyAdmin` to implement the new code,
+which includes the necessary `SharedLockbox` integration.
+The `SharedLockbox` address will be set during the `initialize` function. After this step,
+the `OptimismPortal` will not be able to process deposits and withdrawals until the chain is registered
```
ControlCplusControlV: We should probably specify a downtime consideration here

specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,127 @@
...
+
+    emit ETHUnlocked(msg.sender, _value);
+}
+
+function authorizePortal(address _portal) external {
+    require(msg.sender == superchainConfig, "Unauthorized");
```
ControlCplusControlV: fwiw I would like to use custom errors going forward in new code

ControlCplusControlV (commented):
Left some minor comments, but overall looks great!

smartcontracts:
Dropping this comment here as per request from @tynes. I think the most valuable think to add to this PR would be to clearly define the invariants that need to be maintained when we introduce this component.

Some examples here would be:

1. Only authorized Portal contracts should be able to use the lockbox
1. Portal contracts can only be authorized by <some address>
1. Only safe Portal contracts can be authorized (now we need to define "safe")

I think point (3) is the most important and it would be valuable to explore all of the ways in which this condition could be violated. What is "safe"? We need to be exhaustive.

specs/interop/shared-lockbox.md
agusduha: This is correct, it is a confirmed decision across the organization https://docs.google.com/document/d/1UixbH3G088fD7ecHSOA8ZinzE4UdcvxFl_1GLyM_VcI

specs/interop/shared-lockbox.md
agusduha: We have a few questions about this:

1. What is the motivation behind this change?
2. What about gas costs? Wouldn’t deploying a contract solely for making a transfer increase costs?
3. Is `donateETH` being used anywhere? If not, could we rename it to make its functionality clearer?

specs/interop/shared-lockbox.md
agusduha: In the new implementation code, we’re using custom errors throughout. However, in the specs, we’ve been using require statements to keep the reading simpler.

specs/interop/optimism-portal-interop.md
```diff
@@ -0,0 +1,49 @@
...
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Overview](#overview)
+  - [Interface and properties](#interface-and-properties)
+  - [Integrating `SharedLockbox`](#integrating-sharedlockbox)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
```
tynes: We should be clear that its possible to upgrade to this hardfork and not be part of a cluster. The requirement would be that a chain has its own lockbox

specs/interop/optimism-portal-interop.md
```diff
@@ -0,0 +1,49 @@
...
+
+The `OptimismPortal` contract is upgraded to integrate the `SharedLockbox` and start using the shared ETH liquidity.
+
+### Interface and properties
+
+The `OptimismPortal` contract will add the following storage layout and modified functions:
+
+**`SHARED_LOCKBOX`**
```
tynes: nit: prefer `####` over `** ... **` so that hyperlinks can be generated to specific sections

specs/interop/dependency-set.md
```diff
@@ -6,6 +6,9 @@
 
 - [Chain ID](#chain-id)
 - [Updating the Dependency Set](#updating-the-dependency-set)
+- [Architecture](#architecture)
+  - [L2 DependencyManager](#l2-dependencymanager)
+  - [L1 SuperchainConfigInterop](#l1-superchainconfiginterop)
```
tynes: small nit: we will be merging `SuperchainConfigInterop` into the mainline `SuperchainConfig` as soon as we know interop is the next release. Its better to just call it by the base name `SuperchainConfig` so that we don't go out of date here

specs/interop/optimism-portal-interop.md
```diff
@@ -0,0 +1,119 @@
...
+    - [`_unlockETH`](#_unlocketh)
+- [Events](#events)
+  - [`ETHMigrated`](#ethmigrated)
+- [Invariants](#invariants)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+The `OptimismPortalInterop` contract extends `OptimismPortal2` to integrate with the `SharedLockbox`
```
tynes: nit: we should just call it `OptimismPortal`, we will rename everything to that

specs/interop/superchain-config.md
```diff
@@ -0,0 +1,127 @@
...
+  - [`dependencySetSize`](#dependencysetsize)
+- [Events](#events)
+  - [`DependencyAdded`](#dependencyadded)
+- [Invariants](#invariants)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+The `SuperchainConfigInterop` contract extends `SuperchainConfig` to manage the dependency set and portal authorization
```
tynes: nit: we should just call it `SuperchainConfig`

tynes:
These specs are great!

tynes:
looks like linting is failing

```
specs/interop/superchain-config.md:109 MD012/no-multiple-blanks Multiple consecutive blank lines [Expected: 1; Actual: 2]
```

specs/interop/dependency-set.md
```diff
@@ -47,6 +50,39 @@ It is a known issue that not all software in the Ethereum ecosystem can handle 3
...
+2. The L1 `SuperchainConfig` contract processes this withdrawal and updates the L1-side dependency set
+
+## Architecture
+
+### L2 DependencyManager
+
+The L2 `DependencyManager` is a predeploy contract (0x4200000000000000000000000000000000000029)
+that manages the L2-side dependency set. It:
+
+- Maintains the list of chain IDs in the dependency set
```
ControlCplusControlV: it might be worth noting the chain ID + Superchain config + System Config are all tracked together in unison

specs/interop/predeploys.md
```diff
@@ -816,6 +825,78 @@ The bridging of `SuperchainERC20` using the `SuperchainTokenBridge` will require
...
+
+#### `DependencyAdded`
+
+MUST be triggered when a new dependency is added to the set.
+
+```solidity
+event DependencyAdded(uint256 indexed chainId);
+```
+
+### Invariants
```
ControlCplusControlV: A mention of `ADD_DEPENDENCY_WITHDRAWWAL_GAS_LIMIT` would be good here, as its a constant and assumed to be an invariant

specs/interop/shared-lockbox-upgrade.md
```diff
@@ -0,0 +1,123 @@
...
+  - [L1 Side: Process Addition](#l1-side-process-addition)
+  - [ETH Migration](#eth-migration)
+- [Diagram](#diagram)
+- [Future Considerations / Additional Notes](#future-considerations--additional-notes)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+When a new chain joins the op-governed dependency set, it must integrate with the `SharedLockbox`
```
ControlCplusControlV: I think we should distinguish what op-governed means here, and realistically just say "Superchain" unless we want to mention other interop dependency sets

specs/interop/shared-lockbox-upgrade.md
```diff
@@ -0,0 +1,123 @@
...
+   - Verifying the chain ID isn't already in the dependency set
+   - Adding the chain ID to the dependency set
+   - Getting the chain's portal address from its `SystemConfig`
+
+3. The portal is authorized in the `SharedLockbox`:
+
+   - Verifying the portal uses the correct `SuperchainConfig`
+   - Checking the portal isn't already authorized
+   - Adding the portal to the authorized portals mapping
+   - Triggering ETH migration via `migrateLiquidity()`
```
ControlCplusControlV: I would be very explicit here that this is the step where the eth actually changes hands

specs/interop/shared-lockbox-upgrade.md
```diff
@@ -0,0 +1,123 @@
...
+
+4. The `OptimismPortal` migrates its ETH:
+   - Sets migrated flag to true
```
ControlCplusControlV: I would include the second order effect of why this matters

specs/interop/shared-lockbox-upgrade.md
```diff
@@ -0,0 +1,123 @@
...
+   - `SharedLockbox` emits `ETHLocked` event
+   - Portal emits `ETHMigrated` event
+
+### ETH Migration
+
+After authorization, the `OptimismPortal`'s ETH liquidity is migrated to the `SharedLockbox`:
+
+1. The `OptimismPortal`'s ETH balance is transferred to the `SharedLockbox`
+2. The `OptimismPortal` is configured to use the `SharedLockbox` for all future ETH operations
+3. The migration is marked as complete
```
ControlCplusControlV: why does this matter? What does this actually mean for the system beyond the above mentioned `using SharedLockBox for all future ETH operations.`

specs/interop/shared-lockbox-upgrade.md
```diff
@@ -0,0 +1,123 @@
...
+    Config->>Portal: migrateLiquidity()
+    Portal->>Lockbox: lockETH(balance)
+    Lockbox->>Config: authorizedPortals(portal)
+    Lockbox-->>Lockbox: emits ETHLocked
+    Portal-->>Portal: emits ETHMigrated
+```
+
+## Future Considerations / Additional Notes
+
+- Before calling `addDependency`, it MUST be ensured that the `chainId` and `systemConfig` match
```
ControlCplusControlV: I would be more explicit about what `match` means in this context

specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,138 @@
...
+The `SharedLockbox` contract is proxied and managed by the L1 `ProxyAdmin`.
+
+### Interface and properties
+
+**`lockETH`**
+
+Deposits and locks ETH into the lockbox's liquidity pool.
+
+- The function MUST accept ETH.
+- Only authorized `OptimismPortal` addresses MUST be allowed to interact.
```
ControlCplusControlV: I understand this intuitively, but its worth mentioning why SELFDESTRUCT or other forceful eth send methods (CL rewards address lol) don't matter here

specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,138 @@
...
+- The function MUST NOT revert when called by an authorized `OptimismPortal`
+- The function MUST emit the `ETHLocked` event with the `portal` that called it and the `amount`.
+
+```solidity
+function lockETH() external payable;
+```
+
+**`unlockETH`**
```
ControlCplusControlV: You should should specify where it actually withdraws the given amount of ETH to

specs/interop/shared-lockbox.md
```diff
@@ -0,0 +1,138 @@
...
+  by the `SharedLockbox` of the cluster
+  > With "withdrawable amount", the ETH balance held on `ETHLiquidity` is excluded
+
+### Contract level invariants
+
+- It MUST allow only authorized portals to lock ETH
+
+- It MUST allow only authorized portals to unlock ETH
+
+- It MUST be in paused state if the `SuperchainConfig` is paused
```
ControlCplusControlV: Is there a case where `SuperchainConfig` is paused but this isn't?

specs/interop/superchain-config.md
```diff
@@ -0,0 +1,126 @@
...
+- `dependencySet`: An enumerable set containing the chain IDs of all chains in the dependency set.
+  This set is used to track which chains are part of the interoperable cluster.
+  The set is append-only - chains cannot be removed once added.
+
+- `authorizedPortals`: A mapping that tracks which `OptimismPortal` contracts are authorized to interact
+  with the `SharedLockbox`.
+  When a chain is added to the dependency set, its portal is authorized to lock and unlock ETH from
+  the shared liquidity pool.
+
+## Interface and properties
```
ControlCplusControlV: we should really consider a function to see if a given chainId is within the dependency set

ControlCplusControlV (commented):
some small comments / nits but it looks great overall!

specs/interop/shared-lockbox-upgrade.md
agusduha: hmm I believe this should be common for every cluster so maybe finding a more general definition could be better

specs/interop/shared-lockbox.md
agusduha: The main reasons are for a more restricted interaction that allows better monitoring, do you think is worth mention that here?

specs/interop/shared-lockbox.md
agusduha: That case shouldn't happen

specs/interop/eth-lockbox.md
```diff
@@ -0,0 +1,244 @@
...
+#### `receiveLiquidity`
+
+Receives the ETH liquidity from another `ETHLockbox`.
+
+- Only an authorized `ETHLockbox` can call the function.
+- The function MUST emit the `LiquidityReceived` event with the `lockbox` that is being received from.
+
+```solidity
+function receiveLiquidity() external payable;
+```
```
smartcontracts, line 131: Maybe worth adding a function that exposes the admin address

specs/interop/optimism-portal.md
```diff
@@ -0,0 +1,101 @@
...
+- MUST be called during `depositTransaction` when there is ETH value
+- MUST lock any ETH value in the `ETHLockbox`
+- MUST NOT revert on zero value
+
+#### `_unlockETH`
+
+Called during withdrawal finalization to handle ETH unlocking.
+
+```solidity
+function _unlockETH(Types.WithdrawalTransaction memory _tx) internal;
```
smartcontracts: Maybe easier if this just takes an amount

specs/interop/eth-lockbox.md
agusduha: The admin address or the admin owner address? Because the `Proxy` already has a `admin()` func that can be called, but we should get the owner out of there

specs/interop/optimism-portal.md
agusduha: Yeah this was used for some checks, is outdated now, changing

specs/interop/eth-lockbox.md
smartcontracts: Ah I mean admin owner address

specs/interop/eth-lockbox.md
```diff
@@ -0,0 +1,244 @@
...
+The `ETHLockbox` contract is proxied and managed by the L1 `ProxyAdmin`.
+
+### Interface and properties
+
+#### `lockETH`
+
+Deposits and locks ETH into the lockbox's liquidity pool.
+
+- The function MUST accept ETH.
+- Only authorized `OptimismPortal` addresses MUST be allowed to interact.
```
smartcontracts, line 61: Should we consider checking if the `OptimismPortal` is currently sending a cross-chain message? I don't think we want users to be able to do L2 => L1 messages that trigger the lockbox. We can use the unsafe addresses mechanism but checking if the l2Sender variable is set could be an extra safety net.

agusduha: We were doing a check to verify that the target was not the lockbox in the previous PR, we should do that here too. In the `AnchorStateRegistry` PR you added a `_isUnsafeTarget` function, we can add the lockbox inside that check too so that we avoid calls to it.

specs/interop/eth-lockbox.md
```diff
@@ -0,0 +1,255 @@
...
+
+Authorizes another `ETHLockbox` to migrate its ETH liquidity to the current `ETHLockbox`.
+
+- Only the `ProxyAdmin` owner can call the function.
+- The function MUST emit the `LockboxAuthorized` event with the `lockbox` that is being authorized.
+- The function MUST NOT allow the same `ETHLockbox` to be authorized more than once.
+
+```solidity
+function authorizeLockbox(address _lockbox) external;
+```
```
mds1: Should these also have a variant of the "The `ProxyAdmin` owner of the source lockbox must be the same as the `ProxyAdmin` owner of the destination lockbox" requirement that `migrateLiquidity` has? For `authorizePortal` the portal and lockbox would have the same proxy admin owner, and for `authorizeLockbox` the two lockboxes would have the same proxy admin owner

specs/interop/eth-lockbox.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Invariants
+
+### System level invariants
+
+- The ETH held in the `ETHLockbox` MUST never be less than the amount deposited but not yet withdrawn by the `OptimismPortal`s
+
+- The total withdrawable ETH amount present on all the dependency set's chains MUST NEVER be more than the amount held
+  by the `ETHLockbox` of the cluster
+  > With "withdrawable amount", the ETH balance held on `ETHLiquidity` is excluded
```
mds1, line 201: nit, is L197 in quotes accidentally? 
image

specs/interop/eth-lockbox.md
```diff
@@ -0,0 +1,255 @@
...
+
+MUST be triggered when `receiveLiquidity` is called
+
+```solidity
+event LiquidityReceived(address indexed lockbox);
+```
+
+## Invariants
+
+### System level invariants
```
mds1, line 193: Another one is that all chains in a dependency set, and their lockbox, must have the same `adminOwner()`

specs/interop/eth-lockbox.md
```diff
@@ -0,0 +1,255 @@
...
+
+The merge process is the process of migrating the ETH liquidity from one `ETHLockbox` to another.
+
+- The destination `ETHLockbox` MUST call `authorizeLockbox` with the source `ETHLockbox` as argument.
+
+- The source `ETHLockbox` MUST call `migrateLiquidity` with the destination `ETHLockbox` as argument.
+
+- `migrateLiquidity` MUST call `receiveLiquidity` from the destination `ETHLockbox`.
+
+This process ensures that the ETH liquidity is migrated from the source `ETHLockbox` to the correct destination `ETHLockbox`.
```
mds1: Let's specify if this is intended to be an atomic operation, or separate transactions. Since all involved contracts have the same `adminOwner()` I assume atomic?

specs/interop/optimism-portal.md
```diff
@@ -0,0 +1,99 @@
...
+Migrates the ETH liquidity to the `ETHLockbox`. This function will only be called once by the
+`ProxyAdmin` owner when updating the `OptimismPortal` contract.
+
+```solidity
+function migrateLiquidity() external;
+```
+
+- MUST only be callable by the `ProxyAdmin` owner
+- MUST transfer all ETH balance to the `ETHLockbox`
+- MUST emit an `ETHMigrated` event with the amount transferred
```
mds1, line 50: Given that it "will only be called once", is there a requirement that this is a no-op on subsequent calls? Or should it revert on subsequent calls?

specs/interop/optimism-portal.md
```diff
@@ -0,0 +1,99 @@
...
+
+Called during withdrawal finalization to handle ETH unlocking.
+
+```solidity
+function _unlockETH(uint256 _amount) internal;
+```
+
+- MUST be invoked during withdrawal finalization when there is ETH value
+- MUST unlock the withdrawal value from the `ETHLockbox`
+- MUST NOT revert on zero value
```
mds1: > - MUST be invoked during withdrawal finalization when there is ETH value
> - MUST NOT revert on zero value

These two feel a bit odd/inconsistent—you can call `_unlockETH` with 0 value, but withdrawal finalization will not invoke it if there is zero value?

mds1 (approved):
All comments are small and easy to fix, so approving

specs/interop/eth-lockbox.md
agusduha: okey, we can add a common check for every function

specs/interop/eth-lockbox.md
agusduha: It is intentional, but we can change the style if it is confusing

specs/interop/optimism-portal.md
agusduha: We removed the "will only be called once" invariant because we think is not necessary, there is no problem if its called again

specs/interop/optimism-portal.md
agusduha: Yeah the second one is unnecessary, removing

specs/interop/eth-lockbox.md
agusduha: Added a `l2Sender` check invariant

smartcontracts (approved):

protolambda (approved):

PR #468, opened 2024-11-28
Add opcm upgrades spec

maurelian:
**Description**

1. Updates the spec to match the current implementation.
2. Adds specs for `upgrade()` based on [the design doc](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/l1-upgrades.md).

Note that I have included the `addGameType()` method discussed in the design doc, but serious questions have come up around its feasibility. That content is included for discussion, but can be reverted with this commit: 115c61eb697d6f488df739ae7302e21ec6e2ff19


specs/experimental/op-contracts-manager.md
```diff
@@ -201,16 +306,27 @@ OP Contracts Manager does not consider factors such as these. The EVM supports
...
+### Atomicity of upgrades
+
+Although atomicity of a superchain upgrade is not essential for many types of upgrade, it will
+at times be necessary. It is certainly always desirable for operational reasons.
 
-This contract is upgradeable, and breaking changes are expected, as upgrades
-are required to update the contracts release that is deployed. This is because
-the required inputs to the `deploy` method may change as new contract releases
-are supported. Therefore, if calling this contract from another contract, be
-sure to account for future breaking changes to the ABI.
```
maurelian, line 216: weird diff display here. 
The upgradeability content was removed.

specs/experimental/op-contracts-manager.md
```diff
@@ -153,27 +166,119 @@ This provides the following benefits:
...
+3. Calls `setAnchorState()` on the `AnchorStateRegistry`
+
+Note that in [Standard Configuration chains](../protocol/superchain-configuration.md):
+
+- `DisputeGameFactory.setImplementation()` is authorized to the Upgrade Controller
+- `AnchorStateRegistry.setAnchorState()` is authorized to the Guardian
+
+Since the Guardian is its own Safe controlled by the Security Council, and the Upgrade Controller is
+a 2 of 2 jointly controlled by the Security Council and Optimism Foundation, this is not impossible
+to implement, but will be ugly and require new tooling and processes to do correctly.
```
maurelian, line 267: As mentioned in the description, I'm doubting the value of this method vs. just having reusable superchain-ops tasks for adding a new game. I'm generally in favor of scrapping `addGameType()` for now.

specs/experimental/op-contracts-manager.md
```diff
@@ -18,50 +18,90 @@ of governance approved [contract releases] can be found on the
...
 
-## Interface
+/// @notice Maps an L2 chain ID to an L1 batch inbox address as defined by the standard
+function chainIdToBatchInboxAddress(uint256 _l2ChainId) public pure returns (address);
 
-Version 1.0.0 of the OP Contracts Manager deploys the `op-contracts/v1.6.0`
-contracts release.
+/// @notice Returns the blueprint contract addresses.
+function blueprints() public view returns (Blueprints memory);
+```
```
maurelian, line 88: I moved this section up from below, and edited it to match the current implementation. Except that `latestRelease()` doesn't currently exist on the OPCM, but should be added.

specs/experimental/op-contracts-manager.md
maurelian: As mentioned in the description, I'm doubting the value of this method vs. just having reusable superchain-ops tasks for adding a new game. I'm generally in favor of scrapping `addGameType()` for now.

specs/experimental/op-contracts-manager.md
```diff
@@ -153,27 +166,119 @@ This provides the following benefits:
...
+It has the following interface:
+
+```solidity
+function upgrade(ISystemConfig[] _systemConfigs, NewChainConfig[] _newConfigs) public;
+```
+
+For each chain successfully upgraded, the following event is emitted:
+
+```solidity
+event Upgraded(uint256 indexed l2ChainId, SystemConfig indexed systemConfig);
```
maurelian: This doesn't match the `Deployed` event very closely, which is: 

```
event Deployed(
        uint256 indexed outputVersion, uint256 indexed l2ChainId, address indexed deployer, bytes deployOutput
    );
```

Open questions I have: 
- Is there a need for an `upgradeOutput`? What would go in it?
- Is there any benefit to including the `upgrader` address, since it's a DELEGATECALL it will actually be emitted by that address.
- Is it helpful to emit both `l2ChainId` and `systemConfig`?

specs/experimental/op-contracts-manager.md
maurelian: Of course we could add two separate functions which basically do the different steps if it's deemed valuable, it's just not clear to me that it does much to reduce the operational lift.

ajsutton: The key requirement is having some way to actually deploy the `FaultDisputeGame` contract.  The superchain-ops task is easy enough but it only calls `setImplementation` and requires a deployed `FaultDisputeGame` as input. superchain-ops doesn't handle that because you need to be sure to deploy the right version of the contract and OPCM is designed to be the on-chain source of "the right version" (as I understand it).

An `op-deployer` bootstrap command could be used to do the deployment but we need to be crystal clear on what the process is and get it tested so that it works with every contracts release.  The bootstrap scripts for op-deployer have to be in the contracts release and working so it's critical they're actually tested.

ajsutton: Also for the record, when adding a new game type you currently need to reinitialise the `AnchorStateRegistry` - `setAnchorState` can only set the anchor state to a game of the same type that resolved as valid (allowing you to rollback a bad update).  The superchain-ops template can do that re-initialization already and there's work in place to remove the requirement by having a single anchor state used for all game types (that's updated after the air-gap so it's dependable). We aren't blocking on that for Isthmus (it may be ready, not sure) but happy to leave that to the superchain-ops task.

maurelian: > The key requirement is having some way to actually deploy the `FaultDisputeGame` contract.

Ah, I see. I think that along with https://github.com/ethereum-optimism/design-docs/pull/153, that it should be quite doable to deploy and initialize a new proxy to a `Creator` contract implementation.

> when adding a new game type you currently need to reinitialise the `AnchorStateRegistry`

Got it. If reinitializing is the right approach, then we can also do that in the same function on the basis that it would be called by the Upgrade Controller, and afaict that overcomes all my concerns, thanks for clarifying!

maurelian: Updated the spec in this commit: 34e18e4de0d2

specs/experimental/op-contracts-manager.md
```diff
@@ -18,50 +18,90 @@ of governance approved [contract releases] can be found on the
...
+/// @notice The logic address and initializer selector for an implementation contract.
+struct Implementation {
+  address logic; // Address containing the deployed logic contract.
+  bytes4 initializer; // Function selector for the initializer.
+}
+
+/// @notice Returns the latest approved release of the OP Stack contracts.
+/// @notice Release strings follow semver and are named with the
+/// format `op-contracts/vX.Y.Z`.
+function latestRelease() external view returns (string memory);
```
blmalone: This function does not exist on the OPContractsManager. 

I ran the code through ChatGPT and it spit out the following: 

```solidity
    function version() external pure returns (string memory);
    function OUTPUT_VERSION() external pure returns (uint256);
    function superchainConfig() external view returns (address);
    function protocolVersions() external view returns (address);
    function l1ContractsRelease() external view returns (string memory);
    function blueprints() external view returns (Blueprints memory);
    function implementations() external view returns (Implementations memory);
    function chainIdToBatchInboxAddress(uint256 _l2ChainId) external pure returns (address);
    function deploy(DeployInput calldata _input) external returns (DeployOutput memory);
```

specs/experimental/op-contracts-manager.md
```diff
@@ -18,50 +18,90 @@ of governance approved [contract releases] can be found on the
...
+2. Upgrade the contracts for an existing OP chain from the previous release to the new release.
+3. Orchestrate adding a new game type on a per-chain basis
+
+## Getter Methods
+
+The following interface defines the available getter methods:
+
+```solidity
+/// @notice The logic address and initializer selector for an implementation contract.
+struct Implementation {
```
blmalone: Similar to the comment below, this is now outdated. Perhaps we can remove this section and link to the most recent version of [`OPContractsManager.sol`](https://github.com/ethereum-optimism/optimism/blame/develop/packages/contracts-bedrock/src/L1/OPContractsManager.sol)? Or just include public/external functions I listed below.

specs/experimental/op-contracts-manager.md
blmalone: Ah just saw this comment. I think `l1ContractsRelease` acts the same as the `latestRelease()` function you mentioned.

specs/experimental/op-contracts-manager.md
```diff
@@ -18,50 +18,90 @@ of governance approved [contract releases] can be found on the
...
 
-The OP Contracts Manager is a proxied contract using the standard `Proxy.sol` contract that lives in
-the Optimism monorepo. Therefore the OP Contracts Manager will have the same interface as the
-`Proxy.sol`, in addition to other methods defined in this specification.
+### Interface
 
-The privileged methods of the OP Contracts Manager will be held by the L1 ProxyAdmin owner, as
-specified by the [standard configuration].
+Version 1.0.0 of the OP Contracts Manager deploys the `op-contracts/v1.6.0` contracts release,
+and is deployed at `0x9BC0A1eD534BFb31a6Be69e5b767Cba332f14347`. In the future this will
```
blmalone: I think this shouldn't be called version 1.0.0 of OPCM anymore right? The version of OPCM is tied to the version of the l1 smart contract release that it deploys. 

Also, I think it would be helpful to link to the OPCM addresses stored in op-deployer [here](https://github.com/ethereum-optimism/optimism/blob/564b0c17198bff7c3eebc7c328e1bf57b9a46188/op-deployer/pkg/deployer/standard/standard.go#L202-L217). We've yet to properly added these to superchain-registry and ensure they're kept up to date. 

In fact, the addresses in scr are currently wrong,
- [mainnet](https://github.com/ethereum-optimism/superchain-registry/blob/a54d203e727f1cdeff06b883e239febf92b4b1d6/superchain/configs/mainnet/superchain.toml#L4)
- [sepolia](https://github.com/ethereum-optimism/superchain-registry/blob/a54d203e727f1cdeff06b883e239febf92b4b1d6/superchain/configs/sepolia/superchain.toml#L4)

Unrelated, but we're going to need to find a way to properly store these in superchain-registry.

specs/experimental/op-contracts-manager.md
```diff
@@ -18,50 +18,90 @@ of governance approved [contract releases] can be found on the
...
+#### `deploy`
 
-The `deploy` method is the only non-view method in the contract. It is used to
-deploy the full set of L1 contracts required to setup a new OP Stack chain that
-complies with the [standard configuration]. It has the following interface:
+The `deploy` method is used to deploy the full set of L1 contracts required to setup a new OP Stack
+chain that complies with the [standard configuration]. It has the following interface:
 
 ```solidity
 struct Roles {
```
blmalone: This code is now outdated and the function signature for `deploy` now looks like this: `function deploy(DeployInput calldata _input) external returns (DeployOutput memory)`. 

Again, might be worth linking to the latest code on develop for anyone reading to get up to speed.

specs/experimental/op-contracts-manager.md
```diff
@@ -85,58 +125,31 @@ The `l2ChainId` has the following restrictions:
...
 
-### Batch Inbox Address
+#### Batch Inbox Address
 
 The chain's [Batch Inbox] address is computed at deploy time using the recommend approach defined
 in the [standard configuration]. This improves UX by removing an input, and ensures uniqueness of
 the batch inbox addresses.
 
-### Contract Deployments
+#### Contract Deployments
```
blmalone, line 134: I just started a message on discord about this, I know this isn't part of upgrades but I think it's worth chatting through. 
https://discord.com/channels/1244729134312198194/1273519241076670526/1314283100536311809

blmalone (commented):
I'm about 50% the way through this, I'll finish it off soon.

specs/experimental/op-contracts-manager.md
blmalone, line 134: There's a little more nuance ot how the salts are being calculated, perhaps if [this PR](https://github.com/ethereum-optimism/optimism/pull/13273) gets merged you can link to it.

specs/experimental/op-contracts-manager.md
```diff
@@ -153,13 +166,98 @@ This provides the following benefits:
...
+
+#### `upgrade`
+
+The `upgrade` method is used by the Upgrade Controller to upgrade the full set of L1 contracts for
+all chains that it controls.
+
+It has the following interface:
+
+```solidity
+function upgrade(ISystemConfig[] _systemConfigs, NewChainConfig[] _newConfigs) public;
```
blmalone: The interface in the design doc is slightly different to this, it has the superchain proxy admin as a parameter: https://github.com/ethereum-optimism/design-docs/blob/main/protocol/l1-upgrades.md#release-process

specs/experimental/op-contracts-manager.md
blmalone: 1. Not sure we would need `upgradeOutput` on the basis that all the information you need about an upgrade should be implicit from the OPCM upgrade function itself. 
2. I see your point that it'll be emitted by the upgrader. I don't see any downside of putting it into the event log though. 
3. I think so.

specs/experimental/op-contracts-manager.md
```diff
@@ -153,13 +166,98 @@ This provides the following benefits:
...
+2. For each `_systemConfig`, the list of addresses in the chain is retrieved.
+3. For each address, a two step upgrade is used where:
+   1. the first upgrade is to an `InitializerResetter` which resets the `initialized` value.
+   1. the implementation is updated to the final address and `upgrade()` is called on that address.
+
+This approach requires that all contracts have an `upgrade()` function which sets the `initialized`
+value to `true`. The `upgrade` function body should be empty unless it is used to set a new state
+variable added to that contract since the last upgrade.
+
+#### `NewChainConfig` struct
```
blmalone: For my own understanding, what do you see an example NewChainConfig struct looking like. The design doc has:
```solidity
struct NewChainConfig {
  address newValues;
  address disputeGameBlueprint;
}
```
If contract X's upgrade method needs the value 42 for example. Does this struct store the value 42?

specs/experimental/op-contracts-manager.md
```diff
@@ -153,13 +166,98 @@ This provides the following benefits:
...
+In practice, this struct is likely to be have a unique name for each release of the OP Contracts
+Manager.
+
+#### Requirements on the OP Chain contracts
+
+In general, all contracts used in an OP Chain SHOULD be proxied with a single shared implementation.
+This means that all values which are not constant across OP Chains SHOULD be held in storage rather
+than the bytecode of the implementation.
+
+Any contracts which do not meet this requirement will need to be deployed by the `upgrade()`
```
blmalone, line 242: Do we have an example of a contract that exists today which falls into this category? By this I mean, do some contracts in the OP contract ecosystem require immutable variables which differ between op chains AND this data can't be held in storage for whatever reason?

specs/experimental/op-contracts-manager.md
```diff
@@ -153,13 +166,98 @@ This provides the following benefits:
...
+function, increasing the cost and reducing the number of OP Chains which can be atomically upgraded.
+
+## Adding game types
+
+Because different OP Chains within a Superchain may use different dispute game types, and are
+expected to move from a permissioned to permissionless game over time, an `addGameType()` method is
```
blmalone, line 248: Is this one such example referencing this my previous question?

blmalone: Also is this another way of saying, chains in the interop set can coexist with different dispute game types?

specs/experimental/op-contracts-manager.md
```diff
@@ -153,13 +166,98 @@ This provides the following benefits:
...
+}
+
+function addGameType(ISystemConfig[] _systemConfigs, NewGameConfig[] _newGames) public;
+```
+
+### Implementation
+
+The high level logic of the `addGameType` method is as follows (for each chain):
+
+1. The Upgrade Controller Safe will `DELEGATECALL` to the `OPCM.upgrade()` method.
```
blmalone: Should this be: 
```suggestion
1. The Upgrade Controller Safe will `DELEGATECALL` to the `addGameType` method.
```

blmalone (commented):
This is looking good so far. Have some comments and questions.

specs/experimental/op-contracts-manager.md
maurelian: The `FaultDisputeGame` and `PermissionedDisputeGame` contracts currently contain chain specific immutables ([ex](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/dispute/FaultDisputeGame.sol#L124). 

There is [work](https://github.com/ethereum-optimism/design-docs/pull/161) being done to address that, so this may not be an issue, but it seems worthwhile to document how it would be handled if not.

specs/experimental/op-contracts-manager.md
maurelian: > Is this one such example referencing this my previous question?

Not exactly. While it is the Dispute Game contracts which contain chain specific immutable (AKA are not MCP-L1 compatible), this sentence is referring to the fact that some chains use only the `PermissionedDisputeGame`, whereas others use the `FaultDisputeGame`. In the case of chains which use the `FaultDisputeGame`, they also have the `PermissionedDisputeGame` deployed and registered so that they can fallback to the permissioned game in case there is a bug in the permissionless game (`FaultDisputeGame`).

> Also is this another way of saying, chains in the interop set can coexist with different dispute game types?

No, it's not attempting to make any statements about interop.

specs/experimental/op-contracts-manager.md
maurelian: Nice. @mds1 also has a very nice `interface` shell function you can use for this :)

https://gist.github.com/mds1/3f070676129a095dec372c2d02cedfdd#file-ethrc-sh-L134

specs/experimental/op-contracts-manager.md
maurelian: I think best to just remove this. Even though it appears in the return type of the `implementations()` function (now listed below), it think that the natspec description is sufficient to clarify what the struct is.

specs/experimental/op-contracts-manager.md
maurelian: OK, I've added the current getters here with some natspec to clarify them all.

specs/experimental/op-contracts-manager.md
maurelian: agreed. `l1ContractsRelease() is now listed.

specs/experimental/op-contracts-manager.md
maurelian: > I think this shouldn't be called version 1.0.0 of OPCM anymore right? The version of OPCM is tied to the version of the l1 smart contract release that it deploys.

I may have previously written or said otherwise somewhere, but I do think the OPCM should have both its own `version()` and `l1ContractsRelease()` as independent versions.
Otherwise we'll increment the OPCM version much less frequently, and since the OPCM is its own contract which can vary independently of changes made to other L1 contracts, it feels conservative to be consistent with the practice used in other contracts.

specs/experimental/op-contracts-manager.md
maurelian: Updated with some natspec.

specs/experimental/op-contracts-manager.md
maurelian: Ahh, yes, that part of the design doc is outdated. It had previously included a proposal to consolidate ALL proxy admin contracts into a single `SuperchainProxyAdmin`. I should have updated the interface
when I removed that proposal but did not. But I think that's OK since the spec will superced the design doc and we don't have a requirement of keeping them current.

specs/experimental/op-contracts-manager.md
maurelian: yes, exactly.

specs/experimental/op-contracts-manager.md
maurelian: Updated to reflect the nuance.

specs/experimental/op-contracts-manager.md
blmalone: Agreed.

specs/experimental/op-contracts-manager.md
blmalone: Nice, I'm curious how we know which proxy admin to call for each proxy. Is it by simply getting the owner of each system config contract?

specs/experimental/op-contracts-manager.md
blmalone: Nice.

specs/experimental/op-contracts-manager.md
mds1: 1. Agreed we don't need `upgradeOutput`, unsure what that would include
2. Agreed we may as well include this address, it makes it easier to retrieve the address without having to trace to see the address that emitted that event
3. Seems like we may as well, this could be useful for running post-checks to analyze logs to make sure all expected chains were ugpraded

specs/experimental/op-contracts-manager.md
maurelian: great callout, I think we'll actually need to provide both the list of SystemConfigs and ProxyAdmins. We can use the superchain-registry to get them and ensure they are matching pairs.

maurelian:
```suggestion
function upgrade(ISystemConfig[] _systemConfigs, IProxyAdmin[] _proxyAdmins, NewChainConfig[] _newConfigs) public;
```

specs/experimental/op-contracts-manager.md
maurelian:
```suggestion
event Upgraded(uint256 indexed l2ChainId, SystemConfig indexed systemConfig, address indexed upgrader);
```

specs/experimental/op-contracts-manager.md
maurelian: We agreed we won't make any changes to the current `version()` approach.

blmalone (approved):
LGTM

specs/experimental/op-contracts-manager.md
```diff
@@ -20,146 +20,247 @@ of governance approved [contract releases] can be found on the
...
+    /// @notice Semver version specific to the OPContractsManager
+    function version() external view returns (string memory);
+
+/// @notice Returns the implementation contract addresses.
+function implementations() public view returns (Implementations memory);
 
-### `deploy`
+/// @notice Maps an L2 Chain ID to the SystemConfig address for that chain.
+/// @notice All contracts for a chain can be found from its SystemConfig.
+function systemConfigs(uint256 chainId) external view returns (SystemConfig);
```
tynes: Some arguments have prefixed underscore, others do not

specs/experimental/op-contracts-manager.md
```diff
@@ -20,146 +20,247 @@ of governance approved [contract releases] can be found on the
...
+
+### Interface
 
-function deploy(
-  uint256 l2ChainId,
-  uint32 basefeeScalar,
-  uint32 blobBasefeeScalar,
-  Roles roles
-) external returns (SystemConfig)
+Version 1.0.0 of the OP Contracts Manager deploys the `op-contracts/v1.6.0` contracts release,
```
tynes: This is going to fall out of date pretty quickly, "in the future this will...". Perhaps we want a table of opcm version -> contracts release

specs/experimental/op-contracts-manager.md
```diff
@@ -20,146 +20,247 @@ of governance approved [contract releases] can be found on the
...
+be tracked in the `superchain-registry`.
+
+#### `deploy`
+
+The `deploy` method is used to deploy the full set of L1 contracts required to setup a new OP Stack
+chain that complies with the [standard configuration]. It has the following interface:
+
+```solidity
+/// @notice Deploys a new OP Chain
+/// @param _input DeployInput containing chain specific config information.
```
tynes, line 101: What is the definition of `DeployInput`?

specs/experimental/op-contracts-manager.md
```diff
@@ -20,146 +20,247 @@ of governance approved [contract releases] can be found on the
...
+#### Batch Inbox Address
 
 The chain's [Batch Inbox] address is computed at deploy time using the recommend approach defined
 in the [standard configuration]. This improves UX by removing an input, and ensures uniqueness of
 the batch inbox addresses.
 
-### Contract Deployments
+#### Contract Deployments
 
 All contracts deployed by the OP Contracts Manager are deployed with CREATE2, with a
```
tynes: This would read better if it said 

```
All contracts deployed by the OP Contracts Manager are deployed with CREATE2.

For singletons the following salt is used:

...

For `Proxy` contracts, the following salt is used:

...
```

This removes ambiguity around former/latter, especially for esl speakers

specs/experimental/op-contracts-manager.md
```diff
@@ -20,146 +20,247 @@ of governance approved [contract releases] can be found on the
...
+2. For each `_systemConfig`, the list of addresses in the chain is retrieved.
+3. For each address, a two step upgrade is used where:
+   1. the first upgrade is to an `InitializerResetter` which resets the `initialized` value.
+   1. the implementation is updated to the final address and `upgrade()` is called on that address.
+
+This approach requires that all contracts have an `upgrade()` function which sets the `initialized`
+value to `true`. The `upgrade` function body should be empty unless it is used to set a new state
+variable added to that contract since the last upgrade.
+
+#### `NewChainConfig` struct
```
tynes: What is the definition?

specs/experimental/op-contracts-manager.md
```diff
@@ -20,146 +20,247 @@ of governance approved [contract releases] can be found on the
...
+
+### Interface
+
+#### `addGameType`
+
+The `addGameType` method is used to orchestrate the actions required to add a new game type to one
+or more chains.
+
+```solidity
+struct NewGameConfig {
```
tynes: You should define the game types and the definitions per game type. This spec is not clear enough to enable a second implementation or a reviewer to know that an impl is correct

specs/experimental/op-contracts-manager.md
maurelian: It's [large](https://github.com/ethereum-optimism/optimism/blob/upgrade-against-registry/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L46) and will change everytime a new configurable state var is added. 

I left it out to minimize the maintenance burden of this file.

specs/experimental/op-contracts-manager.md
maurelian: Good catch. There were also duplicate methods. 
178e253dd42414bd3a5cb2c8782435953b54f110

specs/experimental/op-contracts-manager.md
maurelian: This statement isn't new, just relocated from elsewhere in the doc.

I've removed it, as it feels like user facing documentation, not instructions for implementation.

60ecdec71700fbbbe7605edd616b16cd24753f6e

specs/experimental/op-contracts-manager.md
maurelian: good rewrite tyvm: 6de833da8ce2

specs/experimental/op-contracts-manager.md
maurelian: Addressed: e1c854188bee

specs/experimental/op-contracts-manager.md
```diff
@@ -18,148 +18,262 @@ of governance approved [contract releases] can be found on the
...
+
+```solidity
+/// @notice The full set of inputs to deploy a new OP Stack chain.
+struct DeployInput {
+    Roles roles;
+    uint32 basefeeScalar;
+    uint32 blobBasefeeScalar;
+    uint256 l2ChainId;
+    // The correct type is AnchorStateRegistry.StartingAnchorRoot[] memory,
+    // but OP Deployer does not yet support structs.
```
tynes: support arrays? `Roles` is a struct

specs/experimental/op-contracts-manager.md
```diff
@@ -18,148 +18,262 @@ of governance approved [contract releases] can be found on the
...
-### `deploy`
+### Interface
 
-The `deploy` method is the only non-view method in the contract. It is used to
-deploy the full set of L1 contracts required to setup a new OP Stack chain that
-complies with the [standard configuration]. It has the following interface:
+#### `deploy`
 
-```solidity
-struct Roles {
```
tynes, line 67: Where was this definition moved to?

specs/experimental/op-contracts-manager.md
maurelian: I added the following content below: 

---

By way of example, if an upgrade is adding a new variable `address foo` to the `SystemConfig` contract, for
an upgrade named `Example`, the struct could have the following definition:

```solidity
struct NewChainConfigForExampleUpgrade {
  address systemConfigFoo;
}
```

We could create a structure in this document wherein each upgrade is expected to document the specific struct required, but I think this is sufficient instruction for an implementer, and I'm hesitant to add that overhead to each upgrade.

specs/experimental/op-contracts-manager.md
maurelian: Having just learned that the [Proofs contracts refactor](https://github.com/ethereum-optimism/design-docs/pull/153) has been deprioritized, I will need to rewrite this section. Will let you know when that's done.

specs/experimental/op-contracts-manager.md
maurelian: I've restored it: f3262e1f3b43

specs/experimental/op-contracts-manager.md
maurelian: The limitation is that neither the [setters](https://github.com/ethereum-optimism/optimism/blob/upgrade-against-registry/packages/contracts-bedrock/scripts/deploy/DeployOPChain.s.sol#L111), nor [getters](https://github.com/ethereum-optimism/optimism/blob/upgrade-against-registry/packages/contracts-bedrock/scripts/deploy/DeployOPChain.s.sol#L162) on the Input/Output contracts that `op-deployer` uses be a complex/reference type, because those I/O contracts are simulated as predeploys in go code, and that simulation does not have the ability to handle the more complicated encodings.

That however is an implementation detail specific to the monorepo, which would not be required for another implementer to know, so I've removed these comments which had been copied from the impl source.

I've also removed the monorepo specific type wrappers from the `DeployInput` and replaced them with their simple type definitions (ie. `Duration` -> `uint64`, `Claim` -> `bytes32`).

specs/experimental/op-contracts-manager.md
```diff
@@ -18,148 +18,266 @@ of governance approved [contract releases] can be found on the
...
+
+### Implementation
+
+The high level logic of the `addGameType` method is as follows (for each chain):
+
+1. The Upgrade Controller Safe will `DELEGATECALL` to the `OPCM.addGameType()` method.
+1. A new Proxy contract will be deployed, with the implementation set to the `Creator` contract for that game type.
+1. Calls `setImplementation()` on the `DisputeGameFactory`
+1. Calls `upgrade()` on the `AnchorStateRegistry` to set the new game type to add a new entry to the `anchors` mapping.
+   The `upgrade()` method should revert if it would overwrite an existing entry.
```
maurelian: Appreciate support from a proofs SME here to: 
1. Update these instructions, as they are currently written for the [Proofs Contracts Refactor](https://github.com/ethereum-optimism/design-docs/pull/153) work. 
2. Indicate what inputs will need to be provided for each new game type.

specs/protocol/configurability.md
```diff
@@ -379,7 +379,7 @@ Address:
 
 **Description:** Account authorized to change values in the SystemConfig contract. All configuration is stored on L1 and
 picked up by L2 as part of the [derivation](./derivation.md) of the L2 chain.<br/>
-**Administrator:** <br/>
+**Administrator:** [L1 Proxy Admin](#admin-roles)<br/>
```
maurelian, line 382: noticed this was missing. Other changes in this file are either from linting or my IDE's formatting, but are purely cosmetic.

specs/experimental/op-contracts-manager.md
maurelian: This commit should address your comment:
7703c51a000f

specs/experimental/op-contracts-manager.md
maurelian: f3262e1f3b43

specs/experimental/op-contracts-manager.md
maurelian: e1c854188bee

specs/experimental/op-contracts-manager.md
```diff
@@ -18,148 +18,282 @@ of governance approved [contract releases] can be found on the
...
 
-### `Proxy.sol`
+The following interface defines the available getter methods:
+
+```solidity
+/// @notice Returns the latest approved release of the OP Stack contracts are named with the
+///         format `op-contracts/vX.Y.Z`.
+function l1ContractsRelease() external view returns (string memory);
+/// @notice Represents the interface version so consumers know how to decode the DeployOutput struct
+function OUTPUT_VERSION() external view returns (uint256);
```
tynes: Why are we mixing snake case and camel case in the interface? We generally agreed on not doing that because it introduces tech debt if we move from something being an immutable to not. This couples the implementation details to the interface

specs/experimental/op-contracts-manager.md
```diff
@@ -18,148 +18,282 @@ of governance approved [contract releases] can be found on the
...
+All contracts deployed by the OP Contracts Manager are deployed with CREATE2.
+
+For singletons the following salt is used:
 
 ```solidity
-/// @notice The logic address and initializer selector for an implementation contract.
-struct Implementation {
-  address logic; // Address containing the deployed logic contract.
-  bytes4 initializer; // Function selector for the initializer.
+keccak256(abi.encode(l2ChainId, saltMixer))
```
blmalone: The implementation of salts have now been simplied as per this pr: https://github.com/ethereum-optimism/optimism/pull/13273 

Now all salts are computed the _same_ way. So both proxied and non-proxied contracts have their salts computed as: 
```solidity
keccak256(abi.encode(_l2ChainId, _saltMixer, _contractName));
```

blmalone (commented):
one comment about how salts are computed

specs/experimental/op-contracts-manager.md
maurelian: Thanks for following up here! I think that's a great improvement which I've reflected here: ec5563349ed1

specs/experimental/op-contracts-manager.md
maurelian: ewww... good catch. This interface was autogenerated from what's on develop

https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L122. 

I created an issue to track fixing the impl.

Which do you think is the right fix here? I think we should just fix it in the spec now and be done with it.

```suggestion
function outputVersion() external view returns (uint256);
```

specs/experimental/op-contracts-manager.md
```diff
@@ -20,146 +20,271 @@ of governance approved [contract releases] can be found on the
...
-### `Proxy.sol`
+The following interface defines the available getter methods:
 
-The OP Contracts Manager is a proxied contract using the standard `Proxy.sol` contract that lives in
-the Optimism monorepo. Therefore the OP Contracts Manager will have the same interface as the
-`Proxy.sol`, in addition to other methods defined in this specification.
+```solidity
+/// @notice Returns the latest approved release of the OP Stack contracts are named with the
+///         format `op-contracts/vX.Y.Z`.
+function l1ContractsRelease() external view returns (string memory);
```
tynes: This is incredibly version, would far prefer `release()(string)`

maurelian: can do

```suggestion
function release() external view returns (string memory);
```

specs/experimental/op-contracts-manager.md
```diff
@@ -20,146 +20,271 @@ of governance approved [contract releases] can be found on the
...
+  - [Safely using `DELEGATECALL`](#safely-using-delegatecall)
+  - [Atomicity of upgrades](#atomicity-of-upgrades)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-## Deployment
+## Overview
 
-The OP Contracts Manager is a proxied contract deployed at `0xTODO`. It can be deployed as follows:
+The OP Contracts Manager refers to a series of contracts, of which a new singleton is deployed
```
tynes, line 53: It could be good to define upgrade controller early in the doc

specs/experimental/op-contracts-manager.md
```diff
@@ -20,146 +20,271 @@ of governance approved [contract releases] can be found on the
...
+1. Deploy and initialize new `DelayedWethProxy` for the new game type, reusing the existing implementation
+1. Deploy a new `FaultDisputeGame` contract. The source of the constructor args is indicated below this list,
+   the value which is not available onchain is the `absolutePrestate`.
+1. Calls `upgrade()` on the `AnchorStateRegistry` to set the new game type to
+   add a new entry to the `anchors` mapping. The `upgrade()` method should
+   revert if it would overwrite an existing entry.
+1. Read the `DisputeGameFactory` address from the `SystemConfig`.
+1. Call `DisputeGameFactory.setImplementation()` to register the new game.
+
+```solidity
```
tynes: This seems better for a table with name, type and description columns

specs/experimental/op-contracts-manager.md
tynes: There is no definition for the first release of opcm? A spec should have all of the information required to rewrite the code from scratch

tynes (approved):

specs/experimental/op-contracts-manager.md
maurelian: 7404d958896c

specs/experimental/op-contracts-manager.md
maurelian: 19cf6cf2a2c5

specs/experimental/op-contracts-manager.md
maurelian: good call

maurelian: 19cf6cf2a2c5

PR #473, opened 2024-12-05
feat: updated DeputyPauseModule specification

smartcontracts:
Updates the DeputyPauseModule specification to utilize a clearer standardized format for invariants and assumptions.

### DIFF
--- a/specs/protocol/deputy-pause-module.md
+++ b/specs/protocol/deputy-pause-module.md
@@ -1,0 +1,458 @@
+# DeputyPauseModule
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Status](#status)
+- [Overview](#overview)
+- [Context](#context)
+- [Upgradeability](#upgradeability)
+- [Assumptions](#assumptions)
+  - [aSCP-001: Superchain-wide pause authorization process is well-defined](#ascp-001-superchain-wide-pause-authorization-process-is-well-defined)
+  - [aSCP-002: Superchain-wide pause authorization process is robust](#ascp-002-superchain-wide-pause-authorization-process-is-robust)
+  - [aSCP-003: Superchain-wide pause authorization process is fast](#ascp-003-superchain-wide-pause-authorization-process-is-fast)
+  - [aDPM-001: Contract is configured correctly](#adpm-001-contract-is-configured-correctly)
+    - [Mitigations](#mitigations)
+  - [aDPM-002: Deputy key is not compromised](#adpm-002-deputy-key-is-not-compromised)
+    - [Mitigations](#mitigations-1)
+  - [aDPM-003: Deputy key is not deleted](#adpm-003-deputy-key-is-not-deleted)
+    - [Mitigations](#mitigations-2)
+  - [aDPM-004: Ethereum will not censor transactions for extended periods of time](#adpm-004-ethereum-will-not-censor-transactions-for-extended-periods-of-time)
+    - [Mitigations](#mitigations-3)
+  - [aDPM-005: DeputyGuardianModule pause function correctly triggers pause](#adpm-005-deputyguardianmodule-pause-function-correctly-triggers-pause)
+    - [Mitigations](#mitigations-4)
+  - [aDPM-006: OpenZeppelin ECDSA and EIP712 contracts are free of critical bugs](#adpm-006-openzeppelin-ecdsa-and-eip712-contracts-are-free-of-critical-bugs)
+    - [Mitigations](#mitigations-5)
+  - [aDPM-007: Safe contracts are free of critical bugs](#adpm-007-safe-contracts-are-free-of-critical-bugs)
+    - [Mitigations](#mitigations-6)
+  - [aDPM-008: Deputy key is capable of creating signatures](#adpm-008-deputy-key-is-capable-of-creating-signatures)
+    - [Mitigations](#mitigations-7)
+- [System Invariants](#system-invariants)
+  - [iSCP-001: Superchain-wide pause can be activated within a short bounded time of authorization](#iscp-001-superchain-wide-pause-can-be-activated-within-a-short-bounded-time-of-authorization)
+    - [Impact](#impact)
+    - [Dependencies](#dependencies)
+  - [iSCP-002: Superchain-wide pause is not activated outside of the standard process](#iscp-002-superchain-wide-pause-is-not-activated-outside-of-the-standard-process)
+    - [Impact](#impact-1)
+    - [Dependencies](#dependencies-1)
+- [Component Invariants](#component-invariants)
+  - [iDPM-001: Only the Deputy may act through the module](#idpm-001-only-the-deputy-may-act-through-the-module)
+    - [Impact](#impact-2)
+    - [Mitigations](#mitigations-8)
+    - [Dependencies](#dependencies-2)
+  - [iDPM-002: Deputy must only be able to trigger the Superchain-wide pause](#idpm-002-deputy-must-only-be-able-to-trigger-the-superchain-wide-pause)
+    - [Impact](#impact-3)
+    - [Mitigations](#mitigations-9)
+    - [Dependencies](#dependencies-3)
+  - [iDPM-003: Deputy must always be able to act through the module](#idpm-003-deputy-must-always-be-able-to-act-through-the-module)
+    - [Impact](#impact-4)
+    - [Mitigations](#mitigations-10)
+    - [Dependencies](#dependencies-4)
+  - [iDPM-004: Deputy authorizations must not be replayable](#idpm-004-deputy-authorizations-must-not-be-replayable)
+    - [Impact](#impact-5)
+    - [Mitigations](#mitigations-11)
+    - [Dependencies](#dependencies-5)
+  - [iDPM-005: Foundation Safe must be able to change the Deputy account easily](#idpm-005-foundation-safe-must-be-able-to-change-the-deputy-account-easily)
+  - [Impact](#impact-6)
+  - [Mitigations](#mitigations-12)
+- [iDPM-006: Safe must be able to change the DeputyGuardianModule address easily](#idpm-006-safe-must-be-able-to-change-the-deputyguardianmodule-address-easily)
+  - [Impact](#impact-7)
+  - [Mitigations](#mitigations-13)
+  - [Assumptions](#assumptions-1)
+- [Implementation Spec](#implementation-spec)
+  - [constructor](#constructor)
+  - [pause](#pause)
+  - [setDeputy](#setdeputy)
+  - [setDeputyGuardianModule](#setdeputyguardianmodule)
+  - [foundationSafe](#foundationsafe)
+  - [deputyGuardianModule](#deputyguardianmodule)
+  - [superchainConfig](#superchainconfig)
+  - [deputy](#deputy)
+  - [pauseMessageTypehash](#pausemessagetypehash)
+  - [deputyAuthMessageTypehash](#deputyauthmessagetypehash)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Status
+
+Proposed
+
+## Overview
+
+The `DeputyPauseModule` is a [Safe Module][safe-modules] designed to be installed into the Optimism
+Foundation Operations Safe that allows a dedicated Deputy address to create an ECDSA signature that
+authorizes the execution of the Superchain-wide pause mechanism. The `DeputyPauseModule` assumes
+that the Optimism Foundation Operations Safe is the Deputy Guardian that is permitted to act within
+the `DeputyGuardianModule` installed into the Optimism Security Council Safe.
+
+## Context
+
+Additional context and motivation for this component can be found in the corresponding
+[design document][design-doc].
+
+## Upgradeability
+
+The `DeputyPauseModule` is not expected to change frequently and is therefore not a proxied
+contract. The `DeputyPauseModule` allows basic parameters to be modified by the Operations Safe
+directly without needing to deploy a new instance of the contract. Additionally, the address of the
+module itself is only utilized within the Optimism Foundation Operations Safe and can be safely
+replaced.
+
+## Assumptions
+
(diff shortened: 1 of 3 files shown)


specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,387 @@
...
+
+## Overview
+
+The Deputy Guardian Safe (currently the Optimism Foundation Operations Safe) utilizes the
+`DeputyPauseModule` to remove the need for brittle pre-signed pause transactions and to increase
+the reaction speed of the Deputy Guardian Safe for the specific purpose of triggering the
+Superchain-wide pause action. The Deputy Pause Module is explicitly designed to be used with an
+Externally Owned Account and is not designed to function with a smart contract deputy.
+
+## Context
```
maurelian, line 88: This section feels like it contains a lot of design doc like content, which leads to it not feeling very evergreen, ie. once this change is made, a lot of it will become outdated.

I think it would be fine to stick to just describing the motivations and goals for the DPM, without referencing what came before it.

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,387 @@
...
+
+We assume that the Safe contract implementations used by the Security Council Safe and Optimism
+Foundation Operations Safe are free of any critical bugs that would cause their behaviors to
+diverge from their specified behaviors.
+
+#### Mitigations
+
+- Existing audits/safety processes for Safe.
+
+## Invariants
```
maurelian: The impact/mitigations content in this doc is very similar to what we put into an FMA. I appreciate the thoroughness, but I'm worried we're duplicating effort. I think it would be sufficient to put invariants into a bulleted tree with dependencies as children.

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,387 @@
...
+- Takes the address of the `DeputyGuardianModule` as an authorized input.
+- Takes the address of the `SuperchainConfig` as an authorized input.
+- Takes a max signature validity time as an authorized input.
+- Takes the address of the Deputy as an authorized input.
+- Takes a signature from the Deputy over a known EIP-712 message.
+- Must verify that the signature was produced by the provided Deputy address.
+- Must verify that the max signature validity time is greater than 1 day and less than 365 days.
+- Must verify that the Operations Safe exposes some known Safe getter function.
+- Must verify that the `DeputyGuardianModule` exposes some known getter function.
+- Must verify that the `SuperchainConfig` exposes some known getter function.
```
maurelian: I'd love to see this info presented as natspec on an interface, I think that makes it a bit more intuitive to process, especially given that 6 lines are dedicated to input args.

specs/protocol/deputy-pause-module.md
smartcontracts: Yeah fair point, the design doc itself includes a lot of this context

specs/protocol/deputy-pause-module.md
smartcontracts: As much as possible I want to shift-left by having all of these impacts listed as invariants. This way we can understand the security model *before* we write all of the FMA stuff. IMO the FMA is a weak replacement for invariants in this case.

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
...
+
+> **NOTE**: Assumptions are utilized by specific invariants and do not apply globally. Invariants
+> typically only rely on a subset of the following assumptions. Different invariants may rely on
+> different assumptions. Refer to individual invariants for their dependencies.
+
+### aSCP-001: Superchain-wide pause authorization process is well-defined
+
+We assume that the process by which a Superchain-wide pause can be authorized is well-defined such
+that the cases in which the pause should and should not be triggered are apparent.
+
```
maurelian, line 111: This has not always been the case in the past, but I don't think its something that can be addressed here, so this assumption is fine.

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
...
+authorizes the execution of the Superchain-wide pause mechanism. The `DeputyPauseModule` assumes
+that the Optimism Foundation Operations Safe is the Deputy Guardian that is permitted to act within
+the `DeputyGuardianModule` installed into the Optimism Security Council Safe.
+
+## Context
+
+Additional context and motivation for this component can be found in the corresponding
+[design document][design-doc].
+
+## Assumptions
```
maurelian, line 101: I don't know where it goes, but an additional assumption is that "downstream" auth chain is correctly configured, i.e. the full chain of calls from the deputy to the superchainconfig succeeds

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
...
+- Regularly generate valid signatures from the key.
+
+### aDPM-004: Ethereum will not censor transactions for extended periods of time
+
+We assume that Ethereum will not censor transactions for an extended period of time and that any
+transaction submitted to the network can be processed within a reasonable time bound (e.g., 1h).
+
+#### Mitigations
+
+N/A
```
maurelian: High priority fee?

specs/protocol/deputy-pause-module.md
maurelian: Maybe that's `### aDPM-005: DeputyGuardianModule pause function correctly triggers pause`

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
...
+- Regular rotation of the Deputy key.
+
+### aDPM-003: Deputy key is not deleted
+
+We assume that the Deputy key has not been deleted by a malicious actor.
+
+#### Mitigations
+
+- Maintain duplicate copies of the key in multiple secure locations.
+- Regularly generate valid signatures from the key.
```
maurelian: How regularly?
Should the old signatures be deleted/replaced?

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
...
+
+### aDPM-006: OpenZeppelin ECDSA and EIP712 contracts are free of critical bugs
+
+We assume that both the `ECDSA` library and the `EIP712` contract provided by OpenZeppelin V4 at
+commit [ecd2ca2c][oz-v4.7.3] (v4.7.3) are free of any critical bugs that would cause their
+behaviors to diverge from their specified behaviors.
+
+#### Mitigations
+
+- Existing audits/safety processes for OpenZeppelin V4.
```
maurelian, line 184: Also battle testing given the commit is close to 2.5 years old.

I also note that our commit is immediately after this ecdsa [related advisory](https://github.com/OpenZeppelin/openzeppelin-contracts/security/advisories/GHSA-4h98-2769-gh6h).

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
...
+
+### aDPM-007: Safe contracts are free of critical bugs
+
+We assume that the Safe contract implementations used by the Security Council Safe and Optimism
+Foundation Operations Safe are free of any critical bugs that would cause their behaviors to
+diverge from their specified behaviors.
+
+#### Mitigations
+
+- Existing audits/safety processes for Safe.
```
maurelian, line 194: Also battle testing. Our Safe versions are at most 1.3.0 IIRC

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
+# DeputyPauseModule
```
maurelian, line 1: I think this PR should also update the diagram in stage-1.md

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
...
+- Must verify that the signature was produced by the provided Deputy address.
+
+### setDeputyGuardianModule
+
+- Takes an address as an untrusted input.
+- Can only be called by the Foundation Safe as configured in the constructor.
+
+### foundationSafe
+
+- Returns the address of the Operations Safe as set in the constructor.
```
maurelian, line 433: for a moment there I thought we were going to do away with the ops safe? 
However this seems simpler than having to change the Deputy Guardian.

specs/protocol/deputy-pause-module.md
smartcontracts: Yep that's the one

specs/protocol/deputy-pause-module.md
smartcontracts: Yeah generally speaking assumptions are things that *should* be true regardless of whether or not they actually are. Then we can keep track of which assumptions are true and how true they are.

specs/protocol/deputy-pause-module.md
smartcontracts: Oh this is an old line, will remove it

specs/protocol/deputy-pause-module.md
smartcontracts: Good point

specs/protocol/deputy-pause-module.md
smartcontracts: Yeah we decided to keep it for now for a number of reasons

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
...
+
+We assume that the Safe contract implementations used by the Security Council Safe and Optimism
+Foundation Operations Safe are free of any critical bugs that would cause their behaviors to
+diverge from their specified behaviors.
+
+#### Mitigations
+
+- Existing audits/safety processes for Safe.
+
+## System Invariants
```
maurelian, line 204: random place for this comment: we should indicate somewhere that the DPM is not proxied and why, esp. given that we want/wanted to change the DGM to be proxied.

specs/protocol/safe-extensions.md
```diff
@@ -116,46 +112,6 @@ The following security properties must be upheld by the `DeputyGuardianModule`:
...
-
-The Deputy Guardian Safe (currently the Optimism Foundation Safe) utilizes the Deputy Pause Module
-to remove the need for brittle pre-signed pause transactions and to speed up the reaction speed of
-the Deputy Guardian Safe for the specific purpose of triggering the Superchain-wide pause action.
-The Deputy Pause Module is explicitly designed to be used with an Externally Owned Account and is
-not designed to function with a smart contract deputy.
-
-#### Invariants
-
-1. Must enforce that the Deputy account is an EOA.
```
maurelian, line 131: this invariaht did not make it into the new document, I see no mention of 'EOA' there.

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
...
+If this invariant is broken, it would not be possible for the Safe account to easily rotate the
+Deputy account if the account is compromised. This creates operational overhead but is not a
+security risk as we assume that the Safe code does not have bugs and therefore the Safe can always
+remove the module if necessary.
+
+### Mitigations
+
+- Authorized function to change the Deputy address.
+
+## iDPM-006: Safe must be able to change the DeputyGuardianModule address easily
```
maurelian, line 352: I think this is an assumption, esp. given that it's not tested in the impl PR.

maurelian, line 371: I think this is an assumption, esp. given that it's not tested in the impl PR.

maurelian: to be clear, I'm ok with that.

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
...
+#### Dependencies
+
+- [aSCP-001](#ascp-001-superchain-wide-pause-authorization-process-is-well-defined)
+- [iDPM-001](#idpm-001-only-the-deputy-may-act-through-the-module)
+- [iDPM-004](#idpm-004-deputy-authorizations-must-not-be-replayable)
+- [aDPM-002](#adpm-002-deputy-key-is-not-compromised)
+
+## Component Invariants
+
+### iDPM-001: Only the Deputy may act through the module
```
maurelian, line 250: enforced here: https://github.com/ethereum-optimism/optimism/blob/sc/deputy-pause-module/packages/contracts-bedrock/test/safe/DeputyPauseModule.t.sol#L466

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
...
+#### Mitigations
+
+- Signature verification on the pause action.
+- Monitoring for pause triggering actions.
+
+#### Dependencies
+
+- [aDPM-006](#adpm-006-openzeppelin-ecdsa-and-eip712-contracts-are-free-of-critical-bugs)
+
+### iDPM-002: Deputy must only be able to trigger the Superchain-wide pause
```
maurelian, line 277: Not immediately obvious which test enforces this invariant. But the [calldata is hardcoded](https://github.com/ethereum-optimism/optimism/blob/sc/deputy-pause-module/packages/contracts-bedrock/src/safe/DeputyPauseModule.sol#L179) so it's not too worrying. 

This could maybe be covered more directly with a fuzz test over signatures which checks that no calls result to a different contract or have unexpected calldata. But fuzzing signatures makes no sense, and there are other fuzz tests over the signature inputs which check for reverts, so I think we're good for this one.

specs/protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,439 @@
...
+#### Mitigations
+
+- Strict auditing and verification of the `DeputyPauseModule`.
+
+#### Dependencies
+
+- [aDPM-005](#adpm-005-deputyguardianmodule-pause-function-correctly-triggers-pause)
+- [aDPM-007](#adpm-007-safe-contracts-are-free-of-critical-bugs)
+
+### iDPM-003: Deputy must always be able to act through the module
```
maurelian, line 299: Covered in the [DeputyPauseModule_Pause_Test](https://github.com/ethereum-optimism/optimism/blob/sc/deputy-pause-module/packages/contracts-bedrock/test/safe/DeputyPauseModule.t.sol#L377) happy path fuzzing.

specs/protocol/deputy-pause-module.md
smartcontracts: There is a test for `setDeputyGuardianModule`

specs/protocol/safe-extensions.md
smartcontracts: Fixed

maurelian (approved):

PR #480, opened 2024-12-12
docs(dependency-set): specify that dependency is bidirectional

qbzzt:
**Description**

Change the specs to match what I heard from @zainbacchus in our call today (unless I misunderstood).

**Tests**

N/A

**Additional context**

N/A

**Metadata**

N/A

### DIFF
--- a/specs/interop/dependency-set.md
+++ b/specs/interop/dependency-set.md
@@ -21,9 +21,8 @@ social consensus MUST be used to determine the chain that represents the canonic
 particularly impacts the block builder as they SHOULD use the chain id to assist in validation
 of executing messages.
 
-The dependency set is configured on a per chain basis and is a unidirectional relationship. This means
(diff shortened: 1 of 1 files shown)


tynes:
I think this needs a bit more explaining, specifically around why and how. Perhaps this information isn't super legible, its part of https://github.com/ethereum-optimism/specs/pull/465

zainbacchus:
This should be helpful: https://github.com/ethereum-optimism/docs/pull/1171

zainbacchus:
Closing this as this change was reverted. See:https://github.com/ethereum-optimism/docs/pull/1186

qbzzt:
I don't understand. Can chainA be in the dependency set of chainB if chainB is *not* in the dependency set of chainA? In other words, is it possible for me to send an interop message A->B, but not B->A? If so then the change I proposed is wrong. If not, it is necessary for accuracy.

Changing the docs should not affect the specs, IMAO.

PR #494, opened 2025-01-13
feat: AnchorStateRegistry unified anchor state spec

smartcontracts:
Introduces a specification for the unified anchor state version of the AnchorStateRegistry.

### DIFF
--- a/specs/fault-proof/stage-one/anchor-state-registry.md
+++ b/specs/fault-proof/stage-one/anchor-state-registry.md
@@ -1,0 +1,225 @@
+# AnchorStateRegistry
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Overview](#overview)
+- [Definitions](#definitions)
+  - [Dispute Game](#dispute-game)
+  - [Respected Game Type](#respected-game-type)
+  - [Registered Game](#registered-game)
+  - [Respected Game](#respected-game)
+  - [Blacklisted Game](#blacklisted-game)
+  - [Retired Game](#retired-game)
+  - [Proper Game](#proper-game)
+- [Assumptions](#assumptions)
+  - [aASR-001: Dispute Game contracts properly report important properties](#aasr-001-dispute-game-contracts-properly-report-important-properties)
+    - [Mitigations](#mitigations)
+  - [aASR-002: Dispute Game Factory properly reports its created games](#aasr-002-dispute-game-factory-properly-reports-its-created-games)
+    - [Mitigations](#mitigations-1)
+  - [aASR-003: OptimismPortal properly reports respected game type, blacklist, and retirement time](#aasr-003-optimismportal-properly-reports-respected-game-type-blacklist-and-retirement-time)
+    - [Mitigations](#mitigations-2)
+- [Invariants](#invariants)
+  - [iASR-001: Games are represented as Proper Games accurately](#iasr-001-games-are-represented-as-proper-games-accurately)
+    - [Impact](#impact)
+    - [Dependencies](#dependencies)
+- [Function Specification](#function-specification)
+  - [isGameRegistered](#isgameregistered)
+  - [isGameRespected](#isgamerespected)
+  - [isGameBlacklisted](#isgameblacklisted)
+  - [isGameRetired](#isgameretired)
+  - [isGameProper](#isgameproper)
+  - [getAnchorRoot](#getanchorroot)
+  - [anchors](#anchors)
+  - [tryUpdateAnchorState](#tryupdateanchorstate)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+The `AnchorStateRegistry` was designed as a registry where `DisputeGame` contracts could store and
+register their results so that these results could be used as the starting states for new
+`DisputeGame` instances. These starting states, called "anchor states", allow new `DisputeGame`
+contracts to use a newer starting state to bound the size of the execution trace for any given
+game.
+
+We are generally aiming to shift the `AnchorStateRegistry` to act as a unified source of truth for
+the validity of `DisputeGame` contracts and their corresponding root claims. This specification
+corresponds to the first iteration of the `AnchorStateRegistry` that will move us in this
+direction.
+
+## Definitions
+
+### Dispute Game
+
+> See [Fault Dispute Game](fault-dispute-game.md)
+
+A Dispute Game is a smart contract that makes a determination about the validity of some claim. In
+the context of the OP Stack, the claim is generally assumed to be a claim about the value of an
+output root at a given L2 block height.
+
+### Respected Game Type
+
+The `OptimismPortal` contract defines a "respected game type" which is the Dispute Game type that
+the portal allows to be used for the purpose of proving and finalizing withdrawals. This mechanism
+allows the system to use multiple game types simultaneously while still ensuring that the
+`OptimismPortal` contract only trusts respected games specifically.
+
+### Registered Game
+
+A Dispute Game is considered to be a **Registered Game** if the game contract was created by the
+system's `DisputeGameFactory` contract.
+
+### Respected Game
+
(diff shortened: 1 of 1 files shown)


wildmolasses (approved):

specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -0,0 +1,214 @@
...
+output root at a given L2 block height.
+
+### Registered Game
+
+A Dispute Game is considered to be a **Registered Game** if the game contract was created by the
+system's `DisputeGameFactory` contract.
+
+### Respected Game
+
+A Dispute Game is considered to be a **Respected Game** if the game contract's game type is
```
tynes, line 76: This is a self referential definition. A respected game is a respected game if its a respected game. How does the respected game **functionally** impact the system?

smartcontracts: I don't want to bloat the scope of this PR and have to write a spec for the OptimismPortal. The OptimismPortal currently holds the definition for what a "respected" game type is and the AnchorStateRegistry is piggybacking on this definition.

Respected game type functionally impacts the system because it is used as part of the definition for Proper Game, which is used as part of `tryUpdateAnchorState`

smartcontracts: Basically what I'm saying here is that OptimismPortal independently has a concept of a "respected game type" and the AnchorStateRegistry defines the concept of a "Respected Game" as a game that has a game type that matches the "respected game type" defined in the OptimismPortal. So this definition is not circular. It just happens to share the same wording.

tynes: Without this context, its impossible to know its not circular. I recommend using a hyperlink to the place where it is actually defined

smartcontracts: I honestly don't think this is defined anywhere in the specs

smartcontracts: I added some extra context in a separate section called `Respected Game Type`, lmk what you think

specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -0,0 +1,224 @@
...
+We assume that the `FaultDisputeGame` and `PermissionedDisputeGame` contracts properly and
+faithfully report the following properties:
+
+- Game type
+- L2 block number
+- Root claim value
+- Creation timestamp
+- Resolution timestamp
+
+#### Mitigations
```
ajsutton, line 117: nit: You can't mitigate an assumption.   I'd say these are verifications.

And given this section is the same for every assumption, I'm not sure it's adding any actual value.

specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -0,0 +1,224 @@
...
+
+- Existing audit on the `OptimismPortal` contract
+- Integration testing
+
+## Invariants
+
+### iASR-001: Games are represented as Proper Games accurately
+
+When asked if a game is a Proper Game, the `AnchorStateRegistry` must serve a response that is
+identical to the response that would be given by a perfect oracle for this query.
```
ajsutton, line 147: It feels weird that we don't have any invariant about the anchor state being valid.  The rules in `tryUpdateAnchorState` define what updates the anchor state but the single most important invariants are that the anchor state can't be set to something invalid and that it is the latest suitable result.

specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -0,0 +1,224 @@
...
+retirement timestamp (`respectedGameTypeUpdatedAt`) defined in the `OptimismPortal` contract.
+
+### Proper Game
+
+A Dispute Game is considered to be a **Proper Game** if all of the following are true:
+
+- Game is a Registered Game
+- Game is a Respected Game
+- Game is not a Blacklisted Game
+- Game is not a Retired Game
```
ajsutton, line 98: Notably, this definition doesn't consider the outcome of the game in any way. That makes sense in many ways (ie you can prove a withdrawal against an in-progress game) but feels like a key concept is missing here (e.g. you can't finalise a withdrawal against a game that resolved as CHALLENGER_WINS and probably shouldn't be able to prove against it either).

This isn't necessarily wrong - it could be left up to the portal to decide if a game is valid based on its status, but it's notable that `tryUpdateAnchorState` has to make the same determination.  I wonder if there should be a `isValid` method that determines if a game is a proper game and resolved as Defender Wins.

specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -0,0 +1,224 @@
...
+### tryUpdateAnchorState
+
+Allows a `DisputeGame` contract to attempt to set its own result as the anchor state. The
+`AnchorStateRegistry` contract previously stored the anchor state separately for each game type.
+The updated version of this function allows any game type to update the anchor state as long as it
+is the currently respected game type.
+
+- MUST not revert if the caller is a Registered Game and sufficient gas is provided.
+- MUST not update the anchor state if the game is not a Proper Game for any reason.
+- MUST not update the anchor state if the game resolved in favor of the Challenger.
```
ajsutton, line 222: This isn't strict enough.  It must not update if the game is in any state other than defender wins.  You must not update the anchor state with an In Progress game.

ajsutton (commented):
Only real blocker here is that the tryUpdateAnchorState shouldn't allow updating based on an in progress game.  Couple of other queries about where the definition of valid claims should live but happy for that to go either way.

And I'm losing this argument, but all of this content would make way more sense just being a comment in the contract source code so it's kept in one place. It is at least more of a spec and less a description of our implementation than most others though...

specs/fault-proof/stage-one/anchor-state-registry.md
smartcontracts: Technically as of this PR, the anchor state *can* be invalid, which is what causes the poisoning problem. Note that #517 resolves this tension.

specs/fault-proof/stage-one/anchor-state-registry.md
smartcontracts: Yeah I'll probably deal with this at a later point when we start to standardize around the right language. Agree that we can't mitigate an assumption.

specs/fault-proof/stage-one/anchor-state-registry.md
smartcontracts: Yep, this is all defined in #517

specs/fault-proof/stage-one/anchor-state-registry.md
smartcontracts: Good point, updating

specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -0,0 +1,225 @@
...
+
+Allows a `DisputeGame` contract to attempt to set its own result as the anchor state. The
+`AnchorStateRegistry` contract previously stored the anchor state separately for each game type.
+The updated version of this function allows any game type to update the anchor state as long as it
+is the currently respected game type.
+
+- MUST not revert if the caller is a Registered Game and sufficient gas is provided.
+- MUST not update the anchor state if the game is not a Proper Game for any reason.
+- MUST not update the anchor state if the game is not resolved (`IN_PROGRESS`).
+- MUST not update the anchor state if the game resolved in favor of the Challenger.
```
ajsutton, line 222: nit: Should we just say:
```suggestion
- MUST revert if the anchor state if the game is not resolved in favour of the Defender
```

Not sure if solidity enum's a real enough to prevent supplying a uint8 value that doesn't map to any of the possible values, but if that is possible we shouldn't let an unknown status update the anchor state either, so restricting to only update if status is `DEFENDER_WINS` seems simplest and most comprehensive.

ajsutton (approved):
LGTM.

specs/fault-proof/stage-one/anchor-state-registry.md
smartcontracts: Since this already resolved in #517 going to just skip this for now

PR #517, opened 2025-01-16
feat: AnchorStateRegistry incident response changes

smartcontracts:


specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -85,17 +115,89 @@ as blacklisted inside of the `OptimismPortal` contract.
...
+A Dispute Game is considered to be an **Airgapped Game** if the game resolved a result more than
+the airgap delay seconds ago as defined by the `disputeGameFinalityDelaySeconds` variable in the
+`OptimismPortal` contract.
+
+### Finalized Game
+
+A Dispute Game is considered to be a **Finalized Game** if all of the following are true:
+
+- The game is a Resolved Game
+- The game is an Airgapped Game
```
ajsutton: This is the same definition as an Airgapped game, because an Airgapped game already requires the game to have resolved.  And "Finalized Game" is a much better term than "Airgapped Game" because all games are airgapped, just some have completed the air gap time and some haven't.

specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -194,8 +393,38 @@ Determines if a game is a Retired Game.
...
+
+### isGameAirgapped
+
+Determines if a game is an Airgapped Game.
+
+- MUST return `true` if and only if the game has resolved a result more than the airgap delay
+  seconds ago as defined by the `disputeGameFinalityDelaySeconds` variable in the `OptimismPortal`
+  contract.
+
+### isGameFinalized
```
ajsutton: As noted above, these two functions will always return the same result as each other.

ajsutton (approved):
LGTM. I think this brings in the resolution piece that seemed missing from https://github.com/ethereum-optimism/specs/pull/494

And it's starting to justify itself as a doc rather than just a comment in the code, though still debatable. :)

specs/fault-proof/stage-one/anchor-state-registry.md
smartcontracts: Good point, we can get rid of Airgapped and simply keep finalized.

PR #518, opened 2025-01-16
specs: Specify derivation of future tx types

sebastianst:
**Description**

Clarifies how to deal with 7702 and future transactions.

- batcher tx must be of type `0-3`, or `0x7e`, but _not_ `4`, that is, 7702 txs.
- we allow 7702 txs, and possibly other future txs with similarly encoded logs, for user deposits on L1 and SystemConfig configuration tx

**Metadata**

Part of https://github.com/ethereum-optimism/optimism/issues/13627

### DIFF
--- a/specs/protocol/derivation.md
+++ b/specs/protocol/derivation.md
@@ -1300,12 +1300,12 @@ transaction's receipt as invalid.
 - If it's a known transaction type, that is, legacy (first byte of the encoding is in the range `[0xc0, 0xfe]`) or its
 first byte is in the range `[0, 4]` or `0x7e` (_deposited_), then it's not a _future transaction_ and we know how to
 decode the receipt and this process is irrelevant.
-- If a transaction's first byte is in the range `[5, 0x7d]`, it is expected to be a _future_ EIP-2718 transaction, so we can
-proceed to the receipt. Note that we excluded `0x7e` because that's the deposit transaction type, which is known.
-- The _future_ receipt encoding's first byte must be the same byte as the transaction encoding's first byte, or it is discarded
-as invalid, because we require it to be an EIP-2718-encoded receipt to continue.
-- The receipt payload is decoded as if it is encoded as `rlp([status, cumulative_transaction_gas_used, logs_bloom, logs])`,
-which is the encoding of the known non-legacy transaction types.
+- If a transaction's first byte is in the range `[5, 0x7d]`, it is expected to be a _future_ EIP-2718 transaction, so we
+can proceed to the receipt. Note that we excluded `0x7e` because that's the deposit transaction type, which is known.
+- The _future_ receipt encoding's first byte must be the same byte as the transaction encoding's first byte, or it is
+discarded as invalid, because we require it to be an EIP-2718-encoded receipt to continue.
+- The receipt payload is decoded as if it is encoded as `rlp([status, cumulative_transaction_gas_used, logs_bloom,
+logs])`, which is the encoding of the known non-legacy transaction types.
   - If this decoding fails, the transaction's receipt is discarded as invalid.
   - If this decoding succeeds, the `logs` have been obtained and can be processed as those of known transaction types.
 
--- a/specs/protocol/derivation.md
+++ b/specs/protocol/derivation.md
@@ -1017,11 +1017,8 @@ entries.
 
 [deposit-contract-spec]: deposits.md#deposit-contract
 
-Logs for all transaction types are parsed for deposits, also from possibly unknown transactions types like those that
-are considered invalid batcher transaction types, see [L1 Retrieval][#l1-retrieval]. However, they are parsed as if they
-were receipts for one of the known transactions types `0`, `1`, `2`, `3`, `4` and `0x7e`. If they fail this best-effort
-decoding process, they are dropped and considered invalid.
-The intention of this best-effort decoding is to future-proof the protocol for new L1 transaction types.
+Logs are derived from transactions following the future-proof best-effort process described in
+[On Future-Proof Transaction Log Derivation][#on-future-proof-transaction-log-derivation]
 
 ### Network upgrade automation transactions
 
@@ -1287,3 +1284,29 @@ follows:
 
 [extended-attributes]: exec-engine.md#extended-payloadattributesv1
 [Fee Vaults]: exec-engine.md#fee-vaults
+
+## On Future-Proof Transaction Log Derivation
+
+As described in [L1 Retrieval][#l1-retrieval], batcher transactions' types are required to be from a fixed allow-list.
+
+However, we want to allow deposit transactions and `SystemConfig` update events to get derived even from receipts of
+future transaction types, as long as the receipts can be decoded following a best-effort process:
+
+As long as a future transaction type follows the [EIP-2718][https://eips.ethereum.org/EIPS/eip-2718] specification,
+the type can be decoded as the first byte of an encoded transaction or receipt (excluding legacy transactions, which are
+not _future_ transactions). We can then proceed as follows to get the logs of such a future transaction, or discard the
+transaction's receipt as invalid.
+
+- If it's a known transaction type, that is, legacy (first byte of the encoding is in the range `[0xc0, 0xfe]`) or its
+first byte is in the range `[0, 4]`, then it's not a _future transaction_ and we know how to decode the receipt and this
+process is irrelevant.
+- If a transaction's first byte is in the range `[5, 0x7f]`, it is expected to be a _future_ EIP-2718 transaction, so we can
+proceed to the receipt.
+- The receipt encoding's first byte must be the same byte as the transaction encoding's first byte, or it is discarded
+as invalid.
+- The receipt payload is decoded as if it is encoded as `rlp([status, cumulative_transaction_gas_used, logs_bloom, logs])`,
+which is the encoding of the known non-legacy transaction types.
+  - If this decoding fails, the transaction's receipt is discarded as invalid.
+  - If this decoding succeeds, the `logs` have been obtained and can be processed as those of known transaction types.
+
+The intention of this best-effort decoding process is to future-proof the protocol for new L1 transaction types.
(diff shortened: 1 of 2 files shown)


specs/protocol/derivation.md
```diff
@@ -1014,6 +1017,12 @@ entries.
 
 [deposit-contract-spec]: deposits.md#deposit-contract
 
+Logs for all transaction types are parsed for deposits, also from possibly unknown transactions types like those that
+are considered invalid batcher transaction types, see [L1 Retrieval][#l1-retrieval]. However, they are parsed as if they
+were receipts for one of the known transactions types `0`, `1`, `2`, `3`, `4` and `0x7e`. If they fail this best-effort
+decoding process, they are dropped and considered invalid.
+The intention of this best-effort decoding is to future-proof the protocol for new L1 transaction types.
```
sebastianst: I'm not sure if this "best-effort" specification is too vague and not precise enough. It seems like this could be too implementation-dependent...

specs/protocol/system-config.md
```diff
@@ -163,3 +163,9 @@ The contained log events are filtered and processed as follows:
 Note that individual derivation stages may be processing different L1 blocks,
 and should thus maintain individual system configuration copies,
 and apply the event-based changes as the stage traverses to the next L1 block.
+
+Logs for all transaction types are parsed for `SystemConfig` events, also from possibly unknown transactions types like
+those that are considered invalid batcher transaction types, see [L1 Retrieval][derivation.md#l1-retrieval]. However,
+they are parsed as if they were receipts for one of the known transactions types `0`, `1`, `2`, `3`, `4` and `0x7e`. If
+they fail this best-effort decoding process, they are dropped and considered invalid.
+The intention of this best-effort decoding is to future-proof the protocol for new L1 transaction types.
```
sebastianst: I'm not sure if this "best-effort" specification is too vague and not precise enough. It seems like this could be too implementation-dependent...

protolambda (commented):
LGTM.

Maybe consider clarifying "best-effort receipt decoding" as "ignore the tx-type prefix-byte of the RLP payload, and decode the RLP like the receipt-type that is used for for types `0`, `1`, `2`, `3`, `4`.

Also maybe clarify that TX-type != prefix byte. Legacy transactions, and receipts also (!), have a range of prefix byte values, and not really an actual prefix byte but part of the RLP payload itself, that all effectively map to "legacy", and represented as type 0.

geoknee:
I think it would be good to reference https://eips.ethereum.org/EIPS/eip-2718. 

In implementation work so far I have been making a distinction between:
* explicitly supported EIP 2718 tx (fully parsed)
* not explicitly supported EIP 2718 tx (treated as "opaque" and only the tx type and hash is parsed)
* non-EIP 2718 tx (not parsed at all, causes an error / panics). An example is when the first byte is out of range, e.g. `> 0x7f`.

sebastianst:
Thanks @protolambda @geoknee I agree that a specific reference to `EIP-2718` is helpful here. I've completely rewritten the spec and came up with a process that feels much more clearly defined. PTAL.

specs/protocol/system-config.md
sebastianst: I've completely rewritten the spec and came up with a process that feels much more clearly defined.

specs/protocol/derivation.md
```diff
@@ -1278,3 +1284,29 @@ follows:
...
+
+As long as a future transaction type follows the [EIP-2718][https://eips.ethereum.org/EIPS/eip-2718] specification,
+the type can be decoded as the first byte of an encoded transaction or receipt (excluding legacy transactions, which are
+not _future_ transactions). We can then proceed as follows to get the logs of such a future transaction, or discard the
+transaction's receipt as invalid.
+
+- If it's a known transaction type, that is, legacy (first byte of the encoding is in the range `[0xc0, 0xfe]`) or its
+first byte is in the range `[0, 4]`, then it's not a _future transaction_ and we know how to decode the receipt and this
+process is irrelevant.
+- If a transaction's first byte is in the range `[5, 0x7f]`, it is expected to be a _future_ EIP-2718 transaction, so we can
```
protolambda:
```suggestion
- If a transaction's first byte is in the range `[5, 0x7d]` (incl.), it is expected to be a _future_ EIP-2718 transaction, so we can
```

`0x7e` is the deposit tx type.

specs/protocol/derivation.md
```diff
@@ -1278,3 +1284,29 @@ follows:
...
+proceed to the receipt.
+- The receipt encoding's first byte must be the same byte as the transaction encoding's first byte, or it is discarded
+as invalid.
```
protolambda:
```suggestion
- A non-legacy receipt encoding's first byte must be the same byte as the transaction encoding's first byte, or it is discarded
as invalid.
```

See code suggestion. I don't believe this is the case for legacy tx types. There it can be a range of bytes due to the RLP length. And tx length != receipt length.

specs/protocol/derivation.md
sebastianst: Ah yes, forgot to translate this from the old version. Adapted and added a comment about excluding the deposited tx type.

specs/protocol/derivation.md
sebastianst: You're right for legacy transactions, but the process would already have stopped before reaching this step for legacy txs, see previous two steps. Will clarify nonetheless. Note that this is a description of a step that says, "if this is not given, then discard as invalid".

protolambda (approved):

specs/protocol/derivation.md
```diff
@@ -1278,3 +1285,29 @@ follows:
...
+## On Future-Proof Transaction Log Derivation
+
+As described in [L1 Retrieval][#l1-retrieval], batcher transactions' types are required to be from a fixed allow-list.
+
+However, we want to allow deposit transactions and `SystemConfig` update events to get derived even from receipts of
+future transaction types, as long as the receipts can be decoded following a best-effort process:
+
+As long as a future transaction type follows the [EIP-2718][https://eips.ethereum.org/EIPS/eip-2718] specification,
+the type can be decoded as the first byte of an encoded transaction or receipt (excluding legacy transactions, which are
+not _future_ transactions). We can then proceed as follows to get the logs of such a future transaction, or discard the
```
geoknee:
```suggestion
As long as a future transaction type follows the [EIP-2718][https://eips.ethereum.org/EIPS/eip-2718] specification,
the type can be decoded from the first byte of the transaction (or its receipt's) binary encoding. We can then proceed as follows to get the logs of such a future EIP-2718 transaction, or discard the
```

specs/protocol/derivation.md
```diff
@@ -1278,3 +1285,29 @@ follows:
...
+transaction's receipt as invalid.
+
+- If it's a known transaction type, that is, legacy (first byte of the encoding is in the range `[0xc0, 0xfe]`) or its
+first byte is in the range `[0, 4]` or `0x7e` (_deposited_), then it's not a _future transaction_ and we know how to
+decode the receipt and this process is irrelevant.
+- If a transaction's first byte is in the range `[5, 0x7d]`, it is expected to be a _future_ EIP-2718 transaction, so we
```
geoknee: I would prefer not to mix hex and decimal:
```suggestion
- If a transaction's first byte is in the range `[0x05, 0x7d]`, it is expected to be a _future_ EIP-2718 transaction, so we
```

I wonder if a table might display these different cases more clearly?

specs/protocol/derivation.md
```diff
@@ -1278,3 +1285,29 @@ follows:
 
 [extended-attributes]: exec-engine.md#extended-payloadattributesv1
 [Fee Vaults]: exec-engine.md#fee-vaults
+
+## On Future-Proof Transaction Log Derivation
+
+As described in [L1 Retrieval][#l1-retrieval], batcher transactions' types are required to be from a fixed allow-list.
+
+However, we want to allow deposit transactions and `SystemConfig` update events to get derived even from receipts of
+future transaction types, as long as the receipts can be decoded following a best-effort process:
```
geoknee, line 1294: Lifting from this issue https://github.com/ethereum-optimism/optimism/issues/13627, should we:

> Clarify 7702 receipt type can contain valid deposits and system-config events

I'm not sure I understand this well myself at this point. An example might be good.

specs/protocol/derivation.md
```diff
@@ -1278,3 +1285,29 @@ follows:
...
+- If it's a known transaction type, that is, legacy (first byte of the encoding is in the range `[0xc0, 0xfe]`) or its
+first byte is in the range `[0, 4]` or `0x7e` (_deposited_), then it's not a _future transaction_ and we know how to
+decode the receipt and this process is irrelevant.
+- If a transaction's first byte is in the range `[5, 0x7d]`, it is expected to be a _future_ EIP-2718 transaction, so we
+can proceed to the receipt. Note that we excluded `0x7e` because that's the deposit transaction type, which is known.
+- The _future_ receipt encoding's first byte must be the same byte as the transaction encoding's first byte, or it is
+discarded as invalid, because we require it to be an EIP-2718-encoded receipt to continue.
+- The receipt payload is decoded as if it is encoded as `rlp([status, cumulative_transaction_gas_used, logs_bloom,
+logs])`, which is the encoding of the known non-legacy transaction types.
+  - If this decoding fails, the transaction's receipt is discarded as invalid.
```
geoknee, line 1309: Would you agree with this summary:

> WIth this spec change, we are saying that any transaction which is both:
> * EIP 2718 compliant
> * has a receipt whose encoding is_also_ compliant with the existing non-legacy RLP encoding scheme
> 
> Is supported by derivation in the sense that deposits and systemconfig events can be derived from such transaction receipts.
> 

So for example if in the coming years we get tx type 8 coming along with the usual RLP receipt encoding, we don't need to change the spec. If we then get tx type 9 with a different receipt structure, we would need to revisit this part of the spec (unless we are happy to ignore tx type 9).

geoknee (approved):

specs/protocol/derivation.md
sebastianst: It's not so much a list of cases (well the first two kind of are), but steps in a process. So I'd prefer to leave them in this format.

specs/protocol/derivation.md
sebastianst: I already include type `4` here (=7702 SetCode txs), as if we already merged the upstream geth changes that will contain https://github.com/ethereum/go-ethereum/pull/30078. This implies that 7702 txs can contain valid deposits and SystemConfig updates.

specs/protocol/derivation.md
sebastianst: Yes

PR #519, opened 2025-01-16
feat: OptimismPortal changes for incident response improvements

smartcontracts:
**Description**
Specification for the components of the `OptimismPortal` touched by the Incident Response Improvements work. This is a partial specification for the `OptimismPortal`.

### DIFF
--- a/specs/fault-proof/stage-one/optimism-portal.md
+++ b/specs/fault-proof/stage-one/optimism-portal.md
@@ -1,0 +1,251 @@
+# OptimismPortal
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Overview](#overview)
+- [Definitions](#definitions)
+  - [Proof Maturity Delay](#proof-maturity-delay)
+  - [Proven Withdrawal](#proven-withdrawal)
+  - [Finalized Withdrawal](#finalized-withdrawal)
+  - [Valid Withdrawal](#valid-withdrawal)
+  - [Invalid Withdrawal](#invalid-withdrawal)
+  - [Respected Game Type](#respected-game-type)
+  - [Retirement Timestamp](#retirement-timestamp)
+  - [L2 Withdrawal Sender](#l2-withdrawal-sender)
+- [Assumptions](#assumptions)
+  - [aOP-001: Dispute Game contracts properly report important properties](#aop-001-dispute-game-contracts-properly-report-important-properties)
+    - [Mitigations](#mitigations)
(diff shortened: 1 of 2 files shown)


tynes (approved):

tynes:
I see this isn't based on main, won't merge for that reason

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -0,0 +1,251 @@
...
+
+**Severity: Critical**
+
+If this invariant is broken, any number of arbitrarily bad outcomes could happen. Most obviously,
+we would expect all bridge systems relying on the `OptimismPortal` to be immediately compromised.
+
+### iOP-002: Valid Withdrawals can always be finalized in bounded time
+
+We require that [Valid Withdrawals](#valid-withdrawal) can always be
+[finalized](#finalized-withdrawal) within some reasonable, bounded amount of time.
```
ajsutton, line 180: This invariant is broken by the ability to pause withdrawals.

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -0,0 +1,251 @@
...
+### setRespectedGameType
+
+Allows the Guardian to change the [Respected Game Type](#respected-game-type) and to update the
+[Retirement Timestamp](#retirement-timestamp).
+
+- MUST revert if the sender is not the Guardian.
+- If the game type is `type(uint32).max`, MUST update the Retirement Timestamp to the current
+  block timestamp and MUST NOT update the Respected Game Type.
+- If the game type is not `type(uint32).max`, MUST set the Respected Game Type to the provided game
+  type and MUST NOT update the Retirement Timestamp.
```
ajsutton, line 200: This is a very weird API. Not a blocker, but I'm very surprised we didn't just introduce a separate method for updating the retirement timestamp rather than using a magic value to shoehorn it into the same method.

ajsutton (approved):
LGTM generally but probably should clarify the bounded time invariant by adding an assumption that the guardian will unpause withdrawals within a reasonable bounded time or clarify the invariant only applies when not paused.

specs/fault-proof/stage-one/optimism-portal.md
smartcontracts: Yeah, that's correct. We need to clarify the assumption that the Guardian will unpause within a bounded amount of time.

specs/fault-proof/stage-one/optimism-portal.md
smartcontracts: Ahh so the actual reason here was that changing the API or adding a new function meant that we would have to redeploy the `DeputyGuardianModule` which would've been a PITA (and would've required new pre-signed pause transactions).

We *will* be removing this weird API as part of the very next upgrade. See [here](https://github.com/ethereum-optimism/optimism/pull/13824) for a draft implementation of the next upgrade. Specs coming soon, I promise it will fix this!

PR #525, opened 2025-01-17
feat: update spec to use v4 payload methods for isthmus

### LINKED ISSUE #506: EIP-7685: General Purpose Execution Layer Requests
### Description

[EIP-7685](https://eips.ethereum.org/EIPS/eip-7685) introduces a way of sharing execution layer requests with the consensus layer.

Full inclusion of EIP-7685 on L2 requires batch format changes, gossip format changes, and more.

To support EIP-7685 on L2, the following are needed.
- Infer the block content based on fork, and add the missing attributes.
- Sanity-check in the CL that the EL didn't include any requests. (`op-node`)
- EL requests need to be persisted, or recomputed, somewhere if we do want to have them. Since the batch-submitter has to be aware of them, for other nodes to replicate them from the batch data.
    - The EL requests could be stored in the op-node until the batch-submitter consumes them.
- The engine API `engine_newPayloadV4` is required to build prague blocks, but includes a `executionRequests` attribute. We'll need to pass an empty list as argument.

### References

meyer9:
**Description**

- Updates Isthmus fork doc to specify engine v4 methods to be used
  - Required by https://github.com/ethereum-optimism/specs/issues/506
- Specifies that `executionRequests` must be an empty array

**Tests**

**Additional context**

This PR doesn't revert the withdrawal root changes, so there are still some outdated docs here.

**Metadata**

Fixes #506.


specs/protocol/derivation.md
```diff
@@ -744,19 +744,21 @@ To interact with the engine, the [execution engine API][exec-engine] is used, wi
 - `engine_newPayload`
   - [`engine_newPayloadV2`] — executes a Bedrock/Canyon/Delta execution payload to create a block.
   - [`engine_newPayloadV3`] — executes an Ecotone execution payload to create a block.
+  - [`engine_newPayloadV4`] - executes an Isthmus execution payload to create a block.
 
-The current version of `op-node` uses the `v3` Engine API RPC methods as well as `engine_newPayloadV2`, due to
-`engine_newPayloadV3` only supporting Ecotone execution payloads. Both `engine_forkchoiceUpdatedV3` and
+The current version of `op-node` uses the `v4` Engine API RPC methods as well as `engine_newPayloadV3`, due to
```
danyalprout: We'll use both V2 and V3 for pre-Ecotone blocks.

emhane (approved):
this is super helpful, thank you ser. if you could also open an issue for reverting the withdrawal root changes that would be appreciated.

refcell (approved):
This looks great. Nice job @meyer9

specs/protocol/exec-engine.md
```diff
@@ -328,6 +340,13 @@ The [response][GetPayloadV3Response] is extended to:
 
 In Ecotone it MUST be set to the parentBeaconBlockRoot from the L1 Origin block of the L2 block.
 
+### `engine_getPayloadV4`
+
+[`engine_getPayloadV4`][engine_getPayloadV4] retrieves a payload by ID, prepared by `engine_forkchoiceUpdatedV3`
+when called with `payloadAttributes`.
+`engine_getPayloadV4` **must only be called with Isthmus payload.**
+
```
refcell, line 348: Just need to remove this line to make ci pass.

```suggestion
```

specs/protocol/derivation.md
danyalprout: @meyer9 do we want to keep the V2 comment around? The current version will still use V2.

danyalprout (approved):

tynes (approved):

PR #533, opened 2025-01-22
specs: supervisor api

tynes:
**Description**

Take a quick pass at the supervisor api. This is useful because there
are many teams integrating into this right now.

op-geth has already integrated, the op-reth integration is ongoing as
well as the rbuilder integration should be happening soon.

This enables us to point to spec as the source of truth in how the API
is meant to work.

### DIFF
--- a/specs/interop/supervisor.md
+++ b/specs/interop/supervisor.md
@@ -1,0 +1,245 @@
+# Supervisor
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Overview](#overview)
+- [RPC API](#rpc-api)
+  - [Common types](#common-types)
+    - [`Identifier`](#identifier)
+    - [`Message`](#message)
+    - [`ExecutingDescriptor`](#executingdescriptor)
+    - [`HexUint64`](#hexuint64)
+    - [`Int`](#int)
+    - [`ChainID`](#chainid)
+    - [`Hash`](#hash)
+    - [`Bytes`](#bytes)
+    - [`BlockID`](#blockid)
+    - [`BlockRef`](#blockref)
+    - [`DerivedIDPair`](#derivedidpair)
+    - [`ChainRootInfo`](#chainrootinfo)
+    - [`SuperRootResponse`](#superrootresponse)
+    - [`SafetyLevel`](#safetylevel)
+  - [Methods](#methods)
+    - [`supervisor_checkMessage`](#supervisor_checkmessage)
+    - [`supervisor_checkMessages`](#supervisor_checkmessages)
+    - [`supervisor_checkMessagesV2`](#supervisor_checkmessagesv2)
+    - [`supervisor_crossDerivedToSource`](#supervisor_crossderivedtosource)
+    - [`supervisor_localUnsafe`](#supervisor_localunsafe)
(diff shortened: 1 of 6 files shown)


tynes:
This doesn't include all of the methods, see https://github.com/ethereum-optimism/optimism/blob/develop/op-supervisor/supervisor/frontend/frontend.go#L18-L28 h/t @refcell

specs/interop/supervisor.md
```diff
@@ -0,0 +1,70 @@
...
+data from all of the chains in a cluster so that the [safety level](./verifier.md#safety)
+of executing messages can quickly be determined.
+
+## RPC API
+
+### `supervisor_checkMessage`
+
+Checks the safety level of a specific message based on its identifier and message hash.
+This RPC is useful for the block builder to determine if a message is should be included
+in a block.
```
refcell: It looks to me like there are a few more RPC methods defined in the `op-supervisor`. Noting this so they are not forgotten if needed.

The `op-supervisor` defines the set of query methods [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-supervisor/supervisor/frontend/frontend.go#L18-L28).

These methods are then exposed as RPC methods by the server [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-supervisor/supervisor/service.go#L165-L169) under the `supervisor_*` prefix.

refcell: Do we need these @clabby @ajsutton ?

specs/interop/supervisor.md
```diff
@@ -0,0 +1,70 @@
...
+architecture that can be used to implement it. The supervisor is responsible for indexing
+data from all of the chains in a cluster so that the [safety level](./verifier.md#safety)
+of executing messages can quickly be determined.
+
+## RPC API
+
+### `supervisor_checkMessage`
+
+Checks the safety level of a specific message based on its identifier and message hash.
+This RPC is useful for the block builder to determine if a message is should be included
```
emhane:
```suggestion
This RPC is useful for the block builder to determine if a message should be included
```

specs/interop/supervisor.md
```diff
@@ -0,0 +1,70 @@
...
+in a block.
+
+#### Parameters
+
+- identifier: [Identifier](./messaging.md#message-identifier) - The identifier of the message to check.
+- message hash: Hash - Hash of the initiating message.
```
emhane:
```suggestion
- message hash: Hash - Hash of the initiating message payload.
```

specs/interop/supervisor.md
```diff
@@ -0,0 +1,70 @@
...
+  - timestamp: hexutil.Uint64 - The timestamp of the super root
+  - superRoot: Hash - The root of the super root
+  - chains: []object - List of chains included in the super root
+    - chainId: hexutil.Uint64 - The chainid
+    - canonical: Hash - output root at the latest canonical block
+    - pending: hexutil.Bytes - output root preimage
+
+#### Description
+
+Retrieves the super root state at the specified timestamp, which represents the global state across all monitored chains.
```
emhane: would be nice with a sentence about why this is used, like the method above says "to determine if a message should be included in a block"...what's the point of knowing a super root at a certain timestamp?

tynes:
I believe this is now out of date. It would be ideal if we could update this and get it over the line

protolambda (approved):
I just updated the specs to match the latest actual API. Documentation can still improve, but I strongly think it's better to merge than to keep this PR lingering around. Small PRs can then adjust the wording or add extra bits of information where necessary.

PR #534, opened 2025-01-23
specs: add derivation based dep set updates

tynes:
**Description**

Update: We are putting this on hold until we have a better understanding of how we will do upgrades to the proof system

Adds the derivation based dependency update specs.
A prototype implementation can be found in
https://github.com/defi-wonderland/optimism/pull/197

tldr; derivation creates a system tx that updates the dep
set in L2 by calling the `DependencyManager` predeploy


tynes:
The newly added invariants should live in `verifier.md` not in `derivation.md`

refcell (approved):
Looks great.

tynes:
Going to close this as we are descoping this from the initial launch

PR #535, opened 2025-01-23
interop: offchain derivation invariants

tynes:
**Description**

We are moving a bunch of stuff offchain rather than it being onchain. We now need to be explicit with the invariants as defined in the derivation pipeline. Previously we didn't need to do that because it would revert in the smart contract.

This creates a scaffold for where the expiry window should live.


ajsutton (approved):
Makes sense to me.  https://github.com/ethereum-optimism/optimism/issues/13918 has some early notes on what values might be workable for the expiry window, but will follow up with something that lets us play around with the values better.

tynes:
Thank you for the quick review @ajsutton !

PR #536, opened 2025-01-23
interop: fix optimistic block deposit tx attributes

protolambda:
**Description**

This changes the "optimistic block" (the block that is invalidated, and replaced by another block, that then contains a special system deposit tx) spec.

- Correct the L1 Attributes depositor account address. It linked to the `0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001` address. If we want it to be different, that is fine, but then we shouldn't link to the `0001` address.
  - Update: it will now be `0xdeaddeaddeaddeaddeaddeaddeaddeaddead0002`
- Specify the missing `sourceHash` attribute, and how it's constructed.
- Clarify source-hash domains usage in the deposits spec

### DIFF
--- a/specs/interop/derivation.md
+++ b/specs/interop/derivation.md
@@ -128,13 +128,14 @@ to the trimmed transaction list.
 
 ### Optimistic Block Deposited Transaction
 
-An [L1 attributes deposited transaction](../protocol/deposits.md#l1-attributes-deposited-transaction)
-is a deposit transaction sent to the zero address.
+An Optimistic Block Deposited Transaction is a system deposited transaction,
+inserted into the replacement block,
+to signal when a previously derived local-safe block (the "optimistic" block) was invalidated.
 
 This transaction MUST have the following values:
 
-1. `from` is `0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001` (the address of the
-   [L1 Attributes depositor account](../protocol/deposits.md#l1-attributes-depositor-account))
+1. `from` is `0xdeaddeaddeaddeaddeaddeaddeaddeaddead0002`, like the address of the
+   [L1 Attributes depositor account](../protocol/deposits.md#l1-attributes-depositor-account), but incremented by 1
 2. `to` is `0x0000000000000000000000000000000000000000` (the zero address as no EVM code execution is expected).
 3. `mint` is `0`
 4. `value` is `0`
--- a/specs/interop/derivation.md
+++ b/specs/interop/derivation.md
@@ -13,6 +13,7 @@
       - [Deposits-complete Source-hash](#deposits-complete-source-hash)
 - [Replacing Invalid Blocks](#replacing-invalid-blocks)
   - [Optimistic Block Deposited Transaction](#optimistic-block-deposited-transaction)
+    - [Optimistic Block Source-hash](#optimistic-block-source-hash)
 - [Expiry Window](#expiry-window)
 - [Security Considerations](#security-considerations)
   - [Gas Considerations](#gas-considerations)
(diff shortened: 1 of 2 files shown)


specs/interop/derivation.md
```diff
@@ -132,19 +133,31 @@ is a deposit transaction sent to the zero address.
...
 
 This system-initiated transaction for L1 attributes is not charged any ETH for its allocated
 `gasLimit`, as it is considered part of state-transition processing.
 
+[L2 output root]: ../glossary.md#l2-output-root-proposals
+
+#### Optimistic Block Source-hash
+
+The source hash is [computed](../protocol/deposits.md#source-hash-computation)
+with a source-hash domain: `4` (instead of `1`),
```
clabby, line 156: noob question, but why keep making new deposit source domains?

Interop does already introduce one more @ `3` for the special deposit that closes the EVM-aware deposit window in the `L1Block` contract. Is it useful to introduce another one here, rather than create a more generic system tx source domain w/ `3` (or repurpose `2`, currently for upgrades, collapsing `3` + `4` into it, and just calling it `System`)?

What actually uses these identifiers to delineate between deposit sources?

clabby (approved):
`from` update lgtm, small question about deposit source domains.

specs/interop/derivation.md
Inphi: Also, if we're changing the domain then it's no longer an L1 attributes deposited tx based on the [rules for computing the source hash](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/deposits.md#source-hash-computation). You'd have to introduce a new tx type to avoid complicating the spec.

specs/interop/derivation.md
```diff
@@ -132,19 +133,31 @@ is a deposit transaction sent to the zero address.
 
 This transaction MUST have the following values:
 
-1. `from` is `0xdeaddeaddeaddeaddeaddeaddeaddeaddead0002` (the address of the
+1. `from` is `0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001` (the address of the
```
ajsutton: Original thought was that it would be from a unique address which makes it easy to identify this transaction without having to parse the data (and eliminates any risk that you could create a transaction that looks the same, though I think there are other factors that prevent that currently anyway). 

I'm not strongly opinionated either way but it seemed like a good idea at the time (just missed updating the part in brackets).

specs/interop/derivation.md
protolambda: @Inphi It's not meant to be a L1 attributes deposited tx, it's unrelated to L1.

protolambda: @clabby it's a safety thing, to ensure two types of system txs can never accidentally have the same source-hash. It may be a bit overkill, but it's very easy to spec / implement the idea of a unique domain number to combine with it.

protolambda: No tool / thing currently needs it from a non-safety use-case, but like the output-root, if we ever need to unpack the preimage of an unknown type of source-hash, then we can do so safely.

Inphi: In which case, let's fix the blurb on L131 regarding the L1 deposited tx.

specs/interop/derivation.md
tynes: Are we strongly opinionated on what the `from` address is @protolambda? I generally agree with @ajsutton here about the simplicity of identifying the transaction, we would need to be more careful about future changes if we reuse the same `from`, it increases the chances of a bug

tynes:
I'd like to merge this but looks like it has a couple of open questions still

specs/interop/derivation.md
protolambda: Ok, let's change the address to `...0002`. I'll update the implementation too

protolambda:
Will update the specs to resolve the review items :+1:

protolambda:
ready for review

protolambda:
Matching implementation update: https://github.com/ethereum-optimism/optimism/pull/14023

tynes (approved):

PR #553, opened 2025-01-27
Isthmus: EIP-2537

### LINKED ISSUE #509: EIP-2537: Precompile for BLS12-381 curve operations
https://eips.ethereum.org/EIPS/eip-2537

#### op-revm

- [ ] https://github.com/bluealloy/revm/pull/2000 if eip isn't included in isthmus, we will simply have to update the name of that function at a later date to reflect the name of the hf that does include it.

### References

refcell:
### Description

Adds EIP-2537 to the isthmus hardfork.

One thing we still need to add here is specify the byte size at which the bls curve must short-circuit since it reaches the `20M` gas limit.

Closes #509


specs/protocol/isthmus/exec-engine.md
```diff
@@ -131,6 +131,16 @@ is applicable for scenarios where the actual withdrawals root value is not readi
...
 Withdrawals list in the block body is encoded as an empty RLP list.
 
+## EVM Changes
+
+Similar to the `bn256Pairing` precompile in the [granite hardfork](../granite/exec-engine.md),
+[EIP-2537](https://eips.ethereum.org/EIPS/eip-2537) introduces a BLS
+precompile that short-circuits depending on input size in the EVM.
+
+The `BLS12-381` curve operation precompile reverts if its input is
+larger than `<todo>` bytes. This is the input size that consumes
```
refcell: See `<todo>` here.

emhane: how do we get this number?

emhane (approved):
lgtm

specs/protocol/isthmus/exec-engine.md
meyer9: Here's the calculations for those:

Source: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2537.md

- G1 multiplication cost: 12000
- G2 multiplication cost: 22500
- G1 MSM:
  - (k * multiplication_cost * max_discount) // 1000 < 20000000 
  - (k * 12000 * 519) // 1000 < 20000000
  - k < 3211.3, so k <= 3211
  - bytes = 160 * k
  - bytes <= 513760

- G2 MSM:
  - (k * multiplication_cost * max_discount) // 1000 < 20000000 
  - (k * 22500 * 524) // 1000 < 20000000
  - k < 1696.35, so k <= 1696
  - bytes = 288 * k
  - bytes <= 488448

- Pairing check:
  - 32600 * k + 37700 < 20000000
  - k <= 612
  - bytes = 384 * k
  - bytes <= 235008

tynes: > The bn256Pairing precompile execution has additional validation on its input. The precompile reverts if its input is larger than 112687 bytes. This is the input size that consumes approximately 20 M gas given the latest bn256Pairing gas schedule on L2.

https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/granite/exec-engine.md#bn256pairing-precompile-input-restriction

For consistency, we should follow the same reasoning in the above and cap the size to ~20 million gas. Can we fill in the todo before we merge this?

cc @Inphi @ajsutton for visibility

tynes (changes requested):
Lets fill in the todo before merge

specs/protocol/isthmus/exec-engine.md
refcell: TODO filled in with `235008` maximum input byte size where the gas limit is 20M.

This follows @meyer9 's pairing check math.

```
32600 * k + 37700 < 20000000
k <= 612
bytes = 384 * k
bytes <= 235008
```

specs/protocol/isthmus/exec-engine.md
```diff
@@ -131,6 +132,16 @@ is applicable for scenarios where the actual withdrawals root value is not readi
...
 
+## EVM Changes
+
+Similar to the `bn256Pairing` precompile in the [granite hardfork](../granite/exec-engine.md),
+[EIP-2537](https://eips.ethereum.org/EIPS/eip-2537) introduces a BLS
+precompile that short-circuits depending on input size in the EVM.
+
+The `BLS12-381` curve operation precompile reverts if its input is
+larger than `235008` bytes. This is the input size that consumes
+approximately `20M` gas given the latest `BLS12-381` gas schedule on L2.
```
meyer9: Is this the only curve operation we're accelerating? G1_MSM/G2_MSM have variable input length and could cause a large increase in `op-program` runtime if a ton of multiplications are passed in, but I guess the 20M gas limit only matters if they're accelerated precompiles. What's the downside to accelerating all the BLS precompiles?

refcell: Good question - I think other operations could be accelerated. CC @Inphi @ajsutton would appreciate your input here

clabby: We should benchmark the others in the proof for sure, but pairing checks tend to be really expensive in the prover even with the miller loop optimization, so that'd be the first one I'm worried about.

But for example, bn128 add enables the worst-case block construction (at least in the samples that @BrianBland ran in his benchmarks), so we should also look into the others.

tynes: My current understanding is that without a `no_std` rust implementation of these precompiles, we will need to accelerate them. Has the final decision been made here yet?

meyer9: I think accelerating all of them is the right call. We know accelerating makes sense for add (fixed input), hash-to-point (fixed input), and pairing check (very expensive), so multi-scalar-multiplication is the only one that's up in the air.

- Downside to accelerating: precompile is limited to a smaller size per invocation
- Upside to accelerating: no impact on FP program performance and FP program doesn't have to do expensive calculations

I'm still working on getting the benchmarks working with Isthmus is the meantime.

refcell: Discussed with @ajsutton and @clabby today - we cannot support these precompiles in kona for `no_std` without a significant amount of work, so we've largely settled on accelerating all these precompiles.

I'd say as such the priority of the benchmarks can be downgraded to nice-to-have. If your low on bandwidth, it's not critical.

meyer9:
Opened a PR to add size limits and acceleration to this spec: https://github.com/ethereum-optimism/specs/pull/570

tynes (approved):

PR #554, opened 2025-01-27
Isthmus: EIP-7251

### LINKED ISSUE #502: EIP-7251: Increase the MAX_EFFECTIVE_BALANCE
https://eips.ethereum.org/EIPS/eip-7251

### References

refcell:
### Description

Specifies [EIP-7251](https://eips.ethereum.org/EIPS/eip-7251) changes required for L2 Pectra as part of the Isthmus hardfork.

Closes #502


specs/protocol/isthmus/exec-engine.md
```diff
@@ -131,6 +132,13 @@ is applicable for scenarios where the actual withdrawals root value is not readi
 
 Withdrawals list in the block body is encoded as an empty RLP list.
 
+## Block Sealing
+
+[EIP-7251](https://eips.ethereum.org/EIPS/eip-7251) introduces a new consolidation
+[EIP-7685](https://eips.ethereum.org/EIPS/eip-7685) request with type `CONSOLIDATION_REQUEST_TYPE` (or `0x02`).
```
emhane: does it make sense to use the word 'upgrade of' instead of 'consolidation' here?

specs/protocol/isthmus/exec-engine.md
```diff
@@ -131,6 +132,13 @@ is applicable for scenarios where the actual withdrawals root value is not readi
...
+After Isthmus, the OP Stack block sealing code must not add requests of `CONSOLIDATION_REQUEST_TYPE` if the optimism config
+is active. Note, this does not need to activate after any specific hardfork as this type does not exist pre-Pectra.
```
emhane: a bit hard to understand imo, isn't it enough to say
requests of type `CONSOLIDATION_REQUEST_TYPE` must be ignored
?

emhane (approved):

specs/protocol/isthmus/exec-engine.md
refcell: I think it's worthwhile to note that they can always be ignored and this rule to ignore doesn't need to be timestamp based since the type did not exist prior to pectra.

specs/protocol/isthmus/exec-engine.md
refcell: Less familiar with L1 jargon so happy to defer to your opinion here. I tried to keep this as close to the eip as possible

specs/protocol/isthmus/exec-engine.md
emhane: read similar in https://github.com/ethereum-optimism/specs/pull/555, made more sense there. I like how you format the `_not_` in that one.

sebastianst: Agree, it reads easier over there

specs/protocol/isthmus/exec-engine.md
sebastianst: Where is the request of type 0x02 defined? It's not in the eip-7685 spec.
And do we want to specify any behavior for other (possibly unknown) request types?

specs/protocol/isthmus/exec-engine.md
emhane: re-wrote this a bit @refcell . does it make sense?

emhane:
re-request review @sebastianst

specs/protocol/isthmus/exec-engine.md
refcell: Think this reads well. If someone has verbiage changes they can always follow up as well to improve!

tynes (approved):

PR #555, opened 2025-01-27
Isthmus: EIP-6110

### LINKED ISSUE #505: EIP-6110: Validator Deposits on Chain
### Description

[EIP-6110](https://eips.ethereum.org/EIPS/eip-6110) includes validator deposits in the execution layer, removing deposit voting from the consensus layer.

From @protolambda:
> Can consider an upgrade deposit-tx, with same sender as the pre-determined example tx, to ensure the contract is deployed.
> The contract is the same, but effectively no-op, not affecting the block.

### References

refcell:
### Description

Specifies [EIP-6110](https://eips.ethereum.org/EIPS/eip-6110) changes required for L2 Pectra as part of the Isthmus hardfork.

Closes #505

### DIFF
--- a/specs/protocol/isthmus/exec-engine.md
+++ b/specs/protocol/isthmus/exec-engine.md
@@ -17,6 +17,7 @@
     - [Forwards Compatibility Considerations](#forwards-compatibility-considerations)
     - [Client Implementation Considerations](#client-implementation-considerations)
       - [Transaction Simulation](#transaction-simulation)
+- [Deposit Requests](#deposit-requests)
 - [Block Body Withdrawals List](#block-body-withdrawals-list)
 - [Engine API Updates](#engine-api-updates)
   - [Update to `ExecutableData`](#update-to-executabledata)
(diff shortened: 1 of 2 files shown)


emhane (approved):

meyer9 (approved):

sebastianst (approved):

specs/protocol/isthmus/exec-engine.md
```diff
@@ -127,6 +128,17 @@ In response to RPC methods like `eth_simulateV1` that allow simulation of arbitr
 an empty withdrawals root should be included in the header of a block that consists of such simulated transactions. The same
 is applicable for scenarios where the actual withdrawals root value is not readily available.
 
+## Deposit Requests
+
+[EIP-6110] shifts deposit to the execution layer, introducing a new [EIP-7685] deposit request of type
```
tynes, line 133: We should be very clear about the existence of the system contract bytecode. This means the deposit contract. Do we want to have the deposit contract now in the genesis state of the OP Stack as well as deployed as a system transaction? I don't think its really necessary to be honest

 We should also be very clear to say that DepositEvent logs emitted from the deposit contract are ignored. To be clear this is a deposit contract that is deployed to L2 rather than L1

tynes: The deposit requests are populated from the events from the deposit contract

- It is underspecified if we include the deposit contract in the genesis state of the op stack, or if we deploy it via a network upgrade tx to chains upgrading

What are the pros and cons of including it or not including it?

refcell: I guess the pros of shoving it in the genesis state are less of a diff to maintain since we'll need to increase the diff to deploy it via network upgrade txs.

tynes: We always need to handle both the upgrade case and the from genesis case

meyer9: I think we should deploy these contracts similar to Holocene, otherwise the EL block processing code will have to differ more for Optimism. Right now, we still process these contracts the same as L1, but just never send requests to the CL or include them in the block header.

meyer9: Actually, reading a bit more - seems like this just uses the existing deposit contract address. For this EIP, it should be a no-op. (Our contract won't emit any deposit events or include any requests in the block header or payload calls).

refcell: Yes believe @meyer9 is right here - this is what @emhane has done in `op-reth`, making it behave as a no-op.

refcell: Going to land this pr now since we can follow up with changes if needed.

Issue #511, opened 2025-01-15
EIP-7702: Set EOA account code


emhane:
### Description

[EIP-7702](https://eips.ethereum.org/EIPS/eip-7702) adds a new transaction type for EOA account code.

- `op-alloy` modifications needed.
- `kona` modifications needed.
    The fault proof program needs to be updated to read EIP-7702 receipts
    https://github.com/op-rs/kona/blob/main/crates/executor/src/executor/mod.rs#L237-L240
    Calldata should continue to *not* read EIP-7702 txs since eip-7702 txs are not posted to l1
    https://github.com/op-rs/kona/blob/main/crates/derive/src/sources/calldata.rs#L50-L55
- `op-reth` modifications needed.

### References

- [Pectra Devnet 5 EIP List](https://notes.ethereum.org/@ethpandaops/pectra-devnet-5)
- [Proto's list of Pectra EIPs](https://gist.github.com/protolambda/59b534d804c23dc7b3592ad1c38048d4#eip-7685-general-purpose-el-requests)
- Pectra support on the OP Stack tracking issue: https://github.com/ethereum-optimism/optimism/issues/12435

tynes:
We need to update the spec for span batches to support 7702 txs. The span batch serialization is domain specific and has to know about each transaction type it is compressing. Some details can be found in https://github.com/ethereum-optimism/optimism/issues/12435

Right now the derivation pipeline deserializes all L1 transactions. This means that it has to know about the L1 tx types. We generally want to decouple having to update the derivation pipeline from L1 hardforks to remove stressful releases based on L1 timelines. We are planning on retrofitting the spec to only deserialize the transaction types that exist today so it ignores 7702 transactions. If we do not do this before Pectra goes live on L1, then both the proof and the L2 nodes will break. See https://github.com/ethereum-optimism/optimism/issues/13379#issuecomment-2547430177 and https://github.com/ethereum-optimism/optimism/issues/13627 for context

meyer9:
This was merged in!

PR #556, opened 2025-01-27
Isthmus: EIP-7702 Span Batch Updates

### LINKED ISSUE #318: Pectra: Span Batch Support
We will not be able to pull the Paris EVM spec as part of the Pectra hardfork into the OP stack until span batches are updated to support [EIP-7702](https://github.com/ethereum/EIPs/blob/233696d2599e8ad5213a3f26ad1cdf2d8b740be9/EIPS/eip-7702.md) transactions. You can see and example of the lack of support [here](https://github.com/ethereum-optimism/optimism/blob/8741423ea1d5bd9080583783342d15ccde688ea9/op-node/rollup/derive/span_batch_tx.go#L97)

EIP-7702 is a huge UX win for Ethereum and we certainly want to be able to support it as early as possible. It will be a huge differentiator between EVM chains as applications that are built on it will not be able to support the same UX on chains that do not support it.

The span batch spec mentions a [v2](https://github.com/ethereum-optimism/specs/blob/afadeea642d05eda6e111d9098c9a842c2ecf322/specs/protocol/delta/span-batches.md#future-batch-format-extension), we should consider adding EIP-7702 support to span batches v2.

Pectra is targeted to ship in early 2025 and we also want to make sure the spec for 7702 has been finalized before we invest too much time in implementing this

refcell:
### Description

Updates the derivation spec in Isthmus to include the update required to the span batches to support the [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702) transaction type.

Part of #511 , closes https://github.com/ethereum-optimism/specs/issues/318

### DIFF
--- a/specs/protocol/isthmus/derivation.md
+++ b/specs/protocol/isthmus/derivation.md
@@ -61,12 +61,15 @@ Span batches contain the L1 transactions and transaction types that are posted c
 Since [EIP-7702] introduces a new transaction type, the Span Batch must be updated to support the [EIP-7702]
 transaction.
 
-Part of this update is to include the new Span Batch transaction variant.
-That is, the `tx_types` must now include the [EIP-7702] tx type and the `tx_datas` transaction data
-must support a new [EIP-7702] transaction data format. The new [EIP-7702] data format
-is an extension of [EIP-1559] with a new `authorization_list` field.
+This corresponds with a new RLP-encoding of the `tx_datas` list as specified in
+[the Delta span batch spec](../delta/span-batches.md), adding a new transaction type:
 
-The [EIP-7702] transaction data contains the following.
+Transaction type `4` ([EIP-7702]):
+`0x04 ++ rlp_encode(value, max_priority_fee_per_gas, max_fee_per_gas, data, access_list, authorization_list)`
+
+The [EIP-7702] transaction extends [EIP-1559] to include a new `authorization_list` field.
+`authorization_list` is an RLP-encoded list of authorization tuples.
+The [EIP-7702] transaction format is as follows.
 
 - `value`: The transaction value as a `u256`.
 - `max_priority_fee_per_gas`: The maximum priority fee per gas allowed as a `u256`.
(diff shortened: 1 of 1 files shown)


specs/protocol/isthmus/derivation.md
```diff
@@ -51,3 +52,11 @@ cast keccak $(cast concat-hex 0x000000000000000000000000000000000000000000000000
...
 This transaction MUST deploy a contract with the following code hash
 `0x6e49e66782037c0555897870e29fa5e552daf4719552131a0abce779daec0a5d`.
+
+# Span Batch Updates
+
+[Span batches](../delta/span-batches.md) are a span of consecutive L2 blocks than are batched submitted.
+
+Span batches contain the L1 transactions and transaction types that are posted containing the span of L2 blocks.
+Since [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702) introduces a new transaction type, the Span Batch must
+be updated to support the [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702) transaction.
```
sebastianst: Is there a precise specification of how SetCode transactions are encoded in a span batch somewhere already?

sebastianst (approved):

specs/protocol/isthmus/derivation.md
tynes: Ideally this spec contains the exact serialization of a span batch before its merged

tynes (changes requested):
Can we update this with the new span batch spec before merge?

meyer9:
I think the only difference between 1559 and 7702 span batch data is that 7702 transactions include an extra `authorization_list` field (RLP-encoded) that matches the passed transaction. Ref: https://gist.github.com/protolambda/59b534d804c23dc7b3592ad1c38048d4#eip-7702-set-eoa-account-code-for-one-transaction

meyer9:
Added this here if you want to pull it in! https://github.com/ethereum-optimism/specs/pull/561

specs/protocol/isthmus/derivation.md
refcell: Updated. Let me know if this is what you're looking for :)

tynes (approved):

PR #561, opened 2025-01-28
Add encoding for EIP-7702 transactions in span batch

meyer9:
Adds description for EIP-7702 transaction encoding in span batches.

This matches EIP-1559, but adds `authorization_list`.


tynes:
@refcell is this clear enough to implement completely based on this?

tynes:
In the most ideal world the entire span batch spec is rewritten and cleaned up. I dont think that is a requirement here given we want to land pectra in L2 asap

emhane (approved):

meyer9:
> In the most ideal world the entire span batch spec is rewritten and cleaned up. I dont think that is a requirement here given we want to land pectra in L2 asap

Yeah would be nice to have the hardfork change documents as well as a global source of truth with the up-to-date spec. I'll look into adding that in another PR.

PR #566, opened 2025-01-31
specs: clean up interop predeploys

tynes:
**Description**

The interop predeploy specs needed some cleanup and standardization.
More to come.

### DIFF
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -70,35 +70,44 @@
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


emhane (changes requested):
can this be rescoped into smaller prs? will avoid merge conflicts over time as approving reviews are pending, and it is more difficult to sign off a large diff as opposed to a small diff in a single review. for smaller prs, there is the chance that first review can approve changes already.

PR #569, opened 2025-01-31
feat: specification for the AnchorStateRegistry as source of truth

smartcontracts:
Updates the AnchorStateRegistry and OptimismPortal specifications such that the AnchorStateRegistry acts as the source of truth for the validity of dispute game contracts.

### DIFF
--- a/specs/fault-proof/stage-one/anchor-state-registry.md
+++ b/specs/fault-proof/stage-one/anchor-state-registry.md
@@ -143,17 +143,25 @@ A Dispute Game is considered to be a **Proper Game** if it has not been invalida
 the mechanisms defined by the `AnchorStateRegistry` contract. A Proper Game is, in a sense, a
 "clean" game that exists in the set of games that are playing out correctly in a bug-free manner. A
 Dispute Game can be a Proper Game even if it has not yet resolved or resolves in favor of the
-Challenger. A Game that was previously a Proper Game can no longer be a Proper Game if it is later
-invalidated.
+Challenger.
 
-Specifically, a game is considered to be a Proper Game if all of the following are true:
+A Dispute Game that is **NOT** a Proper Game can also be referred to as an **Improper Game** for
+brevity. A Dispute Game can go from being a Proper Game to later *not* being an **Improper Game**
+if it is invalidated by being [blacklisted](#blacklisted-game) or [retired](#retired-game).
+
+**ALL** Dispute Games **TEMPORARILY** become Improper Games while the Superchain-wide pause
+mechanism is active. However, this is a *temporary* condition such that Registered Games that are
+not invalidated by [blacklisting](#blacklisted-game) or [retirement](#retired-game) will become
+Proper Games again once the pause is lifted. The Superchain-wide pause is therefore a way to
+*temporarily* prevent Dispute Games from being used by consumers like the `OptimismPortal` while
+relevant parties coordinate the use of some other invalidation mechanism.
+
+A Game is considered to be a Proper Game if all of the following are true:
 
 - The game is a [Registered Game](#registered-game)
 - The game is **NOT** a [Blacklisted Game](#blacklisted-game)
 - The game is **NOT** a [Retired Game](#retired-game)
-
-Additionally, ALL games will NOT be considered to be Proper Games while the Superchain-wide pause
-is active.
+- The Superchain-wide pause is not active
 
 ### Resolved Game
 
--- a/specs/fault-proof/stage-one/anchor-state-registry.md
+++ b/specs/fault-proof/stage-one/anchor-state-registry.md
@@ -8,9 +8,11 @@
 - [Definitions](#definitions)
   - [Dispute Game](#dispute-game)
   - [Respected Game Type](#respected-game-type)
+  - [Dispute Game Finality Delay (Airgap)](#dispute-game-finality-delay-airgap)
   - [Registered Game](#registered-game)
   - [Respected Game](#respected-game)
   - [Blacklisted Game](#blacklisted-game)
+  - [Retirement Timestamp](#retirement-timestamp)
   - [Retired Game](#retired-game)
   - [Proper Game](#proper-game)
   - [Resolved Game](#resolved-game)
@@ -25,10 +27,8 @@
     - [Mitigations](#mitigations)
   - [aASR-002: DisputeGameFactory properly reports its created games](#aasr-002-disputegamefactory-properly-reports-its-created-games)
     - [Mitigations](#mitigations-1)
-  - [aASR-003: OptimismPortal properly reports respected game type, blacklist, and retirement time](#aasr-003-optimismportal-properly-reports-respected-game-type-blacklist-and-retirement-time)
+  - [aASR-003: Incorrectly resolving games will be invalidated within the airgap delay period](#aasr-003-incorrectly-resolving-games-will-be-invalidated-within-the-airgap-delay-period)
     - [Mitigations](#mitigations-2)
-  - [aASR-004: Incorrectly resolving games will be invalidated within the airgap delay period](#aasr-004-incorrectly-resolving-games-will-be-invalidated-within-the-airgap-delay-period)
-    - [Mitigations](#mitigations-3)
 - [Invariants](#invariants)
   - [iASR-001: Games are represented as Proper Games accurately](#iasr-001-games-are-represented-as-proper-games-accurately)
     - [Impact](#impact)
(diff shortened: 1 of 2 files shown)


specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -230,20 +246,11 @@ created.
...
-We assume that the `OptimismPortal` contract properly and faithfully reports the respected game
-type, blacklist, and retirement timestamp.
-
-#### Mitigations
-
-- Existing audit on the `OptimismPortal` contract
-- Integration testing
-
-### aASR-004: Incorrectly resolving games will be invalidated within the airgap delay period
+### aASR-003: Incorrectly resolving games will be invalidated within the airgap delay period
```
smartcontracts: Consider the pause

specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -102,39 +111,49 @@ system's `DisputeGameFactory` contract.
...
 
-- The game is a Registered Game
-- The game is not a Blacklisted Game
-- The game is not a Retired Game
+- The game is a [Registered Game](#registered-game)
+- The game is **NOT** a [Blacklisted Game](#blacklisted-game)
+- The game is **NOT** a [Retired Game](#retired-game)
+
+Additionally, ALL games will NOT be considered to be Proper Games while the Superchain-wide pause
+is active.
```
clabby: While not a problem, really important to point out in the incident response runbooks that un-pausing after a few days of being paused, without also updating the retirement timestamp, could very quickly get the system into a dangerous state.

clabby: This also kind-of conflicts with "A Game that was previously a Proper Game can no longer be a Proper Game if it is later invalidated," though might be mincing words there. From this, it sounds like games can transition from `proper -> improper -> proper`, if the only thing that made a game temporarily improper was that the superchain was paused.

clabby: > un-pausing after a few days of being paused, without also updating the retirement timestamp, could very quickly get the system into a dangerous state.

Also followed up on here: https://github.com/ethereum-optimism/specs/pull/569/files#diff-7565371fb91f003706461dfdebc80ed23edc8e2d89b4bfeda5c41f22b58d5e2cR263-R264

specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -349,10 +367,68 @@ finalize withdrawals with invalidated games would be considered Critical Severit
...
+- MUST set the value of the `SuperchainConfig` contract that stores the address of the Guardian.
+- MUST set the value of the `DisputeGameFactory` contract that creates Dispute Game instances.
+- MUST set the value of the [Starting Anchor State](#starting-anchor-state).
+- MUST set the value of the initial [Respected Game Type](#respected-game-type).
+- MUST set the value of the [Retirement Timestamp](#retirement-timestamp) to the current block
+  timestamp. NOTE that this is a safety mechanism that invalidates all existing Dispute Game
+  contracts to support the safe transition away from the `OptimismPortal` as the source of truth
+  for game validity. In this way, the `AnchorStateRegistry` does not need to consider the state of
+  the legacy blacklisting/retirement mechanisms within the `OptimismPortal` and starts from a clean
+  slate.
```
clabby, line 393: Are we moving away from the re-initialization pattern we use for upgrades (i.e. using `StorageSetter` to clear `initialized`, and then re-calling `initialize`?) With this, invalidating ongoing withdrawals could be a footgun for ASR proxy upgrades if not.

specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -364,30 +440,31 @@ Determines if a game is a Registered Game.
...
   Check should be a strict comparison that the creation is less than the retirement timestamp.
 
 ### isGameProper
 
 Determines if a game is a Proper Game.
 
-- MUST return `true` if and only if `isGameRegistered(game)` is `true` and
-  `isGameBlacklisted(game)` and `isGameRetired(game)` are both `false`.
+- MUST return `true` if and only if `isGameRegistered(game)` is `true`, `isGameBlacklisted(game)`
+  and `isGameRetired(game)` are both `false`, and `paused()` is `false`.
```
clabby, line 470: Is `paused()` a function that we're exposing on the ASR now? small nit if not:
```suggestion
  and `isGameRetired(game)` are both `false`, and `SuperchainConfig.paused()` is `false`.
```

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -229,13 +272,14 @@ Allows a user to [prove](#proven-withdrawal) a withdrawal transaction.
 
 Checks that a withdrawal transaction can be [finalized](#finalized-withdrawal).
 
+- MUST revert if the withdrawal being finalized has already been finalized.
 - MUST revert if the withdrawal being finalized has not been proven.
+- MUST revert if the withdrawal was proven at a timestamp less than the creation timestamp of the
```
clabby, line 277: Sounds impossible, but ok to keep in the implementation as a sanity check imo.

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -229,13 +272,14 @@ Allows a user to [prove](#proven-withdrawal) a withdrawal transaction.
...
+  dispute game it was proven against, which would signal an unexpected proving bug.
 - MUST revert if the withdrawal being finalized has been proven less than
   [Proof Maturity Delay](#proof-maturity-delay) seconds ago.
-- MUST revert if the withdrawal being finalized was proven against a game that is not a
-  [Valid Game](./anchor-state-registry.md#valid-game).
-- MUST revert if the withdrawal being finalized has already been finalized.
-- MUST otherwise return `true`.
```
clabby, line 238: Was this removed because it just reverts outright now?

clabby (commented):
This looks good to me. Couple small questions before ✅

specs/fault-proof/stage-one/anchor-state-registry.md
smartcontracts: Great points, will make sure this is very clearly called out

specs/fault-proof/stage-one/anchor-state-registry.md
smartcontracts: My understanding is that future upgrades will need a separate `upgrade` function. Let me ping @maurelian to confirm.

specs/fault-proof/stage-one/anchor-state-registry.md
smartcontracts: Yes, it'll be exposed on the ASR.

specs/fault-proof/stage-one/optimism-portal.md
smartcontracts: Yeah I think it's impossible but no harm in keeping the check.

specs/fault-proof/stage-one/optimism-portal.md
smartcontracts: It actually never returned true - I think this was a spec bug

ajsutton (approved):
LGTM.

specs/fault-proof/stage-one/anchor-state-registry.md
clabby: Changes look good here.

specs/fault-proof/stage-one/anchor-state-registry.md
clabby: Sounds good. Can resolve this comment once this is sorted; I don't think this is too bad, just need to make sure the potential footgun is documented. Even if it was hit, though, it'd be safe (just a temp withdrawal liveness failure.)

clabby (approved):

ControlCplusControlV (approved):
:shipit:

specs/fault-proof/stage-one/anchor-state-registry.md
maurelian: @smartcontracts is correct that the way we'll handle adding new state vars in the future is with a `function(uint newVar) public initializer()` method that only sets the new variables. 

However in this case I believe the intent is for this to be an entirely new superchain-wide ASR(?), so it's not being upgraded from a previous state and wouldn't need that.

tynes (approved):

smartcontracts:
Waiting on one more question for @ajsutton

PR #571, opened 2025-02-01
feat: specify requests hash must be empty

### LINKED ISSUE #568: Specify that requests root in block should be an empty hash post-Isthmus
Currently, we specify that the payload API calls should have an empty requests array, but we should probably also specify that blocks should not include any requests in the block header.

meyer9:
**Description**

Specifies that `requests_hash` should be set to `calcRequestsHash([]) = sha256('') = 0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` after Isthmus.

EIP ref: https://eips.ethereum.org/EIPS/eip-7685#execution-layer

**Metadata**

Fixes #568

### DIFF
--- a/specs/protocol/isthmus/exec-engine.md
+++ b/specs/protocol/isthmus/exec-engine.md
@@ -50,16 +50,21 @@ at the given block number.
 
 ### Header Validity Rules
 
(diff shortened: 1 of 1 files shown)


refcell (approved):
Nicely written

tynes (approved):

PR #580, opened 2025-02-06
Update linting

alcueca:
**Description**

The README was not consistent with regards to the linting recipes, for example pointing to the toc check instead of the toc fix, and missing an amount of useful information.

I changed the recipes so that they are now consistent:
 - To check something you do `lint-something`
 - To fix something you do `lint-something-fix`

The README has been updated to match and be more exhaustive.

**Tests**

Tested locally. Spelling checks fail, so maybe we haven't been checking them too hard?


alcueca:
image

mds1 (approved):
LGTM, I don't think my approvals count in this repo, cc ~@clabby~ @tynes ~@smartcontracts~ for review since it looks like you've worked on the lint config and have approvals that count

mds1:
Ah, approval needs to come from one of @ajsutton @mslipper @protolambda @sebastianst @tynes

Justfile
```diff
@@ -10,33 +10,27 @@ deps:
...
 
 # Validates Table of Content Sections with doctoc
-lint-specs-toc-check:
+lint-specs-toc:
     npx doctoc '--title=**Table of Contents**' ./specs && git diff --exit-code ./specs
 
 # Updates Table of Content Sections with doctoc
-lint-specs-toc:
+lint-specs-toc-fix:
     npx doctoc '--title=**Table of Contents**' ./specs
```
ajsutton: I'm not sure why we're changing the meaning of existing targets here - that will just break things for people running them locally.  And the shorter name for the fix is probably better given that's what people running manually probably want to do and only CI really needs the -check variant.

alcueca: I'm fine with using the shorter name for fix, and a longer name for checks. Next commit does that.

What I'm trying to fix here is that we use all possible combinations for naming according to fix/check, which I found quite confusing when trying to work out how this worked. Currently:
- `lint`/`lint-fix`
- `lint-specs-md-check`/`lint-specs-md-fix`
- `lint-specs-toc-check`/`lint-specs-toc`
- `lint-specs-spelling`/`lint-specs-spelling-fix`

As of now, this PR proposes:
- `lint-check`/`lint`
- `lint-specs-md-check`/`lint-specs-md`
- `lint-specs-toc-check`/`lint-specs-toc`
- `lint-specs-spelling-check`/`lint-specs-spelling`

tynes (approved):

alcueca:
I've added an alias in the justfile so that `-check` checks, and both the short version and `-fix` fix. I think that this preserves enough current processes.

@ajsutton, would you be happy for me to merge it now?

PR #587, opened 2025-02-07
isthmus: Clarify SetCode tx type handling in span batches

sebastianst:
See also discussion in
https://github.com/ethereum-optimism/optimism/pull/14197

### DIFF
--- a/specs/protocol/isthmus/derivation.md
+++ b/specs/protocol/isthmus/derivation.md
@@ -14,6 +14,7 @@
   - [GasPriceOracle Enable Isthmus](#gaspriceoracle-enable-isthmus)
   - [EIP-2935 Contract Deployment](#eip-2935-contract-deployment)
 - [Span Batch Updates](#span-batch-updates)
+  - [Activation](#activation)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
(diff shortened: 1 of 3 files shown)


clabby (approved):
Looks good to me. Thanks for the update.

ajsutton (approved):

specs/protocol/isthmus/derivation.md
```diff
@@ -336,7 +337,18 @@ The [EIP-7702] transaction format is as follows.
 - `access_list`: The [EIP-2930] access list.
 - `authorization_list`: The [EIP-7702] signed authorization list.
 
-Span batches with transaction type `4` should only be accepted after Isthmus is enabled.
+## Activation
+
+Singular batches with transactions of type `4` must only be accepted if Isthmus is active at the
+timestamp of the batch. If a singular batch contains a transaction of type `4` before Isthmus is
+active, this batch must be _dropped_. Note that according to Holocene derivation, this will also
+lead to the remaining span batch, and channel that contained it, to get dropped.
```
refcell: Clarifying here that this means _pre_ holocene if a batch within a span batch is invalidated by an EIP-7702 transaction, the whole span batch is dropped correct?

refcell: Nevermind - clarified below.

sebastianst: Yes that's correct. If a pre-Holocene L2 chain uses a Pectra-activated L1 chain, a single SetCode tx in a span batch will lead to dropping the whole span batch, in particular older batches.

We could technically also add a sentence about batch validity to the bedrock derivation specs, because this is really a retrofitted pre-Isthmus rule.

refcell (approved):

specs/protocol/isthmus/derivation.md
refcell: Great thanks - that's how i interpreted it so I think it's clear as is. Approved :)

clabby (approved):
Clarification of pre-Holocene behavior looks good. Thanks :shipit:

axelKingsley (approved):

PR #588, opened 2025-02-10
feat: specify getOperatorFee() return value is capped

leruaa:
**Description**

As per @tynes request, add the fact that the value returned by GasPriceOracle `getOperatorFee()` is capped in the specs.

**Tests**

**Additional context**

**Metadata**


tynes:
Is this also called out as part of the EL spec? I think it should be part of the definition of the formula

tynes (approved):

PR #593, opened 2025-02-13
fix: clarify EVM handling of request queue

meyer9:
**Description**

Clarify that post-processing transactions of three EIPs should be executed, but the results should be ignored. This makes the existing spec more general (consolidation and withdrawal requests are both ignored).

Also moved BLS to a separate section so it's easy to search for in the TOC.


refcell (approved):
Nice!

ajsutton (approved):

PR #596, opened 2025-02-14
fix link on specs

JosepBove:
Fixing a format issue that I found while doing a FMA

### DIFF
--- a/specs/protocol/derivation.md
+++ b/specs/protocol/derivation.md
@@ -1062,7 +1062,7 @@ As described in [L1 Retrieval](#l1-retrieval), batcher transactions' types are r
 However, we want to allow deposit transactions and `SystemConfig` update events to get derived even from receipts of
 future transaction types, as long as the receipts can be decoded following a best-effort process:
 
-As long as a future transaction type follows the [EIP-2718] specification, the
+[EIP-2718](https://eips.ethereum.org/EIPS/eip-2718)
 type can be decoded from the first byte of the transaction's (or its receipt's) binary encoding. We can then proceed as
 follows to get the logs of such a future transaction, or discard the transaction's receipt as invalid.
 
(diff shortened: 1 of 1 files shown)


specs/protocol/derivation.md
```diff
@@ -1062,7 +1062,7 @@ As described in [L1 Retrieval](#l1-retrieval), batcher transactions' types are r
 However, we want to allow deposit transactions and `SystemConfig` update events to get derived even from receipts of
 future transaction types, as long as the receipts can be decoded following a best-effort process:
 
-As long as a future transaction type follows the [EIP-2718][https://eips.ethereum.org/EIPS/eip-2718] specification, the
+As long as a future transaction type follows the [EIP-2718] specification, the
```
refcell:
```suggestion
[EIP-2718](https://eips.ethereum.org/EIPS/eip-2718)
```

JosepBove:
Fixed reviews

specs/protocol/derivation.md
```diff
@@ -1062,7 +1062,7 @@ As described in [L1 Retrieval](#l1-retrieval), batcher transactions' types are r
 However, we want to allow deposit transactions and `SystemConfig` update events to get derived even from receipts of
 future transaction types, as long as the receipts can be decoded following a best-effort process:
 
-As long as a future transaction type follows the [EIP-2718][https://eips.ethereum.org/EIPS/eip-2718] specification, the
+[EIP-2718](https://eips.ethereum.org/EIPS/eip-2718)
```
ajsutton: This seems to have lost the surrounding text.

ajsutton: Ah, I see the previous suggestion, but I don't think the current wording makes sense. That said, I don't think we actually implemented this generic parsing so possibly need to remove this from the spec again.  @sebastianst ?

sebastianst: Yes the surrounding text needs to be added back @JosepBove.

It's correct that we didn't implement the generic handling yet in either of the Go or Rust stack. But OTOH as long as the next L1 upgrade after Pectra isn't imminent, our implementations are technically still following the spec, since no future tx types >= 5 will be seen any time soon. I think we still want to add the generic handling eventually. So I'm fine with either, removing it from the spec or keeping as is and then just have to adapt the implementations in a few months.

ajsutton (approved):

PR #597, opened 2025-02-18
feat: update L2 genesis specs

agusduha:
**Description**

Update L2 genesis specs to match the implementation being developed

### DIFF
--- a/specs/experimental/standard-l2-genesis.md
+++ b/specs/experimental/standard-l2-genesis.md
@@ -12,19 +12,23 @@
   - [Initialization](#initialization)
   - [Interface](#interface)
     - [Fee Vault Config](#fee-vault-config)
-      - [`setBaseFeeVaultConfig`](#setbasefeevaultconfig)
-      - [`setL1FeeVaultConfig`](#setl1feevaultconfig)
-      - [`setSequencerFeeVaultConfig`](#setsequencerfeevaultconfig)
+      - [`setFeeVaultConfig`](#setfeevaultconfig)
+    - [Fee Admin](#fee-admin)
+      - [`feeAdmin`](#feeadmin)
+      - [`setFeeVaultAdmin`](#setfeevaultadmin)
+  - [Invariants](#invariants)
 - [`OptimismPortal`](#optimismportal)
   - [Interface](#interface-1)
     - [`setConfig`](#setconfig)
     - [`upgrade`](#upgrade)
+  - [Invariants](#invariants-1)
 - [SuperchainConfig](#superchainconfig)
   - [Constants](#constants-1)
   - [Interface](#interface-2)
   - [Initialization](#initialization-1)
 - [Predeploys](#predeploys)
-- [Constants](#constants-2)
+  - [L1Block values](#l1block-values)
+  - [L1Block slots](#l1block-slots)
 - [Predeploys](#predeploys-1)
   - [ProxyAdmin](#proxyadmin)
     - [Rationale](#rationale)
@@ -38,12 +42,10 @@
     - [Interface](#interface-4)
       - [`config`](#config)
   - [L2CrossDomainMessenger](#l2crossdomainmessenger)
-    - [Interface](#interface-5)
   - [L2ERC721Bridge](#l2erc721bridge)
-    - [Interface](#interface-6)
   - [L2StandardBridge](#l2standardbridge)
-    - [Interface](#interface-7)
   - [OptimismMintableERC721Factory](#optimismmintableerc721factory)
+  - [OptimismMintableERC20Factory](#optimismmintableerc20factory)
 - [Security Considerations](#security-considerations)
   - [GovernanceToken](#governancetoken)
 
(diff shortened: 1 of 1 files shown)


specs/experimental/standard-l2-genesis.md
```diff
@@ -111,28 +115,42 @@ These actions MAY only be triggered if there is a diff to the value.
...
+This function should be `public` and only callable by the fee admin.
 
 Each function calls `OptimismPortal.setConfig(ConfigType,bytes)` with its corresponding `ConfigType`.
 
-##### `setBaseFeeVaultConfig`
+##### `setFeeVaultConfig`
 
 ```solidity
-function setBaseFeeVaultConfig(address,uint256,WithdrawalNetwork)
+function setFeeVaultConfig(ConfigType,address,uint256,WithdrawalNetwork)
```
tynes, line 118: This MUST revert if the config type doesn't correspond to a fee vault config type

agusduha: Good catch! We are reverting in the implementation but it should be here too

specs/experimental/standard-l2-genesis.md
```diff
@@ -214,22 +240,27 @@ specific configuration out of the initial L2 genesis state. All network specific
...
-| `L1_FEE_VAULT_CONFIG` | `bytes32(uint256(keccak256("opstack.l1feevaultconfig")) - 1)` | The Fee Vault Config for the `L1FeeVault` |
-| `SEQUENCER_FEE_VAULT_CONFIG` | `bytes32(uint256(keccak256("opstack.sequencerfeevaultconfig")) - 1)` | The Fee Vault Config for the `SequencerFeeVault` |
+### L1Block slots
+
+| Name                                | Value                                                                      | Definition                                         |
+| ----------------------------------- | -------------------------------------------------------------------------- | -------------------------------------------------- |
+| `BASE_FEE_VAULT_CONFIG`             | `bytes32(uint256(keccak256("opstack.basefeevaultconfig")) - 1)`            | The Fee Vault Config for the `BaseFeeVault`        |
+| `L1_FEE_VAULT_CONFIG`               | `bytes32(uint256(keccak256("opstack.l1feevaultconfig")) - 1)`              | The Fee Vault Config for the `L1FeeVault`          |
+| `SEQUENCER_FEE_VAULT_CONFIG`        | `bytes32(uint256(keccak256("opstack.sequencerfeevaultconfig")) - 1)`       | The Fee Vault Config for the `SequencerFeeVault`   |
```
tynes, line 247: When we land the operator fee in https://github.com/ethereum-optimism/optimism/pull/12166, we will need to rebase on top of that and then add that to the specs, both here and in the mermaid diagram

specs/experimental/standard-l2-genesis.md
```diff
@@ -290,9 +321,9 @@ via a deposit transaction from the `DEPOSITOR_ACCOUNT`.
 
 ##### `setIsthmus`
```
tynes: This is fine to leave for now but for clarity, we will need to modify this, unclear which hardfork this will land in exactly right now

specs/experimental/standard-l2-genesis.md
```diff
@@ -60,30 +62,31 @@ configurability.
...
+| `BATCHER`             | `uint8(0)` | `abi.encode(address)`                                                             | Modifies the account that is authorized to progress the safe chain   |
+| `FEE_SCALARS`         | `uint8(1)` | `(uint256(0x01) << 248) \| (uint256(_blobbasefeeScalar) << 32) \| _basefeeScalar` | Modifies the fee scalars                                             |
+| `GAS_LIMIT`           | `uint8(2)` | `abi.encode(uint64 _gasLimit)`                                                    | Modifies the L2 gas limit                                            |
+| `UNSAFE_BLOCK_SIGNER` | `uint8(3)` | `abi.encode(address)`                                                             | Modifies the account that is authorized to progress the unsafe chain |
+| `EIP_1559_PARAMS`     | `uint8(4)` | `uint256(uint64(uint32(_denominator))) << 32 \| uint64(uint32(_elasticity))`      | Modifies the EIP-1559 denominator and elasticity                     |
+| `FEE_VAULT_ADMIN`     | `uint8(5)` | `abi.encode(address)`                                                             | Modifies the fee vault admin                                         |
```
tynes: This doesn't seem right, why is fee vault admin added here? These are generally values that go into L1 attributes tx and we dont want to put the fee vault admin in there. The `ConfigUpdate` events influence L2 consensus

specs/experimental/standard-l2-genesis.md
```diff
@@ -160,15 +178,15 @@ The following fields are included:
...
+  is `200_000`, `isCreation` is `false` and the `data` is `abi.encodeCall(L1Block.setConfig, (_type, _value))`
 
 #### `upgrade`
 
 The `upgrade` function MUST only be callable by the `UPGRADER` role as defined
 in the [`SuperchainConfig`](#superchainconfig).
 
 ```solidity
-function upgrade(bytes memory _data) external
+function upgrade(uint32 _gasLimit, bytes memory _data) external
```
maurelian: How will the _gasLimit value be set? Will it be hardcoded in the SystemConfig?

maurelian: Also, is `upgrade()` the right name for this? If "`to` is `Predeploys.L1Block`", them I think something like `updateConfig()` might be more appropriate.

specs/experimental/standard-l2-genesis.md
```diff
@@ -180,10 +198,18 @@ event TransactionDeposited(address indexed from, address indexed to, uint256 ind
 The following fields are included:
 
 - `from` is the `DEPOSITOR_ACCOUNT`
-- `to` is `Predeploys.ProxyAdmin`
+- `to` is `Predeploys.L2_PROXY_ADMIN`
 - `version` is `uint256(0)`
 - `opaqueData` is the tightly packed transaction data where `mint` is `0`, `value` is `0`, the `gasLimit`
-   is `200_000`, `isCreation` is `false` and the `data` is the data passed into `upgrade`.
+  is `200_000`, `isCreation` is `false` and the `data` is the data passed into `upgrade`.
```
maurelian: the gasLimit presumably needs to be updated if it becomes an input.

I'm also curious what the current thinking is around the correct behaviour of `useGas()`:

[CleanShot 2025-02-19 at 15 56 14@2x](https://github.com/user-attachments/assets/619d6f39-764f-435a-8f63-eb714fe3621e)

Note that I don't agree with the recommended fix, as it will enables anyone to DoS a system deposit.

specs/experimental/standard-l2-genesis.md
0xOneTony: Fixed here https://github.com/ethereum-optimism/specs/pull/597/commits/6cf6decd5a057f4361fdcd8e34c1b3f425d2ecea

specs/experimental/standard-l2-genesis.md
0xOneTony: We believe that we shouldnt add the check in `useGas` since this can cause further DoS issues by front-running the chain-operator. To compensate for this we agree with adding a test to make sure we dont exceed the buffer gas limit like mentioned here https://github.com/ethereum-optimism/design-docs/pull/97#discussion_r1811253276

emhane (changes requested):
is it possible to scope this into smaller prs, one per section. would make reviewing a lot easier and allow making incremental progress by merging the loc without remarks now already.

PR #599, opened 2025-02-19
chore: polish message passing docs

will-2012:
The previous description seems to mean the opposite.


maurelian (approved):
good edit, thanks.

PR #603, opened 2025-02-24
feat: proveWithdrawalTransaction for Super Roots

smartcontracts:
Updates the OptimismPortal to specify the behavior of the proveWithdrawalTransaction when proving against Super Roots.

### DIFF
--- a/specs/fault-proof/stage-one/optimism-portal.md
+++ b/specs/fault-proof/stage-one/optimism-portal.md
@@ -12,10 +12,10 @@
   - [Valid Withdrawal](#valid-withdrawal)
   - [Invalid Withdrawal](#invalid-withdrawal)
   - [L2 Withdrawal Sender](#l2-withdrawal-sender)
+  - [Block Output](#block-output)
   - [Output Root](#output-root)
-  - [Output Root Hash](#output-root-hash)
+  - [Super Output](#super-output)
   - [Super Root](#super-root)
-  - [Super Root Hash](#super-root-hash)
 - [Assumptions](#assumptions)
   - [aOP-001: Dispute Game contracts properly report important properties](#aop-001-dispute-game-contracts-properly-report-important-properties)
     - [Mitigations](#mitigations)
@@ -111,13 +111,17 @@ The **L2 Withdrawal Sender** is the address of the account that triggered a give
 transaction on L2. The `OptimismPortal` is expected to expose a variable that includes this value
 when [finalizing](#finalized-withdrawal) a withdrawal.
 
-### Output Root
+### Block Output
 
-An **Output Root** is a data structure that wraps the key hash elements of a given L2 block. An
-Output Root has the following structure:
+A **Block Output**, commonly called an **Output**, is a data structure that wraps the key hash
+elements of a given L2 block.
+
+The structure of the Block Output is versioned (32 bytes). The current Block Output version is
+`0x0000000000000000000000000000000000000000000000000000000000000000` (V0). A V0 Block Output has
+the following structure:
 
 ```solidity
-struct OutputRoot {
+struct BlockOutput {
   bytes32 version;
   bytes32 stateRoot;
   bytes32 messagePasserStorageRoot;
@@ -133,45 +137,40 @@ Where:
   block this Output Root corresponds to
 - `blockHash` is the block hash of the L2 block this Output Root corresponds to
 
-### Output Root Hash
+### Output Root
 
-An **Output Root Hash** is the hash of an [Output Root](#output-root), computed as:
+An **Output Root** is a commitment to a [Block Output](#block-output). A detailed description of
+this commitment can be found [here](../../protocol/proposals.md#l2-output-commitment-construction).
 
-```solidity
-keccak256(abi.encode(OutputRoot))
-```
+### Super Output
 
-Output Root Hashes are used as the Root Claims for the `FaultDisputeGame` and
-`PermissionedDisputeGame` contracts.
+A **Super Output** is a data structure that commits all of the [Block Outputs](#block-output) for
+all chains within the Superchain Interop Set at a given timestamp.
 
-### Super Root
-
-A **Super Root** is a data structure that commits all of the [Output Roots](#output-root) for all
-chains within the Superchain Interop Set at a given timestamp. It has the following structure:
+The structure of the Super Output is versioned (1 byte). The current version is `0x01` (V1). A V1
+Super Output has the following structure:
 
 ```solidity
-struct OutputRootHashWithChainId {
+struct OutputRootWithChainId {
   uint256 chainId;
-  bytes32 outputRootHash;
+  bytes32 outputRoot;
 }
 
-struct SuperRoot {
+struct SuperOutput {
   uint64 timestamp;
   OutputRootHashWithChainid[] outputRoots;
 }
 ```
 
-The Super Root is versioned and currently has a version byte of `0x01`.
+### Super Root
 
-### Super Root Hash
-
-A **Super Root Hash** is the hash of a [Super Root](#super-root), computed as:
+A **Super Root** is a commitment to a [Super Output](#super-output), computed as:
 
 ```solidity
 keccak256(encodeSuperRoot(SuperRoot))
 ```
 
-Where `encodeSuperRoot` is:
+Where `encodeSuperRoot` for the V1 Super Output is:
 
 ```solidity
 function encodeSuperRoot(SuperRoot memory root) returns (bytes) {
@@ -349,14 +348,14 @@ that uses dispute games that argue over [Super Roots](#super-root).
 - MUST revert if the withdrawal is being proven against a game that has resolved in favor of the
   Challenger.
 - MUST revert if `superRootsActive` is `false`.
-- MUST revert if the proof provided by the user of the contents of the Super Root that the dispute
+- MUST revert if the proof provided by the user of the preimage of the Super Root that the dispute
   game argues about is invalid. This proof is verified by hashing the user-provided Super Root
-  contents and comparing them to the Super Root in the referenced dispute game.
+  preimage and comparing them to the Super Root in the referenced dispute game.
 - MUST revert if the pointer index of the Output Root inside of the Super Root provided by the user
   is beyond the size of the Output Roots array or points to a chain ID other than the chain ID
   stored within the `SystemConfig` contract for the `OptimismPortal`.
-- MUST revert if the proof provided by the user of the contents of the Output Root is invalid. This
-  proof is verified by hashing the user-provided contents and comparing them to the Output Root.
+- MUST revert if the proof provided by the user of the preimage of the Output Root is invalid. This
+  proof is verified by hashing the user-provided preimage and comparing them to the Output Root.
 - MUST revert if the provided merkle trie proof that the withdrawal was included within the root
   claim of the provided dispute game is invalid.
 - MUST otherwise store a record of the withdrawal proof that includes the hash of the proven  
@@ -376,8 +375,8 @@ that uses dispute games that argue over [Output Roots](#output-root).
 - MUST revert if the withdrawal is being proven against a game that has resolved in favor of the
   Challenger.
 - MUST revert if `superRootsActive` is `true`.
-- MUST revert if the proof provided by the user of the contents of the Output Root that the dispute
-  game argues about is invalid. This proof is verified by hashing the user-provided contents and
+- MUST revert if the proof provided by the user of the preimage of the Output Root that the dispute
+  game argues about is invalid. This proof is verified by hashing the user-provided preimage and
   comparing them to the root claim of the referenced dispute game.
 - MUST revert if the provided merkle trie proof that the withdrawal was included within the root
   claim of the provided dispute game is invalid.
--- a/specs/fault-proof/stage-one/optimism-portal.md
+++ b/specs/fault-proof/stage-one/optimism-portal.md
@@ -145,7 +145,8 @@ this commitment can be found [here](../../protocol/proposals.md#l2-output-commit
 ### Super Output
 
 A **Super Output** is a data structure that commits all of the [Block Outputs](#block-output) for
-all chains within the Superchain Interop Set at a given timestamp.
+all chains within the Superchain Interop Set at a given timestamp. A Super Output can also commit
+to a single Block Output to maintain compatibility with chains outside of the Interop Set.
 
 The structure of the Super Output is versioned (1 byte). The current version is `0x01` (V1). A V1
 Super Output has the following structure:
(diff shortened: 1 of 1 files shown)


specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -105,6 +111,64 @@ The **L2 Withdrawal Sender** is the address of the account that triggered a give
...
+  OutputRootHashWithChainid[] outputRoots;
+}
+```
+
+### Super Root Hash
+
+A **Super Root Hash** is the hash of a [Super Root](#super-root), computed as:
+
+```solidity
+keccak256(abi.encode(SuperRoot))
```
clabby: This is incorrect. The super root is not an ABI encoded structure.

Reference encoding + hashing implementation: https://github.com/op-rs/kona/blob/57724bc382267371205f861f1807c138a5f834f4/crates/protocol/interop/src/root.rs#L63-L84

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -105,6 +111,64 @@ The **L2 Withdrawal Sender** is the address of the account that triggered a give
...
+An **Output Root Hash** is the hash of an [Output Root](#output-root), computed as:
+
+```solidity
+keccak256(abi.encode(OutputRoot))
+```
+
+Output Root Hashes are used as the Root Claims for the `FaultDisputeGame` and
+`PermissionedDisputeGame` contracts.
+
+### Super Root
```
clabby: We should note the [current version byte](https://github.com/op-rs/kona/blob/main/crates/protocol/interop/src/constants.rs#L12-L15)

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -262,6 +359,10 @@ Allows a user to [prove](#proven-withdrawal) a withdrawal transaction.
   [Respected Game](./anchor-state-registry.md#respected-game).
 - MUST revert if the withdrawal is being proven against a game that has resolved in favor of the
   Challenger.
+- MUST revert if `superRootsActive` is `true`.
+- MUST revert if the proof provided by the user of the contents of the Output Root that the dispute
+  game argues about is invalid. This proof is verified by hashing the user-provided contents and
+  comparing them to the root claim of the referenced dispute game.
```
clabby: Nice add - surprised this was overlooked before.

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -251,9 +321,36 @@ has not been initialized then this value will be `address(0)` and should not be
...
+- MUST revert if the withdrawal is being proven against a game that is not a
+  [Respected Game](./anchor-state-registry.md#respected-game).
+- MUST revert if the withdrawal is being proven against a game that has resolved in favor of the
+  Challenger.
+- MUST revert if `superRootsActive` is `false`.
+- MUST revert if the proof provided by the user of the contents of the Super Root that the dispute
+  game argues about is invalid. This proof is verified by hashing the user-provided Super Root
+  contents and comparing them to the Super Root in the referenced dispute game. The user provides
+  a specific Output Root index to extract and the chain ID of this Output Root must match the chain
+  ID within the `SystemConfig` contract.
```
clabby: Can we split up the first / second part of this invariant? They are checked separately, would be a bit easier to read.

clabby (changes requested):

specs/fault-proof/stage-one/optimism-portal.md
mds1: Is the output root an ABI encoded structure? Referring to a few lines up in this PR: https://github.com/ethereum-optimism/specs/pull/603/files?diff=unified&w=0#diff-473596af2fca6e2289f873e13448687868604a5a7230afb3001c6f0b6603b918R138-R142

smartcontracts: Can we simply the spec to make this an abi encoded thing? Way easier to represent onchain

smartcontracts: @mds1 yes, the abi struct encoding for the output root is equivalent to the way it's currently encoded in the contracts

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -105,6 +111,79 @@ The **L2 Withdrawal Sender** is the address of the account that triggered a give
...
+keccak256(abi.encode(OutputRoot))
+```
+
+Output Root Hashes are used as the Root Claims for the `FaultDisputeGame` and
+`PermissionedDisputeGame` contracts.
+
+### Super Root
+
+A **Super Root** is a data structure that commits all of the [Output Roots](#output-root) for all
+chains within the Superchain Interop Set at a given timestamp. It has the following structure:
```
Inphi: there will be multiple interop clusters, including single-chain clusters. So the spec shouldn't restrict the interop set to just the superchain's. I suggest rewording this to something like:
```suggestion
chains within an Interop cluster at a given timestamp. It has the following structure:
```

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -105,6 +111,79 @@ The **L2 Withdrawal Sender** is the address of the account that triggered a give
...
+}
+```
+
+Where:
+
+- `version` is a version identifier that describes the structure of the Output Root
+- `stateRoot` is the state root of the L2 block this Output Root corresponds to
+- `messagePasserStorageRoot` is the storage root of the `L2ToL1MessagePasser` contract at the L2
+  block this Output Root corresponds to
+- `blockHash` is the block hash of the L2 block this Output Root corresponds to
```
ajsutton, line 138: Strictly speaking, this is the structure for version 0 of output roots.  Other versions may have completely different structures - the only guarantee is that the first 32 bytes will be the version. Everything after that is version dependent.

Output roots are already defined in https://specs.optimism.io/protocol/proposals.html#l2-output-commitment-construction although potentially that could be replaced by this definition given the page is mostly about the L2OO. We just shouldn't define output roots in two different places and the existing definition is probably the better definition.

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -105,6 +111,79 @@ The **L2 Withdrawal Sender** is the address of the account that triggered a give
...
+
+### Output Root Hash
```
ajsutton: This is actually an output root.

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -105,6 +111,79 @@ The **L2 Withdrawal Sender** is the address of the account that triggered a give
 transaction on L2. The `OptimismPortal` is expected to expose a variable that includes this value
 when [finalizing](#finalized-withdrawal) a withdrawal.
 
+### Output Root
+
+An **Output Root** is a data structure that wraps the key hash elements of a given L2 block. An
+Output Root has the following structure:
```
ajsutton: This is actually an output.  The output root is the hash of the output.

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -105,6 +111,79 @@ The **L2 Withdrawal Sender** is the address of the account that triggered a give
...
+  block this Output Root corresponds to
+- `blockHash` is the block hash of the L2 block this Output Root corresponds to
+
+### Output Root Hash
+
+An **Output Root Hash** is the hash of an [Output Root](#output-root), computed as:
+
+```solidity
+keccak256(abi.encode(OutputRoot))
+```
```
ajsutton: I _think_ this is equivalent, but output roots are not actually abi encoded - they're concatenated so that they aren't language specific. Given the concatenation is simple to express, it seems better to avoid adding the requirement to understand the entirety of solidity abi encoding to work out how this is actually structured.

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -105,6 +111,79 @@ The **L2 Withdrawal Sender** is the address of the account that triggered a give
...
+
+Output Root Hashes are used as the Root Claims for the `FaultDisputeGame` and
+`PermissionedDisputeGame` contracts.
+
+### Super Root
```
ajsutton: The super root, is the 32 byte hash.  Strictly speaking the preimage of that is the `Super` which does read a bit weird but is consistent with the naming of output/output root.

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -105,6 +111,79 @@ The **L2 Withdrawal Sender** is the address of the account that triggered a give
...
+  bytes32 outputRootHash;
+}
+
+struct SuperRoot {
+  uint64 timestamp;
+  OutputRootHashWithChainid[] outputRoots;
+}
+```
+
+The Super Root is versioned and currently has a version byte of `0x01`.
```
ajsutton: This structure only applies to v1 super roots.  Like outputs, the one guarantee is that the first byte (vs 32 bytes) is the version.

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -251,9 +336,37 @@ has not been initialized then this value will be `address(0)` and should not be
...
+- MUST revert if the withdrawal is being proven against a game that is not a
+  [Proper Game](./anchor-state-registry.md#proper-game).
+- MUST revert if the withdrawal is being proven against a game that is not a
+  [Respected Game](./anchor-state-registry.md#respected-game).
+- MUST revert if the withdrawal is being proven against a game that has resolved in favor of the
+  Challenger.
+- MUST revert if `superRootsActive` is `false`.
+- MUST revert if the proof provided by the user of the contents of the Super Root that the dispute
+  game argues about is invalid. This proof is verified by hashing the user-provided Super Root
+  contents and comparing them to the Super Root in the referenced dispute game.
```
ajsutton: We could probably clear up some confusion about the Super vs Super Root naming here by saying the _preimage_ of the Super Root rather than the contents of it.  The Super Root is a 32 byte hash and its the preimage the user is providing.

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -251,9 +336,37 @@ has not been initialized then this value will be `address(0)` and should not be
...
+- MUST revert if the pointer index of the Output Root inside of the Super Root provided by the user
+  is beyond the size of the Output Roots array or points to a chain ID other than the chain ID
+  stored within the `SystemConfig` contract for the `OptimismPortal`.
+- MUST revert if the proof provided by the user of the contents of the Output Root is invalid. This
+  proof is verified by hashing the user-provided contents and comparing them to the Output Root.
```
ajsutton: Using preimage here too would probably be better than contents.

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -262,6 +375,10 @@ Allows a user to [prove](#proven-withdrawal) a withdrawal transaction.
   [Respected Game](./anchor-state-registry.md#respected-game).
 - MUST revert if the withdrawal is being proven against a game that has resolved in favor of the
   Challenger.
+- MUST revert if `superRootsActive` is `true`.
+- MUST revert if the proof provided by the user of the contents of the Output Root that the dispute
+  game argues about is invalid. This proof is verified by hashing the user-provided contents and
+  comparing them to the root claim of the referenced dispute game.
```
ajsutton: As above, preimage of the Output Root is probably clearer than contents since the Output Root is a 32 byte hash.

ajsutton (approved):
Approving because the key content is right. There's just some confusion around the fact that an Output Root is actually the 32 byte hash and would be good to reference the existing output root definition or at least bring that text over here since it's a more accurate description.

specs/fault-proof/stage-one/optimism-portal.md
tynes: Perhaps better to link to a single location that contains a clear definition

specs/fault-proof/stage-one/optimism-portal.md
tynes: Better to do this with a `||` or `++` to denote concatenation, could use pythonish `Bytes32(...)` types

specs/fault-proof/stage-one/optimism-portal.md
smartcontracts: Changed to use common link

specs/fault-proof/stage-one/optimism-portal.md
smartcontracts: Renamed to "Block Output" aka "Output" for short.

specs/fault-proof/stage-one/optimism-portal.md
smartcontracts: Renamed to "Output Root"

specs/fault-proof/stage-one/optimism-portal.md
smartcontracts: Changed to just use the common link instead

specs/fault-proof/stage-one/optimism-portal.md
smartcontracts: Renamed to "Super Output" for consistency

specs/fault-proof/stage-one/optimism-portal.md
smartcontracts: Clarified that this structure only applies to V1 Super Outputs

specs/fault-proof/stage-one/optimism-portal.md
smartcontracts: Right now we only intend to have a single interop set of more than one chain so I just went ahead and added some context that clarifies that the Super Output contains either >1 chain if it's for the Interop Set or 1 chain otherwise.

If things change in the future we can update the spec accordingly.

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -105,6 +111,79 @@ The **L2 Withdrawal Sender** is the address of the account that triggered a give
...
+
+```solidity
+struct OutputRootWithChainId {
+  uint256 chainId;
+  bytes32 outputRoot;
+}
+
+struct SuperOutput {
+  uint64 timestamp;
+  OutputRootHashWithChainid[] outputRoots;
```
smartcontracts:
```suggestion
  OutputRootWithChainid[] outputRoots;
```

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -105,6 +111,79 @@ The **L2 Withdrawal Sender** is the address of the account that triggered a give
...
+
+Where `encodeSuperRoot` for the V1 Super Output is:
+
+```solidity
+function encodeSuperRoot(SuperRoot memory root) returns (bytes) {
+  require(root.outputRoots.length > 0); // Super Root must have at least one Output Root.
+  return concat(
+    0x01, // Super Root version byte
+    root.timestamp,
+    [outputRoot.chainId || outputRoot.outputRootHash for outputRoot in root.outputRoots]
```
smartcontracts: Use concat

mds1 (approved):

clabby (approved):

PR #612, opened 2025-03-10
interop: access-list spec and RPC

protolambda:
**Description**

Specifies how access-list entries are used, formatted and checked.

This also includes a new supervisor RPC endpoint, to check the access-list entries.

And also includes an optional timeout attribute, so we can ensure that a transaction with an old message will stick around for a while, and not suddenly disappear due to message-expiry.

### DIFF
--- a/specs/interop/predeploys.md
+++ b/specs/interop/predeploys.md
@@ -6,6 +6,10 @@
 
 - [Overview](#overview)
 - [CrossL2Inbox](#crossl2inbox)
+  - [Access-list](#access-list)
+    - [type 1: Lookup identity](#type-1-lookup-identity)
+    - [type 2: ChainID extension](#type-2-chainid-extension)
+    - [type 3: Checksum](#type-3-checksum)
   - [Functions](#functions)
     - [validateMessage](#validatemessage)
   - [`ExecutingMessage` Event](#executingmessage-event)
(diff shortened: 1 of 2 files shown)


specs/interop/predeploys.md
```diff
@@ -131,8 +271,9 @@ A simple implementation of the `validateMessage` function is included below.
 
 ```solidity
     function validateMessage(Identifier calldata _id, bytes32 _msgHash) external {
-        // We need to know if this is being called on a depositTx
-        if (IL1BlockInterop(Predeploys.L1_BLOCK_ATTRIBUTES).isDeposit()) revert NoExecutingDeposits();
+        // TODO calldata load the inputs
+        // TODO compute checksum
+        // TODO check if checksum is known access-list entry
```
protolambda: My understanding is that we can get rid of the `isDeposit()` check now that we've access-lists.

I left the implementation as TODO. I wrote some solidity/yul snippets above in the spec that can largely be used here for implementation, but would like to agree on spec of those before we duplicate any work here in solidity.

specs/interop/predeploys.md
```diff
@@ -86,6 +90,142 @@ To ensure safety of the protocol, the [Message Invariants](./messaging.md#messag
...
+Execution of messages is statically pre-declared in transactions,
+to ensure the cross-chain validity can be verified outside the single-chain EVM environment constraints.
+
+After pre-verification of the access-list, the `CrossL2Inbox` can allow messages
+to execute when there is a matching pre-verified access-list entry.
+
+Each executing message is declared with 3 typed access-list entries:
+- 1: Lookup identity
+- 2: ChainID extension
+- 3: Checksum
```
protolambda: Since we last discussed the entry formatting, I realized it's good to have typing for each of them, and that the chain-ID can really just be an extension entry, and we don't need it by default if it already fits in the lookup identity entry.

specs/interop/predeploys.md
```diff
@@ -86,6 +90,142 @@ To ensure safety of the protocol, the [Message Invariants](./messaging.md#messag
...
+Not every access-list entry may be executed:
+access-list content must not be used by applications to interpret results of transactions,
+the `ExecutingMessage` event describes in detail what is executed.
+
+To prevent cross-contamination of access-list contents,
+the checksum entry commits to the contents of the other entries.
+The checksum will be invalid if the wrong entries are interpreted with it.
+
+The `CrossL2Inbox` only checks the checksum is present in the access-list:
+the presence of other needed entries is enforced during pre-validation.
```
protolambda, line 131: This is really key, and keeps execution relatively cost efficient, since we only pay for 1 cold `SLOAD` per message.

Besides, block-builders that fill access-lists for themselves can optimize this way, and not include unnecessary access-list data. The checksum per message is enough in that case.

specs/interop/supervisor.md
```diff
@@ -74,6 +75,8 @@ Specifically, this helps apply message-expiry rules on message checks.
 
 Object:
 - `timestamp`: `HexUint64`
+- `timeout`: `HexUint64` or `null` - optional, requests verification to still hold at `timestamp+timeout`
+  (message expiry may drop previously valid messages).
```
skeletor-spaceman: would love an explanation on this change

specs/interop/supervisor.md
```diff
@@ -243,3 +246,26 @@ Parameters:
...
 - key: `ChainID`
 - value: `BlockID`
+
+#### `supervisor_checkAccessList`
+
+Verifies if an access-list references only valid messages.
+Message execution in the [`CrossL2Inbox`] that is statically declared in the access-list will not revert.
+
+Only the [`CrossL2Inbox`] subset of the access-list in the transaction is required,
+storage-access by other addresses is not included.
```
skeletor-spaceman, line 232: beware that this is also considered a valid `access-list` in our tests:
```
[
  CROSS_L2_INBOX: [
    ...
  ],
  CROSS_L2_INBOX: [
    ...
  ]
]
```

protolambda: Good catch. Yes let's call that out in the spec. We need the joined set of all entries applied to the crossL2Inbox.

specs/interop/predeploys.md
tynes: Lets get rid of the `isDeposit` check as well as the deposit context. We don't need either of them at this point

protolambda: Updated to include a part of the wonderland code, with reference to the relevant spec sections for checksum computation and info about the is-warm check.

specs/interop/supervisor.md
protolambda: Added additional explanation

specs/interop/predeploys.md
```diff
@@ -86,6 +90,103 @@ To ensure safety of the protocol, the [Message Invariants](./messaging.md#messag
...
+- 2: Chain-ID extension
+- 3: Checksum
+
+The type of entry is encoded in the first byte.
+Type 0 is reserved, so valid access-list entries are always non-zero.
+
+Note that the access-list entries may be de-duplicated:
+the same message may be executed multiple times.
+
+The access-list content not always a multiple of 3.
```
```suggestion
The access-list content might not always be a multiple of 3.
```

specs/interop/predeploys.md
```diff
@@ -130,15 +231,23 @@ hash comparison.
...
-        // We need to know if this is being called on a depositTx
-        if (IL1BlockInterop(Predeploys.L1_BLOCK_ATTRIBUTES).isDeposit()) revert NoExecutingDeposits();
+function validateMessage(Identifier calldata _id, bytes32 _msgHash) external {
+  bytes32 checksum = calculateChecksum(_id, _msgHash);
 
-        emit ExecutingMessage(_msgHash, _id);
-    }
+  (bool _isSlotWarm,) = _isWarm(checksum);
+
+  if (!_isSlotWarm) revert NotWarm();
```
```suggestion
  if (!_isSlotWarm) revert NonDeclaredExecutingMessage();
```
I'd advise to use a more suitable error :)

specs/interop/predeploys.md
```diff
@@ -130,15 +231,23 @@ hash comparison.
...
+
+  emit ExecutingMessage(_msgHash, _id);
 }
 ```
 
+`calculateChecksum` implements the checksum computation (including type-byte) as defined
+in the [access-list checksum computation](#type-3-checksum) spec.
+
+`_isWarm` checks that the access-list prepared the `checksum` storage key to be warm,
+no other contract function may warm up this storage without message validation.
```
```suggestion
`_isWarm` checks that the access-list prepared the `checksum` storage key to be warm.
**No other contract function may warm up this storage without message validation.**
```
this is extremely critical. we should add the relevant in-line comments and nat-spec in `CrossL2Inbox.col`

skeletor-spaceman (changes requested):
minor changes required
looking "fire"

skeletor-spaceman (approved):

tynes:
This looks good to me, should wait on merge for design review

clabby (approved):

tynes (approved):

PR #619, opened 2025-03-12
Isthmus: Fix operator fee codehash

sebastianst:
The old verification recipe was wrong. This operator fee vault code hash is already correct in the Go implementation, see https://github.com/ethereum-optimism/optimism/blob/develop/op-e2e/actions/upgrades/isthmus_fork_test.go#L35


refcell (approved):
Very nice

protolambda (approved):

Issue #541, opened 2025-01-23
feat(interop-proof): Specify dispute game modifications


clabby:
## Overview

The dispute game is being modified as a part of the interop proof to allow for uploading a new local preimage kind, renaming some variables, and removing the L2 block number challenge. These changes need to be specified in order to inform the updates to the `op-challenger`.

ajsutton:
https://specs.optimism.io/interop/fault-proof.html describes the difference to the pre-interop game.

PR #620, opened 2025-03-13
interop: Describe super fault dispute game differences to fault dispute game

ajsutton:
**Description**

Add an initial definition of the super fault dispute game. Mostly works by referencing the existing fault dispute game and explaining the differences, plus going into detail on how the state transition function works for the top half of the super game.

**Metadata**

https://github.com/ethereum-optimism/specs/issues/541


specs/interop/fault-proof.md
```diff
@@ -4,24 +4,110 @@
...
+there is no possible valid proposal. The dispute game contract MUST prevent a game from being created where the root
+claim is the invalid state.
+
+When the prestate is the invalid state, the post state is also the invalid state.
+
+### Local Safe Block Derivation
+
+The first 127 steps derive the next local safe block for one chain in the super root using
+the [derivation process](../protocol/derivation.md). Chains are processed in the order they appear in the super root
+(ascending order of chain ID). The `Step` is incremented after each step.
```
Inphi, line 76: > ascending order of chain

This is somewhat new to me. Is this basically what the "chain index" will be?
I think it'll be less ambiguous to refer to the ordering of chains in the super output instead. Though I'd imagine in practice both orders will be the same.

Inphi (approved):
Excellent. It's easy to read with the diff against the pre-interop system.

specs/interop/fault-proof.md
ajsutton: Yeah the chain index is the index of the chain in the super output (and dependency set - the ordering rules are the same). So the chain with the lowest chain ID is chain index 0.

PR #625, opened 2025-03-14
feat: specification for Stage 1 changes

smartcontracts:
**Description**
Specification for the changes described by https://github.com/ethereum-optimism/design-docs/pull/202

### DIFF
--- a/specs/fault-proof/stage-one/anchor-state-registry.md
+++ b/specs/fault-proof/stage-one/anchor-state-registry.md
@@ -152,7 +152,7 @@ brevity. A Dispute Game can go from being a Proper Game to later *not* being an 
 if it is invalidated by being [blacklisted](#blacklisted-game) or [retired](#retired-game).
 
 **ALL** Dispute Games **TEMPORARILY** become Improper Games while the
-[Pause Mechanism](../../protocol/superchain-config.md#pause-mechanism) is active. However, this is
+[Pause Mechanism](../../protocol/stage-1.md#pause-mechanism) is active. However, this is
 a *temporary* condition such that Registered Games that are not invalidated by
 [blacklisting](#blacklisted-game) or [retirement](#retired-game) will become Proper Games again
 once the pause is lifted. The Pause Mechanism is therefore a way to *temporarily* prevent Dispute
@@ -164,7 +164,7 @@ A Game is considered to be a Proper Game if all of the following are true:
 - The game is a [Registered Game](#registered-game)
 - The game is **NOT** a [Blacklisted Game](#blacklisted-game)
 - The game is **NOT** a [Retired Game](#retired-game)
-- The [Pause Mechanism](../../protocol/superchain-config.md#pause-mechanism) is not active
+- The [Pause Mechanism](../../protocol/stage-1.md#pause-mechanism) is not active
 
 ### Resolved Game
 
--- a/specs/fault-proof/stage-one/anchor-state-registry.md
+++ b/specs/fault-proof/stage-one/anchor-state-registry.md
@@ -42,6 +42,8 @@
   - [iASR-004: Invalidation functions operate correctly](#iasr-004-invalidation-functions-operate-correctly)
     - [Impact](#impact-3)
     - [Dependencies](#dependencies-3)
+  - [iASR-005: The Anchor Game is recent enough to be fault provable](#iasr-005-the-anchor-game-is-recent-enough-to-be-fault-provable)
+    - [Impact](#impact-4)
 - [Function Specification](#function-specification)
   - [constructor](#constructor)
   - [initialize](#initialize)
@@ -372,6 +374,20 @@ finalize withdrawals with invalidated games would be considered Critical Severit
 #### Dependencies
 
 - [aASR-003](#aasr-003-incorrectly-resolving-games-will-be-invalidated-before-they-have-valid-claims)
+
+### iASR-005: The Anchor Game is recent enough to be fault provable
+
+We require that the Anchor Game corresponds to an L2 block with an L1 origin timestamp that is no
+older than 6 months from the current timestamp. This time constraint is necessary because the fault
+proof VM must walk backwards through L1 blocks to verify derivation, and processing 7 months worth
+of L1 blocks approaches the maximum time available to challengers in the dispute game process.
+
+#### Impact
+
+**Severity: High**
+
+If this invariant is broken, challengers will be unable to participate in fault proofs within the
+allotted response time, and resolution would require intervention from the Proxy Admin Owner.
 
 ## Function Specification
 
--- a/specs/fault-proof/stage-one/anchor-state-registry.md
+++ b/specs/fault-proof/stage-one/anchor-state-registry.md
@@ -149,19 +149,20 @@ A Dispute Game that is **NOT** a Proper Game can also be referred to as an **Imp
 brevity. A Dispute Game can go from being a Proper Game to later *not* being an **Improper Game**
 if it is invalidated by being [blacklisted](#blacklisted-game) or [retired](#retired-game).
 
-**ALL** Dispute Games **TEMPORARILY** become Improper Games while the Superchain-wide pause
-mechanism is active. However, this is a *temporary* condition such that Registered Games that are
-not invalidated by [blacklisting](#blacklisted-game) or [retirement](#retired-game) will become
-Proper Games again once the pause is lifted. The Superchain-wide pause is therefore a way to
-*temporarily* prevent Dispute Games from being used by consumers like the `OptimismPortal` while
-relevant parties coordinate the use of some other invalidation mechanism.
+**ALL** Dispute Games **TEMPORARILY** become Improper Games while the
+[Pause Mechanism](../../protocol/superchain-config.md#pause-mechanism) is active. However, this is
+a *temporary* condition such that Registered Games that are not invalidated by
+[blacklisting](#blacklisted-game) or [retirement](#retired-game) will become Proper Games again
+once the pause is lifted. The Pause Mechanism is therefore a way to *temporarily* prevent Dispute
+Games from being used by consumers like the `OptimismPortal` while relevant parties coordinate the
+use of some other invalidation mechanism.
 
 A Game is considered to be a Proper Game if all of the following are true:
 
 - The game is a [Registered Game](#registered-game)
 - The game is **NOT** a [Blacklisted Game](#blacklisted-game)
 - The game is **NOT** a [Retired Game](#retired-game)
-- The Superchain-wide pause is not active
+- The [Pause Mechanism](../../protocol/superchain-config.md#pause-mechanism) is not active
 
 ### Resolved Game
 
@@ -261,10 +262,10 @@ have [Valid Claims](#valid-claim).
 
 Proper Games that resolve in favor the Defender will be considered to have Valid Claims after the
 [Dispute Game Finality Delay](#dispute-game-finality-delay-airgap) has elapsed UNLESS the
-Superchain-wide pause mechanism is active. Therefore, in the absence of the Superchain-wide pause
-mechanism, parties responsible for game invalidation have exactly the Dispute Game Finality Delay
-to invalidate a withdrawal after it resolves incorrectly. If the Superchain-wide pause is active,
-then any incorrectly resolving games must be invalidated before the pause is deactivated.
+Pause Mechanism is active. Therefore, in the absence of the Pause Mechanism, parties responsible
+for game invalidation have exactly the Dispute Game Finality Delay to invalidate a withdrawal after
+it resolves incorrectly. If the Pause Mechanism is active, then any incorrectly resolving games
+must be invalidated before the pause is deactivated.
 
 #### Mitigations
 
@@ -381,7 +382,7 @@ finalize withdrawals with invalidated games would be considered Critical Severit
 ### initialize
 
 - MUST only be triggerable once.
-- MUST set the value of the `SuperchainConfig` contract that stores the address of the Guardian.
+- MUST set the value of the `SystemConfig` contract that stores the address of the Guardian.
 - MUST set the value of the `DisputeGameFactory` contract that creates Dispute Game instances.
 - MUST set the value of the [Starting Anchor State](#starting-anchor-state).
 - MUST set the value of the initial [Respected Game Type](#respected-game-type).
@@ -394,7 +395,7 @@ finalize withdrawals with invalidated games would be considered Critical Severit
 
 ### paused
 
-Returns the value of the Superchain-wide `paused()` flag in the `SuperchainConfig` contract.
+Returns the value of `paused()` from the `SystemConfig` contract.
 
 ### respectedGameType
 
--- a/specs/protocol/deputy-pause-module.md
+++ b/specs/protocol/deputy-pause-module.md
@@ -20,14 +20,12 @@
     - [Mitigations](#mitigations-2)
   - [aDPM-004: Ethereum will not censor transactions for extended periods of time](#adpm-004-ethereum-will-not-censor-transactions-for-extended-periods-of-time)
     - [Mitigations](#mitigations-3)
-  - [aDPM-005: DeputyGuardianModule pause function correctly triggers pause](#adpm-005-deputyguardianmodule-pause-function-correctly-triggers-pause)
+  - [aDPM-005: OpenZeppelin ECDSA and EIP712 contracts are free of critical bugs](#adpm-005-openzeppelin-ecdsa-and-eip712-contracts-are-free-of-critical-bugs)
     - [Mitigations](#mitigations-4)
-  - [aDPM-006: OpenZeppelin ECDSA and EIP712 contracts are free of critical bugs](#adpm-006-openzeppelin-ecdsa-and-eip712-contracts-are-free-of-critical-bugs)
+  - [aDPM-006: Safe contracts are free of critical bugs](#adpm-006-safe-contracts-are-free-of-critical-bugs)
     - [Mitigations](#mitigations-5)
-  - [aDPM-007: Safe contracts are free of critical bugs](#adpm-007-safe-contracts-are-free-of-critical-bugs)
-    - [Mitigations](#mitigations-6)
   - [aDPM-008: Deputy key is capable of creating signatures](#adpm-008-deputy-key-is-capable-of-creating-signatures)
-    - [Mitigations](#mitigations-7)
+    - [Mitigations](#mitigations-6)
 - [System Invariants](#system-invariants)
   - [iSCP-001: Pause can be activated within a short bounded time of authorization](#iscp-001-pause-can-be-activated-within-a-short-bounded-time-of-authorization)
     - [Impact](#impact)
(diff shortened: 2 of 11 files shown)


specs/protocol/superchain-config.md
```diff
@@ -5,60 +5,257 @@
...
+The **Guardian** is a dedicated role that in the OP Stack that is permitted to trigger certain
+actions to maintain the security of an OP Chain or set of OP Chains in case of a bug in the
+protocol. In the Superchain, the Guardian role is held by the Optimism Security Council.
 
-## Configuration data flow
+### Pause Deputy
 
-All contracts which read from the `SuperchainConfig` contract hold its address as storage values
-in the proxy account, and call directly to it when reading configuration data.
+The **Pause Deputy** is a dedicated role managed assigned by the [Guardian](#guardian) that can
```
ajsutton:
```suggestion
The **Pause Deputy** is a dedicated role assigned by the [Guardian](#guardian) that can
```

specs/protocol/superchain-config.md
```diff
@@ -5,60 +5,245 @@
...
+### pausable
+
+Allows any user to check if the [Pause Mechanism](#pause-mechanism) can be triggered for a specific
+[Pause Identifier](#pause-identifier). The pausable status of a specific identifier is independent
+of the pausable status of the global pause (zero address identifier).
+
+- MUST return true if the pausable flag for the given identifier is 0 (ready to pause).
+- MUST return false if the pausable flag for the given identifier is 1 (used/unavailable).
+
+### paused
```
smartcontracts, line 108: Instead of handling the conditional for the global pause, I felt it would be better if these functions were simpler and just took a single pause identifier as a parameter. Then, inside of the `SystemConfig`, I handle checking for both the local and global pause identifiers.

specs/protocol/stage-1.md
```diff
@@ -125,11 +120,7 @@ flowchart LR
    subgraph GuardianSystem[Guardian System]
       subgraph GuardianSafe[Guardian Safe]
          GS[Guardian Safe]
-         DGM[Deputy Guardian Module]
-      end
-      subgraph FndOpsSafe[Foundation Ops Safe]
-         FndOps[Foundation Ops Safe]
-         DPM[Deputy Pause Module]
+         DGM[Deputy Pause Module]
```
ajsutton: Should that be DPM not DGM now?

specs/protocol/superchain-config.md
```diff
@@ -5,60 +5,245 @@
...
 
-#### Paused identifiers
+- When the identifier is the zero address (`0x0000000000000000000000000000000000000000`), the pause
+  applies globally to all chains sharing the `SuperchainConfig` contract.
+- When the identifier is a non-zero address, the pause applies specifically to the chain or set of
+  chains associated with that identifier.
 
-When the system is paused the `Paused(string identifier)` event is emitted. This allows for the
-caller to provide additional information to be used during incident response.
+The identifier is typically expected to be an `ETHLockbox` address, as chains within the Superchain
```
ajsutton: I wonder if "typically expected" is too weak here.  It is either the lockbox address or the 0 address. That's all SystemConfig should be checking and it's weird to leave open the possibility of setting completely random identifiers.

specs/protocol/superchain-config.md
```diff
@@ -5,60 +5,245 @@
...
+only cause a *temporary* [Withdrawal Liveness](#withdrawal-liveness) failure. This is distinct from
+the [iSUPC-001][iSUPC-001] which allows the [Guardian](#guardian) to cause an indefinite liveness failure.
+
+#### Impact
+
+**Severity: High**
+
+If this invariant were broken, the Pause Deputy would be able to cause a
+[Withdrawal Safety](#withdrawal-safety) failure, which would be a violation of the definition of
+Stage 1 as of [January 2025][stage-1]. Because the Pause Deputy is generally assumed to be a
```
ajsutton: This seems to have been copied from the above invariant.  For this one they could only cause liveness failures if it were broken.

ajsutton (approved):
Few small things but mostly looks good.

specs/protocol/stage-1.md
smartcontracts: Good catch

specs/protocol/superchain-config.md
smartcontracts: SGTM

specs/protocol/superchain-config.md
smartcontracts: Tweaked the phrasing. Trying to capture things like what @lucadonnoh brought up where pausing for excessively long periods of time can actually lead to a safety failure.

ajsutton: Ah ok, that makes sense.

ajsutton (approved):

specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -372,6 +375,20 @@ finalize withdrawals with invalidated games would be considered Critical Severit
 
 - [aASR-003](#aasr-003-incorrectly-resolving-games-will-be-invalidated-before-they-have-valid-claims)
 
+### iASR-005: The Anchor Game is recent enough to be fault provable
+
+We require that the Anchor Game corresponds to an L2 block with an L1 origin timestamp that is no
+older than 6 months from the current timestamp. This time constraint is necessary because the fault
+proof VM must walk backwards through L1 blocks to verify derivation, and processing 7 months worth
+of L1 blocks approaches the maximum time available to challengers in the dispute game process.
```
mds1, line 383: What constrains this processing time, i.e. what would need to be improved/changed to allow older anchor games? I don't think we need to support older anchor games, just want to make sure I understand the trade off

specs/protocol/deputy-pause-module.md
```diff
@@ -104,28 +96,27 @@ replaced.
...
+We assume that the process by which a pause can be authorized is well-defined such that the cases
+in which the pause should and should not be triggered are apparent.
 
-### aSCP-002: Superchain-wide pause authorization process is robust
+### aSCP-002: Pause authorization process is robust
 
-We assume that the process by which a Superchain-wide pause can be authorized correctly accounts
-for all cases in which such a pause would be necessary.
+We assume that the process by which a pause can be authorized correctly accounts for all cases in
+which such a pause would be necessary.
```
mds1: This sounds pretty similar to the first assumption, what makes these unique from one another?

specs/protocol/deputy-pause-module.md
```diff
@@ -104,28 +96,27 @@ replaced.
 > typically only rely on a subset of the following assumptions. Different invariants may rely on
 > different assumptions. Refer to individual invariants for their dependencies.
 
-### aSCP-001: Superchain-wide pause authorization process is well-defined
+### aSCP-001: Pause authorization process is well-defined
```
mds1, line 99: SCP = superchain pause? maybe we should have a legend for these acronyms

specs/protocol/deputy-pause-module.md
```diff
@@ -203,44 +183,44 @@ We assume that the Deputy key configured is actually capable of creating signatu
...
 
-### iSCP-001: Superchain-wide pause can be activated within a short bounded time of authorization
+### iSCP-001: Pause can be activated within a short bounded time of authorization
 
-It is an important system-level invariant that the Superchain-wide pause functionality can be
-activated within a short, bounded amount of time of its authorization by the standard process that
-approves the usage of this pause.
+It is an important system-level invariant that the pause functionality can be activated within a
+short, bounded amount of time of its authorization by the standard process that approves the usage
+of this pause.
```
mds1: Should we define a specific upper bound instead of just saying a "short" amount of time? Maybe 30 minutes? (really it can probably be even shorter since this is only considering post-authorization)

specs/protocol/superchain-config.md
```diff
@@ -5,60 +5,243 @@
...
+  - [Guardian](#guardian)
+  - [Pause Deputy](#pause-deputy)
+  - [Pause Identifier](#pause-identifier)
+  - [Withdrawal Liveness](#withdrawal-liveness)
+  - [Withdrawal Safety](#withdrawal-safety)
+  - [Pause Mechanism](#pause-mechanism)
+- [Invariants](#invariants)
+  - [iSUPC-001: The Guardian can only cause a Withdrawal Liveness failure](#isupc-001-the-guardian-can-only-cause-a-withdrawal-liveness-failure)
+    - [Impact](#impact)
+  - [iSUPC-002: The Pause Deputy can only cause a temporary Withdrawal Liveness failure](#isupc-002-the-pause-deputy-can-only-cause-a-temporary-withdrawal-liveness-failure)
```
mds1: Is it intentional that the guardian "can only cause a **_[permanent]_** Withdrawal Liveness failure" but the pause deputy "can only cause a **_temporary_** Withdrawal Liveness failure"? This is because the SC is the guardian and they can continually pause -> reset -> pause?

specs/protocol/superchain-config.md
```diff
@@ -5,60 +5,243 @@
...
+- When the identifier is the zero address (`0x0000000000000000000000000000000000000000`), the pause
+  applies globally to all chains sharing the `SuperchainConfig` contract.
+- When the identifier is a non-zero address, the pause applies specifically to the chain or set of
+  chains associated with that identifier.
 
-When the system is paused the `Paused(string identifier)` event is emitted. This allows for the
-caller to provide additional information to be used during incident response.
+The identifier must be an `ETHLockbox` address or `address(0)`. This allows for targeted pausing of
+either specific chains, the interop set (which shares an `ETHLockbox` contract), or the all chains
+that share the same `SuperchainConfig` when `address(0)` is used as the identifier.
```
mds1: Does pausing revert if the identifier is an unknown address? e.g. do we call a method on the identifier to verify it's likely a lockbox?

specs/fault-proof/stage-one/anchor-state-registry.md
smartcontracts: Processing time is, in the worst case, approximately proportional to the distance in L1 blocks between the L1 origin of the anchor state and the current L1 block at the time that the new dispute game is created. Our best bet in improving this is an architectural change to the proof system or faster proof systems altogether.

specs/protocol/superchain-config.md
smartcontracts: Yes this is intentional. The Guardian (i.e., a majority of the security council) can cause a permanent liveness failure if they choose to.

specs/protocol/superchain-config.md
smartcontracts: For simplicity I figured we wouldn't, we maybe could? I'm not strongly opinionated here.

specs/protocol/deputy-pause-module.md
smartcontracts: Oh we should kill that and just always refer to the Pause Mechanism and link it.

specs/protocol/deputy-pause-module.md
smartcontracts: They're pretty similar. The idea here is that the first assumption is saying "we know all of the circumstances in which the pause should or should not be used" and the second is saying "if such a circumstance occurs, the pause *will* be used".

specs/protocol/deputy-pause-module.md
smartcontracts: Ah this does include the authorization process. I think 30 mins is too low. I agree we need an SLA though.

specs/fault-proof/stage-one/anchor-state-registry.md
```diff
@@ -261,10 +264,10 @@ have [Valid Claims](#valid-claim).
...
 Proper Games that resolve in favor the Defender will be considered to have Valid Claims after the
 [Dispute Game Finality Delay](#dispute-game-finality-delay-airgap) has elapsed UNLESS the
-Superchain-wide pause mechanism is active. Therefore, in the absence of the Superchain-wide pause
-mechanism, parties responsible for game invalidation have exactly the Dispute Game Finality Delay
-to invalidate a withdrawal after it resolves incorrectly. If the Superchain-wide pause is active,
-then any incorrectly resolving games must be invalidated before the pause is deactivated.
+Pause Mechanism is active. Therefore, in the absence of the Pause Mechanism, parties responsible
+for game invalidation have exactly the Dispute Game Finality Delay to invalidate a withdrawal after
+it resolves incorrectly. If the Pause Mechanism is active, then any incorrectly resolving games
+must be invalidated before the pause is deactivated.
```
maurelian, line 270: No strong opinion, but why don't we just retire everything when a pause is triggered?

specs/protocol/superchain-config.md
```diff
@@ -5,60 +5,243 @@
...
 
-#### Scope of pausability
+OP Chains are expected to integrate with the `SuperchainConfig` via their `SystemConfig` contract,
+which will check for the status of the pause by passing along the address of the `ETHLockbox`
+being used within that system as the Pause Identifier.
 
-The pause applies specifically to withdrawals of assets from the L1 bridge contracts. The L2 bridge contracts
-are not pausable, on the basis that issues on L2 can be addressed more easily by a hard fork in the consensus
-layer.
+### Withdrawal Liveness
```
maurelian: This liveness and safety content is redundant (though I think better) than what is in the [stage-1 doc](https://github.com/ethereum-optimism/specs/blob/5aae4a4fb8e9149c5c2441f1d5cc265195e741db/specs/protocol/stage-1.md#L30-L41). I think we should dedupe and have one of these two docs reference each other.

specs/protocol/superchain-config.md
```diff
@@ -5,60 +5,243 @@
...
+### paused
+
+Allows any user to check if the system is currently paused for a specific
+[Pause Identifier](#pause-identifier).
+
+- MUST return true if the pause timestamp for the given identifier is non-zero AND not expired
+  (current time < pause timestamp + expiry duration).
+- MUST return false if the pause timestamp is 0 for the given identifier.
+- MUST return false if the pause has expired (current time > pause timestamp + expiry duration) for
+  the given identifier.
```
maurelian: Nit, the case of `current time == pause timestamp + expiry duration` is unspecified.

This might be simpler (and even here the last line is arguably unnecessary):
```suggestion
- MUST return true if the pause timestamp for the given identifier is non-zero AND not expired
  (current time < pause timestamp + expiry duration). MUST return false otherwise.
- MUST return false if the pause timestamp is 0 for the given identifier.
```

specs/protocol/system-config.md
```diff
@@ -5,164 +5,304 @@
...
-The gas limit of the L2 blocks is configured through the system config.
-Changes to the L2 gas limit are fully applied in the first L2 block with the L1 origin that introduced the change,
-as opposed to the 1/1024 adjustments towards a target as seen in limit updates of L1 blocks.
+Changes to the L2 gas limit are fully applied in the first L2 block with the L1 origin that
+introduced the change, as opposed to the 1/1024 adjustments towards a target as seen in limit
+updates of L1 blocks.
 
-The gas limit may not be set to a value larger than `200_000_000`. This is to ensure that the L2 blocks are fault
-provable and of reasonable size to be processed by the client software. Over time, this value will be increased.
+The gas limit may not be set to a value larger than `200,000,000`. This is to ensure that L2 blocks
```
maurelian: I think we should avoid duplicating info that belongs in `configurability.md` here.

specs/protocol/system-config.md
```diff
@@ -5,164 +5,304 @@
...
+- MUST set the start block to the current block number if it hasn't been set already.
+
+### minimumGasLimit
+
+Returns the minimum L2 gas limit that can be safely set for the system to operate, calculated as
+the sum of the maximum resource limit and the system transaction maximum gas.
+
+### maximumGasLimit
+
+Returns the maximum L2 gas limit that can be safely set for the system to operate (200,000,000).
```
maurelian: same

specs/fault-proof/stage-one/anchor-state-registry.md
smartcontracts: @ajsutton felt pretty strongly about this and I'm inclined to agree - the pause is something we want to be able to use without being too worried about the consequences. If we attach more strings to the pause then we might hesitate to do so when it comes down to it.

specs/protocol/superchain-config.md
smartcontracts: Will dedupe, will move most of the invariants into the stage-1 doc.

specs/fault-proof/stage-one/anchor-state-registry.md
maurelian: OK, I recall that conversation now, thanks for clarifying.

specs/protocol/deputy-pause-module.md
mds1: Got it, I'd suggest tweaking the phrasing a bit to make that more clear. Feel free to resolve this thread afterwards

specs/protocol/deputy-pause-module.md
mds1: kill what? I was referring to the SCP acronym in `aSCP-001`

specs/protocol/deputy-pause-module.md
mds1: 30 minutes seems ok (arguably even long) if it's post-authorization (which was my assumption, given "iSCP-001: Pause can be activated within a short bounded time of authorization"), but agreed it should be longer if that time includes authorization

specs/protocol/superchain-config.md
mds1: Reverting is nice since it's a little safer, but I think we can only probabilistically revert (i.e. I think we can't know for certainty if an indentifier is actually an ETH Lockbox that's part of an L2 using this superchain config?). Checking bytecode feels excessive, so the best we can do is call a method like `paused` or `proxyAdminOwner` to check if it's likely the address is a lockbox at all

specs/protocol/superchain-config.md
```diff
@@ -5,60 +5,243 @@
...
 
-The pause applies specifically to withdrawals of assets from the L1 bridge contracts. The L2 bridge contracts
-are not pausable, on the basis that issues on L2 can be addressed more easily by a hard fork in the consensus
-layer.
+### Withdrawal Liveness
 
-When the Pause is activated, the following methods are disabled:
+**Withdrawal Liveness** is the ability for users to execute valid withdrawals out of any contract
+that stores ETH or tokens within an OP Chain's set of smart contracts. We tend to refer to
+Withdrawal Liveness in the context of the `OptimismPortal` and the rest of the Standard Bridge
```
mds1: typo? unsure what you meant by "the rest of the Standard Bridge"
```suggestion
Withdrawal Liveness in the context of the `OptimismPortal` and the `StandardBridge`
```

specs/protocol/deputy-pause-module.md
smartcontracts: Oh sorry I misread this

smartcontracts: Going to keep as-is for now, this is relatively experimental, the actual acronyms don't matter that much for now. If we use this more then I'll figure out a better naming scheme

maurelian (approved):

specs/protocol/superchain-config.md
smartcontracts: I'd like to leave this as-is for now, I'm worried that a revert here adds risk without major benefit, especially if there are contracts on different versions that have different functions. We can keep legacy functions in place but I'm not confident that explicitly limiting this to ETHLockbox addresses gives us any significant benefit.

mds1 (approved):

specs/protocol/stage-1.md
```diff
@@ -5,96 +5,202 @@
...
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 ## Overview
 
-This document outlines configuration requirements (including roles and other parameters)
-for an implementation of the OP Stack to satisfy the Stage 1 decentralization requirements as defined
-by
-L2 Beat [[1](https://medium.com/l2beat/introducing-stages-a-framework-to-evaluate-rollups-maturity-d290bb22befe), [2](https://medium.com/l2beat/stages-update-security-council-requirements-4c79cea8ef52)].
+This document describes the requirements necessary for an OP Stack implementation to satisfy the
+Stage 1 decentralization requirements [as defined by L2Beat][stage-1]. It also defines the specific
```
lucadonnoh:
```suggestion
Stage 1 decentralization requirements [as defined by L2BEAT][stage-1]. It also defines the specific
```

specs/protocol/stage-1.md
```diff
@@ -5,96 +5,202 @@
...
+that share the same `SuperchainConfig` when `address(0)` is used as the identifier.
 
-The following diagram outlines the control relationships between the contracts in the system.
+OP Chains are expected to integrate with the `SuperchainConfig` via their `SystemConfig` contract,
+which will check for the status of the pause by passing along the address of the `ETHLockbox`
+being used within that system as the Pause Identifier.
+
+### Stage 1 Rollup
+
+L2Beat defines that a system can be considered a "Stage 1 Rollup" system if it has the following
```
lucadonnoh:
```suggestion
L2BEAT defines that a system can be considered a "Stage 1 Rollup" system if it has the following
```

specs/protocol/stage-1.md
```diff
@@ -5,96 +5,202 @@
...
+properties:
+
+- The system has a Security Council ([as defined by L2Beat][security-council]).
```
lucadonnoh:
```suggestion
- The system has a Security Council ([as defined by L2BEAT][security-council]).
```

specs/protocol/stage-1.md
```diff
@@ -115,41 +221,96 @@ flowchart LR
...
+[`ProtocolVersions`contract](../protocol/superchain-upgrades.md#protocolversions-l1-contract) is
+listed as "Out of Protocol", because the decision to follow the version signals in the contract is
+optional. It is included here for completeness, but it does not impact
+[Withdrawal Safety](#withdrawal-safety) or [Withdrawal Liveness](#withdrawal-liveness).
+
+## Invariants
+
+### iS1-001: A permanent Withdrawal Liveness or Withdrawal Safety failure requires 75% of the Security Council (w/o bugs)
+
+This is the core invariant for Stage 1 and aligns with [the definition by L2Beat][stage-1] from
```
lucadonnoh:
```suggestion
This is the core invariant for Stage 1 and aligns with [the definition by L2BEAT][stage-1] from
```

specs/protocol/stage-1.md
lucadonnoh: btw having a security council is not strictly necessary, in the sense that you could just have long delays for everything that provide an exit window. a security council becomes necessary for emergency actions that need to be applied fast enough

lucadonnoh (approved):

mslipper (approved):

PR #636, opened 2025-03-19
feat(isthmus): Operator fee spec improvements

### LINKED ISSUE #44: chore(specs): Improve operator fee specs
## Overview

The [operator fee specs](https://specs.optimism.io/protocol/isthmus/exec-engine.html#operator-fee) are fairly bare right now, not mentioning the semantic details of how the EVM actually charges the fee.

We should update the spec to be very specific on:
1. Which transactions the operator fee is charged for. (i.e., "the operator fee is only charged on non-deposit transactions")
1. How operator fee gas is bought (i.e., "operator fee gas is bought using the worst-case operator fee, akin to how standard gas is bought using the worst-case cost of the transaction ...")
1. How refunds work (i.e., "operator fee refunds are granted to transactions that have bought operator fee gas, and the refund is calculated as follows ...")
    3.i) Specific notes on refunds not being applied to transactions that don't buy operator fee gas.

The end goal is that the spec clearly notes the invariants that the L2 EVM must hold, as well as any assumptions / implementation details that are inherited from the L1 EVM.

clabby:
## Overview

Adds additional clarity to the operator fee specification. Now covers:
* Which transactions the operator fee applies to
* How exactly the EVM is meant to charge the operator fee to L2-originating transactions.
* A callout that the transaction pool + tx pre-validation steps in the EVM must abide by the new worst-case tx fee formula, in order to correctly reject transactions that do not have enough balance to pay for the worst-case tx fee.

### DIFF
--- a/specs/protocol/isthmus/exec-engine.md
+++ b/specs/protocol/isthmus/exec-engine.md
@@ -27,7 +27,11 @@
   - [`engine_newPayloadV4` API](#engine_newpayloadv4-api)
 - [Fees](#fees)
   - [Operator Fee](#operator-fee)
-    - [Configuring Parameters](#configuring-parameters)
+    - [Fee Formula](#fee-formula)
+    - [Deposit Operator Fees](#deposit-operator-fees)
+    - [EVM Fee Semantics](#evm-fee-semantics)
+    - [Transaction Pool Changes](#transaction-pool-changes)
+    - [Configuring Operator Fee Parameters](#configuring-operator-fee-parameters)
   - [Fee Vaults](#fee-vaults)
   - [Receipts](#receipts)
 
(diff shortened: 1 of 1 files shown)


specs/protocol/isthmus/exec-engine.md
```diff
@@ -208,17 +212,57 @@ and the `operatorFeeConstant`.
...
+#### Fee Formula
+
+$$
+\text{operatorFee} = (\text{gas} \times \text{operatorFeeScalar} \div 10^6) + \text{operatorFeeConstant}
+$$
 
 Where:
 
-- `gasUsed` is amount of gas used by the transaction.
+- `gas` is the amount of gas that the transaction used. When calculating the amount of gas that is bought at the
```
teddyknox, line 226: Are you defining `gas` generally or as a specific term in this context?

clabby: Generally. Semantically, the value that's passed into this function isn't necessarily the actual gas used post-execution. Pre-execution, we always assume that the gas cost will be the worst-case (using `gas_limit` when calculating it,) and then just refund the diff when we finish execution.

specs/protocol/isthmus/exec-engine.md
```diff
@@ -208,17 +212,57 @@ and the `operatorFeeConstant`.
 
 ### Operator Fee
 
-The operator fee, is set as follows:
+The operator fee is integrated directly into the EVM, alongside the standard gas fee and the OP Stack specific L1 data
+fee. This fee must follow the exact same semantics as existing fees charged in the EVM.
```
refcell: > This fee must follow the exact same semantics as existing fees charged in the EVM.

Not sure what this means. The operator fee is not applied in the exact same way tx costs are applied - fees are accrued to a new operator fee recipient while the charge and refund apply to the tx sender. Maybe there's something we could point to here to make this more clear, otherwise it should be deleted as it doesn't add anything concrete.

clabby: I see what you mean - what I'm trying to convey here is that the existing semantics of "charge worst-case to caller before execution, refund remaining after execution, and send spent gas to a fee beneficiary" are the same. Maybe some updates to highlight that the only diff is the new fee recipient will be more clear?

specs/protocol/isthmus/exec-engine.md
```diff
@@ -208,17 +212,57 @@ and the `operatorFeeConstant`.
...
+   gas that was unused (i.e., the caller should only be charged the _effective_ operator fee.) The refund should be
+   calculated as $\text{opFeeRefund} = \text{opFeeWorstCase} - \text{opFeeActual}$, where:
+    - $\text{opFeeWorstCase}$ is as described in #1 + #2.
+    - $\text{opFeeActual}$ is the amount of the operator fee that was actually used. This value is computed using the
+      [fee formula](#fee-formula) with `gas` set to the `gas_limit - gas_used + refunded_gas`
+1. After execution, when rewarding the fee beneficiaries, send the _spent operator fee_ to the operator fee vault. This
+   value is exactly $\text{opFeeActual}$ as described above.
+
+Implementations must ensure that by charging and refunding the operator fee during transaction execution, over the range
+of allowed configuration values, no ETH is minted or destroyed.
```
refcell: Let's simplify this.

e.g. `Implementations must ensure ETH is neither minted nor destroyed as a result of the operator fee.`

specs/protocol/isthmus/exec-engine.md
```diff
@@ -208,17 +212,57 @@ and the `operatorFeeConstant`.
...
+
+#### Transaction Pool Changes
+
+To account for the additional fee factored into transaction validity mentioned above, the transaction pool must reject
+transactions that do not have enough ETH to cover the worst-case cost of the transaction fee. This worst-case cost of a
```
refcell: Preference for `balance` instead of `ETH`.

specs/protocol/isthmus/exec-engine.md
refcell: I actually prefer the route you take comparing the fee pattern.

Maybe it's enough just to remove "exact same". Might I suggest the following.

```suggestion
fee. This follows the semantics of existing fees charged in the EVM.
```

clabby: Perf - will update!

specs/protocol/isthmus/exec-engine.md
```diff
@@ -208,17 +212,56 @@ and the `operatorFeeConstant`.
...
+   worst-case transaction fee.
+1. When buying gas prior to execution, charge the account the worst-case operator fee. To compute this value, use the
+   [fee formula](#fee-formula) with `gas` set to the `gas_limit` of the transaction.
+1. After execution, when issuing refunds, transactions that bought operator fee gas should be refunded the operator fee
+   gas that was unused (i.e., the caller should only be charged the _effective_ operator fee.) The refund should be
+   calculated as $\text{opFeeRefund} = \text{opFeeWorstCase} - \text{opFeeActual}$, where:
+    - $\text{opFeeWorstCase}$ is as described in #1 + #2.
+    - $\text{opFeeActual}$ is the amount of the operator fee that was actually used. This value is computed using the
+      [fee formula](#fee-formula) with `gas` set to the `gas_limit - gas_used + refunded_gas`
+1. After execution, when rewarding the fee beneficiaries, send the _spent operator fee_ to the operator fee vault. This
```
tynes: nit: link to the fee vault section in this document

specs/protocol/isthmus/exec-engine.md
```diff
@@ -208,17 +212,56 @@ and the `operatorFeeConstant`.
 
 ### Operator Fee
 
-The operator fee, is set as follows:
+The operator fee is integrated directly into the EVM, alongside the standard gas fee and the OP Stack specific L1 data
+fee. This fee follows the same semantics of existing fees charged in the EVM, just with a new fee beneficiary account.
```
tynes: It would be useful to have a link or description of what you mean by same semantics. That context may not be obvious to all readers

specs/protocol/isthmus/exec-engine.md
clabby: Good shout - done

specs/protocol/isthmus/exec-engine.md
clabby: This is described below, but we can link to the relevant section of the yellow paper. I'm unaware of any other good plain-english descriptions of the semantics, unless you have any suggestions 😄 

relevant section of the yellow paper is on pg. 8, under section 5: "_Gas and Payment_": https://ethereum.github.io/yellowpaper/paper.pdf

clabby: There's also the https://github.com/ethereum/execution-specs, if you feel a python impl is better to link to.

tynes: Linking to the yellow paper would be great

tynes (approved):

tynes:
This looks good to me, looks clear enough to implement directly from the spec. There is the open question of whether or not we want python psuedocode in our specs, sometimes we do it and sometimes we don't

refcell (approved):

specs/protocol/isthmus/exec-engine.md
```diff
@@ -208,17 +212,56 @@ and the `operatorFeeConstant`.
...
+   [fee formula](#fee-formula) with `gas` set to the `gas_limit` of the transaction, and add it to the existing
+   worst-case transaction fee.
+1. When buying gas prior to execution, charge the account the worst-case operator fee. To compute this value, use the
+   [fee formula](#fee-formula) with `gas` set to the `gas_limit` of the transaction.
+1. After execution, when issuing refunds, transactions that bought operator fee gas should be refunded the operator fee
+   gas that was unused (i.e., the caller should only be charged the _effective_ operator fee.) The refund should be
+   calculated as $\text{opFeeRefund} = \text{opFeeWorstCase} - \text{opFeeActual}$, where:
+    - $\text{opFeeWorstCase}$ is as described in #1 + #2.
+    - $\text{opFeeActual}$ is the amount of the operator fee that was actually used. This value is computed using the
+      [fee formula](#fee-formula) with `gas` set to the `gas_limit - gas_used + refunded_gas`
```
sebastianst: maybe we want to clarify that `refunded_gas` are gas refunds according to https://eips.ethereum.org/EIPS/eip-3529 ?

sebastianst (approved):

PR #639, opened 2025-03-21
broken link  Linux/MIPS Syscall 

Hopium21:
I recommend replacing the broken link to Linux/MIPS Syscall with this updated Wikipedia reference:

✅ New link: https://en.wikipedia.org/wiki/System_call

If you have a better suggestion or a more accurate reference, please share it, and I'll update the link accordingly. 🚀

### DIFF
--- a/specs/fault-proof/cannon-fault-proof-vm.md
+++ b/specs/fault-proof/cannon-fault-proof-vm.md
@@ -125,8 +125,9 @@ behavior in typical MIPS implementations, FPVM must raise an exception when step
 
 ## Syscalls
 
-Syscalls work similar to [Linux/MIPS](https://web.archive.org/web/20220529105937/https://www.linux-mips.org/wiki/Syscall), including the
-syscall calling conventions and general syscall handling behavior.
+Syscalls work similar to 
+[Linux/MIPS](https://web.archive.org/web/20220529105937/https://www.linux-mips.org/wiki/Syscall), 
+including the syscall calling conventions and general syscall handling behavior.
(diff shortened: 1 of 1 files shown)


Inphi:
The wiki isn't really informative. I found an old archive of the linux-mips.org here - https://web.archive.org/web/20220529105937/https://www.linux-mips.org/wiki/Syscall. Let's use that instead. Thanks.

Hopium21:
> The wiki isn't really informative. I found an old archive of the linux-mips.org here - https://web.archive.org/web/20220529105937/https://www.linux-mips.org/wiki/Syscall. Let's use that instead. Thanks.

Done

Inphi (approved):

specs/fault-proof/cannon-fault-proof-vm.md
```diff
@@ -125,7 +125,7 @@ behavior in typical MIPS implementations, FPVM must raise an exception when step
 
 ## Syscalls
 
-Syscalls work similar to [Linux/MIPS](https://www.linux-mips.org/wiki/Syscall), including the
+Syscalls work similar to [Linux/MIPS](https://web.archive.org/web/20220529105937/https://www.linux-mips.org/wiki/Syscall), including the
```
Inphi: @Hopium21 could you break up this line to satisfy the markdown linter check.

tynes:
Looks like the linter is still broken

Hopium21:
> Looks like the linter is still broken

All done! Let me know if anything else is needed 🙂

PR #669, opened 2025-04-18
rollup-node-p2p: Specify Unsafe Chain Consensus Behaviors

axelKingsley:
Recently I've had trouble reasoning about the behavior of the network when the Sequencer *would like* to issue a reorg. This lead to the realization that we don't have any specific branch selection rules, which in itself creates some behaviors around consensus of the unsafe chain, L1 consolidation, and batcher configuration.

I called all these points out here in the P2P spec.


specs/protocol/rollup-node-p2p.md
```diff
@@ -348,6 +348,21 @@ A node may apply the block to their local engine ahead of L1 availability, if it
...
+blocks available on the network, the unsafe chain will not have consensus until one of the branches is
+made available for Safe Chain Derivation from the L1, at which point all nodes will sync to the Safe Chain,
+and will continue their unsafe chains from that point.
+
+Because the data which is posted to the L1 is determined by the Batcher, this means that the winning branch
+will be selected by whatever Node the Batcher reads from. This is most relevant when the Sequencer attempts
+to reorg the chain be reissuing new block for an old height - if Nodes hae already processed the old block,
+they will reject the reorg until the L1 corrects them. If the Batcher is not listening to the Sequencer,
+the Sequencer's reorg will be ineffective, as it will be set back to the original chain when the L1 data
+corrects them.
```
axelKingsley: While I'm happy to document this, I would much rather this behave differently as well. Specifically, I'd like for there to be consensus on the unsafe chain when selecting between branches.

Some ways I could see us doing this:
- *latest* block wins, instead of first block wins. Could have some thrash to it
- System transaction or Header somewhere in the block specifies which revision this is, and nodes follow the latest revision (avoids equivocation from causing temporary chain splits)

I tried not to get too far into the weeds about the downstream impact of not having consensus rules for branch selection, but it is important to note that in the current configuration, an Unsafe Reorg issued by the Sequencer results in *at minimum* a halt of the unsafe head until the next L1 publishing, where *none* of the nodes can follow the chain until consolidation. If the batcher is not connected to the Sequencer, not only will there be a halt, but the attempted reorg won't even succeed, instead the Sequencer is pushed back to the old chain after spending ~10 minutes ineffectively building an unsafe chain which will never be taken by another node.

specs/protocol/rollup-node-p2p.md
```diff
@@ -348,6 +349,21 @@ A node may apply the block to their local engine ahead of L1 availability, if it
...
+
+There is no consensus rule for selecting branches when multiple blocks of the same height are available.
+All nodes accept the first valid block which extends their chain. This means that if there are multiple valid
+blocks available on the network, the unsafe chain will not have consensus until one of the branches is
+made available for Safe Chain Derivation from the L1, at which point all nodes will sync to the Safe Chain,
+and will continue their Unsafe Chains from that point.
+
+Because the data which is posted to the L1 is determined by the Batcher, this means that the winning branch
+will be selected by whatever Node the Batcher reads from. This is most relevant when the Sequencer attempts
+to reorg the chain by reissuing new block for an old height - if Nodes hae already processed the old block,
```
axelKingsley:
```suggestion
to reorg the chain by reissuing new block for an old height - if Nodes have already processed the old block,
```

specs/protocol/rollup-node-p2p.md
```diff
@@ -348,6 +349,21 @@ A node may apply the block to their local engine ahead of L1 availability, if it
 - The subsequent forkchoice-update ensures this block is recognized as "unsafe"
   (see [fork choice updated](derivation.md#engine-api-usage))
 
+#### Branch selection
+
+There is no consensus rule for selecting branches when multiple blocks of the same height are available.
+All nodes accept the first valid block which extends their chain. This means that if there are multiple valid
+blocks available on the network, the unsafe chain will not have consensus until one of the branches is
+made available for Safe Chain Derivation from the L1, at which point all nodes will sync to the Safe Chain,
+and will continue their Unsafe Chains from that point.
```
ajsutton: This is how I often express this but I think it's probably overly negative. I think it's more along the lines of:

Since the sequencer is not expected to equivocate, the fork choice rule for unsafe blocks operates on a "first seen" basis where the canonical unsafe chain is only re-orged if the safe head changes.

ie there is a fork choice rule, it's just very simple and says to follow the chain you first see (that builds on the safe head). And then it is worth going on to confirm that nodes which see sequencer blocks in different orders will not reach consensus until the safe head advances as you have said.

ajsutton (commented):
Looks good generally and I think it is worth explicitly capturing this, even if we do then go on to change it.  Just suggest tweaking the wording a little.

specs/protocol/rollup-node-p2p.md
axelKingsley: Sure, I reworded this more along your lines, and removed some of the rabbit-hole comments around the batcher.

specs/protocol/rollup-node-p2p.md
```diff
@@ -348,6 +349,16 @@ A node may apply the block to their local engine ahead of L1 availability, if it
 - The subsequent forkchoice-update ensures this block is recognized as "unsafe"
   (see [fork choice updated](derivation.md#engine-api-usage))
 
+#### Branch selection
+
+Nodes expect that the sequencer will not equivocate, and therefore the fork choice rule for unsafe blocks
+is a "first block wins" model, where the unsafe chain will not change once it has been extended, unless
```
emhane, line 355:
```suggestion
is a FCFS model, where the unsafe chain will not change once it has been extended, unless
```
can this be expressed by standardised FCFS term instead? blocks pass through some kind of validation before being added right, so think "serve", in first come first _serve_, isn't the wrong word here, but open to that being debated

axelKingsley: I don't agree with "FCFS" being used here, because a later block won't ever be served if an earlier block was. Where FCFS describes an ordering, "First Block Wins" describes a selection.

I guess if you define "Serve" as "Consider this block" you could make the case, but if a competing block has already landed, no other block of the same height will be accepted, even if they are equally as valid as the original block.

emhane: ok fair enough. was just an attempt to resort to standardised terminology wherever possible to lower barriers of entry as much as possible to devs familiar with the particular strategy but not with OP specs. nonetheless, for this particular situation, it does not seem to be possible based on your explanation, hence I rest my case.

ajsutton (approved):
LGTM

PR #671, opened 2025-04-18
specs: add expiry window

tynes:
**Description**

Going with an expiry window of 30 days.

See https://github.com/ethereum-optimism/optimism/issues/10981 for
context.

App devs can rely on
https://github.com/ethereum-optimism/design-docs/blob/25ef5537e39b63cddf1c83479cee9f0e02431dce/protocol/resend-messages.md#L4
to get around this issue

### DIFF
--- a/specs/interop/derivation.md
+++ b/specs/interop/derivation.md
@@ -106,10 +106,6 @@ The expiry window is the time period after which an initiating message is no lon
 | -------- | ----- |
 | `EXPIRY_WINDOW` | `2592000` (30 days) |
 
-The expiry window ensures that the proof can execute in a reasonable amount of time.
-The proof program needs to walk back and re-execute to reproduce the consumed logs. This means
-that very old logs are more expensive to prove.
-
 ## Security Considerations
 
 ### Depositing an Executing Message
--- a/specs/interop/derivation.md
+++ b/specs/interop/derivation.md
@@ -13,6 +13,7 @@
 - [Expiry Window](#expiry-window)
 - [Security Considerations](#security-considerations)
   - [Depositing an Executing Message](#depositing-an-executing-message)
+  - [Expiry Window](#expiry-window-1)
   - [Reliance on History](#reliance-on-history)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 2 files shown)


specs/interop/derivation.md
```diff
@@ -105,7 +104,11 @@ The expiry window is the time period after which an initiating message is no lon
 
 | Constant | Value |
 | -------- | ----- |
-| `EXPIRY_WINDOW` | `TODO` |
+| `EXPIRY_WINDOW` | `2592000` (30 days) |
+
+The expiry window ensures that the proof can execute in a reasonable amount of time.
+The proof program needs to walk back and re-execute to reproduce the consumed logs. This means
+that very old logs are more expensive to prove.
```
protolambda: I don't believe it has to re-execute. It can just prove by receipts-hash, since the older L2 blocks are trusted.

protolambda: I may be wrong here though, I did not implement the op-program behavior there. Worth checking with proofs team.

ajsutton: Yeah the expensive part is just walking back since it has to be done by block hash and we can't just do a `blockByNumber` call. There's no re-execution involved.

tynes: Ok, happy to delete this

specs/interop/derivation.md
```diff
@@ -123,10 +122,3 @@ When fully executing historical blocks, a dependency on historical receipts from
 needing to execute increasingly long chain histories.
 
 [eip-4444]: https://eips.ethereum.org/EIPS/eip-4444
-
-### Expiry Window
-
-The expiry window ensures that the proof can execute in a reasonable amount of time.
-There is currently no way to prove old history with a sublinear proof size. The proof
-program needs to walk back and re-execute to reproduce the consumed logs. This means
-that very old logs are more expensive to prove.
```
emhane, line 132: can we keep this section and instead write
> App devs can rely on
https://github.com/ethereum-optimism/design-docs/blob/25ef5537e39b63cddf1c83479cee9f0e02431dce/protocol/resend-messages.md#L4
to get around this issue

from the pr description in this section?

emhane (changes requested):

specs/interop/derivation.md
```diff
@@ -105,7 +104,7 @@ The expiry window is the time period after which an initiating message is no lon
 
 | Constant | Value |
 | -------- | ----- |
-| `EXPIRY_WINDOW` | `TODO` |
+| `EXPIRY_WINDOW` | `2592000` (30 days) |
```
emhane:
```suggestion
| `EXPIRY_WINDOW` | `2592000 secs` (30 days) |
```
units!

emhane (changes requested):

specs/interop/derivation.md
clabby: agree. https://github.com/ethereum-optimism/specs/pull/681

clabby (approved):

specs/interop/derivation.md
clabby: addressed after 681 got merged in.

emhane (approved):

PR #672, opened 2025-04-22
Add Tentative Specs for Interop Upgrade Txs

### LINKED ISSUE #637: Update Interop Network Upgrade Transactions Spec
The network upgrade transactions are not fully specified for interop. We need to make sure that they are defined and fully sufficient.

The following list contains the upgrade transactions (but double check):
- deploy CrossL2Inbox
- deploy L2ToL2CrossDomainMessenger
- ~~deploy ETHLiquidity~~
- ~~deploy SuperchainTokenBridge~~
- ~~deploy SuperchainETHBridge~~
- ~~mint liquidity into ETHLiquidity `type(uint248).max`~~
- set 1967 proxy for CrossL2Inbox
- set 1967 proxy for L2ToL2CrossDomainMessenger
- ~~set 1967 proxy for ETHLiquidity~~
- ~~set 1967 proxy for SuperchainTokenBridge~~
- ~~set 1967 proxy for SuperchainETHBridge~~

We do not need to deploy the `OptimismMintableERC20TokenFactory` because that was defined as part of the [liquidity migration](https://github.com/ethereum-optimism/design-docs/blob/9e919c5b173fe8fc89949b012f6f70a0bc3247f6/protocol/superchain-erc20/liquidity-migration.md) project

Update as of 04/28/25: the initial release of interop will ship with only the `CrossL2Inbox` and the `L2ToL2CrossDomainMessenger` to simplify the rollout

teddyknox:
**Description**

Conforms to the outline described in #637. Is based on contracts @ [71c460ec7c7c05791ddd841b97bcb664a1f0c753](https://github.com/ethereum-optimism/optimism/tree/71c460ec7c7c05791ddd841b97bcb664a1f0c753), with the plan to update these specs once the interop contracts release is done. Includes specs for the following upgrade transactions:

* deploy CrossL2Inbox
* deploy L2ToL2CrossDomainMessenger
* set 1967 proxy for CrossL2Inbox
* set 1967 proxy for L2ToL2CrossDomainMessenger

**Tests**

**Additional context**

Used the scripts #673 to generate these specs, so that they will be easy to update with specs associated with the contracts release for interop later.

**Metadata**


specs/interop/derivation.md
```diff
@@ -87,18 +98,449 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+# 0x50684989256294e3c64949ea1cf5bad586c7e6b91b8b7f21ee9ef7086efe60db
+```
+
+### Mint mint liquidity into ETHLiquidity
+
+This transaction mints liquidity into the ETHLiquidity contract.
+
+A deposit transaction is derived with the following attributes:
+
+- `from`: `0x4200000000000000000000000000000000000024`
```
teddyknox: @protolambda am I correct to set the `from` address here to the `SuperchainETHBridge` proxy address, such that `msg.sender != Predeploys.SUPERCHAIN_ETH_BRIDGE` ([source](https://github.com/ethereum-optimism/optimism/blob/46db2fc4e69c2e5e310edfc9b72939900a240e6a/packages/contracts-bedrock/src/L2/ETHLiquidity.sol#L40)?

tynes:
nice work with the automation script @teddyknox

specs/interop/derivation.md
tynes: This should be a simple send of value to the contract where the deposit transaction has a `mint` and `value` set to `type(uint248).max`. Realizing now for this to work, the contract needs a `receive` function. Another option would be to deploy a dummy contract that uses self destruct to transfer the value without calling it, this pattern exists inside of `op-faucet`. This would prevent us needing to modify the source code of the `ETHLiquidity` contract

emhane (commented):
can this pr be rescoped into one pr per network upgrade transactions sub section? hopeful there will be a better chance of incrementally merging this pr and thereby speed up merging it entirely. it is easier to sign off small diff on first review, than large diff.

specs/interop/derivation.md
```diff
@@ -87,17 +91,190 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+
+With MaxResourceLimit set to `20_000_000` and SystemTxMaxGas set to `1_000_000`,
+this mean every upgrading chain MUST have their L2GasLimit set to greater than
+`24_320_000`.
+
+The upgrade transaction details below are based on a nightly release at commit
+hash `71c460ec7c7c05791ddd841b97bcb664a1f0c753`, and will be updated once a
+contracts release is made.
+
+Explicitly not upgraded as part of Interop is the OptimismMintableERC20Factory
```
tynes: nit: i think its fine to remove this paragraph and not mention the contracts that aren't included, and then we can follow up and move the specs for what is not being released into the `experimental` directory

specs/static/bytecode/interop-e-t-h-liquidity-deployment.txt
```diff
@@ -0,0 +1 @@
...94565b610146565b337342000000000000000000000000000000000000241461010f576040517f82b4290000000000000000000000000000000000000000000000000000000000815260040160405180910390fd5b60405134815233907f875e07afd7ce17c6531b1a6b7b34829dcd8b7e6639448afbd6a8e29fa1422b829060200160405180910390a2565b3373420000000000000000000000000000000000002414610193576040517f82b4290000000000000000000000000000000000000000000000000000000000815260040160405180910390fd5b80336040516101a190610215565b73ffffffffffffffffffffffffffffffffffffffff90911681526020016040518091039082f09050801580156101db573d6000803e3d6000fd5b505060405181815233907f85719716ac5bd2744ae7ed3d16702129383049b97123b506320e7a5826ebbbba9060200160405180910390a250565b6059806102ae83390190565b600060208083528351808285015260005b8181101561024e57858101830151858201604001528201610232565b81811115610260576000604083870101525b50601f017fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe016929092016040019392505050565b6000602082840312156102a657600080fd5b503591905056fe608060405260405160593803806059833981016040819052601e91602a565b806001600160a01b0316ff5b600060208284031215603b57600080fd5b81516001600160a01b0381168114605157600080fd5b939250505056fea164736f6c634300080f000a
```
tynes: we should be able to remove this for now

specs/static/bytecode/interop-superchain-e-t-h-bridge-deployment.txt
```diff
@@ -0,0 +1 @@
...73ffffffffffffffffffffffffffffffffffffffff85169133917fed98a2ff78833375c368471a747cdf0633024dde3f870feb08a934ac5be83402910160405180910390a392915050565b60598061072683390190565b73ffffffffffffffffffffffffffffffffffffffff811681146105ab57600080fd5b50565b6000806000606084860312156105c357600080fd5b83356105ce81610589565b925060208401356105de81610589565b929592945050506040919091013590565b6000815180845260005b81811015610615576020818501810151868301820152016105f9565b81811115610627576000602083870101525b50601f017fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe0169290920160200192915050565b60208152600061066d60208301846105ef565b9392505050565b6000806040838503121561068757600080fd5b823561069281610589565b946020939093013593505050565b600080604083850312156106b357600080fd5b82516106be81610589565b6020939093015192949293505050565b83815273ffffffffffffffffffffffffffffffffffffffff8316602082015260606040820152600061070360608301846105ef565b95945050505050565b60006020828403121561071e57600080fd5b505191905056fe608060405260405160593803806059833981016040819052601e91602a565b806001600160a01b0316ff5b600060208284031215603b57600080fd5b81516001600160a01b0381168114605157600080fd5b939250505056fea164736f6c634300080f000a
```
tynes: we should be able to remove this for now

specs/static/bytecode/interop-superchain-token-bridge-deployment.txt
```diff
@@ -0,0 +1 @@
...5b91508473ffffffffffffffffffffffffffffffffffffffff163373ffffffffffffffffffffffffffffffffffffffff168773ffffffffffffffffffffffffffffffffffffffff167f0247bfe63a1aaa59e073e20b172889babfda8d3273b5798e0e9ac4388e6dd11c8787604051610656929190918252602082015260400190565b60405180910390a450949350505050565b5f81518084528060208401602086015e5f6020828601015260207fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe0601f83011685010191505092915050565b602081525f6106c56020830184610667565b9392505050565b73ffffffffffffffffffffffffffffffffffffffff811681146106ed575f80fd5b50565b5f805f8060808587031215610703575f80fd5b843561070e816106cc565b9350602085013561071e816106cc565b9250604085013561072e816106cc565b9396929550929360600135925050565b5f805f8060808587031215610751575f80fd5b843561075c816106cc565b9350602085013561076c816106cc565b93969395505050506040820135916060013590565b5f8060408385031215610792575f80fd5b825161079d816106cc565b6020939093015192949293505050565b5f602082840312156107bd575f80fd5b815180151581146106c5575f80fd5b83815273ffffffffffffffffffffffffffffffffffffffff83166020820152606060408201525f6108006060830184610667565b95945050505050565b5f60208284031215610819575f80fd5b505191905056fea164736f6c6343000819000a
```
tynes: we should be able to remove this for now

specs/interop/derivation.md
```diff
@@ -87,17 +91,190 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+which will be handled as part of the [liquidity
+migration](https://github.com/ethereum-optimism/design-docs/blob/9e919c5b173fe8fc89949b012f6f70a0bc3247f6/protocol/superchain-erc20/liquidity-migration.md)
+project.
+
+<!-- To regenerate the deploy/upgrade tx docs below, run `./scripts/upgrades/gen_interop_upgrade_tx_specs.sh | pbcopy` and then paste here -->
+
+### CrossL2Inbox Deployment
+<!-- Generated with: ./scripts/run_gen_predeploy_docs.sh --optimism-repo-path ../optimism --fork-name Interop --contract-name CrossL2Inbox --from-address 0x4220000000000000000000000000000000000000 --from-address-nonce 0 --git-commit-hash 71c460ec7c7c05791ddd841b97bcb664a1f0c753 --eth-rpc-url https://optimism.rpc.subquery.network/public --proxy-address 0x4200000000000000000000000000000000000022 --copy-contract-bytecode true -->
+
+The `CrossL2Inbox` contract is upgraded.
```
tynes: This is the deployment rather than a contract upgrade

specs/interop/derivation.md
```diff
@@ -87,17 +91,190 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+
+```bash
+cast keccak $(cast concat-hex 0x0000000000000000000000000000000000000000000000000000000000000002 $(cast keccak "Interop: CrossL2Inbox Proxy Update"))
+# 0x88c6b48354c367125a59792a93a7b60ad7cd66e516157dbba16558c68a46d3cb
+```
+
+### L2ToL2CrossDomainMessenger Deployment
+<!-- Generated with: ./scripts/run_gen_predeploy_docs.sh --optimism-repo-path ../optimism --fork-name Interop --contract-name L2ToL2CrossDomainMessenger --from-address 0x4220000000000000000000000000000000000001 --from-address-nonce 0 --git-commit-hash 71c460ec7c7c05791ddd841b97bcb664a1f0c753 --eth-rpc-url https://optimism.rpc.subquery.network/public --proxy-address 0x4200000000000000000000000000000000000023 --copy-contract-bytecode true -->
+
+The `L2ToL2CrossDomainMessenger` contract is upgraded.
```
tynes: nit: not upgrade

specs/interop/derivation.md
```diff
@@ -87,17 +91,190 @@ The interop network upgrade timestamp defines the timestamp at which all functio
 the consensus rules for an OP Stack based network. On the interop network upgrade block, a set of deposit transaction
 based upgrade transactions are deterministically generated by the derivation pipeline in the following order:
 
-- L1 Attributes Transaction calling `setL1BlockValuesIsthmus`
```
teddyknox, line 90: @tynes should we update [L1Block.sol](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L2/L1Block.sol) to have a `setL1BlockValuesInterop()` method and update the specs here?

emhane:
@ethereum-optimism/security-reviewers this pr is high priority to review cc @yoonchung8822

specs/interop/derivation.md
tynes: We no longer need `setL1BlockValuesInterop` given modifications to the design, so any mentions of that should be deleted from the specs

specs/interop/derivation.md
```diff
@@ -87,17 +91,185 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+<!-- To regenerate the deploy/upgrade tx docs below, run `./scripts/upgrades/gen_interop_upgrade_tx_specs.sh | pbcopy` and then paste here -->
+
+### CrossL2Inbox Deployment
+<!-- Generated with: ./scripts/run_gen_predeploy_docs.sh --optimism-repo-path ../optimism --fork-name Interop --contract-name CrossL2Inbox --from-address 0x4220000000000000000000000000000000000000 --from-address-nonce 0 --git-commit-hash 71c460ec7c7c05791ddd841b97bcb664a1f0c753 --eth-rpc-url https://optimism.rpc.subquery.network/public --proxy-address 0x4200000000000000000000000000000000000022 --copy-contract-bytecode true -->
+
+The `CrossL2Inbox` contract is deployed.
+
+A deposit transaction is derived with the following attributes:
+
+- `from`: `0x4220000000000000000000000000000000000000`
```
tynes, line 119: It is important that the `from` for the deployment has not been used yet so that we can compute the address correctly (or if it has been used, then we use the correct nonce, but generally we have always used a fresh `from` each time)

tynes (approved):
I didn't validate everything but can confirm this is the scope that we want. Will leave it to the implementers to fully validate with tests to ensure its correctness

specs/interop/derivation.md
```diff
@@ -87,17 +91,185 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
-  - Update L2ToL2CrossDomainMessenger Proxy ERC-1967 Implementation Slot
-  - Update the OptimismMintableERC20Factory Proxy ERC-1967 Implementation Slot
-
-The execution payload MUST set `noTxPool` to `true` for this block.
+The execution payload MUST set `noTxPool` to `true` for this block, excluding user transactions from the block.
+
+This upgrade requires `3_320_000` in gas. In order to upgrade successfully the following invariant MUST be maintained:
+
+```plaintext
+MaxResourceLimit + SystemTxMaxGas + UpgradeGasUsage ≤ L2GasLimit
```
Ethnical, line 99: Interesting, newbie question:
Just to make sure how do we deal with the `Deposit transaction`? 
My attacking scenario would be: 
Is a DoS possible to make a massive deposit transaction on L1, and break the invariants higher since the deposit transaction seems to not be included here? Thus to be bigger than the `L2GasLimit` and make the upgrade fail? 
Or the deposit transaction are excluded here somehow? 
because for me: "The execution payload MUST set `noTxPool` to `true`" should only affect L2 geth transaction but not the deposit one.

specs/interop/derivation.md
```diff
@@ -87,17 +91,185 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+
+A deposit transaction is derived with the following attributes:
+
+- `from`: `0x4220000000000000000000000000000000000001`
+- `to`: `null`
+- `mint`: `0`
+- `value`: `0`
+- `nonce`: `0`
+- `gasLimit`: `1100000`
+- `data`: `0x0x6080604052348015600e575f80fd5b...` ([full bytecode](../static/bytecode/interop-l2-to-l2-cross-domain-messenger-deployment.txt))
```
Ethnical: Looks a typo here the `0x` for the data: 
```suggestion
- `data`: `0x6080604052348015600e575f80fd5b...` ([full bytecode](../static/bytecode/interop-l2-to-l2-cross-domain-messenger-deployment.txt))
```

specs/interop/derivation.md
```diff
@@ -87,17 +91,185 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+
+A deposit transaction is derived with the following attributes:
+
+- `from`: `0x4220000000000000000000000000000000000000`
+- `to`: `null`
+- `mint`: `0`
+- `value`: `0`
+- `nonce`: `0`
+- `gasLimit`: `420000`
+- `data`: `0x0x6080604052348015600e575f80fd5b...` ([full bytecode](../static/bytecode/interop-cross-l2-inbox-deployment.txt))
```
Ethnical: Looks a typo here the `0x` for the data:
```suggestion
- `data`: `0x6080604052348015600e575f80fd5b...` ([full bytecode](../static/bytecode/interop-cross-l2-inbox-deployment.txt))
```

specs/interop/derivation.md
```diff
@@ -87,17 +91,185 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+
+```bash
+cast concat-hex $(cast sig "upgradeTo(address)") $(cast abi-encode "upgradeTo(address)" 0x0D0eDd0ebd0e94d218670a8De867Eb5C4d37cadD)
+# 0x3659cfe60000000000000000000000004200000000000000000000000000000000000023
+```
+
+Verify `sourceHash`:
+
+```bash
+cast keccak $(cast concat-hex 0x0000000000000000000000000000000000000000000000000000000000000002 $(cast keccak "Interop: L2ToL2CrossDomainMessenger Proxy Update"))
```
Ethnical, line 270: I forgot why we need to have the concat-hex with `0x2`? This is the version of something each time? 
```
cast concat-hex 0x0000000000000000000000000000000000000000000000000000000000000002
```

specs/interop/derivation.md
```diff
@@ -87,17 +91,185 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+This transaction updates the L2ToL2CrossDomainMessenger Proxy ERC-1967
+implementation slot to point to the new L2ToL2CrossDomainMessenger deployment.
+
+A deposit transaction is derived with the following attributes:
+
+- `from`: `0x0000000000000000000000000000000000000000`
+- `to`: `0x4200000000000000000000000000000000000023` (L2ToL2CrossDomainMessenger Proxy)
+- `mint`: `0`
+- `value`: `0`
+- `gasLimit`: `50,000`
```
Ethnical, line 255: I would like to know how do we verify that the `gasLimit` is correct each time? 
I suppose we have tests that perform the upgrade and then we copy the `gasLimit`? If yes, where are the tests to make sure the `gasLimit` is copied correctly for each: 
```
- `gasLimit`: `50,000`
- `gasLimit`: `1100000`
- `gasLimit`: `50,000`
- `gasLimit`: `420000`
```

Ethnical (commented):
I don't know who is the DRI but at least few comments to help us to understand more here.

specs/interop/derivation.md
```diff
@@ -87,17 +91,185 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+- `gasLimit`: `420000`
+- `data`: `0x0x6080604052348015600e575f80fd5b...` ([full bytecode](../static/bytecode/interop-cross-l2-inbox-deployment.txt))
+- `sourceHash`: `0x6e5e214f73143df8fe6f6054a3ed7eb472d373376458a9c8aecdf23475beb616`,  
+  computed with the "Upgrade-deposited" type, with `intent = "Interop: CrossL2Inbox Deployment"`
+
+This results in the Interop CrossL2Inbox contract being deployed to
+`0x691300f512e48B463C2617b34Eef1A9f82EE7dBf`, to verify:
+
+```bash
+cast compute-address --nonce=0 0x4220000000000000000000000000000000000000
```
Ethnical, line 133: NIT: Here, how do we ensure that the nonce has not being used in the past already? 
I assume before upgrading the contract we perform a check to know that no collision are possible?

tynes:
I double checked the commit used is good

specs/interop/derivation.md
tynes: You can grab the nonce from an isthmus supported network to know because the address has no known private key and is impersonated only during an upgrade

specs/interop/derivation.md
tynes: There are tests in the source code that assert the post state of the txs and would catch the revert

specs/interop/derivation.md
tynes: This is a way to namespace so there is no hash collision accidentally

specs/interop/derivation.md
tynes: The sum of gas used by deposits cannot be larger than the L2 block gas limit. The L1 contracts don't know when network upgrade transactions (issued as deposits) are going to be inserted. We rely on L1 contracts to prevent deposits from using too much L2 gas. There are protocol solutions to fix this (issue network upgrade txs over multiple blocks) but the most simple is to say that the L2 block gas limit must be larger than a certain amount for the upgrade to happen successfully. As long as that is true, then there is no way to dos. Its up to the chain operator to ensure that the gas limit is high enough, and its not an unreasonable gas limit

tynes: The sum of gas used by deposits cannot be larger than the L2 block gas limit. The L1 contracts don't know when network upgrade transactions (issued as deposits) are going to be inserted. We rely on L1 contracts to prevent deposits from using too much L2 gas. There are protocol solutions to fix this (issue network upgrade txs over multiple blocks) but the most simple is to say that the L2 block gas limit must be larger than a certain amount for the upgrade to happen successfully. As long as that is true, then there is no way to dos. Its up to the chain operator to ensure that the gas limit is high enough, and its not an unreasonable gas limit

tynes: The sum of gas used by deposits cannot be larger than the L2 block gas limit. The L1 contracts don't know when network upgrade transactions (issued as deposits) are going to be inserted. We rely on L1 contracts to prevent deposits from using too much L2 gas. There are protocol solutions to fix this (issue network upgrade txs over multiple blocks) but the most simple is to say that the L2 block gas limit must be larger than a certain amount for the upgrade to happen successfully. As long as that is true, then there is no way to dos. Its up to the chain operator to ensure that the gas limit is high enough, and its not an unreasonable gas limit

specs/interop/derivation.md
teddyknox: That is the case here I believe.

specs/interop/derivation.md
sebastianst: @Ethnical this is the exact same situation as for Isthmus' deposit transactions - we need to tell chain operators to have a gas limit of at least 25M.

specs/interop/derivation.md
sebastianst: Do I see correctly that you took the last used deployer address from Isthmus, `0x4210000000000000000000000000000000000005`, as a starting point, then bumped the `1` to `2` & reset the `5` to `0`? And then the same address `+1` for the next deployment?
This seems valid to me. We don't have a clear pattern defined to use for deployment addresses yet, so this seems as good as any.

specs/interop/derivation.md
```diff
@@ -87,17 +91,185 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+implementation slot to point to the new CrossL2Inbox deployment.
+
+A deposit transaction is derived with the following attributes:
+
+- `from`: `0x0000000000000000000000000000000000000000`
+- `to`: `0x4200000000000000000000000000000000000022` (CrossL2Inbox Proxy)
+- `mint`: `0`
+- `value`: `0`
+- `gasLimit`: `50,000`
+- `data`: `0x3659cfe60000000000000000000000004200000000000000000000000000000000000022`
```
sebastianst: This is using as argument `0x4200000000000000000000000000000000000022`, _not_ `0x691300f512e48B463C2617b34Eef1A9f82EE7dBf`

specs/interop/derivation.md
```diff
@@ -87,17 +91,185 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+- `sourceHash`: `0x88c6b48354c367125a59792a93a7b60ad7cd66e516157dbba16558c68a46d3cb`
+  computed with the "Upgrade-deposited" type, with `intent = "Interop: CrossL2Inbox Proxy Update"`
+
+Verify data:
+
+```bash
+cast concat-hex $(cast sig "upgradeTo(address)") $(cast abi-encode "upgradeTo(address)" 0x691300f512e48B463C2617b34Eef1A9f82EE7dBf)
+# 0x3659cfe60000000000000000000000004200000000000000000000000000000000000022
```
sebastianst: This gives `0x3659cfe6000000000000000000000000691300f512e48b463c2617b34eef1a9f82ee7dbf`, seems this was called with argument `4200000000000000000000000000000000000022` instead.

specs/interop/derivation.md
```diff
@@ -87,17 +91,185 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+implementation slot to point to the new L2ToL2CrossDomainMessenger deployment.
+
+A deposit transaction is derived with the following attributes:
+
+- `from`: `0x0000000000000000000000000000000000000000`
+- `to`: `0x4200000000000000000000000000000000000023` (L2ToL2CrossDomainMessenger Proxy)
+- `mint`: `0`
+- `value`: `0`
+- `gasLimit`: `50,000`
+- `data`: `0x3659cfe60000000000000000000000004200000000000000000000000000000000000023`
```
sebastianst: Same error, not argument `0x0D0eDd0ebd0e94d218670a8De867Eb5C4d37cadD`

specs/interop/derivation.md
```diff
@@ -87,17 +91,185 @@ The interop network upgrade timestamp defines the timestamp at which all functio
...
+- `sourceHash`: `0xe54b4d06bbcc857f41ae00e89d820339ac5ce0034aac722c817b2873e03a7e68`
+  computed with the "Upgrade-deposited" type, with `intent = "Interop: L2ToL2CrossDomainMessenger Proxy Update"`
+
+Verify data:
+
+```bash
+cast concat-hex $(cast sig "upgradeTo(address)") $(cast abi-encode "upgradeTo(address)" 0x0D0eDd0ebd0e94d218670a8De867Eb5C4d37cadD)
+# 0x3659cfe60000000000000000000000004200000000000000000000000000000000000023
```
sebastianst: Gives `0x3659cfe60000000000000000000000000d0edd0ebd0e94d218670a8de867eb5c4d37cadd`

sebastianst (commented):
I think I found some issues with the proxy `updateTo` txs

sebastianst (commented):
Otherwise, specs lgtm. There's only so much we can review statically, we need to have e2e test of the transition block.

specs/interop/derivation.md
teddyknox: Good catch. I've updated the PR and the script to correct the bug.

tynes:
There is consensus on using a git commit rather than a git tag for the contract bytecode given that is what we have done historically. After standard L2 genesis we are able to have a more reliable L2 predeploys release method

specs/interop/derivation.md
sebastianst: Interop uses deployment addresses `0x422000000000000000000000000000000000000N` with `N` being `0` or `1` for the two deployments. Isthmus used `0x421000000000000000000000000000000000000N`

sebastianst (approved):
This is good to merge as a tentative spec of the upgrade txs. We will need to adapt the transaction payloads if we need to change the contracts for some reason.

PR #673, opened 2025-04-22
Add script for generating upgrade tx specs

teddyknox:
**Description**

Adds a set of scripts that make it easy to generate specifications for upgrade/proxy update transactions.
**Tests**

I manually compared the result of this script to previous specifications for upgrade/proxy update transactions, given the same input.

**Additional context**

Used to generate #672.


emhane (commented):
can we have a workflow for manually dispatching this script in ci pls?

teddyknox:
@emhane as a test? Or what would be the purpose?

emhane:
> @emhane as a test? Or what would be the purpose?

for creating new specs when needed and, yes, for not relying on manual test to check its validity

scripts/gen_predeploy_docs.py
```diff
@@ -0,0 +1,448 @@
...
+            "git_commit_hash": args.git_commit_hash,
+            "contract_code_hash": contract_code_hash,
+            "source_hash": source_hash,
+            "deployed_address": deployed_address,
+            "command": "./scripts/run_gen_predeploy_docs.sh " + ' '.join(sys.argv[1:]),
+            "forge_artifact_path_data": forge_artifact_path_val,
+        }
+        if args.proxy_address:
+            proxy_intent = f"{args.fork_name}: {args.contract_name} Proxy Update"
+            proxy_data = compute_proxy_update_data(args.proxy_address)
```
sebastianst: need to pass deployed address instead?

tynes:
I would love to merge this but ideally it uses `uv` instead of `pip`

scripts/gen_predeploy_docs.py
```diff
@@ -0,0 +1,447 @@
...
+### {{ params.contract_name }} Proxy Update
+
+This transaction updates the {{ params.contract_name }} Proxy ERC-1967
+implementation slot to point to the new {{ params.contract_name }} deployment.
+
+A deposit transaction is derived with the following attributes:
+
+- `from`: `0x0000000000000000000000000000000000000000`
+- `to`: `{{ params.proxy_address }}` ({{ params.contract_name }} Proxy)
+- `mint`: `0`
```
geoknee, line 71: Any reason we don't print the `nonce` here?

teddyknox: I'm omitting the nonce in the proxy update section because it’s a deposit call from 0x000…000, not a CREATE. The Bedrock deposit transaction format doesn’t carry an account nonce, and nonce isn’t needed to derive anything there.

teddyknox:
@geoknee I've updated it to use `uv`.

scripts/gen_predeploy_docs.py
```diff
@@ -0,0 +1,447 @@
+import argparse
+import subprocess
+import sys
+import os
+import re
+import json
+
+
+# Questions:
+# - Should proxy update transactions originate from the same address as the contract deployment?
```
geoknee:
```suggestion
```
I think not, they are always from the zero address.

scripts/upgrades/gen_interop_upgrade_tx_specs.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash -e
```
geoknee, line 1: I think we need to add to the top level readme to instruct specs authors on how to use this tool. For example, they should copy this file (or some other template we provide) and make appropriate changes before running it.

scripts/upgrades/gen_interop_upgrade_tx_specs.sh
```diff
@@ -0,0 +1,38 @@
...
+
+GIT_COMMIT_HASH=71c460ec7c7c05791ddd841b97bcb664a1f0c753
+FROM_ADDRESS_NONCE=0
+FROM_ADDRESS=0x4220000000000000000000000000000000000000
```
geoknee, line 6: It would be great if there was some way we could ensure that this address has not yet been used, or at least that it's nonce is equal to the provided nonce. Otherwise we may have specs bugs and upgrade transactions will (I think) potentially revert. 

We apparently have a _convention_ that each deployment will come from a unique address with a zero nonce. This convention guarantees the upgrade transaction does not revert, so if we check the convention is adhered to we should be fine. 

In lieu of automation around this, we can instruct the user to choose an unused address or suggest they just increment the previous address which was used (going in order through the hardforks).

scripts/upgrades/gen_interop_upgrade_tx_specs.sh
```diff
@@ -0,0 +1,38 @@
...
+FORK_NAME=Interop
+
+# Array of contract names and their corresponding proxy addresses
+declare -a contracts=(
+    "CrossL2Inbox:0x4200000000000000000000000000000000000022"
+    "L2ToL2CrossDomainMessenger:0x4200000000000000000000000000000000000023"
+)
```
geoknee, line 49: Would be cool to provide a template or similar, such that the user does not need to go and look up a predeploy address. We could add all predeploys and comment them out, and the user just needs to uncomment the appropriate contracts.

scripts/upgrades/gen_interop_upgrade_tx_specs.sh
geoknee: Some automation would also help to catch typos / validate this input data.

scripts/gen_predeploy_docs.py
```diff
@@ -0,0 +1,447 @@
...
+
+SOURCE_HASH_PREFIX = "0x0000000000000000000000000000000000000000000000000000000000000002"
+
+JINJA_TEMPLATE = """### {{ params.contract_name }} Deployment
+<!-- Generated with: {{ params.command }} -->
```
geoknee, line 11: Super nit, but this tool generates MD that upsets the linter in this repo. In a few places, but particularly here where the line length is really long. Be nice if we can smooth that over somehow. Perhaps we just tell the linter to not care about comments.

scripts/upgrades/gen_interop_upgrade_tx_specs.sh
geoknee: This is non blocking and could be done in a followup.

geoknee: As a case in point, early use of this tool https://github.com/ethereum-optimism/specs/pull/778 actually reused a deployer address. Some automation could have save us!

geoknee:
@teddyknox check out https://github.com/ethereum-optimism/specs/pull/778 for a few more pointers.

scripts/gen_predeploy_docs.py
teddyknox: Removed the question.

scripts/gen_predeploy_docs.py
teddyknox: I updated the command string to be broken between args over many lines, hopefully this fixes the linter error.

scripts/upgrades/gen_interop_upgrade_tx_specs.sh
teddyknox: Done!

scripts/upgrades/gen_interop_upgrade_tx_specs.sh
teddyknox: Agreed that some automation would be nice here. I'll add a comment in the python script noting that we want to do this.

scripts/upgrades/gen_interop_upgrade_tx_specs.sh
teddyknox: Done

geoknee (approved):

PR #676, opened 2025-04-23
isthmus: clarify genesis withdrawalsRoot value and block topic selection

geoknee:
See https://github.com/ethereum-optimism/optimism/issues/15458

### DIFF
--- a/specs/protocol/isthmus/exec-engine.md
+++ b/specs/protocol/isthmus/exec-engine.md
@@ -2,7 +2,6 @@
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
-
 **Table of Contents**
 
 - [Overview](#overview)
--- a/specs/protocol/isthmus/exec-engine.md
+++ b/specs/protocol/isthmus/exec-engine.md
@@ -123,7 +123,8 @@ even if a `withdrawalsRoot` is present and a MPT root is present in the header, 
 Any implementation that calculates output root should be careful not to use the header `withdrawalsRoot`.
 
 After Isthmus activation, if there was never any withdrawal contract storage, a MPT root of an empty list
-can be set as the `withdrawalsRoot`
+can be set as the `withdrawalsRoot`. Note that by default there _is_ nonzero storage in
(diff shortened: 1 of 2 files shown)


specs/protocol/isthmus/exec-engine.md
```diff
@@ -123,7 +123,8 @@ even if a `withdrawalsRoot` is present and a MPT root is present in the header,
 Any implementation that calculates output root should be careful not to use the header `withdrawalsRoot`.
 
 After Isthmus activation, if there was never any withdrawal contract storage, a MPT root of an empty list
-can be set as the `withdrawalsRoot`
+can be set as the `withdrawalsRoot`. Note that by default there _is_ nonzero storage in
+the [`L2ToL1MessagePasser`][l2-to-l1-mp] at genesis, so the `withdrawalsRoot` will not be the MPT root of an empty list.
```
sebastianst: Can't we strengthen this wording to say that it's impossible to have an empty storage root because the `L2ToL1MessagePasser` is guaranteed to have storage because it stores the proxy owner and the implementation address? This is not optional, it is guaranteed by upgrade transactions that are executed as part of consensus.

geoknee: Yes I updated the wording, PTAL. I don't know that it has to do with upgrade transactions though?

specs/protocol/isthmus/exec-engine.md
```diff
@@ -122,8 +122,10 @@ they are not reflected in the `withdrawalsRoot`. Hence, prior to Isthmus activat
 even if a `withdrawalsRoot` is present and a MPT root is present in the header, it should not be used.
 Any implementation that calculates output root should be careful not to use the header `withdrawalsRoot`.
 
-After Isthmus activation, if there was never any withdrawal contract storage, a MPT root of an empty list
-can be set as the `withdrawalsRoot`
+Note that there is always nonzero storage in the [`L2ToL1MessagePasser`][l2-to-l1-mp],
+because it is a [proxied predeploy](../../protocol/predeploys.md) -- from genesis it
+stores an implementation address and owner address. So from Isthmus,
+the `withdrawalsRoot` will not always be non-nil and never be the MPT root of an empty list.
```
sebastianst:
```suggestion
the `withdrawalsRoot` will always be non-nil and never be the MPT root of an empty list.
```
?

sebastianst (approved):
lgtm, but there's probably one typo to fix

PR #681, opened 2025-04-24
chore: Rework security section for the expiry window

clabby:
## Overview

Adds the security section back in for the expiry window, noting why the expiry window exists and a link to the design doc for replaying messages that have expired.

### DIFF
--- a/specs/interop/derivation.md
+++ b/specs/interop/derivation.md
@@ -13,6 +13,7 @@
 - [Expiry Window](#expiry-window)
 - [Security Considerations](#security-considerations)
   - [Depositing an Executing Message](#depositing-an-executing-message)
+  - [Expiry Window](#expiry-window-1)
   - [Reliance on History](#reliance-on-history)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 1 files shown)


specs/interop/derivation.md
```diff
@@ -115,6 +116,14 @@ It is possible to deposit an invalid executing message, forcing the sequencer to
...
 instability. A future upgrade will enable deposits to trigger executing messages.
 
+### Expiry Window
+
+The expiry window ensures that the proof can execute in a reasonable amount of time.
+[`EIP-2935`](https://eips.ethereum.org/EIPS/eip-2935) introduced the capability to traverse history with sub-linear
+complexity, however deep lookups remain expensive. App developers and users, in the event that they encounter a message
+that has expired but has yet to be relayed, can [resend the message](https://github.com/ethereum-optimism/design-docs/blob/25ef5537e39b63cddf1c83479cee9f0e02431dce/protocol/resend-messages.md#L4)
+in order to complete the process.
+
```
emhane:
```suggestion
### Expiry Window

The expiry window ensures that the proof can execute in a reasonable amount of time. [`EIP-2935`][eip-2935] introduced the capability to traverse history with sub-linear complexity, however deep lookups remain expensive. App developers and users, in the event that they encounter a message that has expired but has yet to be relayed, can [resend the message][resend-messages] in order to complete the process.

[eip-2935]: https://eips.ethereum.org/EIPS/eip-2935
[resend-messages]: https://github.com/ethereum-optimism/design-docs/blob/25ef5537e39b63cddf1c83479cee9f0e02431dce/protocol/resend-messages.md#L4
```
perhaps links should go at end of page along side the `eip-4444` link, think that's what we usually do

emhane (changes requested):

emhane (approved):

PR #684, opened 2025-04-24
interop: supervisor_dependencySet API

protolambda:
**Description**

See https://github.com/ethereum-optimism/optimism/issues/15048

### DIFF
--- a/specs/interop/supervisor.md
+++ b/specs/interop/supervisor.md
@@ -23,7 +23,7 @@
     - [`SupervisorChainSyncStatus`](#supervisorchainsyncstatus)
     - [`SuperRootResponse`](#superrootresponse)
     - [`SafetyLevel`](#safetylevel)
-    - [`DependencySetConfig`](#dependencysetconfig)
+    - [`DependencySetConfigV1`](#dependencysetconfigv1)
   - [Methods](#methods)
     - [`supervisor_crossDerivedToSource`](#supervisor_crossderivedtosource)
     - [`supervisor_localUnsafe`](#supervisor_localunsafe)
@@ -32,7 +32,7 @@
     - [`supervisor_finalizedL1`](#supervisor_finalizedl1)
     - [`supervisor_superRootAtTimestamp`](#supervisor_superrootattimestamp)
     - [`supervisor_syncStatus`](#supervisor_syncstatus)
-    - [`supervisor_dependencySet`](#supervisor_dependencyset)
+    - [`supervisor_dependencySetV1`](#supervisor_dependencysetv1)
     - [`supervisor_allSafeDerivedAt`](#supervisor_allsafederivedat)
     - [`supervisor_checkAccessList`](#supervisor_checkaccesslist)
       - [Access-list contents](#access-list-contents)
@@ -207,10 +207,12 @@ Corresponds to a verifier [SafetyLevel](./verifier.md#safety).
 - `safe`: matching cross-safe, named `safe` to match the RPC label.
 - `finalized`
 
-#### `DependencySetConfig`
+#### `DependencySetConfigV1`
 
 Defines an interop dependency set.
 Includes information to determine the active set of chains at any particular block time.
+
+This type is versioned: the format of this configuration may change over time.
 
 `OBJECT`:
 - `dependencies`: `MAP`:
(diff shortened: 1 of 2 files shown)


specs/interop/supervisor.md
```diff
@@ -181,6 +184,30 @@ Corresponds to a verifier [SafetyLevel](./verifier.md#safety).
...
+Includes information to determine the active set of chains at any particular block time.
+
+`OBJECT`:
+- `dependencies`: `MAP`:
+  - key: `ChainID`
+  - value: `ConfigDependency`: `OBJECT`:
+    - `chainIndex`: `ChainIndex`: the unique short identifier of this chain.
+    - `activationTime`: `uint64`: when the chain becomes part of the dependency set.
+      This is the minimum timestamp of the inclusion of an executing message.
+    - `historyMinTime`: `uint64`: the lower bound of data to store.
```
emhane:
```suggestion
    - `historyMinTime`: `uint64`: the lower bound of duration to store data.
```

specs/interop/supervisor.md
```diff
@@ -181,6 +184,30 @@ Corresponds to a verifier [SafetyLevel](./verifier.md#safety).
...
+      This is the minimum timestamp of an initiating message to be accessible to others.
+      This is set to 0 when all data since genesis is executable.
```
emhane: assume we will never want to store no data at all then

specs/interop/supervisor.md
```diff
@@ -234,6 +261,16 @@ Parameters: (none)
 
 Returns: `SupervisorSyncStatus`
 
+#### `supervisor_dependencySet`
+
+Retrieves the full dependency set configuration,
+to determine which chains are available at any particular block.
+The returned configuration may change with new chain upgrades. 
+
+Parameters: (none)
```
emhane, line 282: I would have instinctively though this should take a block number as parameter ..

specs/interop/supervisor.md
```diff
@@ -181,6 +184,30 @@ Corresponds to a verifier [SafetyLevel](./verifier.md#safety).
...
 - `finalized`
 
+#### `ChainIndex`
+
+`STRING`:
+Decimal number, representing a `uint32`.
+
+A unique short identifier for a chain, within the dependency set.
+
+#### `DependencySetConfig`
```
tynes: Why are we exposing the config format here? We are exposing implementation details of the db (chain index + history min time) as well as the message expiry override time? What would the most useful API be from first principles?

I would guess its something like:

```js
{
  timestamp: { expiryWindow: x, chains: [chainid, chainid] },
  timestamp2: { expiryWindow: y, chains: [chainid, chainid, chainid] }
}
```

protolambda: This PR is old. The dependency set approach has changed a lot since. I will likely just replace it with an endpoint to get the list of chain IDs in the dependency set.

emhane:
@dhyaniarun1993 @itschaindev @sadiq1971

protolambda:
For readers: this PR is not accurate, the format changed last 2 weeks. Will need to update this after other higher priority work is out of the way.

specs/interop/supervisor.md
protolambda: Update: this was simplified

specs/interop/supervisor.md
protolambda: Based on slack discussion I think everyone is in rough consensus that the whole config type like we have today, as-is, is the simplest. The config does not contain any hardfork times anymore, so it's pretty minimal, and it's dumbed down enough that there's no block-number/timestamp argument needed.

protolambda:
Rebased and updated to match https://github.com/ethereum-optimism/optimism/pull/16156

specs/interop/supervisor.md
```diff
@@ -204,6 +207,17 @@ Corresponds to a verifier [SafetyLevel](./verifier.md#safety).
...
+
+Defines an interop dependency set.
+Includes information to determine the active set of chains at any particular block time.
+
+`OBJECT`:
+- `dependencies`: `MAP`:
+  - key: `ChainID`
+  - value: `ConfigDependency`: `OBJECT`: empty object. This may be extended in the future.
+- `overrideMessageExpiryWindow`: `uint64`: changes the default protocol expiry window time.
```
sebastianst, line 221: So this isn't something we only use in tests, but want to have in the specs?

sebastianst (approved):

specs/interop/supervisor.md
protolambda: I already suggested to change it to just be `messageExpiryWindow`, but there is no support/momentum on all the usages of this config code to change it.

specs/interop/supervisor.md
```diff
@@ -121,7 +123,8 @@ Regular JSON number, always integer. Assumed to always fit in 51 bits.
 #### `ChainID`
 
 `STRING`:
-Hex-encoded big-endian number, variable length up to 256 bits, prefixed with `0x`.
+Decimal-encoded string, variable length up to 256 bits.
+This supports decoding of hex-encoded values when prefixed with `0x`.
```
emhane, line 127: @sadiq1971

PR #689, opened 2025-04-30
chore: Remove custom gas token

clabby:
## Overview

Removes custom gas token from the spec. This feature is no longer being worked on as a part of the standard configuration.


maurelian:
I have no objection to this, but will hold off on approving for now in case others wish to opine

mslipper (approved):
@tynes I'm going to approve and merge this since you're working on a new version of the CGT spec.

PR #690, opened 2025-04-30
chore: Remove Alt DA

clabby:
## Overview

Removes Alt DA from the spec. This feature is no longer being worked on as a part of the standard configuration.


maurelian:
My understanding from recent conversations with @tessr is that we definitely need to keep this around and maintained for a possible future when it may be more of a priority

tynes:
I do not think we should merge this, its still in the code and teams rely on it

PR #691, opened 2025-04-30
interop: messages may only be initiated AFTER the activation time.

ghost:
Previously messages could be initiated >= activation but now it's strictly greater than.

This is changed in the code in the WIP PR https://github.com/ethereum-optimism/optimism/pull/15614

### DIFF
--- a/specs/interop/messaging.md
+++ b/specs/interop/messaging.md
@@ -122,17 +122,18 @@ do not count as executing messages.
 ## Messaging Invariants
 
 - [Timestamp Invariant](#timestamp-invariant): The timestamp at the time of inclusion of the initiating message MUST
-  be less than or equal to the timestamp of the executing message as well as greater than or equal to the Interop Start Timestamp.
+  be less than or equal to the timestamp of the executing message as well as greater than the Interop Start Timestamp.
 - [ChainID Invariant](#chainid-invariant): The chain id of the initiating message MUST be in the dependency set
 - [Message Expiry Invariant](#message-expiry-invariant): The timestamp at the time of inclusion of the executing
   message MUST be lower than the initiating message timestamp (as defined in the [`Identifier`]) + `EXPIRY_TIME`.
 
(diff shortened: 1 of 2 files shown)


tynes (approved):

specs/interop/messaging.md
```diff
@@ -122,17 +122,18 @@ do not count as executing messages.
...
 
-The timestamp invariant ensures that initiating messages have at least the same timestamp as the Interop upgrade timestamp
-and cannot come from a future block than the block of its executing message. Note that since
-all transactions in a block have the same timestamp, it is possible for an executing transaction to be
-ordered before the initiating message in the same block.
+The timestamp invariant ensures that initiating messages have a timestamp greater than the Interop upgrade timestamp
+and cannot come from a future block than the block of its executing message. This means that messages can only be
+initiated in blocks that come after the block with the Interop upgrade timestamp. Messages in the block with the
+Interop upgrade timestamp itself cannot be initiated. Note that since all transactions in a block have the same
+timestamp, it is possible for an executing transaction to be ordered before the initiating message in the same block.
```
protolambda:
```suggestion
The timestamp invariant ensures that initiating messages have a timestamp greater than the Interop upgrade timestamp
and cannot come from a future block than the block of its executing message.

This means that messages can only be initiated in blocks that come after the block with the Interop upgrade timestamp.
Contract log events in the block with the upgrade timestamp are thus not valid initiating messages.

This same block with the upgrade timestamp only includes deposit-type transactions
(from the system, and possibly from L1): this block can thus not include executing messages,
even if only executing the initiating messages of previously Interop-activated chains.

Note that since all transactions in a block have the same timestamp, it is possible for an executing transaction
to be ordered before the initiating message in the same block.
However, cyclic message dependencies are not allowed and
this is verified with rules complementary to the timestamp invariant.
```

protolambda (approved):

PR #692, opened 2025-04-30
feat(interop): Transition block sequencer + derivation rules

clabby:
## Overview

Further specifies the rules for the interop transition block that the sequencer + derivation pipeline must adhere to.

### Rationale

The current spec only notes that the sequencer must define `no_tx_pool = true` when building the activation block for the interop hardfork. However, this is just policy, and no where in the protocol currently enforces this. This change adds an extra clause to derivation where, if a fork transition block >= interop contains user space transactions, the batch is to be dropped per Holocene rules.


teddyknox (approved):
Thanks for the quick turnaround.

sebastianst (approved):
LGTM, thanks!

This is also similar to the text that we almost added for the same rule for Isthmus in https://github.com/ethereum-optimism/specs/pull/616

specs/interop/derivation.md
```diff
@@ -43,6 +44,17 @@ because it includes invalid executing messages MUST be replaced by a deposits on
...
+## Omit User Transactions in Fork Activation Block
+
+With the interop hardfork, fork activation blocks no longer include any user transactions. Sequencers, when building
+the fork activation block, MUST set `noTxPool` to `true` in the execution payload attributes for this block, instructing
+the builder to exclude user transactions.
+
+The derivation pipeline MUST enforce that the sequencer has not included any user transactions in the batch covering
+the hardfork's activation block. If the sequencer does include any user transactions within the hardfork activation
+block, that batch, and the remaining span batch it originated from, MUST be dropped per the
+[rules defined in the holocene hardfork](../protocol/holocene/derivation.md).
```
protolambda, line 56:
```suggestion
block, that batch, and the remaining span batch it originated from, MUST be dropped following the
batch-dropping behavior introduced in the [Holocene hardfork](../protocol/holocene/derivation.md#span-batches).
```

protolambda (approved):

specs/interop/derivation.md
protolambda: Moved to: https://github.com/ethereum-optimism/specs/pull/693

PR #694, opened 2025-05-01
interop: Add section describing the properties of the interop Activat…

ghost:
Currently the specs have various rules and constraints on what the first block after/at Interop activation must adhere to. This change creates a section for this Activation Block that collates the various rules.


specs/interop/derivation.md
```diff
@@ -44,11 +45,23 @@ because it includes invalid executing messages MUST be replaced by a deposits on
 block height. This guarantees progression of the chain, ensuring that an infinite loop of processing
 the same block in the proof system is not possible.
 
+## Activation Block
+
+The activation block is the block that has a timestamp equal to the interop network upgrade timestamp.
```
teddyknox: IIUC it could also be the first block that is greater than the network upgrade timestamp.

specs/interop/derivation.md
```diff
@@ -44,11 +45,23 @@ because it includes invalid executing messages MUST be replaced by a deposits on
...
+This block has several special properties and constraints:
+
+- It MUST NOT include any user transactions and the execution payload MUST set `noTxPool` to `true`.
+- Messages MAY NOT be initiated or executed in this block.
```
teddyknox: I think messages may not be initiated (though things will still work if they are), but must not be executed. cc @protolambda

Though strictly speaking this line may not be necessary given the above point and given we enumerate the exact transactions in the block lower in this document and they do not initiate or execute messages.

specs/interop/derivation.md
```diff
@@ -44,11 +45,23 @@ because it includes invalid executing messages MUST be replaced by a deposits on
...
+- The derivation pipeline MUST enforce that the sequencer has not included any user transactions in
+the batch covering the upgrade's activation block. If the sequencer includes any user transactions
```
teddyknox: > The derivation pipeline MUST enforce that the sequencer has not included any user transactions in **the batch** covering the upgrade's activation block.

Maybe I misunderstand but I thought that user transactions were disabled for the upgrade block only, not all other blocks within the upgrade block's batch? Makes sense however that if the upgrade block does contain user transactions the batch and remaining span batch must be dropped.

specs/interop/derivation.md
```diff
@@ -44,11 +45,23 @@ because it includes invalid executing messages MUST be replaced by a deposits on
...
+within the activation block, that batch and the remaining span batch it originated from MUST be dropped.
+
 ## Omit User Transactions in Fork Activation Block
 
-With the interop upgrade, fork activation blocks no longer include any user transactions. Sequencers, when building
-the fork activation block, MUST set `noTxPool` to `true` in the execution payload attributes for this block, instructing
-the builder to exclude user transactions.
+With the interop upgrade, fork activation blocks no longer include any user transactions, as specified in the
```
teddyknox: > With the interop upgrade, fork activation blocks no longer include any user transactions

Might modify to be more specific, the current phrasing suggests that all fork activation blocks in the future will no longer include user transactions, which may end up being the case but this hasn't been codified in any way.

specs/interop/derivation.md
```diff
@@ -44,11 +45,23 @@ because it includes invalid executing messages MUST be replaced by a deposits on
...
+[Activation Block](#activation-block) section. Sequencers, when building the fork activation block, MUST set
+`noTxPool` to `true` in the execution payload attributes for this block, instructing the builder to exclude user
```
teddyknox: Can this line be dropped given that you added the above section?

teddyknox (approved):

specs/interop/derivation.md
ghost: @protolambda or @tynes could you clarify this? If the activation block may have a timestamp greater than the activation time then our guards against allowing messages to be initialized in this block need to be updated to rely on the anchor block time instead of the configured activation time.

specs/interop/derivation.md
ghost: We explicitly do not want messages to be initiated in the activation block.

specs/interop/derivation.md
ghost: My understanding is that this is new behavior introduced in Holocene.

specs/interop/derivation.md
ghost: Yeah that's probably best. The motivation for this PR is to consolidate these rules so I'll remove it.

specs/interop/derivation.md
ghost: That language is untouched by this PR it's just on the same line. I agree it could be more specific but I don't want that to hold up getting this spec change in.

specs/interop/derivation.md
protolambda: Yes, technically the activation block does not always have the exact equal timestamp. Good catch @teddyknox!

I'll make some amendments to change it to the activation block. We need to get this spec change in + the implementation change in, to unblock RC betanet.

I'll notify the rust team too when ready, since the `>=` to `>` check change that was adopted in the implementation isn't valid with activation block.

specs/interop/derivation.md
protolambda: We should clarify that log events are acceptable, but that they may not be consumed as initiating messages yet.

specs/interop/derivation.md
protolambda: The naming is confusing due to legacy. The "singular batch" that we read from the span-batch is a single block. Only that block is what we don't include additional txs in. The other blocks in the span batch may contain txs.

specs/interop/derivation.md
protolambda: I'll make it more specific

specs/interop/derivation.md
protolambda: I think it's better to omit the `noTxPool` instruction from the text above, since the text above applies to the activation block, not specifically to the block-inputs process.

protolambda: Hmm, the below "The derivation pipeline MUST enforce that the sequence" is still very much a duplicate though. Will simplify that.

protolambda:
Added a commit that implements some resolutions for the above + fixes lint, to unblock the other work.
Looking to get this PR merged asap so we can unblock RC Betanet.

specs/interop/derivation.md
teddyknox: My bad @tyler-smith, I hadn't checked if this was really a new line. Not trying to nit, I believe I authored the line originally by copying it from the Isthmus derivation.md.

protolambda:
@clabby is giving this a sanity-check, so we know it's not a problem on the rust side. Should be a simple thing though, to handle the activation block exclusion from init messages, without forcing the fork timestamp to exactly match.
Ready to merge if @clabby gives a :+1: 
Will add a do-not-merge for now.

specs/interop/derivation.md
```diff
@@ -34,26 +34,40 @@ executing messages.
...
+  - `t > execution_timestamp - expiry_window`, where `expiry_window` is the expiry window in seconds.
 
 L2 blocks that produce invalid executing messages MUST not be allowed to be considered safe.
 They MAY optimistically exist as unsafe blocks for some period of time. An L2 block that is invalidated
 because it includes invalid executing messages MUST be replaced by a deposits only block at the same
 block height. This guarantees progression of the chain, ensuring that an infinite loop of processing
 the same block in the proof system is not possible.
 
-## Omit User Transactions in Fork Activation Block
+## Activation Block
```
protolambda, line 49: Specs doesn't have a do-not-merge label.
Adding this comment to block instead until @clabby has checked

clabby:
This seems like it's not a problem. We can add a simple calculation to convert the timestamp -> a block number and do the check that way, we do have the rollup config w/ the genesis time and block time available to us where we do this check.

specs/interop/derivation.md
clabby: https://github.com/ethereum-optimism/specs/pull/694#issuecomment-2847971979

protolambda (approved):

PR #696, opened 2025-05-02
specs: update expiry window

tynes:
Per the decision doc, we have agreed to update the expiry window to be 7
days.

- User experience of expired messages is minimal given the ability to re-emit.
  - The autorelayer handling cross-chain message passing dramatically reduces the
    probability of any message even reaching the 7 day expiry
  - We can clearly document how resendMessage() is only available for initiating
    messages that rely on the CDM, and that there is a significant risk of expiry
    if using a lower-level abstraction (e.g. CrossL2Inbox)
- While we *believe* that there is work to scale the proof system to support more
  than 2 chains, we have not benchmarked this yet – if it turns out that we can
  support more than 2 without needing to break up consolidation, reducing the expiry
  window proactively accelerates this timeline. There is no harm in shortening the expiry window.

**Description**

**Tests**

**Additional context**

**Metadata**

### DIFF
--- a/specs/interop/derivation.md
+++ b/specs/interop/derivation.md
@@ -287,7 +287,7 @@ The expiry window is the time period after which an initiating message is no lon
 
 | Constant | Value |
 | -------- | ----- |
-| `EXPIRY_WINDOW` | `2592000 secs` (30 days) |
+| `EXPIRY_WINDOW` | `604800 secs` (7 days) |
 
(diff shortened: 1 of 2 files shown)


tynes:
@clabby @Inphi can we update the implementations to have the new 7 day default?

clabby (approved):

Inphi (approved):

PR #701, opened 2025-05-12
feat: governance proposal validator specs

0xOneTony:
**Description**

Following the design doc https://github.com/ethereum-optimism/design-docs/pull/260, this PR details the specs for the new `ProposalValidator` contract

### DIFF
--- a/specs/experimental/gov-proposal-validator.md
+++ b/specs/experimental/gov-proposal-validator.md
@@ -34,7 +34,7 @@ The `ProposalValidator` manages the proposal lifecycle through three main functi
 - `approveProposal`: Handles proposal approvals
 - `moveToVote`: Transitions approved proposals to voting phase
 
-The contract also integrates with EAS (Ethereum Attestation Service) to verify authorized proposers for specific proposal types. For detailed flows of each proposal, see [Design Document Link].
+The contract also integrates with EAS (Ethereum Attestation Service) to verify authorized proposers for specific proposal types. For detailed flows of each proposal, see https://github.com/ethereum-optimism/design-docs/pull/260.
 
 ## Roles
 
@@ -50,10 +50,9 @@ The contract has a single `owner` role (Optimism Foundation) with permissions to
 
 `submitProposal`
 
-Submits a proposal for approval and voting. Based on the `ProposalType` provided this will require different validation checks.
+Submits a proposal for approval and voting. Based on the `ProposalType` provided this will require different validation checks and actions.
 
-For `ProtocolOrGovernorUpgrade` , `MaintenanceUpgradeProposals`, and `CouncilMemberElections` types:
-
+- MUST only be called for `ProtocolOrGovernorUpgrade` , `MaintenanceUpgradeProposals`, or `CouncilMemberElections` types
 - MUST be called by an approved address
 - MUST check if the proposal is a duplicate
 - MUST provide a valid proposal type configurator
@@ -82,8 +81,9 @@ Note: `MaintenanceUpgradeProposals`type can move straight to voting if all submi
 
 `submitFundingProposal`
 
-Submits a `GovernanceFund` or `CouncilBudget` proposal type that transfers OP tokens.
+Submits a `GovernanceFund` or `CouncilBudget` proposal type that transfers OP tokens for approval and voting.
 
+- MUST only be called for `GovernanceFund` or `CouncilBudget` proposal type
 - CAN be called by anyone
 - MUST check if the proposal is a duplicate
 - MUST provide a valid proposal type configurator
@@ -122,21 +122,21 @@ Checks if the provided proposal is ready to move for voting. Based on the Propos
 
 For `ProtocolOrGovernorUpgrade`:
 
-- MUST check if provided data produce the same `proposalHash`
-- proposal MUST have gathered X amount of approvals by top delegates
+- MUST check if provided data produces a valid `proposalHash`
+- Proposal MUST have gathered X amount of approvals by top delegates
 - MUST check if proposal has already moved for voting
 - MUST emit `ProposalMovedToVote` event
 
 For `MaintenanceUpgradeProposals`:
 
-- This type does not require any checks and is being forwarded to the Governor contracts, this should happen atomic.
+- This type does not require any checks and is being forwarded to the Governor contracts, this should happen atomically.
 - MUST emit `ProposalMovedToVote` event
 
 For `CouncilMemberElections`, `GovernanceFund` and `CouncilBudget`:
 
 - MUST check if provided data produce the same `proposalHash`
-- proposal MUST have gathered X amount of approvals by top delegates
-- proposal MUST be moved to vote during a valid voting cycle
+- Proposal MUST have gathered X amount of approvals by top delegates
+- Proposal MUST be moved to vote during a valid voting cycle
 - MUST check if proposal has already moved for voting
 - MUST check if the total amount of tokens that can possible be distributed during this voting cycle does not go over the `VotingCycleData.votingCycleDistributionLimit`
 - MUST emit `ProposalMovedToVote` event
@@ -183,16 +183,16 @@ Sets the start and the duration of a voting cycle.
 - MUST NOT change an existing voting cycle
 - MUST emit `VotingCycleSet` event
 
-`setDistributionThreshold`
-
 ```solidity
 function setVotingCycleData(uint256 _cycleNumber, uint256 _startBlock, uint256 _duration, uint256 _distributionLimit) external
 ```
 
+`setDistributionThreshold`
+
 Sets the maximum distribution threshold a proposal can request.
 
 - MUST only be called by the owner of the contract
-- MUST change the previous threshold to the new
+- MUST change the previous threshold to the new one
 - MUST emit `DistributionThresholdSet` event
 
 ```solidity
@@ -204,7 +204,7 @@ function setDistributionThreshold(uint256 _threshold) external
 Sets the number of approvals a specific proposal type should have before being able to move for voting.
 
 - MUST only be called by the owner of the contract
-- MUST change the previous value to the new
+- MUST change the previous value to the new one
 - MUST emit `ProposalTypeApprovalThresholdSet` event
 
 ```solidity
@@ -215,7 +215,7 @@ function setProposalTypeApprovalThreshold(uint8 _proposalTypeId, uint256 _approv
 
 `ATTESTATION_SCHEMA_UID`
 
-The schema UID that is used to verify attestation for approved addresses that can submit proposals for specific `ProposalTypes`. 
+The EAS' schema UID that is used to verify attestation for approved addresses that can submit proposals for specific `ProposalTypes`.
 
 ```solidity
 /// Schema { approvedProposer: address, proposalType: uint8 }
@@ -224,7 +224,7 @@ bytes32 public immutable ATTESTATION_SCHEMA_UID;
 
 `TRANSFER_SIGNATURE`
 
-The 4byte signature of ERC20.transfer, will be used for creating the calldata for funding proposals
+The 4bytes signature of ERC20.transfer, will be used for creating the calldata for funding proposals
 
 ```solidity
 bytes4 public constant TRANSFER_SIGNATURE = 0xa9059cbb;
@@ -270,12 +270,12 @@ A mapping that stores each submitted proposals’ data based on its `proposalHas
 mapping(bytes32 => ProposalSubmissionData) private _proposals;
 ```
 
-`_proposaTypesData` 
+`_proposalTypesData`
 
 A mapping that stores data related to each proposal type.
 
 ```solidity
-mapping(ProposalType => ProposalTypeData) private _proposaTypesData;
+mapping(ProposalType => ProposalTypeData) private _proposalTypesData;
 ```
 
 ### Structs
@@ -307,7 +307,7 @@ struct ProposalSubmissionData {
 A struct that holds data for each proposal type.
 
 - `requiredApprovals`:  The number of approvals each proposal type requires in order to be able to move for voting.
-- `validProposalTypeConfigurators`: The accepted proposal type configurators that can be used for each proposal type.
+- `validProposalTypeConfigurators` : The accepted proposal type configurators that can be used for each proposal type.
 
 ```solidity
 struct ProposalTypeData {
@@ -392,7 +392,7 @@ event ProposalMovedToVote(uint256 indexed proposalHash, address indexed executor
 
 ## EAS Integration
 
-`ProposalValidator` integrates Ethereum Attestation Service (EAS) to handle proposer authorization for specific `ProposalType`s. This removes the need for maintaining separate allowlists or custom registries, reducing contract complexity and offloading identity tracking to a proven system.
+`ProposalValidator` integrates Ethereum Attestation Service (EAS) to handle proposer authorization for specific `ProposalType`s, removing the need for adding custom logic to the contract.
 
 ### Why EAS?
 
@@ -426,7 +426,7 @@ A proposal is uniquely identified by the tuple:
 These elements are ABI-encoded and hashed:
 
 ```solidity
-keccak256(abi.encode(targets, values, calldatas, keccak256(bytes(description))));
+keccak256(abi.encode(targets, values, calldatas, description));
 ```
 
 This hash serves as a unique identifier for the proposal. The contract stores submitted proposals in:
@@ -441,9 +441,9 @@ This mechanism guarantees that proposals with the same intent and execution logi
 
 ## Invariants
 
-- It MUST allow only the `owner` to set the `minimumVotingPower`, `votingCycle` and `distributionThreshold`
-- It MUST allow only eligible addresses to approve a proposal
-- It MUST allow only authorizeds addresses to submit proposals for types `ProtocolOrGovernorUpgrade`, `MaintenanceUpgradeProposals`, and `CouncilMemberElections`
+- It MUST allow only the `owner` to set the `minimumVotingPower`, `votingCycleData,` `distributionThreshold`, and `proposalTypeApprovalThreshold`
+- It MUST only allow eligible addresses to approve a proposal
+- It MUST only allow authorizeds addresses to submit proposals for types `ProtocolOrGovernorUpgrade`, `MaintenanceUpgradeProposals`, and `CouncilMemberElections`
 - It MUST NOT transfer any tokens or ETH for `ProtocolOrGovernorUpgrade`, `MaintenanceUpgradeProposals`, and `CouncilMemberElections` proposal types
 - It MUST emit the following events:
     - `ProposalSubmitted`
--- a/specs/experimental/gov-proposal-validator.md
+++ b/specs/experimental/gov-proposal-validator.md
@@ -36,10 +36,10 @@ The `ProposalValidator` manages the proposal lifecycle through three main action
 
 - `Submit Proposal`: Records new proposals
 - `Approve Proposal`: Handles proposal approvals
-- `Move to Vote`: Transitions approved proposals to voting phase
+- `Move to Vote`: Transitions approved proposals to the Governor for voting
 
-The contract also integrates with EAS (Ethereum Attestation Service) to verify authorized proposers for specific proposal
-types. For detailed flows of each proposal, see [design docs](https://github.com/ethereum-optimism/design-docs/pull/260).
+The contract also integrates with EAS (Ethereum Attestation Service) to verify authorized proposers and approvals for specific
+proposal types. For detailed flows of each proposal, see [design docs](https://github.com/ethereum-optimism/design-docs/pull/260).
 
 ## Roles
 
@@ -60,40 +60,42 @@ Submits a Protocol/Governor Upgrade or a Maintenance Upgrade proposal to move fo
 
 `ProtocolOrGovernorUpgrade`: If all submission checks pass for this proposal type, the `ProposalValidator`
 will store the submission proposal data and will be able to accept approvals by top delegates before being
-able to move for voting.
+able to move to the Governor for voting.
 
 `MaintenanceUpgrade`: This proposal type can move straight to voting if all submission checks pass, unlike
 the rest of the proposals where they need to collect a number of approvals by top delegates in order
 to move to vote. This call should be atomic.
 
 - MUST be called by an `owner` approved address
+- MUST only be called for `ProtocolOrGovernorUpgrade` or `MaintenanceUpgrade` proposal type
 - MUST check if the proposal is a duplicate
 - MUST use the `Optimistic` Voting Module
 - MUST provide a valid againstThreshold
 - MUST provide a valid attestation UID
 - MUST NOT execute any operations
 - MUST emit `ProposalSubmitted` and `ProposalVotingModuleData` events
-- MUST store submission proposal data which are defined by the `ProposalSubmissionData` struct
-- MUST move straight to vote IF `_proposalType` is `MaintenanceUpgrade`
+- MUST store submission proposal data which are defined by the `ProposalData` struct
+- MUST call `Governor.proposeWithModule` IF `_proposalType` is `MaintenanceUpgrade`
 
 ```solidity
 function submitUpgradeProposal(
     uint248 _againstThreshold,
     string memory _proposalDescription,
     bytes32 _attestationUid,
-    ProposalType _proposalType
+    ProposalType _proposalType,
+    uint256 _votingCycle
 ) external returns (bytes32 proposalHash_);
 ```
 
 **Optimistic Voting Module**
 
-Maintenance Upgrade proposals use the `Optimistic` voting module.
+Both proposals use the `Optimistic` voting module.
 
 For the `ProposalSettings` of the voting module, these are:
-- `uint248 againstThreshold`: Should be provided by the proposer. This value will be the percentage
+- `uint248 againstThreshold`: Should be provided by the caller. This value will be the percentage
 that will be used to calculate the fraction of the votable supply that the proposal will need in votes in order
 to pass.
-- `bool isRelativeToVotableSupply`: Should always be `true`.
+- `bool isRelativeToVotableSupply`: If voting power should be relative to the votable supply. Should always be `true`.
 ---
 
 `submitCouncilMemberElectionsProposal`
@@ -107,14 +109,15 @@ Submits a Council Member Elections proposal for approval and voting.
 - MUST provide a valid attestation UID
 - MUST NOT execute any operations
 - MUST emit `ProposalSubmitted` and `ProposalVotingModuleData` events
-- MUST store submission proposal data which are defined by the `ProposalSubmissionData` struct
+- MUST store submission proposal data which are defined by the `ProposalData` struct
 
 ```solidity
 function submitCouncilMemberElectionsProposal(
     uint128 _criteriaValue,
     string[] _optionDescriptions,
     string memory _proposalDescription,
-    bytes32 _attestationUid
+    bytes32 _attestationUid,
+    uint256 _votingCycle
 ) external returns (bytes32 proposalHash_);
 ```
 
@@ -133,7 +136,7 @@ For the `ProposalOptions` of the voting module, these are:
 
 `submitFundingProposal`
 
-Submits a `GovernanceFund` or `CouncilBudget` proposal type that transfers OP tokens for approval and voting.
+Submits a `GovernanceFund` or `CouncilBudget` proposal type, for approval and voting, that transfers OP tokens.
 
 - MUST only be called for `GovernanceFund` or `CouncilBudget` proposal type
 - CAN be called by anyone
@@ -143,8 +146,8 @@ Submits a `GovernanceFund` or `CouncilBudget` proposal type that transfers OP to
 - MUST use the `Predeploys.GOVERNANCE_TOKEN` and `IERC20.transfer` signature to create the `calldata`
 for each option
 - MUST NOT request to transfer more than `proposalDistributionThreshold` tokens
-- MUST emit `ProposalSubmitted` event
-- MUST store submission proposal data which are defined by the `ProposalSubmissionData` struct
+- MUST emit `ProposalSubmitted` and `ProposalVotingModuleData` events
+- MUST store submission proposal data which are defined by the `ProposalData` struct
 
 ```solidity
 function submitFundingProposal(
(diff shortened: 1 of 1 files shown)


specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+    uint8 _proposalTypeConfiguration,
+    bytes32 _attestationUid
+) external returns (bytes32 proposalHash_);
+```
+
+`submitFundingProposal`
+
+Submits a `GovernanceFund` or `CouncilBudget` proposal type that transfers OP tokens for approval and voting.
+
+- MUST only be called for `GovernanceFund` or `CouncilBudget` proposal type
```
ben-chain, line 141: Just to be explicit -- these two types should only split out/separate to the extent that it's useful for frontends to differentiate between them.  Other than that, all onchain logic should be identical between these two types.  I don't see anything that suggests otherwise yet but felt worth a call-out.

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+Submits a proposal for approval and voting. Based on the `ProposalType` provided this will require different validation checks and actions.
+
+- MUST only be called for `ProtocolOrGovernorUpgrade` , `MaintenanceUpgradeProposals`, or `CouncilMemberElections` types
+- MUST be called by an approved address
+- MUST check if the proposal is a duplicate
+- MUST provide a valid proposal type configurator
+- MUST provide a valid attestation UID
+- MUST NOT transfer any tokens or change any allowances
+- MUST emit `ProposalSubmitted` event
+- MUST store proposal data
```
ben-chain: Let's try to specify what exactly we mean by "proposal data" in this context. I don't think it requires all arguments to `submitProposal` to be stored. Once I review the actual functions, I will try to return to this comment and specify what I think it would be.

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+### Public Functions
+
+`submitProposal`
+
+Submits a proposal for approval and voting. Based on the `ProposalType` provided this will require different validation checks and actions.
+
+- MUST only be called for `ProtocolOrGovernorUpgrade` , `MaintenanceUpgradeProposals`, or `CouncilMemberElections` types
+- MUST be called by an approved address
+- MUST check if the proposal is a duplicate
+- MUST provide a valid proposal type configurator
```
ben-chain: This should not be necessary I believe.  There is a 1:1 mapping of `ProposalType` as defined in this spec to the `ProposalTypesConfigurator` uint in the existing governor.  This is defined in column G [here](https://docs.google.com/spreadsheets/d/1HCAQQmBbWYpXni2zZ3Vnyv-bM_f5pHFFrpItI0TXdAk/edit?gid=1779711153#gid=1779711153)

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+
+- The user MUST use the `submitFundingProposal` that uses specific `calldata` pre-defined by the owner
+
+Note: `MaintenanceUpgradeProposals`type can move straight to voting if all submission checks pass, unlike the rest of the proposal where they need to collect a number of approvals by top delegates in order to move to vote. This call should be atomic.
+
+```solidity
+ function submitProposal(
+    address[] memory _targets,
+    uint256[] memory _values,
+    bytes[] memory _calldatas,
```
ben-chain:
```suggestion
```

In the future this may change, but for MVP, none of these proposal types are ever expected to create a call from the governor. So instead what we want to do is either:
 - enforce that `submitProposal` stores `[0x00..00],[0],[0x]` for any proposal submitted via this function
 - enforce that `moveToVote` only passes those zero values as calldata to the `OptimismGovernor.propose` function
 Depending on which feels cleaner/easier :)

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+    string memory _description,
+    ProposalType _proposalType,
+    uint8 _proposalTypeConfiguration,
```
ben-chain:
```suggestion
```

See above comment, `_proposalType` is sufficient to infer this value

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+```
+
+`submitFundingProposal`
+
+Submits a `GovernanceFund` or `CouncilBudget` proposal type that transfers OP tokens for approval and voting.
+
+- MUST only be called for `GovernanceFund` or `CouncilBudget` proposal type
+- CAN be called by anyone
+- MUST check if the proposal is a duplicate
+- MUST provide a valid proposal type configurator
```
ben-chain: Same as for `submitProposal`
```suggestion
```

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+- MUST emit `ProposalSubmitted` event
+- MUST store proposal data
+
+```solidity
+function submitFundingProposal(
+	address _to,
+	uint256 _amount,
+	string memory _description,
+    ProposalType _proposalType,
+	uint8 _proposalTypeConfiguration
```
ben-chain:
```suggestion
    ProposalType _proposalType
```

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+) external returns (bytes32 proposalHash_);
+```
+
+`approveProposal`
+
+Approves a proposal before being moved for voting, used by the top delegates.
+
+- MUST check if proposal hash corresponds to a valid proposal
+- MUST check if caller is has enough voting power to call the function and approve a proposal
```
ben-chain: One question I don't know the answer to is at what time this needs to hold true (i.e. what is the correct `blockNumber` argument to `GovToken.getPastVotes`?). I will follow up with the governance team to check how we should handle. I think it cannot be `block.number`, because then a single delegate can transfer between their own wallets and basically sybil.

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+```
+
+`moveToVote`
+
+Checks if the provided proposal is ready to move for voting. Based on the Proposal Type different checks are being validated. If all checks pass then `OptimismGovernor.propose` is being called to forward the proposal for voting. This function can be called by anyone
+
+For `ProtocolOrGovernorUpgrade`:
+
+- MUST check if provided data produces a valid `proposalHash`
+- Proposal MUST have gathered X amount of approvals by top delegates
```
ben-chain: Realizing I never specified before but X=4 here for all proposal types

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+function approveProposal(bytes32 _proposalHash) external
+```
+
+`moveToVote`
+
+Checks if the provided proposal is ready to move for voting. Based on the Proposal Type different checks are being validated. If all checks pass then `OptimismGovernor.propose` is being called to forward the proposal for voting. This function can be called by anyone
+
+For `ProtocolOrGovernorUpgrade`:
+
+- MUST check if provided data produces a valid `proposalHash`
```
ben-chain: I see, so is this what we mean by `MUST store proposal data` in the submission functions above?  Store a commitment to them?

I am fine with this but let's discuss with Agora whether literally storing everything is easier, I am not worried about the L2 gas costs of these transactions

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+- This type does not require any checks and is being forwarded to the Governor contracts, this should happen atomically.
+- MUST emit `ProposalMovedToVote` event
+
+For `CouncilMemberElections`, `GovernanceFund` and `CouncilBudget`:
+
+- MUST check if provided data produce the same `proposalHash`
+- Proposal MUST have gathered X amount of approvals by top delegates
+- Proposal MUST be moved to vote during a valid voting cycle
+- MUST check if proposal has already moved for voting
+- MUST check if the total amount of tokens that can possible be distributed during this voting cycle does not go over the `VotingCycleData.votingCycleDistributionLimit`
```
ben-chain: Let's discuss on the call tomorrow the implementation complexity of this approach.  We are comfortable with a per-proposal-only cap if it simplifies things a lot.

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+    address[] memory _targets,
+    uint256[] memory _values,
+    bytes[] memory _calldata,
+    string memory description
+)
+    external
+    returns (uint256 governorProposalId_)
+```
+
+`canSignOff`
```
ben-chain: nit: can we call this `canApproveProposal`, feels a bit more obvious what it is than "sign off"

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+- MUST check if proposal has already moved for voting
+- MUST check if the total amount of tokens that can possible be distributed during this voting cycle does not go over the `VotingCycleData.votingCycleDistributionLimit`
+- MUST emit `ProposalMovedToVote` event
+
+```solidity
+function moveToVote(
+    address[] memory _targets,
+    uint256[] memory _values,
+    bytes[] memory _calldata,
+    string memory description
```
ben-chain:
```suggestion
    string memory description,
    ProposalType memory _proposalType,
    address[] memory candidates, // only for elections
    address _recipient, // only for funding proposals
    uint256 _amount // only for funding proposals
```

I think it will be something like this, may want to break out a few of these functions/signature overload or whatever

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+
+- MUST check if provided data produce the same `proposalHash`
+- Proposal MUST have gathered X amount of approvals by top delegates
+- Proposal MUST be moved to vote during a valid voting cycle
+- MUST check if proposal has already moved for voting
+- MUST check if the total amount of tokens that can possible be distributed during this voting cycle does not go over the `VotingCycleData.votingCycleDistributionLimit`
+- MUST emit `ProposalMovedToVote` event
+
+```solidity
+function moveToVote(
```
ben-chain: I can't think of a reason that we would need it right now, but this smells like a place where traditionally there would be some sort of nonce stored and incremented during proposal submission and then provided to `moveToVote`. I can't think of what would go wrong without it though. It would mean you can't submit _identical_ proposals multiple times, but changing the description is also very easy. Let's discuss live.

specs/experimental/gov-proposal-validator.md
ben-chain: Actually, I am realizing that there is an exception to the above for Approval votes (elections) because we need to specify a number of options. The options should still be no-ops but there needs to be some submission step here which I don't fully understand.  Effectively, what I'm saying above that we just enforce no-op still applies, but those no-ops will within [this struct](https://github.com/voteagora/optimism-governor/blob/3f8a2ea29d2c91a176758c72a213d499424ee2fc/src/modules/ApprovalVotingModule.sol#L37-L39), which is embedded in an array in the [overall proposal struct](https://github.com/voteagora/optimism-governor/blob/3f8a2ea29d2c91a176758c72a213d499424ee2fc/src/modules/ApprovalVotingModule.sol#L47) for this type..  I will follow up and figure out exactly how this logic works, I am having trouble understanding.  For reference, here is an [example election proposal](https://dashboard.tenderly.co/tx/0xc966c14adcaf45fca9c57db889a2b69639ec60f97b8b891b7950331960700535/logs); I'm struggling to decode the data, need a night's sleep first 😅

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,458 @@
...
+
+- MUST only be called by the owner of the contract
+- MUST change the previous threshold to the new one
+- MUST emit `DistributionThresholdSet` event
+
+```solidity
+function setDistributionThreshold(uint256 _threshold) external
+```
+
+`setProposalTypeApprovalThreshold`
```
ben-chain: Oh I see, we made this configurable.  I think that's okay, it would be nice if we can default to 4 in the constructor so we don't have to remember to call this unless we change in the future

ben-chain (commented):
I have reviewed this up to about the halfway mark which covers most of the business logic.  Will return to the rest, but this feels like enough meat to start making improvements.

Main open item I need to figure out is how we populate the election calldata. Need to investigate how Agora does this and will return here

specs/experimental/gov-proposal-validator.md
0xOneTony: Definitely, since these 2 types require the same checks there wont be any difference when it comes to smart contract logic

specs/experimental/gov-proposal-validator.md
0xOneTony: I agree, we will make this more descriptive on what the "proposal data" are. In the meantime I will give some context here. We use this mapping to store each proposal data
https://github.com/ethereum-optimism/specs/blob/2465cca9fc3b0ef018d7ec78e40d93f67ae99e84/specs/experimental/gov-proposal-validator.md?plain=1#L265-L271

where `ProposalSubmissionData` are defined here
https://github.com/ethereum-optimism/specs/blob/2465cca9fc3b0ef018d7ec78e40d93f67ae99e84/specs/experimental/gov-proposal-validator.md?plain=1#L283-L303

Do you think we could remove any of the structs properties?

specs/experimental/gov-proposal-validator.md
0xOneTony: Thank you for pointing this out, we initially thought to not have this, and tie the `ProposalTypeConfigurtator` to each `ProposalType` like you mentioned, however we were not sure if these are possible to change in the future or have a `ProposalType` to be able to choose between more than 1 `ProposalTypeConfigurator` thats why we defined it as is. 
If these `ProposalTypeConfigurators` WILL NOT change any time in the future we could use the mapping and remove the check, this would make things much more simple for us.

specs/experimental/gov-proposal-validator.md
0xOneTony: We are aware of the current X=4 that is being used, we are wondering if this could possibly change in the future, thats why we made it X and settable by the `owner`.
If this is not going to change we can create a constant and remove the setter function aswell

specs/experimental/gov-proposal-validator.md
0xOneTony: yes exactly what you mentioned, also left some context on the other comment here https://github.com/ethereum-optimism/specs/pull/701#discussion_r2088511275

specs/experimental/gov-proposal-validator.md
0xOneTony: I agree, will change it!

specs/experimental/gov-proposal-validator.md
ben-chain: Thanks yes, I made this comment before reading further down, it makes more sense with the struct specified.

> Do you think we could remove any of the structs properties?

`proposalTypeConfigurator`, pending other discussions on this topic :)

specs/experimental/gov-proposal-validator.md
ben-chain: I think it is very low probability that we break 1:1 invariant any time soon.  Further, it is always possible to add additional `ProposalType`s to implement the same functionality and maintain the 1:1 invariant.  IMO this might be a better pattern anyway, so my intuition is the simpler thing is better.

specs/experimental/gov-proposal-validator.md
ben-chain: I have discussed this a bit with the gov team, no conclusion yet.  How it is done today offchain involves "forward-looking" approvals, where the delegate approves in advance of the `blockNumber` passed to `getPastVotes`.  I would like to avoid this for the onchain version since it's going to introduce complexity

specs/experimental/gov-proposal-validator.md
ben-chain: Probably fine to leave this settable, I guess it's not a lot of complexity.

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,581 @@
...
+- Set minimum voting power threshold for delegate approvals
+- Set the minimum approvals for each supported proposal type
+- Configure voting cycle parameters
+- Set maximum token distribution limits for proposals
+
+## Interface
+
+### Public Functions
+
+`submitProtocolOrGovernorUpgradeProposal`
```
ben-chain: fwiw if this is killing you I'm fine with just `submitUpgradeProposal` or `submitRegularUpgradeProposal` or something :P

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,581 @@
...
+- `Submit Proposal`: Records new proposals
+- `Approve Proposal`: Handles proposal approvals
+- `Move to Vote`: Transitions approved proposals to voting phase
+
+The contract also integrates with EAS (Ethereum Attestation Service) to verify authorized proposers for specific proposal
+types. For detailed flows of each proposal, see [design docs](https://github.com/ethereum-optimism/design-docs/pull/260).
+
+## Roles
+
+The contract has a single `owner` role (Optimism Foundation) with permissions to:
```
ben-chain, line 46: Something I didn't think to ask -- do we read this storage address to check for the EAS attestations as well?  That should probs be in this list if so.  If not, curious how it's done otherwise

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,581 @@
...
+- MUST use "Threshold" criteria type for the Voting Module
+- MUST provide a valid attestation UID
+- MUST NOT execute any operations
+- MUST emit `ProposalSubmitted` and `ProposalVotingModuleData` events
+- MUST store submission proposal data which are defined by the `ProposalSubmissionData` struct
+
+```solidity
+function submitProtocolOrGovernorUpgradeProposal(
+    uint128 _criteriaValue,
+    string[] memory _optionDescriptions,
```
ben-chain: I believe the options should always be yes/no here for upgrades, so we can omit from this method and hardcode in the function.  Following up w/ gov team if there is a third abstain option, but I believe there is not.

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,581 @@
...
+    string memory _proposalDescription,
+    bytes32 _attestationUid
+) external returns (bytes32 proposalHash_);
+```
+
+Approval Voting Module
```
ben-chain:
```suggestion
**Approval Voting Module**
```
nit

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,581 @@
...
+
+## Interface
+
+### Public Functions
+
+`submitProtocolOrGovernorUpgradeProposal`
+
+Submits a Protocol or Governor Upgrade proposal for approval and voting.
+
+- MUST be called by an approved address
```
ben-chain:
```suggestion
- MUST be called by an `owner`-approved address
```
clarity nit assuming my question on the `owner` role is yes

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,581 @@
...
+- `string[] optionDescriptions`: The strings of the different options that can be voted.
+
+`submitMaintenanceUpgradeProposal`
+
+Submits a Maintenance Upgrade proposal to move for voting. `MaintenanceUpgradeProposals` type can move
+straight to voting if all submission checks pass, unlike the rest of the proposals where they
+need to collect a number of approvals by top delegates in order to move to vote. This call should be
+atomic.
+
+- MUST be called by an approved address
```
ben-chain:
```suggestion
- MUST be called by an `owner`-approved address
```
clarity nit assuming my question on the `owner` role is yes

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,581 @@
...
+) external returns (bytes32 proposalHash_);
+```
+
+Approval Voting Module
+
+Protocol or Governor Upgrade proposals use the `Approval` voting module.
+This requires the user who submits the proposal to provide some additional data related to the proposal.
+
+For the `ProposalSettings` of the voting module, these are:
+- `uint128 criteriaValue`: Since the passing criteria type is always "Threshold" for this proposal type
```
ben-chain: This does not definitively answer how we will populate this field.  I believe it will always just be a fixed % of the votably supply so it should be calculable by querying the VotableSupply contract and multiplying by a fraction.  Does that sound right to you?

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,581 @@
...
+- MUST use the `Optimistic` Voting Module
+- MUST provide a valid attestation UID
+- MUST NOT execute any operations
+- MUST emit `ProposalSubmitted` and `ProposalVotingModuleData` events
+- MUST store submission proposal data which are defined by the `ProposalSubmissionData` struct
+
+```solidity
+function submitMaintenanceUpgradeProposal(
+    uint248 _againstThreshold,
+    bool _isRelativeToVotableSupply,
```
ben-chain: We haven't documented these values here, but I believe they can be omitted from this method. I believe that this should always be `12%, true` -- meaning, if more than 12% of votable supply chooses veto option, the proposal is rejected. I think hardcoding should be fine

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,581 @@
...
+```
+
+Approval Voting Module
+
+Funding proposals use the `Approval` voting module but unlike the Protocol or Governor upgrade proposals,
+funding proposals need to execute token transfers.
+This requires the user who submits the proposal to provide some additional data related to the proposal.
+
+For the `ProposalSettings` of the voting module, these are:
+- `uint128 criteriaValue`: Since the passing criteria type is always "Threshold" for this proposal type
```
ben-chain: I believe that this should be a fixed % of votable supply.  So I think we can remove it as an argument and just pass fixed % of the votable supply oracle value

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,581 @@
...
+For `ProtocolOrGovernorUpgrade` type:
+
+- MUST also check that is called by the proposer that submitted the proposal
+- `_optionsRecipients` and `_optionsAmounts` MUST be empty
+
+For `MaintenanceUpgradeProposals` type:
+
+- This type does not require any checks and is being forwarded to the Governor contracts, this should happen atomically.
+
+For `CouncilMemberElections`, `GovernanceFund` and `CouncilBudget` types:
```
ben-chain: Elections do not do fund transfers.  So I believe that the `_optionsAmounts/recipients` should be checked as zero in the `CouncilMemberElections` case, same as upgrades case

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,581 @@
...
+- Can be called by anyone
+- MUST return TRUE if the delegates' voting power is above the `minimumVotingPower`
+
+```solidity
+function canApproveProposal(address _delegate) public view returns (bool canSignOff_)
+```
+
+`setMinimumVotingPower`
+
+Sets the minimum voting power a delegate must have in order to be eligible to approve a proposal.
```
ben-chain: Technically if the `canApproveProposal` check is based on the start of the voting cycle, wouldn't this need to be indexed by voting cycle or something?

If the associated complexity is significant, we can think about other options.  But if there's a voting cycle storage object, it might just be that you `setMinimumVotingPower(_power, _cycleId)` or something here instead

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,581 @@
...
+
+- `startingBlock`: The block number/timestamp that the voting cycle starts
+- `duration`: The duration of the specific voting cycle
+- `votingCycleDistributionLimit`: A more general distribution amount limit tied to the voting cycle
+
+```solidity
+struct VotingCycleData {
+    uint256 startingBlock;
+    uint256 duration;
+    uint256 votingCycleDistributionLimit;
```
ben-chain, line 463: So I guess we could put the `approvalThreshold` here to do it per-cycle?  I guess the issue would then be the OF needs to set it quickly, or we might have the wrong value used in a `canApprove` check.

Maybe we can block delegate approvals until the threshold is set. So the flow would be:
- initial voting cycles have a new `approvalThreshold` field, all are initialized to 0
- once a cycle has started and we know what the threshold should be and set to that value
- `approveProposal` requires the above step was complete/nonzero value

Does that feel doable?  Any other ideas?

ben-chain (commented):
Looking good!  I think the biggest concern I realized in this review is the delegate approval threshold being dynamic / changing each voting cycle as delegations change.  I proposed one way to do it that seems ok, but let's discuss in discord how painful this is to implement in practice

specs/experimental/gov-proposal-validator.md
0xOneTony: perfect! I prefer the sorter version aswell just wanted to be as verbose but I think `submitUpgradeProposal` is fine aswell

specs/experimental/gov-proposal-validator.md
0xOneTony: beautiful! will adjust to the 2 options

specs/experimental/gov-proposal-validator.md
0xOneTony: @ben-chain Perfect this makes things more simple for the specific proposal, do we know if these numbers could change in the future?

specs/experimental/gov-proposal-validator.md
0xOneTony: Correct, for the `CouncilMemberElections` we will be checking to make sure these fields are empty, will  update the document to mention this

specs/experimental/gov-proposal-validator.md
0xOneTony: Yes we do read the address to make sure that the `attester` is the `owner` address

specs/experimental/gov-proposal-validator.md
0xOneTony: Yes, sounds right to me. Is this % fixed? If not would the approved address that submits the proposal know the "correct" %? I am concerned how we could validate the validity of the value if it is not a fixed one.

specs/experimental/gov-proposal-validator.md
0xOneTony: Agreed to keep this as is for the MVP

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,588 @@
...
+
+**Approval Voting Module**
+
+Protocol or Governor Upgrade proposals use the `Approval` voting module.
+This requires the user who submits the proposal to provide some additional data related to the proposal.
+
+For the `ProposalSettings` of the voting module, these are:
+- `uint128 criteriaValue`: Since the passing criteria type is always "Threshold", for this proposal type,
+this value will be the percentage that will be used to calculate the fraction of the votable supply that
+the proposal will need in votes in order to pass.
```
0xOneTony: @ben-chain If the % is fixed we can even remove this as an input and hardcode it

specs/experimental/gov-proposal-validator.md
ben-chain: Whoops, I was incorrect -- it should still be hardcoded, but options are yes/no/abstain

specs/experimental/gov-proposal-validator.md
ben-chain: Hmmmm, actually, this may not be fixed.  I think it needs to remain as an input.  There may be a small set (~3) of permissible values instead of one.  I need to follow up with the team tomorrow, let me return here then.

On validation -- since this is method is OF-only, I think it is okay not to validate. But if there's an easy win with the three values maybe we should; I'll follow up tomorrow.

specs/experimental/gov-proposal-validator.md
ben-chain: Same story as previous comment -- I will follow up tomorrow.  It is definitely always `true` but there might be a few options to %.

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,574 @@
...
+- Set maximum token distribution limits for proposals
+- Checks the address of the attester for the submit proposal attestations with the address of the owner
+
+## Interface
+
+### Public Functions
+
+`submitUpgradeProposal`
+
+Submits a Protocol/Governor Upgrade or a Maintenance Upgrade proposal to move for voting.
```
ben-chain, line 59: This should have a method spec, right?

I think it would be the same exact signature and requirements, minus `can move straight to voting if all submission checks pass`

ben-chain (commented):
This is looking great.  I'd like us to fill in the non-maintenance upgrade method for completion, but everything else looks spot on, no comments.

specs/experimental/gov-proposal-validator.md
0xOneTony: The idea is to have the same function for both of these types since most of the logic is the same, however we improved the text and params here! https://github.com/ethereum-optimism/specs/pull/701/commits/d23b08240930ee3f9f7fe4a8669431c725230145

mds1 (approved):
Approving on behalf of Ben to unblock merge

specs/experimental/gov-proposal-validator.md
```diff
@@ -0,0 +1,625 @@
...
+the rest of the proposals where they need to collect a number of approvals by top delegates in order
+to move to vote. This call should be atomic.
+
+- MUST be called by an `owner` approved address
+- MUST only be called for `ProtocolOrGovernorUpgrade` or `MaintenanceUpgrade` proposal type
+- MUST check if the proposal is a duplicate
+- MUST use the `Optimistic` Voting Module
+- MUST provide a valid againstThreshold
+- MUST provide a valid attestation UID
+- MUST NOT execute any operations
```
alcueca, line 75: What's the definition of `operations` here?

0xOneTony: This should be removed from here since these proposals are using the `Optimistic` voting module. It is also mentioned for the Council Member Elections proposals but it should stay there (will rephrase it to be more clear). Basically the `Approval` voting module has this `ProposalOption` struct https://github.com/voteagora/optimism-governor/blob/3f8a2ea29d2c91a176758c72a213d499424ee2fc/src/modules/ApprovalVotingModule.sol#L35-L41
that can include some arbitary `calldata`, so the meaning of `operations` is basically that these `calldata` should be empty and thus not execute any further operations.
An other example of the `Approval` voting module which for this case should include some calldata is for the funding proposal types where we have to execute a transfer call if the proposal passes.

PR #705, opened 2025-05-19
fix: add missing DerivedIDPair spec

dhyaniarun1993:
**Description**

- Added the missing DerivedIDPair pair spec from the supervisor spec

### DIFF
--- a/specs/interop/supervisor.md
+++ b/specs/interop/supervisor.md
@@ -154,7 +154,7 @@ Describes a block.
 
 #### `DerivedIDPair`
 
-A pair of block identifiers, linking an `derived` block (typically an L2 block) to the `source` (typically an L1 block) from which it was derived.
+A pair of block identifiers, linking a `derived` block (typically an L2 block) to the `source` (typically an L1 block) from which it was derived.
 
 `OBJECT`:
 
--- a/specs/interop/supervisor.md
+++ b/specs/interop/supervisor.md
@@ -139,7 +139,7 @@ representing a bytes list, prefixed with `0x`.
 Describes a block.
 
 `OBJECT`:
-- `hash`: `HASH` - block hash
+- `hash`: `Hash` - block hash
 - `number`: `Int` - block number
 
 #### `BlockRef`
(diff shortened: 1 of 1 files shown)


itschaindev (approved):
lgtm

emhane (approved):
lgtm

specs/interop/supervisor.md
```diff
@@ -154,6 +154,13 @@ Describes a block.
 
 #### `DerivedIDPair`
 
+A pair of block identifiers, linking an L2 block (`derived`) to the L1 block (`source`) from which it was derived.
+
+`OBJECT`:
+
+- `source`: `BlockID` - The L1 block ID from which the L2 block is derived.
+- `derived`: `BlockID` - The L2 block ID derived from the source L1 block.
```
axelKingsley: Where possible I'd like to avoid labeling source/derived as "L1" or "L2" specific

dhyaniarun1993: Make sense. I will update

specs/interop/supervisor.md
```diff
@@ -154,6 +154,13 @@ Describes a block.
 
 #### `DerivedIDPair`
 
+A pair of block identifiers, linking an L2 block (`derived`) to the L1 block (`source`) from which it was derived.
```
axelKingsley: Sorry for the nit-picking, but could we adjust the description too?

```suggestion
A pair of block identifiers, linking an `derived` block (typically an L2 block) to the `source` (typically an L1 block) from which it was derived.
```

There is some abstract future where we could accept log messages even from the `L1`, and in that case we would probably model a `DerivedIDPair` as `L1:L1` (identical on both sides), because the L1 is derived from itself, so to speak.

Feel free to push back if you think I'm introducing too much nuance here.

dhyaniarun1993: Ahh. Sorry, I missed it

specs/interop/supervisor.md
```diff
@@ -154,6 +154,13 @@ Describes a block.
 
 #### `DerivedIDPair`
 
+A pair of block identifiers, linking a `derived` block (typically an L2 block) to the `source` (typically an L1 block) from which it was derived.
```
emhane: this line is too wide, needs to be 120 or less. fixing this should make ci pass.

sebastianst (approved):

PR #707, opened 2025-05-21
interop: Update upgrade transactions to only initialize CrossL2InboxProxy when multiple chains are in the dependency set

ajsutton:
**Description**

Update upgrade transactions to only initialize CrossL2InboxProxy when multiple chains are in the dependency set

**Metadata**

https://github.com/ethereum-optimism/optimism/issues/16041


specs/interop/derivation.md
```diff
@@ -127,199 +129,212 @@ based upgrade transactions are deterministically generated by the derivation pip
...
+
+With MaxResourceLimit set to `20_000_000` and SystemTxMaxGas set to `1_000_000`,
+this mean every upgrading chain MUST have their L2GasLimit set to greater than
+`24_320_000`.
+
+## CrossL2InboxProxy Initialization
+
+On the first block at or after the interop  [activation block](#activation-block), where the current chain is in a
+dependency set with two or more active chains, a deposit transaction based upgrade transaction is deterministically
+generated by the derivation pipeline:
```
ajsutton, line 240: We don't actually have a way to add chains currently so the implementation is literally just checking this condition at the interop activation block.  We could change the spec to match the more restrictive implementation but I think this version actually says what we want the protocol rules to be, there just isn't a way to add a chain yet so in practice you only need to check the dep set size at the activation block.

tynes: This assumes that no other chains building on OP Stack will modify the dep set before we add support for adding more chains. This is technically possible under the stage 0 model

tynes (approved):

specs/interop/derivation.md
ajsutton: Modifying the dependency set is definitely not supported by the current software implementations. I don't believe it can be done even on a stage 0 chain without requiring code changes to op-supervisor.

Issue #582, opened 2025-02-06
interop: specify op-node managed-mode RPC


protolambda:
- Specify the authentication works like engine API auth
- Specify the RPC namespace itself: methods used by the supervisor to update the op-node, and watch (or pull) interop state changes.

emhane:
@teddyknox are you working on this?

teddyknox:
@emhane not at the moment, it's been deprioritized. Unassigning myself.

PR #708, opened 2025-05-21
spec: managed node <> supervisor rpc

### LINKED ISSUE #670: Interop: Specify Supervisor Managed Mode
Managed Mode is an interface that `op-node` satisfies in order to provide a stable derivation source for the `op-supervisor`.

We would like for Kona to be able to act in the same way `op-node` does, by providing a stable source for logs and derivation events.

To that end, we need to create a specification around the existing Managed Mode which Kona may follow.

Managed Mode exchanges are documented in the [README](https://github.com/ethereum-optimism/optimism/blob/develop/op-supervisor/README.md), but we need to track:

- API surface needed to handle orchestration from the Supervisor
- API surface implicitly needed by the Supervisor (like fetching logs)
- Subscription Events which must be emitted from Kona/Node to the Supervisor.

itschaindev:
Closes ethereum-optimism/specs#670 and ethereum-optimism/specs#582

@emhane

### DIFF
--- a/specs/SUMMARY.md
+++ b/specs/SUMMARY.md
@@ -78,7 +78,7 @@
     - [Sequencer](./interop/sequencer.md)
     - [Verifier](./interop/verifier.md)
     - [Supervisor](./interop/supervisor.md)
-    - [Control Flow](./interop/control-flow.md)
+    - [Managed Mode](./interop/managed-mode.md)
     - [Fault Proof](./interop/fault-proof.md)
     - [Token Bridging](./interop/token-bridging.md)
     - [ETH Liquidity](./interop/eth-liquidity.md)
--- a/specs/SUMMARY.md
+++ b/specs/SUMMARY.md
@@ -78,6 +78,7 @@
     - [Sequencer](./interop/sequencer.md)
     - [Verifier](./interop/verifier.md)
     - [Supervisor](./interop/supervisor.md)
+    - [Control Flow](./interop/control-flow.md)
     - [Fault Proof](./interop/fault-proof.md)
     - [Token Bridging](./interop/token-bridging.md)
     - [ETH Liquidity](./interop/eth-liquidity.md)
--- a/specs/interop/control-flow.md
+++ b/specs/interop/control-flow.md
@@ -1,0 +1,330 @@
+# Control flow between Supervisor and Managed node
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Authentication](#authentication)
+- [Node `->` Supervisor](#node---supervisor)
+  - [Reset](#reset)
+  - [UnsafeBlock](#unsafeblock)
+  - [DerivationUpdate](#derivationupdate)
+  - [DerivationOriginUpdate](#derivationoriginupdate)
+  - [ExhaustL1](#exhaustl1)
+  - [ReplaceBlock](#replaceblock)
+- [Supervisor `->` Node](#supervisor---node)
+  - [Control Signals](#control-signals)
+    - [interop_pullEvent](#interop_pullevent)
+    - [interop_anchorPoint](#interop_anchorpoint)
+    - [interop_invalidateBlock](#interop_invalidateblock)
+    - [interop_provideL1](#interop_providel1)
+    - [interop_reset](#interop_reset)
+  - [DB](#db)
+    - [interop_updateCrossSafe](#interop_updatecrosssafe)
+    - [interop_updateCrossUnsafe](#interop_updatecrossunsafe)
+    - [interop_updateFinalized](#interop_updatefinalized)
+  - [Sync Methods](#sync-methods)
+    - [interop_fetchReceipts](#interop_fetchreceipts)
+    - [interop_l2BlockRefByTimestamp](#interop_l2blockrefbytimestamp)
+    - [interop_blockRefByNumber](#interop_blockrefbynumber)
+    - [interop_chainID](#interop_chainid)
+    - [interop_outputV0AtTimestamp](#interop_outputv0attimestamp)
+    - [interop_pendingOutputV0AtTimestamp](#interop_pendingoutputv0attimestamp)
+- [Types](#types)
+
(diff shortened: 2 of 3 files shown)


specs/SUMMARY.md
```diff
@@ -78,6 +78,7 @@
     - [Sequencer](./interop/sequencer.md)
     - [Verifier](./interop/verifier.md)
     - [Supervisor](./interop/supervisor.md)
+    - [Control Flow](./interop/control-flow.md)
```
emhane: I would just call this file managed-mode.md and section Managed Mode. since it's the term we are all familiar with now, wdyt?

protolambda: +1, the label is useful, and we may add more modes later based on needs of users.

specs/interop/control-flow.md
```diff
@@ -0,0 +1,330 @@
...
+## Authentication
+
+Supervisor [initiates the connection](https://github.com/ethereum-optimism/optimism/issues/13182) to the node in order
+to manage it. As supervisor does the dial in, for our context, it is the client and the node is the server. For the web
+socket authentication, only the initial request is authenticated and a live socket is opened. Subsequent requests need
+not have the authentication parameter.
+
+The authentication is performed using `JWT`. The construction of jwt is similar to the one
+[used in Engine API](https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md). The path of
+the file containing the jwt secret should be provided as part of the L2 config parameter. If no input is provided, a new
```
protolambda:
```suggestion
the file containing the jwt secret should be provided to the supervisor instance. If no input is provided, a new
```

specs/interop/control-flow.md
```diff
@@ -0,0 +1,330 @@
...
+Both cases should be handled because `op-node` and `op-supervisor` can be using different client implementations.
+In both the case, we use `event` based signaling of new information. Ideally, this should be done using bi-directional
+RPC using web sockets.
+
+## Authentication
+
+Supervisor [initiates the connection](https://github.com/ethereum-optimism/optimism/issues/13182) to the node in order
+to manage it. As supervisor does the dial in, for our context, it is the client and the node is the server. For the web
+socket authentication, only the initial request is authenticated and a live socket is opened. Subsequent requests need
+not have the authentication parameter.
```
protolambda:
```suggestion
socket authentication, only the initial request is authenticated and a live socket is opened.
No re-authentication is needed during the lifetime of this socket.
```

specs/interop/control-flow.md
```diff
@@ -0,0 +1,330 @@
...
+UnsafeBlock: BlockRef //L2's BlockRef
+```
+
+New L2 unsafe block was processed, updating local-unsafe head.
+
+### DerivationUpdate
+
+```javascript
+DerivationUpdate: DerivedBlockRefPair {
+    Source: L1BlockRef
```
protolambda: I suggest using `BlockRef` everywhere, it's used for both L1 and L2. We avoided using the expanded `L2BlockRef` in the API, and the L1/L2 distinction is a legacy thing of the op-node that we should not enshrine here.

specs/interop/control-flow.md
```diff
@@ -0,0 +1,330 @@
...
+payload: Hash -> Receipts //L2 block hash
+```
+
+Fetches all transaction receipts in a given L2 block.
+
+#### interop_l2BlockRefByTimestamp
+
+```javascript
+payload: uint64 -> L2BlockRef
+```
```
protolambda: We should change this to `BlockRef`. Using the `L2BlockRef` in the API is completely unnecessary, and a mistake in the op-supervisor implementation.

specs/interop/control-flow.md
```diff
@@ -0,0 +1,330 @@
...
+
+#### interop_pullEvent
+
+```javascript
+payload: none -> One of the event from node events (previous section).
+```
+
+When websocket is not available, supervisor can use this method to get the next event from node.
+
+#### interop_anchorPoint
```
protolambda: This method is being removed soon, in favor of a config that is loaded by the supervisor client. cc @sebastianst is working on this. See https://github.com/ethereum-optimism/optimism/pull/16038

specs/interop/control-flow.md
```diff
@@ -0,0 +1,330 @@
...
+    Hash: 32 Bytes
+    Number: uint64
+    ParentHash: 32 Bytes
+    Time: uint64
+}
+
+L1BlockRef = BlockRef
+
+L2BlockRef {
+    Hash: 32 Bytes
```
protolambda: Let's use `Hash` instead for all hash type values, instead of `32 Bytes`

specs/interop/control-flow.md
```diff
@@ -0,0 +1,330 @@
...
+```
+
+Returns the optimistic output of the invalidated block from replacement
+
+**Note:** All events should return relevant error(s) in case of unsuccessful calls.
+
+## Types
+
+```javascript
+Hash: keccack256 hash
```
protolambda:
```suggestion
Hash: 0x prefixed, hex encoded, fixed-length string representing 32 bytes
```

protolambda (commented):
Looks good. A few things are changing or need adjustments. And I think we'll consolidate this + the op-supervisor RPC spec + other RPCs into a shared rpc directory at some point after this PR.

specs/interop/control-flow.md
itschaindev: Should we also do the same in the rust implementation?

specs/interop/control-flow.md
itschaindev: Should I totally remove it or make an additional comment about its removal?

emhane: better make a comment about it being deprecated referencing to use config in \<link to spec pr by seb\> instead, since it seems this pr will merge first https://github.com/ethereum-optimism/optimism/pull/16038#pullrequestreview-2861126941

emhane: ah seb's pr isn't even a spec pr yet, it's an implementation poc/production impl. so yes, defo add a comment, will take some time before that is ported to specs I presume.

dhyaniarun1993 (approved):
lgtm

emhane (approved):
lgtm

protolambda (approved):

PR #716, opened 2025-06-02
Fix broken EAS.sol link to official Ethereum Attestation Service repository

VolodymyrBg:
Replaced the outdated and broken link to EAS.sol in the predeploys documentation with the correct URL pointing to the official Ethereum Attestation Service (EAS) contract in the ethereum-attestation-service/eas-contracts repository. This ensures that readers can access the latest and authoritative implementation of the EAS contract.

### DIFF
--- a/specs/protocol/predeploys.md
+++ b/specs/protocol/predeploys.md
@@ -341,7 +341,7 @@ protocol.
 
 ## EAS
 
-[Implementation](https://github.com/ethereum-attestation-service/eas-contracts/blob/master/contracts/EAS.sol)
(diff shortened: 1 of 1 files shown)


specs/protocol/predeploys.md
```diff
@@ -341,7 +341,7 @@ protocol.
 
 ## EAS
 
-[Implementation](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/EAS/EAS.sol)
```
tynes, line 344: The link should be https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts-bedrock/src/vendor/eas given that we have vendored the contracts and do not use the upstream code

mslipper (approved):

PR #739, opened 2025-07-17
fix(supervisor/rpc): clarify null representation for BlockRef and BlockID in `syncStatus` method

dhyaniarun1993:
This PR updates the Supervisor API specification to clarify how missing or non-existent block references should be represented in API responses.

- Adds a note to `SupervisorSyncStatus` specifying that if `minSyncedL1` does not exist, it must be represented as a `BlockRef` with all-zero hash and parentHash, number: 0, and timestamp: 0.
- Adds a note to `SupervisorChainSyncStatus` specifying that for the fields localSafe, crossUnsafe, safe, and finalized, if the referenced block does not exist, the `BlockID` must be represented with an all-zero hash and number: 0.


emhane (approved):
thanks!

PR #746, opened 2025-08-06
Add Block Calldata Limit Spec

teddyknox:
Documented [here](https://www.notion.so/oplabs/Calldata-Limit-Jovian-247f153ee16280868071c9a2a6e9e783), a feature of Jovian to mitigate calldata spam attacks and their detriments.

### DIFF
--- a/specs/protocol/jovian/block-calldata-limit.md
+++ b/specs/protocol/jovian/block-calldata-limit.md
@@ -1,0 +1,36 @@
+# Block Calldata Limit
+
+In order to mitigate calldata spam attacks, Jovian introduces a feature that enforces a maximum estimated L1 calldata footprint per block and changes the EIP-1559 base fee calculation to compute over a multidimensional resource usage context. Noteably, this approach preserves EVM-equivalence in transaction gas accounting, such that wallets and other tooling do not need to update their gas estimation logic properly estimate gas consumption.
+
+## Terms
+
+- **`BlockCalldataLimit`**: Protocol constant numerically the same as the block gas limit (made possible by scaling)
+- **`txCalldataFootprint`**: `sizeEstimate * calldataFootprintCost`
+- **`blockCalldataFootprint`**: Sum over all txs’ calldata footprints
+- **(Tx) Size Estimate**: The Fjord size estimate protocol constants, scaled back to 1
+    - `minTransactionSize` is `100`
+    - `intercept` is `-42_585_600`
+    - `fastlzCoef` is `836_500`
+- **`CalldataFootprintCost`**: Protocol constant representing cost per size estimate, set to `400`.
+
+### Maximum Block Size
+
+In order for a block to be valid, a block's `blockCalldataFootprint` MUST be below the `blockCalldataLimit`.
+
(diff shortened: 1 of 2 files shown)


specs/protocol/jovian/block-calldata-limit.md
```diff
@@ -0,0 +1,36 @@
...
+- **`blockCalldataFootprint`**: Sum over all txs’ calldata footprints
+- **(Tx) Size Estimate**: The Fjord size estimate protocol constants, scaled back to 1
+    - `minTransactionSize` is `100`
+    - `intercept` is `-42_585_600`
+    - `fastlzCoef` is `836_500`
+- **`CalldataFootprintCost`**: Protocol constant representing cost per size estimate, set to `400`.
+
+### Maximum Block Size
+
+In order for a block to be valid, a block's `blockCalldataFootprint` MUST be below the `blockCalldataLimit`.
```
axelKingsley, line 18: If the block *is not* valid, what is done? Is there a block replacement like in other cases?

specs/protocol/jovian/block-calldata-limit.md
```diff
@@ -0,0 +1,36 @@
...
+```py
+def block_gas_metered()
+    tx_total_gas_used = 0
+    block_da_bytes_estimate = 0
+    for tx in block_txs:
+        tx_total_gas_used += tx.gas_used
+        block_da_bytes_estimate += max(min_transaction_size, intercept + fastlz_coef * tx.fastlz_size / 1e6)
+    block_calldata_footprint = block_da_bytes_estimate * calldata_footprint_cost
+    gas_metered = max(block_calldata_footprint, tx_total_gas_used)
+    return gas_metered
```
axelKingsley, line 33: Just want to make sure I understand the logic here (and in the feature generally):

- There is some existing feature which prevents a given block from using too much gas. For the sake of argument, a block can only use 100 gas, and past that limit it is not valid.
  - When a block has 10 transactions using 8 gas each, the block uses 80 gas and would be considered "80% full" for the purposes of 1559 Fee Updates
- There is this new feature which seeks to have the volume of calldata treated identically to the gas limit. For the sake of argument, a block can only use 150b calldata, and past that limit it is not valid.
  - Because there is a different denominator (100 gas, 150 calldata), there is a scaler which adjusts them to match. In the example it would be like a 0.666x scaler on calldata. This is the `calldata_footprint_cost`
  - When the block has 10 transactions using 10 calldata each, the block uses 100 calldata and would be considered "66% full" because the denominator is 150.
- In order for Calldata and Gas to be used interchangeably for block rejection and Fee Updates, the *larger* of the two is returned. And it is always returned post-scale, so that it resembles a gas value, even when it was actually a calldata one.

geoknee: This sounds correct to me. One subtlety is that we don't deal with raw calldata, but our best estimate of the "DA footprint" of that calldata. The more compressible the calldata is, the lower the footprint.

geoknee:
Probably superceded by #748

PR #747, opened 2025-08-11
feat: Add minimum base fee specs

niran:
Converted the [minimum base fee design doc](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/minimum-base-fee.md) to specs for U17.


specs/protocol/jovian/exec-engine.md
```diff
@@ -0,0 +1,70 @@
...
+    parentBeaconBlockRoot: DATA (32 bytes)
+    transactions: array of DATA
+    noTxPool: bool
+    gasLimit: QUANTITY or null
+    eip1559Params: DATA (8 bytes) or null
+    minBaseFeeLog2: QUANTITY or null
+}
+```
+
+- Pre-Jovian: `eip1559Params` is 8 bytes and encodes `denominator` and `elasticity`.
```
geoknee: Perhaps make it clearer that this remains unchanged with Jovian?

specs/protocol/jovian/system-config.md
```diff
@@ -0,0 +1,32 @@
...
+| `minBaseFeeLog2` | `uint8` | `0`     | Log2 of the minimum base fee in wei. `0` disables the minimum base fee entirely.        |
+
+The configuration is updated via a new method on `SystemConfig`:
+
+```solidity
+function setMinBaseFeeLog2(uint8 minBaseFeeLog2) external onlyOwner;
+```
+
+Upon update, the contract emits a `ConfigUpdate` event with a new `UpdateType` value `MINIMUM_BASE_FEE`, enabling nodes
+to derive the configuration from L1 logs.
```
geoknee: We should specify numerical values for:
* version
* type
* value
for the `ConfigUpdate` event, and add this to the table here https://specs.optimism.io/protocol/system-config.html#system-config-updates.

Lets avoid using `MINIMUM_BASE_FEE` especially as it does not refer to the log2 aspect. We want to avoid any confusion about relationship between the storage value and the event value (both should be the log2 of the minimum base fee).

specs/protocol/jovian/system-config.md
```diff
@@ -0,0 +1,32 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Minimum Base Fee Configuration
+
+Jovian adds a new configuration value to `SystemConfig` to control the minimum base fee used by the EIP-1559 fee market
+on OP Stack chains. The value represents the log2 of the minimum base fee in wei.
+
+| Name             | Type    | Default | Meaning                                                                                 |
+|------------------|---------|---------|-----------------------------------------------------------------------------------------|
+| `minBaseFeeLog2` | `uint8` | `0`     | Log2 of the minimum base fee in wei. `0` disables the minimum base fee entirely.        |
```
geoknee: When we say `0` disables the feature, we should be more precise about how that is implemented. For example, this magic value can be totally opaque to the smart contracts layer, and it is only the _execution clients_ that need to detect it as a special case. Of course, the smart contract setter is the user interface to the feature so it needs to document the meaning of the `0` value. But in terms of specification, let's be clear that it is only downstream code that does the disabling.

specs/protocol/jovian/exec-engine.md
```diff
@@ -0,0 +1,70 @@
...
+| `minBaseFeeLog2` | `u8`               | `[9, 10)`   |
+
+Constraints:
+
+- `version` MUST be `1` (incremented from Holocene's `0`).
+- `denominator` MUST be non-zero.
+- There MUST NOT be any data beyond these 10 bytes.
+
+The `minBaseFeeLog2` field encodes the base-2 logarithm of the minimum base fee in wei. A value of `0` disables the
+minimum base fee entirely.
```
geoknee: I think some pseudo-code here would be good:
```
	if minBaseFeeLog2 > 0 {
		minBaseFee = 2**minBaseFeeLog2
		if baseFee < minBaseFee {
			baseFee = minBaseFee
		}
	}
```

specs/protocol/jovian/exec-engine.md
```diff
@@ -0,0 +1,70 @@
...
+
+Jovian introduces a configurable minimum base fee to reduce the duration of priority-fee auctions on OP Stack chains.
+
+The minimum base fee is configured via `SystemConfig` (see `./system-config.md`) and enforced by the execution engine
+via the block header `extraData` encoding and the Engine API `PayloadAttributesV3` parameters.
+
+### Minimum Base Fee in Block Header
+
+Like [Holocene's dynamic EIP-1559 parameters](../holocene/exec-engine.md#dynamic-eip-1559-parameters), Jovian encodes
+fee parameters in the `extraData` field of each L2 block header. The format is extended to include an additional byte
```
geoknee: We are [committed to feature/fork separation](https://github.com/ethereum-optimism/design-docs/pull/300/files), so I was initially going to suggest we strip out mention of "Jovian" from the body of the `## Minimum Base Fee` section to mirror that philosophy.

However, I realise that we cannot easily specify this feature without presuming it will go into a certain hardfork (i.e. Jovian). That is because we essentially need to "lock" or reserve slots in the block and engine API schemas and specify them with absolute positions. I think it would be very messy indeed to try and do that in some hardfork agnostic way with relative positions or some such.

Therefore let's try to keep to the feature/fork separation where possible in the implementation, to make a remapping of the feature to a different hardfork easier (in the unlikely event). But we can continue to assume it is scoped for Jovian (for now) where it is infeasible to do otherwise.

specs/protocol/jovian/exec-engine.md
```diff
@@ -0,0 +1,70 @@
...
+
+- [Minimum Base Fee](#minimum-base-fee)
+  - [Minimum Base Fee in Block Header](#minimum-base-fee-in-block-header)
+  - [Minimum Base Fee in `PayloadAttributesV3`](#minimum-base-fee-in-payloadattributesv3)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Minimum Base Fee
+
+Jovian introduces a configurable minimum base fee to reduce the duration of priority-fee auctions on OP Stack chains.
```
geoknee: You could link out to the design doc here in case the user wants to know more about the motivation.

specs/protocol/jovian/exec-engine.md
sebastianst: Just chatted with George about this. I think the feature/fork separation is more of an implementation detail and shouldn't be as visible in the specs and we should just write the specs for any feature assuming it goes into a specific fork. If it then were to be removed from a fork, we'd then move that feature's sections from the specs into another/the next fork. So let's keep it as is for now.

specs/protocol/jovian/system-config.md
axelKingsley: Does the Execution Client even need to special-case this? An enforced minimum of 0 seems to be a no-op that doesn't need any branched logic.

We could just say that "At 0, this feature is effectively disabled, because all fees are higher than 0"

specs/protocol/exec-engine.md
```diff
@@ -299,6 +304,8 @@ or a zero `bytes32` if the Dencun functionality with `parentBeaconBlockRoot` is
 Starting with Holocene, the `eip1559Params` field must encode the EIP1559 parameters. It must be `null` before.
 See [Dynamic EIP-1559 Parameters](holocene/exec-engine.md#dynamic-eip-1559-parameters) for details.
 
+Starting with Jovian, the `minBaseFeeLog2` field is added (uint8). It must be `null` before Jovian. See [Jovian Minimum Base Fee](jovian/exec-engine.md#minimum-base-fee) for details.
```
axelKingsley: Couldn't it be 0 before jovian? That way the value can be specified even pre-jovian without any impact when the HF activates.

specs/protocol/jovian/exec-engine.md
```diff
@@ -0,0 +1,70 @@
...
+| `version`        | `u8`               | `[0, 1)`    |
+| `denominator`    | `u32 (big-endian)` | `[1, 5)`    |
+| `elasticity`     | `u32 (big-endian)` | `[5, 9)`    |
+| `minBaseFeeLog2` | `u8`               | `[9, 10)`   |
+
+Constraints:
+
+- `version` MUST be `1` (incremented from Holocene's `0`).
+- `denominator` MUST be non-zero.
+- There MUST NOT be any data beyond these 10 bytes.
```
axelKingsley: If I understand correctly, `version`, `denominator`, `elasticity` are all existing aspects of the `extraData` format. Are these definitions redundant with another section of the spec? If so, maybe we just link to that and then only talk about the new byte we are extending with this feature.

specs/protocol/jovian/exec-engine.md
axelKingsley: +1, I would avoid duplicating claims made in other parts of the spec where possible, and instead just saying something like 
```suggestion
- Pre-Jovian: `eip1559Params` is [defined here](../other-spec.md)
- Jovian: As defined above, plus adds a separate `minBaseFeeLog2`...
```

specs/protocol/jovian/system-config.md
geoknee: The problem is we are setting the log2 to `0`, implying an actual minimum base fee of `1`. So we do need some special case logic to map `minBaseFeeLog2 = 0` to `minBaseFee = 0` to truly disable the feature.

axelKingsley: Ah right, `0` configured value != `0` effective value.

We just discussed offline that this value could *always* have an offset of `1`, allowing for the 0 value to map to 0. I am not strongly opinionated on if we do that.

geoknee: Interesting suggestion, let's see what @niran makes of it!

niran: This issue came up with the design doc, and it turns out that [EIP 1559 actually doesn't allow the base fee to fall that low](https://github.com/ethereum-optimism/design-docs/pull/295#discussion_r2178810649). The theoretical minimum base fee is `BASE_FEE_MAX_CHANGE_DENOMINATOR - 1` in wei, but even that can't be reached because of the system transaction in "empty" blocks. There are several live OP Stack chains that illustrate this behavior with mostly empty blocks, but a base fee around 252 wei.

I think the offset is the conceptually superior way to implement this, but it would make the documentation to explain the values much messier. I'd like to keep it as is.

specs/protocol/exec-engine.md
niran: Good question! These fork boundary conditions are where I have the least context. It was defined this way to be consistent with the Holocene specs. I'm fine with making that change.

specs/protocol/jovian/exec-engine.md
niran: ddcaa69fe4b63a01aba73d81748181f549a8dff4

specs/protocol/jovian/system-config.md
niran: 43ab65c304ef67284f2f0d569d267b67f80d0edb

specs/protocol/jovian/exec-engine.md
niran: 08124edff723c0b1eb87ef42491e835bff1a947e

specs/protocol/jovian/exec-engine.md
niran: 6dd14b494061e8ca874a04da9547e9a6edb3b9e7

specs/protocol/jovian/exec-engine.md
niran: e18495f26c5dcf6111b6dbfa2c7b199b94b6828e

specs/protocol/exec-engine.md
geoknee: We touched on this in design review somewhat. It was decided that we prefer to activate the feature in a hard fork in the usual way, where the new data field does not exist before the fork activates. The feature (i.e. additional data in the block header, even if zero) might break tools that want to parse the block header, so best to do that in a coordinated and well communicated way. So I think the spec is good as is, here.

axelKingsley:
Hey @niran , sorry for the late breaking questions, but @geoknee and I were talking about this feature, and I was wondering if there were other encoding schemes which might be more useful.

I noticed in the design doc that only a small window of log2 values were used as an example, and so I went and pulled the full range up with a quick spreadsheet. Due to the exponential nature of the encoding, seems that there's a ton of unhelpful values.

For example, 
- the first 10 values are basically meaningless, imposing a minimum of 512 wei or below.
- value 19 costs 0.0005 gwei
- value 29 costs 0.5 gwei
- value 39 costs 549 gwei
- value 60 costs over *1 Eth*!
- value 256 costs 5*E+58 Eth

And since we are using Log2, any other exponentiation (like base10) would produce even larger ranges. Seems like we are giving people 256 choices, with only ~10 good options in a somewhat arbitrary range inside it.

Would it make sense to make a more comprehensive encoding such that the value range is useful throughout? Or alternatively, are there any ways to use less than a byte?

One encoding I could imagine would be:
bytes 0-4: how many wei base
bytes 5-8: log10 exponentiation

This has the following advantages:
- 0 value maps to a true 0 value, so no branching required
- minimum base fee always appears 10s based
- easily scale the minimum within a given band by changing the base without changing the exponent (ie go from 3k gwei to 7k gwei just by changing the base)
- exponentiation still gives a range all the way up to 0.015 Eth at maximum (maximum base 15, maximum exponent E+15)

niran:
If you're suggesting expanding the field to 8 bytes, we probably don't need any fancy encoding. `MAX_UINT64` is over 18 ETH. (Reading the proposal again, it seems more likely that you were talking about bits, so 0-15 wei times 10^(0-15), which covers one significant digit from 0 wei to 10,000,000 gwei plus a little more flexibility for values that start with 1. I think that's probably sufficient.)

The log2 approach came from the framing that we're trying to let chain operators shave off doublings for the base fee to rise from 0, so they'd just specify how many doublings to skip. Representing the value as a byte does make it seem weird for there to be so few usable options, but I think the ratio of usable options for any other data type would be worse, just harder to count. But representing the value in a simpler way is generally better if we're fine with taking up more space in block headers (and in blobs). The extra data field can hold 32 bytes, of which Holocene used 9.

I think moving from one byte to eight bytes is viable. An alternative is to switch from exponential **values** to exponential **deltas** between values, and to limit the range of possible values. This would make all 256 values useful. The downside is that we'd need to do fixed point decimal math to implement this. Here's what that looks like with a range of 0 to 1000 Gwei as possible minimum base fee values:

| Parameter         | Value                  |
|-------------------|------------------------|
| RANGE_START       | 0                      |
| RANGE_END         | 1e12 (1000 Gwei)       |
| INCREMENT_FACTOR  | 1.104623                    |

```
minBaseFee = RANGE_START + (RANGE_END - RANGE_START) * (INCREMENT_FACTOR^minBaseFeeConstant - 1) / (INCREMENT_FACTOR^255 - 1)
```

Here's a sampling of a tenth of the possible values:

n | value
-- | --
0 | 0 wei
10 | 16 wei
20 | 60 wei
30 | 180 wei
40 | 502 wei
50 | 1374 wei
60 | 3734 wei
70 | 10115 wei
80 | 27377 wei
90 | 74066 wei
100 | 200353 wei
110 | 541939 wei
120 | 0.001466 gwei
130 | 0.003965 gwei
140 | 0.010725 gwei
150 | 0.029008 gwei
160 | 0.078463 gwei
170 | 0.212228 gwei
180 | 0.574042 gwei
190 | 1.552690 gwei
200 | 4.199770 gwei
210 | 11.359686 gwei
220 | 30.726079 gwei
230 | 83.108981 gwei
240 | 224.796101 gwei
250 | 608.036418 gwei
255 | 1,000.000000 gwei

So we have three options:
1. Stick with `minLogFeeBase2`.
2. Use @axelKingsley's four bit base + four bit log10 format.
3. Squeeze the exponential curve to fit meaningful values within a specified range.

I think the simplest option that does the trick is probably option 2.

niran:
Ok, after thinking about it more, the base + log10 format seems much better. Starting to make that change.

axelKingsley:
❤️ thank you @niran for the detailed introspection!

To be clear, I would not want us to move to 8 bytes, I find that to be wasteful. Rather, I wanted the single byte to have a better concentration. Your description "four bit base + four bit log10 format" is spot on.

And than you for the "doubling" line of thinking, that helps me understand the reasoning there.

specs/protocol/jovian/exec-engine.md
```diff
@@ -0,0 +1,72 @@
...
+Constraints:
+
+- `version` MUST be `1` (incremented from Holocene's `0`).
+- There MUST NOT be any data beyond these 10 bytes.
+
+The `minBaseFeeFactors` field encodes minimum base fee in wei as a 4-bit significand followed by a
+4-bit exponent to the power of 10. A value of `0` disables the minimum base fee entirely.
+
+```javascript
+if minBaseFeeFactors > 0 {
```
geoknee: I think with this new encoding, it would be possible to not have this outer `if` statement. Any `minBaseFeeFactors` value which implies a `0` significand would be sufficient to effectively disable the feature. That said, I'm fine with keeping it if you think it is clearer.

specs/protocol/jovian/exec-engine.md
```diff
@@ -0,0 +1,72 @@
...
+  significand = minBaseFeeFactors >> 4 & 0x0F
+  exponent = minBaseFeeFactors & 0x0F
+  minBaseFee = significand * 10**exponent
```
geoknee: Could we include a quick word on the maximum value that can be set? By my reckoning this is $1.5 x 10^{16}$ wei or (when the `minBaseFeeFactors=0xff`).

specs/protocol/jovian/exec-engine.md
```diff
@@ -0,0 +1,72 @@
...
+    suggestedFeeRecipient: DATA (20 bytes)
+    withdrawals: array of WithdrawalV1
+    parentBeaconBlockRoot: DATA (32 bytes)
+    transactions: array of DATA
+    noTxPool: bool
+    gasLimit: QUANTITY or null
+    eip1559Params: DATA (8 bytes) or null
+    minBaseFeeFactors: QUANTITY or null
+}
+```
```
geoknee, line 68: Let's add a section on "rationale" like this one https://github.com/ethereum-optimism/specs/blob/780c878cef5d91456c50fdbdf882da4832210e45/specs/protocol/holocene/exec-engine.md#rationale.

This came up when I was discussing with @sebastianst. An alternative design could store this consensus information in L2 state instead of the block header, but this would require us to reach into the state when sealing a block.

specs/protocol/jovian/exec-engine.md
niran: 7bb6f5c60a5e56b1523d8b513218affee26ec214

specs/protocol/jovian/exec-engine.md
niran: b19ce9287882a4a3ecff48eed7827b4c1113c717

specs/protocol/jovian/exec-engine.md
niran: aa32b369530070c6df84a5e580b7b7c1b2e74603

specs/protocol/jovian/system-config.md
axelKingsley: I believe this topic is closed by way of adopting an 8-byte format for the base min fee, which means input values of 0 will map directly to a minimum of 0.

axelKingsley:
To close the loop on the formatting of the value in the header:

After some deliberation, we've decided to go with an 8-byte encoding of the Min Base Fee, which gives us maximum control and eliminates all the value spread issues we've been discussing.

I believe everyone in this PR was already involved in this decision, just reporting it here for clarity.

niran:
The specs have been updated to use the final u64 format as implemented in ethereum-optimism/op-geth#648 and ethereum-optimism/optimism#16852.

axelKingsley (approved):

PR #748, opened 2025-08-13
Jovian: DA Footprint

### LINKED ISSUE #17087: [DA Footprint Limit] Write spec

axelKingsley:
Draft, starting with iteratively constructing the terms such that a top-to-bottom read fully explains the whole concept.

TODO:
- [x] PR feedback
- [x] "Justification" section to explain why this feature exists?
- [x] glossary integrations, cross-links

Closes https://github.com/ethereum-optimism/optimism/issues/17087

### DIFF
--- a/specs/protocol/jovian/calldata-block-limit.md
+++ b/specs/protocol/jovian/calldata-block-limit.md
@@ -15,7 +15,7 @@ Calldata is the `data` field attached to a transaction submitted to the OP Chain
 da_footprint = max(min_transaction_size, intercept + fastlz_coef * tx.fastlz_size / 1e6)
 ```
 
-The protocol constants `min_transaction_size`, `intercept`, `fastlz_coef` are defined in prior Hard Forks.
+The protocol constants `min_transaction_size`, `intercept`, `fastlz_coef` are defined in [Fjord LZ Estimation](../fjord/exec-engine#fjord-l1-cost-fee-changes-fastlz-estimator).
 
 ## `block_da_footprint`
 `block_da_footprint` represents the cumulative total of `da_footprint` for all transactions included in the given block.
--- a/specs/protocol/jovian/calldata-block-limit.md
+++ b/specs/protocol/jovian/calldata-block-limit.md
@@ -1,0 +1,51 @@
+
+# Calldata Limit
+Starting with Jovian, the `gasUsed` field of a block may instead represent the utilization of calldata within the block.
+The purpose of this feature is to take into account the throughput of the DA Layer (the L1) when creating blocks that feature a high volume of `calldata`
+
+# Terms 
+
+## `calldata`
+Calldata is the `data` field attached to a transaction submitted to the OP Chain. The transaction author fully controls the content and size of the calldata at the time of authoring.
+
+## `da_bytes_estimate`
+`da_bytes_estimate` is the size of the `calldata` present in a transaction with respect to how much space it is expected to consume on the DA Layer. The DA Bytes Estimate may be smaller thatn the raw Calldata because the data is compressed when published to L1.
+
+```py
+da_bytes_estimate = max(min_transaction_size, intercept + fastlz_coef * tx.fastlz_size / 1e6)
+```
+
+The protocol constants `min_transaction_size`, `intercept`, `fastlz_coef` are defined in prior Hard Forks.
+
+## `block_da_bytes_estimate`
+`block_da_bytes_estimate` represents the cumulative total of `da_bytes_estimate` for all transactions included in the given block.
+
+```py
+block_da_bytes_estimate = 0
+for tx in block_txs:
+    block_da_bytes_estimate += tx.da_bytes_estimate
+```
+
+## `calldata_footprint_cost`
+`calldata_footprint_cost` is a coefficient value which normalizes the `block_da_bytes_estimate` such that it can be compared to the `gasLimit` of the block.
+The value of `calldata_footprint_cost` is 400, which is an arbitrary coefficient selected by analysis of real chain data.
+
+```py
+calldata_footprint_cost = da_bytes_estimate * calldata_footprint_cost
+```
+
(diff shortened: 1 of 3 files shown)


specs/protocol/jovian/calldata-block-limit.md
```diff
@@ -0,0 +1,51 @@
...
+Starting with Jovian, the `gasUsed` field of a block may instead represent the utilization of calldata within the block.
+The purpose of this feature is to take into account the throughput of the DA Layer (the L1) when creating blocks that feature a high volume of `calldata`
+
+# Terms 
+
+## `calldata`
+Calldata is the `data` field attached to a transaction submitted to the OP Chain. The transaction author fully controls the content and size of the calldata at the time of authoring.
+
+## `da_bytes_estimate`
+`da_bytes_estimate` is the size of the `calldata` present in a transaction with respect to how much space it is expected to consume on the DA Layer. The DA Bytes Estimate may be smaller thatn the raw Calldata because the data is compressed when published to L1.
```
geoknee:
```suggestion
`da_bytes_estimate` is the size of the `calldata` present in a transaction with respect to how much space it is expected to consume on the DA Layer. The DA Bytes Estimate may be smaller than the raw Calldata because the data is compressed when published to L1.
```

specs/protocol/jovian/calldata-block-limit.md
```diff
@@ -0,0 +1,51 @@
+
+# Calldata Limit
+Starting with Jovian, the `gasUsed` field of a block may instead represent the utilization of calldata within the block.
+The purpose of this feature is to take into account the throughput of the DA Layer (the L1) when creating blocks that feature a high volume of `calldata`
```
geoknee:
```suggestion
The purpose of this feature is to take into account the throughput of the DA Layer (the L1) when creating blocks that feature a high volume of `calldata`.
```

specs/protocol/jovian/calldata-block-limit.md
```diff
@@ -0,0 +1,51 @@
...
+Calldata is the `data` field attached to a transaction submitted to the OP Chain. The transaction author fully controls the content and size of the calldata at the time of authoring.
+
+## `da_bytes_estimate`
+`da_bytes_estimate` is the size of the `calldata` present in a transaction with respect to how much space it is expected to consume on the DA Layer. The DA Bytes Estimate may be smaller thatn the raw Calldata because the data is compressed when published to L1.
+
+```py
+da_bytes_estimate = max(min_transaction_size, intercept + fastlz_coef * tx.fastlz_size / 1e6)
+```
+
+The protocol constants `min_transaction_size`, `intercept`, `fastlz_coef` are defined in prior Hard Forks.
```
geoknee: Can link to Fjord here https://specs.optimism.io/protocol/fjord/exec-engine.html#fjord-l1-cost-fee-changes-fastlz-estimator

specs/protocol/jovian/calldata-block-limit.md
```diff
@@ -0,0 +1,51 @@
...
+for tx in block_txs:
+    block_da_bytes_estimate += tx.da_bytes_estimate
+```
+
+## `calldata_footprint_cost`
+`calldata_footprint_cost` is a coefficient value which normalizes the `block_da_bytes_estimate` such that it can be compared to the `gasLimit` of the block.
+The value of `calldata_footprint_cost` is 400, which is an arbitrary coefficient selected by analysis of real chain data.
+
+```py
+calldata_footprint_cost = da_bytes_estimate * calldata_footprint_cost
```
geoknee: Something is not right here, you have the same name on both sides :-) . I think you need to give this expression a name like `calldata_pseudo_gas`.

While we are on naming, why not use "footprint" instead of da_estimate?

So then we have 

```
da_footprint_pseudo_gas = da_footprint_estimate * da_footprint_cost
```

Thinking on this some more, I don't know if we actually want to use "calldata" in the variable names or even the feature name. We are operating solely with the estimated DA footprint of a transaction. This is typically some function of its calldata, but it is almost never going to be equal to its calldata. Hence my suggested terminology above.

specs/protocol/jovian/calldata-block-limit.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+## `block_gas_limit`
+`block_gas_limit` represents the maximum allowable gas accumulated by all transactions in a given block. Above this limit, the block is not valid by protocol rules.
+
+## `block_da_bytes_limit`
+`block_da_bytes_limit` represents the maximum allowable size of a block's `block_da_bytes_estimate`.
+This is equal to `(block_gas_limit / calldata_footprint_cost)`, but in practice the `block_da_bytes_estimate` is normalized to be comparable with the `block_gas_limit`, so `block_da_bytes_limit` is not used directly.
```
geoknee: I think this is a useful "virtual" quantity, helps us get a better feel for the feature.

specs/protocol/jovian/calldata-block-limit.md
axelKingsley: I really like this, avoiding calldata in the name allows for this da footprint to be something more complex later.

and yes, the equation is wrong, ideally this section just says `footprint_cost = 400`, and then the *next* section shows this equation.

will realign on these terms

geoknee:
@axelKingsley this PR needs a lint and conflict resolution 🙏

geoknee:
We should add information about how the parameter is configured (via SystemConfig and L1Block contract, we can use previous specs to inform how this goes).

geoknee:
* Also nice to have a crisp justification for storing the parameter info in the L1 Block (versus the block header as we did for min base fee).
* Should be clear about _whether_ and _how_ this feature depends on min base fee.

geoknee:
Taking this forward here https://github.com/ethereum-optimism/specs/pull/767, we will add @axelKingsley as a co-author to that PR.

PR #750, opened 2025-08-25
feat: u16a spec

smartcontracts:
Proposed specification for U16a. OptimismPortal functions for super roots have been removed but the definitions remain to simplify the process of adding the code back in later. Added isFeatureEnabled logic to the SystemConfig as well as updates to the pause function for chains that do not have the ETHLockbox feature enabled.


specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -323,13 +322,19 @@ see this as a critical system risk.
 
 - MUST only be callable by the ProxyAdmin or its owner.
 - MUST only be triggerable once.
-- MUST set the value of the `SystemConfig` contract.
-- MUST set the value of the `AnchorStateRegistry` contract.
-- MUST set the value of the `ETHLockbox` contract.
```
mbaxter, line 328: Where is the EthLockbox address configured now?

specs/fault-proof/stage-one/optimism-portal.md
```diff
@@ -460,39 +435,18 @@ Allows a user to [finalize](#finalized-withdrawal) a withdrawal transaction.
 - MUST revert if the withdrawal being finalized does not pass `checkWithdrawal`.
 - MUST mark the withdrawal as finalized.
 - MUST set the L2 Withdrawal Sender variable correctly.
-- MUST unlock ETH from the ETHLockbox if the withdrawal includes an ETH value.
+- MUST unlock ETH from the ETHLockbox if the withdrawal includes an ETH value ONLY IF the
+  OptimismPortal has an ETHLockbox configured AND the ETHLockbox system feature is active.
```
mds1: I think it will be more clear to leave all the original text, and just convert the applicable parts into conditional statements:
- If v4.1.0 and no ETH Lockbox then ....
- If v4.1.0 and ETH Lockbox then ....
- If v5.0.0 then ....

Sentences like this currently don't describe the behavior of what to do when there is no ETHLockbox, since a valid 4.1.0 chain really has two valid configs you can run in prod. This will also make it easier to update the spec for v5.0.0 because we just delete the 4.1.0 branches

specs/protocol/system-config.md
```diff
@@ -194,6 +207,18 @@ In version `0`, the following update types are supported:
 - MUST set the SuperchainConfig contract address to the provided value.
 - MUST clear the old dispute game factory address from storage (now derived from OptimismPortal).
 
+### setFeatureEnabled
+
+- Used to enable or disable a [Customizable Feature](#customizable-feature).
+- Takes a bytes32 feature string and a boolean as an input.
+- MUST only be triggerable by the ProxyAdmin or its owner.
+- MUST toggle the feature flag on or off, based on the value of the boolean.
```
mds1: Will any features need more config than a bool, e.g. should we use a uint8 or uint256 here? The default can be that 0=disabled, 1=enabled, >=2 is invalid, but certain features can allow values of >=2 for more granular configs

specs/protocol/system-config.md
```diff
@@ -245,13 +270,15 @@ Returns the block number at which the op-node can start searching for logs.
...
+chain's `ETHLockbox` address or the chain's `OptimismPortal` address as the
+[Pause Identifier](./stage-1.md#pause-identifier).
 
-- MUST return true if `SuperchainConfig.paused(optimismPortal().ethLockbox())` returns true OR if
-  `SuperchainConfig.paused(address(0))` returns true.
+- MUST return true if `SuperchainConfig.paused(optimismPortal().ethLockbox())` returns true AND the
+  system is configured to use the `ETHLockbox` contract.
+- MUST return true if `SuperchainConfig.paused(optimismPortal())` returns true AND the system is
+  NOT configured to use the `ETHLockbox` contract.
+- MUST return true if `SuperchainConfig.paused(address(0))` returns true.
```
mds1: What if it has an ETH Lockbox and `SuperchainConfig.paused(optimismPortal())` returns true? I think we want to update the pause method to prevent that misconfiguration

specs/fault-proof/stage-one/optimism-portal.md
smartcontracts: The ETHLockbox address either exists in the contract because of U16 or doesn't exist at all (there's no way to enable ETHLockbox in U16a).

specs/protocol/system-config.md
smartcontracts: A change to uint8 should be backwards compatible with a boolean, it'll get represented as 0 or 1. We can start with boolean for now and switch to uint8 later if we find a use case.

specs/protocol/system-config.md
smartcontracts: We'd ignore `SuperchainConfig.paused(optimismPortal())` if ETHLockbox is enabled.

specs/protocol/system-config.md
mds1: It would be a breaking change to the solidity ABI and function signatures though, but I agree starting with a bool should be ok and we can always change this later

specs/protocol/system-config.md
mds1: So it would be possible to call `pause(portal)` on a chain that has an ETH lockbox, and that would silently be a no-op? It feels safer to revert in that case

specs/protocol/system-config.md
smartcontracts: Hmm good point, I *think* it should actually be fine for bool => uint8 specifically because they both sit in one byte, switching to uint256 could cause more problems.

specs/protocol/system-config.md
smartcontracts: We have no clean way to revert since `pause(portal)` is in the SuperchainConfig and we've abstracted this so the SuperchainConfig has no concept of the ETHLockbox or OptimismPortal.

Our two options here would either be to:
- Keep the code as-is, pause(portal) does nothing if you use ETHLockbox and vice-versa
- Respect pause(portal) but we need to reduce the pause time to 2 months instead of 3 months because in the worst case we'd now have 3 pauses (2*3 = 6 months, fits in Stage 1 requirement)

specs/protocol/system-config.md
mds1: Storage slot wise uint8 and bool are compatible, but it still does change the function selectors (`bytes4(keccak256(setFeatureEnabled(bytes32,bool))` -> `bytes4(keccak256(setFeatureEnabled(bytes32,uint8))`), though I'm not sure if that will matter. Probably not much since chain operators will just call a different method now?

specs/protocol/system-config.md
mds1: I see. I would say we keep the code as-is because that's simpler then changing durations. Then we just make sure this limitation is documented in a code comment, and make sure our off-chain pause tooling accounts for this—the tooling should revert if you pass in a portal address for a chain that has a lockbox, and the tooling should error if the pause state is false afterwards, this way there's two total spots to catch this misue

specs/protocol/system-config.md
mds1: If we agree that changing selectors doesn't matter, we can just start introducing this as a boolean and if we have to, we can move to uint8 in a future version and just deal with the breaking change to the interface. I can't imagine us needing more than 256 config values, so that should be fine.

smartcontracts: Sounds good, let's move forward with boolean for now, I think we could always emulate more complex behavior with multiple features if we really needed to avoid breaking the ABI

specs/protocol/system-config.md
```diff
@@ -245,13 +270,15 @@ Returns the block number at which the op-node can start searching for logs.
 
 ### paused
 
-This function integrates with the [Pause Mechanism](./stage-1.md#pause-mechanism) by using the
-chain's `ETHLockbox` address as the [Pause Identifier](./stage-1.md#pause-identifier). Returns the
-current pause state of the system by checking if the `SuperchainConfig` is paused for this chain's
-`ETHLockbox`.
+This function integrates with the [Pause Mechanism](./stage-1.md#pause-mechanism) by using either the
```
Inphi: the Pause Mechanism spec in that document should be updated to allow the `OptimismPortal` address to be used as an identifier. It currently says only the lock box and 0 are valid identifiers.

Inphi (approved):
Looks good.

specs/protocol/system-config.md
smartcontracts: Will do

specs/protocol/stage-1.md
```diff
@@ -117,13 +117,21 @@ identifier determines which systems or chains are affected by the pause:
...
   chains associated with that identifier.
 
-The identifier must be an `ETHLockbox` address or `address(0)`. This allows for targeted pausing of
-either specific chains, the interop set (which shares an `ETHLockbox` contract), or the all chains
-that share the same `SuperchainConfig` when `address(0)` is used as the identifier.
-
-OP Chains are expected to integrate with the `SuperchainConfig` via their `SystemConfig` contract,
-which will check for the status of the pause by passing along the address of the `ETHLockbox`
-being used within that system as the Pause Identifier.
+(-v4.1.0) The identifier must be an `ETHLockbox` address or `address(0)`. This allows for targeted
```
AmadiMichael, line 120: it stated here that `The identifier must be an ETHLockbox address or address(0). This allows for targeted pausing of either specific chains, the interop set (which shares an ETHLockbox contract), or all
chains that share the same SuperchainConfig when address(0) is used as the identifier.`. It says how to pause a set of chains in the interop set using the ETHLockbox, and how to pause chains that share the same superchainConfig via `address(0)`. It does  say how to pause specific chains but not how

AmadiMichael: oh i have just seen it further down

AmadiMichael: though i do not understand the use of (-v4.1.0) and (+v4.1.0) here

mds1: Looks like the way to read that syntax is (-v4.1.0) means that spec item is removed in v4.1.0 but was present before, and (+v4.1.0) means it's new to v4.1.0

specs/protocol/stage-1.md
```diff
@@ -117,13 +117,21 @@ identifier determines which systems or chains are affected by the pause:
...
+pausing of either specific chains, the interop set (which shares an `ETHLockbox` contract), or all
+chains that share the same `SuperchainConfig` when `address(0)` is used as the identifier.
+
+(-v4.1.0) OP Chains are expected to integrate with the `SuperchainConfig` via their `SystemConfig`
+contract, which will check for the status of the pause by passing along the address of the
+`ETHLockbox` being used within that system as the Pause Identifier.
+
+(+v4.1.0) The identifier must be an `ETHLockbox` address, an `OptimismPortal` address, or
+`address(0)`. This allows for targeted pausing of specific chains, the interop set (which shares an
+`ETHLOckbox` contract), or all chains that share the same `SuperchainConfig` when `address(0)` is
```
mds1:
```suggestion
`ETHLockbox` contract), or all chains that share the same `SuperchainConfig` when `address(0)` is
```

mds1 (approved):

specs/protocol/stage-1.md
smartcontracts: Yes, sorry this is the intended usage. I landed on it after trying a few different things like (Added in v4.1.0) but it gets to be a lot of text all over the place

PR #752, opened 2025-08-26
TimelockGuard

alcueca:
Specs for [TimelockGuard](https://github.com/ethereum-optimism/design-docs/pull/324)

### DIFF
--- a/specs/governance/timelock-guard.md
+++ b/specs/governance/timelock-guard.md
@@ -6,149 +6,221 @@
 
 - [Overview](#overview)
 - [Definitions](#definitions)
-  - [`quorum(safe)`](#quorumsafe)
+  - [Quorum](#quorum)
+  - [Blocking Threshold](#blocking-threshold)
+  - [Active Owner](#active-owner)
+  - [Honest Owner](#honest-owner)
+  - [Malicious Owner](#malicious-owner)
+  - [Full Key Control](#full-key-control)
+  - [Joint Key Control](#joint-key-control)
+  - [Temporary Key Control](#temporary-key-control)
+  - [Multisig Liveness Failure](#multisig-liveness-failure)
+  - [Multisig Safety Failure](#multisig-safety-failure)
   - [Scheduled Transaction](#scheduled-transaction)
-  - [`scheduling_time(safe, tx)`](#scheduling_timesafe-tx)
-  - [`delay_period(safe)`](#delay_periodsafe)
+  - [Scheduling Time](#scheduling-time)
+  - [Delay Period](#delay-period)
   - [Rejected Transaction](#rejected-transaction)
-  - [`rejecting_owners(safe, tx)`](#rejecting_ownerssafe-tx)
-  - [`cancellation_threshold(safe)`](#cancellation_thresholdsafe)
-  - [Nested Safe Setup](#nested-safe-setup)
-  - [Nested Cancellation](#nested-cancellation)
+  - [Rejecting Owners](#rejecting-owners)
+  - [Cancellation Threshold](#cancellation-threshold)
 - [Assumptions](#assumptions)
+  - [aTG-001: Dishonest Users Don't Have Permanent Key Control Over a Quorum of Keys](#atg-001-dishonest-users-dont-have-permanent-key-control-over-a-quorum-of-keys)
+    - [Severity: Critical](#severity-critical)
 - [Invariants](#invariants)
+  - [iTG-001: Honest Users Can Recover From Temporary Key Control Over a Quorum of Keys](#itg-001-honest-users-can-recover-from-temporary-key-control-over-a-quorum-of-keys)
+    - [Severity: Critical](#severity-critical-1)
+- [Requirements and Contraints](#requirements-and-contraints)
+  - [Nested Cancellation](#nested-cancellation)
+  - [Module Execution](#module-execution)
+  - [Replayability Conflicts](#replayability-conflicts)
 - [Function Specification](#function-specification)
-  - [constructor](#constructor)
-  - [enable(delay_period)](#enabledelay_period)
+  - [enable](#enable)
   - [disable](#disable)
-  - [cancellationThreshold(safe)](#cancellationthresholdsafe)
-  - [delayPeriod(safe)](#delayperiodsafe)
-  - [setDelayPeriod(delay_period)](#setdelayperioddelay_period)
+  - [cancellationThreshold](#cancellationthreshold)
+  - [delayPeriod](#delayperiod)
   - [scheduleTransaction](#scheduletransaction)
   - [checkTransaction](#checktransaction)
+  - [checkModuleTransaction](#checkmoduletransaction)
+  - [checkPendingTransactions](#checkpendingtransactions)
   - [rejectTransaction](#rejecttransaction)
+  - [rejectTransactionWithSignature](#rejecttransactionwithsignature)
   - [cancelTransaction](#canceltransaction)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 ## Overview
 
-The `TimelockGuard` implements a mandatory delay on Gnosis Safe contracts as a Safe Guard. Safe Guards are external contracts that Safes call into before and after the execution of a transaction to enquire if the execution fulfils conditions implemented in the Guard, and revert if they don't. The `TimelockGuard` is a Safe Guard that implements a mandatory delay on the execution of transactions.
+The `TimelockGuard` implements a mandatory delay on Gnosis Safe contracts as a Safe Guard. Safe Guards are external
+contracts that Safes call into before and after the execution of a transaction to enquire if the execution fulfils
+conditions implemented in the Guard, and revert if they don't. The `TimelockGuard` is a Safe Guard that implements
+a mandatory delay on the execution of transactions.
 
 ## Definitions
-The following list defines variables that will be used in TimelockGuard. None of the variable names are compulsory, and if better names are found, they should be used.
+
+The following list defines variables that will be used in TimelockGuard. None of the variable names are compulsory,
+and if better names are found, they should be used.
 
 ### Quorum
+
 The number of owners required to execute a transaction.
 
 ### Blocking Threshold
-The minimum number of owners not willing to execute a transaction, that by not approving the transaction guarantee that `quorum` is not met. It is defined as `min(quorum, total_owners - quorum + 1)`.
+
+The minimum number of owners not willing to execute a transaction, that by not approving the transaction guarantee
+that `quorum` is not met. It is defined as `min(quorum, total_owners - quorum + 1)`.
 
 ### Active Owner
+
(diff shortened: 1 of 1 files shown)


specs/governance/timelock-guard.md
```diff
@@ -0,0 +1,136 @@
...
+  - [scheduleTransaction](#scheduletransaction)
+  - [checkTransaction](#checktransaction)
+  - [rejectTransaction](#rejecttransaction)
+  - [cancelTransaction](#canceltransaction)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+The `TimelockGuard` implements a mandatory delay on Gnosis Safe contracts as a Safe Guard. Safe Guards are external contracts that Safes call into before and after the execution of a transaction to enquire if the execution fulfils conditions implemented in the Guard, and revert if they don't. The `TimelockGuard`
```
smartcontracts: Looks like this overview wasn't finished

specs/governance/timelock-guard.md
```diff
@@ -0,0 +1,136 @@
...
+Given a scheduled transaction, the minimum time in seconds after `scheduling_time` that execution is allowed.
+
+### Rejected Transaction
+Given a transaction scheduled to execute from given safe, owners for the safe reject the transaction to signal they think the transaction should not be executed after the `delay_period`.
+
+### `rejecting_owners(safe, tx)`
+Given a transaction scheduled to execute from given safe, the owners for the safe that rejected the transaction.
+
+### `cancellation_threshold(safe)`
+The number of owners that must reject a given transaction for it not to be executed after the `delay_period`. It is equal to `blocking_minority`, where `blocking_minority = min(quorum(safe), total_owners(safe) - quorum + 1)`.
```
smartcontracts: We should rename `blocking_minority` to `blocking_threshold` since it may not always be a minority

specs/governance/timelock-guard.md
```diff
@@ -0,0 +1,136 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+The `TimelockGuard` implements a mandatory delay on Gnosis Safe contracts as a Safe Guard. Safe Guards are external contracts that Safes call into before and after the execution of a transaction to enquire if the execution fulfils conditions implemented in the Guard, and revert if they don't. The `TimelockGuard` 
+
+## Definitions
+
+### `quorum(safe)`
```
smartcontracts: Generally prefer to use plain text here since it's easier to reference elsewhere in the doc, so "Quorum" instead of quorum(safe) and "Delay Period" instead of `delay_period(safe)`.

specs/governance/timelock-guard.md
```diff
@@ -0,0 +1,136 @@
...
+### `delay_period(safe)`
+Given a scheduled transaction, the minimum time in seconds after `scheduling_time` that execution is allowed.
+
+### Rejected Transaction
+Given a transaction scheduled to execute from given safe, owners for the safe reject the transaction to signal they think the transaction should not be executed after the `delay_period`.
+
+### `rejecting_owners(safe, tx)`
+Given a transaction scheduled to execute from given safe, the owners for the safe that rejected the transaction.
+
+### `cancellation_threshold(safe)`
```
smartcontracts: Should also define Blocking Threshold separately

specs/governance/timelock-guard.md
```diff
@@ -0,0 +1,136 @@
...
+The number of owners that must reject a given transaction for it not to be executed after the `delay_period`. It is equal to `blocking_minority`, where `blocking_minority = min(quorum(safe), total_owners(safe) - quorum + 1)`.
+
+Alternatively, the `cancellation_threshold` can start at 1 for each safe, increasing by 1 with each consecutive cancelled by the given safe transaction up to `blocking_minority`, resetting to 1 with each successfully executed transaction by the given safe.
+
+### Nested Safe Setup
+A nested safe setup is a safe (parent safe) in which one or more owners are a Gnosis Safe (child safes). There can be multiple levels of nesting. The one parent safe that is not a child of any other safe is called a root safe.
+
+### Nested Cancellation
```
smartcontracts: I think this is more of an invariant than a thing we need to define

specs/governance/timelock-guard.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Assumptions
+
+## Invariants
+
+## Function Specification
+
+### constructor
+
+### enable(delay_period)
```
smartcontracts: Maybe just `configure(TimelockGuardConfig)` where `TimelockGuardConfig` is a struct that at least includes a `delay` uint

specs/governance/timelock-guard.md
```diff
@@ -0,0 +1,136 @@
...
+Returns the `cancellation_threshold` for a given safe.
+
+- MUST return 0 if the TimelockGuard is not enabled for the safe.
+
+### delayPeriod(safe)
+Returns the `delay_period` for a given safe.
+
+- MUST return 0 if the TimelockGuard is not enabled for the safe.
+
+### setDelayPeriod(delay_period)
```
smartcontracts: Can remove this if we just have a single `configure` function

specs/governance/timelock-guard.md
```diff
@@ -0,0 +1,136 @@
...
+Called by a Safe, set its `delay_period`.
+
+- MUST revert if the TimelockGuard is not enabled for the safe.
+
+### scheduleTransaction
+Called by anyone using signatures from Safe owners, registers a transaction in the TimelockGuard for execution after the `delay_period`.
+
+- MUST revert if the TimelockGuard is not enabled for the safe.
+- MUST take the same parameters as `execTransaction`.
```
smartcontracts, line 226: I need to think a bit about how to make this work for module txns too

specs/governance/timelock-guard.md
```diff
@@ -0,0 +1,136 @@
...
+- MUST revert if an identical transaction has already been scheduled.
+- MUST revert if an identical transaction was cancelled.
+
+To allow for identical transactions to be scheduled more than once, but requiring different signatures for each one, a `salt` parameter can be included in `data` with the sole purpose of differentiating otherwise identical transactions.
+
+### checkTransaction
```
smartcontracts, line 234: A function to see pending txns would be nice too

specs/governance/timelock-guard.md
```diff
@@ -0,0 +1,136 @@
...
+
+- MUST revert if the TimelockGuard is not enabled for the safe.
+
+### scheduleTransaction
+Called by anyone using signatures from Safe owners, registers a transaction in the TimelockGuard for execution after the `delay_period`.
+
+- MUST revert if the TimelockGuard is not enabled for the safe.
+- MUST take the same parameters as `execTransaction`.
+- MUST revert if an identical transaction has already been scheduled.
+- MUST revert if an identical transaction was cancelled.
```
smartcontracts, line 228: Would be good to describe where events need to be emitted

specs/governance/timelock-guard.md
```diff
@@ -0,0 +1,136 @@
...
+
+### checkTransaction
+Called by anyone, and also by the Safe in `execTransaction`, verifies if the transaction was scheduled and the delay period has passed.
+
+- MUST revert if the TimelockGuard is not enabled for the safe.
+- MUST take the exact parameters from the Safe API
+- MUST revert if `scheduling_time(safe, tx) + delay_period(safe) < block.timestamp`
+- MUST revert if the scheduled transaction was cancelled
+
+### rejectTransaction
```
smartcontracts, line 252: We have three versions of this function:

- `rejectTransaction(transactingSafe, transactionHash)` - uses `msg.sender`
- `rejectTransaction(transactingSafe, transactionHash, signature)` - uses the signature
- `rejectTransaction(transactingSafe, transactionHash, otherSafe, owners[])` - optional if we want to support recursion

specs/governance/timelock-guard.md
```diff
@@ -0,0 +1,136 @@
...
+
+- MUST revert if the TimelockGuard is not enabled for the safe.
+- MUST revert if not called by an owner of the safe.
+- MUST be able to reference a transaction scheduled in a different safe. The transaction might not exist.
+
+### cancelTransaction
+Called by anyone, verify that the `cancellation_threshold` has been met for the Safe to cancel a given scheduled transaction.
+
+- MUST revert if the TimelockGuard is not enabled for the safe.
+- MUST revert if `scheduling_time(safe, tx) + delay_period(safe) >= block.timestamp`
```
smartcontracts: Should also check that each owner is actually an owner on the safe, and that the owner set is unique

specs/governance/timelock-guard.md
alcueca: It seems that the Guard system is completely separate for regular transactions and transactions through modules.

The TimelockGuard would need to reuse its logic to be both a Guard and a ModuleGuard, and implement both `checkTransaction` and `checkModuleTransaction`.

Then the TimelockGuard would need to be enabled as both a Guard, and a ModuleGuard.

alcueca: Added some idea on how to do it. Resolving this comment so that we continue dicussing the matter over the module guard proposal.

specs/governance/timelock-guard.md
alcueca: Added events with some description of minimal parameters.

specs/governance/timelock-guard.md
alcueca: This revealed a bit of an issue. The TimelockGuard doesn't know which transactions have been executed, for that it would need to peek into the storage of the Safe. more importantly, if a module (UnorderedExecutionModule) overrides the replayability mechanism, the TimelockGuard needs to be tailored to also check in the module storage, if we want the TimelockGuard to be execution-aware.

specs/governance/timelock-guard.md
alcueca: We check that the owner is an owner of a valid safe for the transaction in `rejectTransaction` to avoid spamming.

As for the set of `owners` being unique, the specs don't include the arguments for the `cancelTransaction` function, and don't require that a list of `owners` is provided as an argument.

Also, the owner set won't necessarily be unique. It is possible and valid in a nested setup for the same address to be an owner in two different safes, and to reject and be counted in both.

specs/governance/timelock-guard.md
alcueca: Removed `setDelay` so that `enable` both enables and configures, and a module can be re-enabled to re-configure.

I also added disclaimers to both definitions and the function specification to allow the developer to choose better function and variable names if they can think of any (I don't think we need to decide on `enable` vs `configure`).

I don't think I need to require which structs should be used.

RamsesAupart (approved):

RamsesAupart (approved):

alcueca:
Superseded by [PR761](https://github.com/ethereum-optimism/specs/pull/761)

PR #753, opened 2025-08-26
feat(supervisor): supervisor_localSafe added

dhyaniarun1993:


emhane (approved):
thanks!

emhane:
@protolambda I see this method isn't implemented in Go, any reason for omitting it?

dhyaniarun1993:
> @protolambda I see this method isn't implemented in Go, any reason for omitting it?

@emhane It's implemented  [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-supervisor/supervisor/frontend/frontend.go#L42)

PR #756, opened 2025-09-05
explicitly require `PayloadAttributes.MinBaseFee` to be `NULL` pre jovian

### LINKED ISSUE #755: Be explicit about the requirement for `PayloadAttributes.MinBaseFee` to be `nil` pre Jovian.
If we go this way, the specs should be updated to be explicit about the requirement to be `nil` pre Jovian.

_Originally posted by @geoknee in https://github.com/ethereum-optimism/op-geth/pull/666#discussion_r2321238041_

geoknee:
Closes #755

### DIFF
--- a/specs/protocol/jovian/exec-engine.md
+++ b/specs/protocol/jovian/exec-engine.md
@@ -49,7 +49,7 @@ extended by future upgrades.
 
 ### Minimum Base Fee in `PayloadAttributesV3`
 
-The Engine API [`PayloadAttributesV3`](../exec-engine.md#extended-payloadattributesv3) is extended in with a new
+The Engine API [`PayloadAttributesV3`](../exec-engine.md#extended-payloadattributesv3) is extended with a new
 field `minBaseFee`. The existing `eip1559Params` remains 8 bytes (Holocene format).
 
 ```text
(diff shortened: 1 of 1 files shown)


specs/protocol/jovian/exec-engine.md
```diff
@@ -49,7 +49,7 @@ extended by future upgrades.
 
 ### Minimum Base Fee in `PayloadAttributesV3`
 
-The Engine API [`PayloadAttributesV3`](../exec-engine.md#extended-payloadattributesv3) is extended in Jovian with a new
+The Engine API [`PayloadAttributesV3`](../exec-engine.md#extended-payloadattributesv3) is extended in with a new
```
joshklop: Accidental diff?

joshklop (approved):
Approving with the assumption that the typo will be resolved

specs/protocol/jovian/exec-engine.md
geoknee: Actually intentional -- we are extending this for the entire history of the chain, not just for Jovian. For example, we would send a message across the API with this new format even for _pre_ Jovian blocks. They would just take on the default value (in this case it is NULL, but in an alternative implementation it could have been `0`) for the new field until the fork activates.

joshklop: Ah, in that case I think we'll want to remove the `in`. Right now it says

> is extended in with a new

geoknee: Totally, good catch thanks [34f92ab](https://github.com/ethereum-optimism/specs/pull/756/commits/34f92ab7ed3fa1dc5441718ce82ab52f4f3bc91b)

PR #761, opened 2025-09-07
Safer Safes - Safe Extensions v2

alcueca:
Specs for [LivenessModuleV2](https://github.com/ethereum-optimism/design-docs/pull/320) and [TimelockGuard](https://github.com/ethereum-optimism/design-docs/pull/324), combined into a single contract as **Safe Extensions V2**, replacing Safe Extensions.


specs/protocol/safe-extensions-v2.md
```diff
@@ -0,0 +1,489 @@
...
+Challenges an enabled `safe` to prove that it is not in a liveness failure state.
+
+- MUST only be executable by `fallback` owner of the challenged `safe`.
+- MUST revert if the `safe` hasn't enabled the contract as a module.
+- MUST revert if the `safe` hasn't configured the contract as a module.
+- MUST revert if a challenge for the `safe` exists.
+- MUST set `challenge_start_time` to the current block time.
+- MUST emit the `ChallengeStarted` event.
+
+### `respond`
```
alcueca: You could put the `respond` logic in `checkAfterExecution`, which would make that any successful transaction by the safe clears the liveness challenge, but then you would be forced to install the contract as a module and a guard for liveness.

Right now, if we don't want to enable the timelock, we just don't enable the contract as a guard. If we would want to use `checkAfterExecution` then we would have to always enable the contract as a module and as a guard, and if the `timelock_delay` is 0, then `checkTransaction` would always succeed.

specs/protocol/safe-extensions-v2.md
```diff
@@ -0,0 +1,489 @@
...
+Removes the liveness module configuration by a previously enabled `safe`.
+
+- The contract MUST NOT be enabled as a module on the `safe`.
+- MUST erase the existing `liveness_challenge_period` and `fallback_owner` data related to the calling `safe`.
+- If a challenge exists, it MUST be cancelled, including emitting the appropriate events.
+- MUST emit a `ModuleCleared` event.
+
+### `configureTimelockGuard`
+
+Configure the contract as a timelock guard by setting the `timelock_delay`.
```
alcueca: Note that you can change the delay on the timelock, and scheduled transactions will be affected as a result.

This makes the mandatory delay not mandatory at all, since the quorum can change the delay to zero and execute whathever they want immediately.

alcueca: Fixed it myself.

alcueca: Oh, I'm a fool. You can only change the delay through a transaction that is in itself subject to the timelock delay, so at most you could make transactions that are executable already not executable by setting a longer delay, but not the opposite.

We can change the logic to record scheduling time instead of execution time. No preference here from me.

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,504 @@
...
+Called by anyone, and also by the Safe in `execTransaction`, verifies if a given transaction was scheduled and the
+delay period has passed.
+
+- MUST revert if the contract is not enabled as a guard for the safe.
+- MUST take the exact parameters from the `ITransactionGuard.checkTransaction` interface.
+- MUST revert if `execution_time(safe, tx) < block.timestamp`.
+- MUST revert if the scheduled transaction was cancelled.
+
+### `checkPendingTransactions`
+
```
JosepBove, line 491: Must never revert

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,504 @@
...
+- MUST revert if the `safe` hasn't enabled the contract as a module.
+- MUST revert if the `safe` hasn't configured the contract as a module.
+- MUST revert if there isn't a successful challenge for the given `safe`.
+- MUST reset `challenge_start_time` to 0 to enable the fallback to start a new challenge.
+- MUST emit the `ChallengeExecuted` event.
+
+### `cancellationThreshold`
+
+Returns the `cancellation_threshold` for a given safe.
+
```
JosepBove, line 467: Must never revert

words.txt
```diff
@@ -202,6 +202,8 @@ remotechainid
 replayability
 replayable
 reposted
+reputational
+restablished
```
JosepBove: Double checked with GPT:

“Restablished” is not a standard word in English.

The correct word is “reestablished” (with two e’s in the middle), which means to establish again


Let's update it here and in the changes of the specs

specs/protocol/safe-extensions-v2.md
JosepBove: I prefer to have them separated for the reason that you mentioned

specs/protocol/safe-extensions-v2.md
JosepBove: There's no preference for me either here.

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,507 @@
...
+- MUST revert if there isn't a challenge for the calling `safe`.
+- MUST reset `challenge_start_time` to 0.
+- MUST emit the `ChallengeCancelled` event.
+
+### `changeOwnershipToFallback`
+
+With a successful challenge, removes all current owners from an enabled `safe`, appoints `fallback` as its sole owner,
+and sets its quorum to 1.
+
+- MUST be executable by anyone.
```
JosepBove: It looks like smartcontracts disagrees here: https://github.com/ethereum-optimism/optimism/pull/17272#discussion_r2326104783

Are we okay with having it executable for just the fallback address or we need to start a discussion?

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,507 @@
...
+
+### `viewLivenessModuleConfiguration`
+
+Returns the `liveness_challenge_period` and `fallback_owner` for a given `safe`.
+
+- MUST never revert.
+
+### `viewTimelockGuardConfiguration`
+
+Returns the `timelock_delay` for a given `safe`.
```
maurelian: I'm thinking about what this should return if the delay is configured, but the guard is not enabled. 

I think we should just returns the value as configured, as we don't want to obfuscate the current config, 
but perhaps we should add a boolean value to this function that indicates if the guard is enabled.

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,507 @@
...
+  - [`checkTransaction`](#checktransaction)
+  - [`checkPendingTransactions`](#checkpendingtransactions)
+  - [`rejectTransaction`](#rejecttransaction)
+  - [`rejectTransactionWithSignature`](#rejecttransactionwithsignature)
+  - [`cancelTransaction`](#canceltransaction)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+This document describes extensions to the Security Council and Guardian Safe contracts, which
+provide additional functionality and security guarantees on top of those provided by the Safe
+contract.
```
maurelian: Open questions for this document to address: 

1. what versions of the Safe are supported (we also need to think about a testing plan for confirming that)
2. under what conditions if any (ie. a pause call) should a transaction not be subject to the delay.

specs/protocol/safer-safes.md
smartcontracts: I see no real benefit to having this be executable by anyone. Keeping it only executable by the fallback owner gives more control to the timeline on which we execute this with no meaningful downside.

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,507 @@
...
+*Note:* If we want to exclude executed transactions from this list, we would need to implement the
+`checkAfterExecution` hook and store the executed state for transactions.
+
+### `rejectTransaction`
+
+Called by a Safe owner, signal the rejection of a scheduled transaction.
+
+- MUST revert if the contract is not enabled as a guard for the safe.
+- MUST revert if not called by an owner of the safe scheduled to execute the transaction, or in one of its child safes.
+- MUST be able to reference a transaction scheduled in a different safe. The transaction might not exist.
```
maurelian: I don't understand what this requirement means?

specs/protocol/safer-safes.md
alcueca: Elsewhere I suggested adding an `initialized` or `configured` boolean to the per-safe guard configuration, because using `timelock_delay != 0` as a marker of the guard having been configured clashes with the non-irrational concept of a zero delay.

specs/protocol/safer-safes.md
alcueca: 1. Let's code this guard for the latest version, and add additional support in the guard only if we can't upgrade some Safe.
2. The delay is mandatory in all conditions, at least in this version. We are not installing the timelock in the upgrade related safes. A timelock bypass is a question for the Reactive Security Countil feature.

specs/protocol/safer-safes.md
alcueca: If you have a nested setup, transactions will be scheduled in the parent safe, but rejected in the child safes.

I think I wrote this without considering that the guard is a singleton, and was worried that the timelock guard in the child safe wouldn't have access to the storage for the parent safe timelock guard.

Given that the timelock guard is a singleton, and all transactions have an unique identifier regardless of the safe scheduled for execution, this doesn't make much sense.

specs/protocol/safer-safes.md
alcueca: The latest version is 1.5.0, but in this implementation I think that we are only going to use functionality that was available since 1.0.0.

specs/protocol/safer-safes.md
alcueca: Strongly agree with @smartcontracts here. It makes no sense to make this executable by anyone, since we assume the fallback owner is alive.

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,509 @@
...
+
+### Liveness Challenge
+
+A challenge to a multisig to prove that it is not in a liveness failure state.
+
+### Liveness Challenge Period
+
+The `liveness_challenge_period(safe)` is the minimum period of time that must pass without a response since a liveness
+challenge for a liveness failure state to be assumed, which is equal to the sum of the `timelock_delay(safe)` plus the
+`liveness_challenge_period(safe)`.
```
alcueca, line 177: There is a problem in this definition, as it is circular.

The contract should be provided with a `liveness_challenge_period_delta`, which will be added to the `timelock_delay` in order to get the actual `liveness_challenge_period`.

This will cascade into the specs below, notably in that `liveness_challenge_period_delta` is a configuration parameter, but `liveness_challenge_period` is a calculated value. I think we would need a `getLivenessChallengePeriod` function, and only `liveness_challenge_period_delta` would be shown in `viewLivenessConfiguration`.

The alternative is to make sure externally that we never configure a `timelock_delay` that is greater than the `liveness_challenge_period`. Otherwise a liveness challenge couldn't be responded. That doesn't seem a major issue to me, the risk might be worth it considering [the simpler implementation](https://github.com/ethereum-optimism/optimism/pull/17402).

Discuss.

specs/protocol/safer-safes.md
maurelian: > was worried that the timelock guard in the child safe wouldn't have access to the storage for the parent safe timelock guard.

Yes, the same guard will be used on all Safes, but I don't currently see any functionality that would take advantage of that. 

As is, in order to cancel a transaction, a child-safe would need to have a quorum of signers on an transaction that cancels the scheduled parent-safe transaction. My sense though is that there might be an idea that a signer on a child safe should be able directly cancel the parent-safe's transaction, is that meant to be?

specs/protocol/safer-safes.md
maurelian: >  I think we would need a getLivenessChallengePeriod function, and only liveness_challenge_period_delta would be shown in viewLivenessConfiguration

This feels workable to me. I also think we can make that work with the simpler implementation just by adding an override to `getLivenessChallengePeriod()`. 

I will try it out.

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,509 @@
...
+
+*Note:* If we want to exclude executed transactions from this list, we would need to implement the
+`checkAfterExecution` hook and store the executed state for transactions.
+
+### `rejectTransaction`
+
+Called by a Safe owner, signal the rejection of a scheduled transaction.
+
+- MUST revert if the contract is not enabled as a guard for the safe.
+- MUST revert if not called by an owner of the safe scheduled to execute the transaction, or in one of its child safes.
```
alcueca: @maurelian, note that the rejection of a transaction can be signalled by the owners of child safes to the parent safe that is executing the transaction, and that there is an arbitrary level of nesting.

specs/protocol/safer-safes.md
alcueca: To cancel a transaction, `cancellation_threshold` of owners of the safe in which the transaction is scheduled must signal they want to cancel.

We allow nested safes, so if one or more of the owners are a child safe, they can only signal they want to cancel by having their own owners signal they want to cancel meeting the `cancellation_threshold`of the child safe.

We allow an arbitrary level nesting, so this would be best done with a recursive algorithm, imo.

1. Owners of any safe can signal `reject(safe, txHash)` where `scheduled[safe][txHash]` exists and hasn't been cancelled. There is the question on whether we want to make sure at this point that the account is an owner of a safe involved in `txHash` (`isOwnerRecursive`). Seems more trouble than is worth. The only risk if we don't is having the events polluted since anyone can enable the module.

2. Anyone can call `cancel(parentSafe, txHash)`, this will check that at least `cancellation_threshold` owners of `parentSafe` have called `reject(parentSafe, txHash)`. If any of the `parentSafe` owners is a child safe, it will call `isRejected(childSafe, txHash)` on the child safe to find out if the child safe is rejecting, by checking if the owners of the child safe are rejecting in sufficient numbers. As you see, `isRejected` is recursive.

@smartcontracts had an alternative design to support nested safes, which he might want to share. I think this one is simpler.

specs/protocol/safer-safes.md
alcueca: I'm not sure it is worth it, tbh. We can just validate in the superchain-ops task that `liveness_challenge_period > timelock_delay`.

specs/protocol/safer-safes.md
alcueca: Included a [task in the tracker](https://www.notion.so/oplabs/SaferSafes-Chain-Ops-268f153ee1628006b5d6dd0bcf070875) to identify all safe versions in range and explore upgrading them to 1.5.0 as part of adding SaferSafes to them.

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,509 @@
...
+   authorized to execute transactions via the Safe. This means the liveness module must properly implement
+   auth conditions internally.
+
+The extension contract specificationis compatible with the 
+[V1.5.0](https://github.com/safe-global/safe-contracts/releases/tag/v1.5.0) Safe contract version.
+
+When enabled as a guard, the extension contract implements a mandatory delay on the execution of transactions. When
+enabled as a module, the extension contract ensures a multisig remains operable by allowing challenges when it becomes
+unresponsive. If the multisig fails to prove liveness within a set period, ownership transfers to a trusted fallback
+ owner to prevent deadlock.
```
blockchaindevsh: Is the trusted fallback owner an EOA or another Safe contract?

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,509 @@
+# Safe Contract Extensions V2
```
maurelian, line 1: To add: The Timelock Guard will need a module exclusion list so that txns that originate from the Liveness module don't get stuck behind the timelock guard.

specs/protocol/safer-safes.md
alcueca: The trusted fallback should have stronger security guarantees than the safe implemented the module, as a best practice.

For example, our finance multisigs might have the Foundation Operational Safe as fallback. The FOS might have the Security Council multisig as fallback. The Security Council might have some core multisig at the Ethereum Foundation or a top-tier security services company.

blockchaindevsh: Are all operations protected by time locks? Without them, a top-level module could execute harmful actions before the fallback modules have the opportunity to respond.

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,559 @@
...
+- MUST revert if the contract is not configured for the `safe`.
+- MUST revert if `sum(rejecting_owners(safe, tx)) < cancellation_threshold(safe)`.
+- MUST emit a `TransactionCancelled` event, with at least `safe` and a transaction identifier.
+
+#### `checkTransaction`
+
+Can be called by anyone. It will be called by the `safe` in `execTransaction` if the contract is enabled as a guard.
+It verifies if a given transaction was scheduled and the delay period has passed, and reverts if not.
+
+- MUST revert if the contract is not enabled as a guard for the `safe`.
```
maurelian, line 551: We should remove this as not workable, because the Safe won't call this function if no guard is enabled.

```suggestion
```

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,559 @@
...
+- MUST revert if the contract is not configured for the `safe`.
```
maurelian: While implementing this requirement, I realized that it makes it impossible to configure the guard. 
ie. if the installation path is meant to be:
1. enable guard
2. configure guard

Then 2 is not possible with this requirement.

```suggestion
- MUST immediately execute the transaction if the contract is not configured for the `safe`.
```

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,559 @@
...
+- MUST take the exact parameters from the `ITransactionGuard.checkTransaction` interface.
+- MUST revert if `execution_time(safe, tx) < block.timestamp`.
+- MUST revert if the scheduled transaction was cancelled.
+
```
maurelian, line 545: What should happen if the Safe is configured with a delay of zero? 
From a philosophical perspective (ie. the one where you can call `ERC20.transfer(alice, 0)` because zero is a valid uint), it feels to me like we should do the normal checks, thus requiring the transaction to be scheduled before it can be executed, even though there is no delay.

From an operational perspective, it feels like it would be dumb to add that extra step... and I care about ops more than philosophy.

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,559 @@
...
+#### `checkAfterExecution`
+
+Can be called by anyone. It will be called by the `safe` in `execTransaction` if the contract is enabled as a guard.
+It updates state after the transaction has been executed.
+
+- MUST take the exact parameters from the `ITransactionGuard.checkTransaction` interface.
+- MUST revert if the contract is not enabled as a guard for the `safe`. // TODO: I'm not sure if we want to ever revert here
+- MUST revert if the contract is not configured for the `safe`. // TODO: I'm not sure if we want to ever revert here
```
maurelian: Agree we should not revert here.

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,559 @@
...
+- MUST set `cancellation_threshold(safe)` to 1.
```
maurelian, line 558: IMO we shold do this in the `checkTransaction()` function. It's simpler and follows the checks, effects and interactions pattern.

specs/protocol/safer-safes.md
alcueca: Actually, the default is for transactions that originate from modules to not get stuck in guards.

specs/protocol/safer-safes.md
alcueca: This timelock implementation doesn't delay modules, because that is what we need right now at optimism.

We have two modules, the deputy pause module, and the liveness module, and none of them should be delayed by the timelock.

When we develop a module that should be delayed, we will update the timelock accordingly.

alcueca: Leaving this open to remember to add this in the docs

specs/protocol/safer-safes.md
alcueca: True, removed the whole after-execution function from the spec.

specs/protocol/safer-safes.md
alcueca: Removed function from the spec

specs/protocol/safer-safes.md
alcueca: I think that because the execution of the function is the same in both implementations, we can skip the need to schedule a transaction if the delay is zero.

specs/protocol/safer-safes.md
alcueca: But other contracts could call it. To no effect, probably, but just in case.

specs/protocol/safer-safes.md
alcueca: Added to the specs that it doesn't check the scheduling if the delay is 0

specs/protocol/safer-safes.md
alcueca: Removed the requirement and added a note at the end. Maybe we need some better explanations around the fact the enabling/disabling the guard in the safe doesn't affect operations, and that setting the timelock to zero or a value different than zero is what really enables/disables it.

Also, `clearTimelockGuard` is the same as `configureTimelockGuard` with a delay of zero.

And I'm not sure if we want to keep the `.configured` field in the struct, or radically simplify and remove both that struct field and the `clearTimelockGuard` function.

specs/protocol/safer-safes.md
blockchaindevsh: The deputy pause module and the liveness module are designed in a way that prevents them from executing harmful actions, so a time lock isn’t necessary—am I understanding this correctly?

maurelian: > The deputy pause module and the liveness module are designed in a way that prevents them from executing harmful actions

That is correct. Both of those modules are, (and I would say that modules in general should be), very constrained in the kinds of actions they allow. Typically that means the modules can only cause the Safe to make calls to certain addresses with calldata that is constrained so that it can only call certain functions on those addresses.

Regardless, when developing any new module we will need to consider how it should interact with the guard.

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,556 @@
...
+  - [`rejectTransactionWithSignature`](#rejecttransactionwithsignature)
+  - [`cancelTransaction`](#canceltransaction)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+This document describes extensions to the Security Council and Guardian Safe contracts, which
+provide additional functionality and security guarantees on top of those provided by the Safe
+contract.
+
+These extensions are developed using a singleton contract that can be enabled simultaneously as a
+([module](https://docs.safe.global/advanced/smart-account-modules) and a
```
maurelian, line 92: TOC needs an update

specs/protocol/safer-safes.md
```diff
@@ -0,0 +1,556 @@
...
+Configure the contract as a timelock guard by setting the `timelock_delay` and `safety_delay`.
+
+- MUST allow an arbitrary number of `safe` contracts to use the contract as a guard.
+- MUST revert if the contract is not enabled as a guard for the `safe`.
+- MUST revert if `timelock_delay` is longer than 1 year.
+- MUST revert if `safety_delay` is longer than 1 year.
+- MUST set the caller as a `safe`.
+- MUST take `timelock_delay` as a parameter and store it as related to the `safe`.
+- MUST take `safety_delay` as a parameter and store it as related to the `safe`.
+- MUST set the `cancellation_threshold` to 1.
```
maurelian, line 514: I would like to remove this requirement, it seems reasonable that a Safe should be able to configure itself whether or not the guard is enabled.

```suggestion
```

mds1:
Closing this in favor of a replacement spec PR based on current specs/main and the latest SaferSafes implementation on optimism/develop. The old branch is behind main, touches removed interop files, and no longer matches the current contract behavior after the merged SaferSafes audit fixes.

PR #766, opened 2025-09-11
jovian: add more min base fee details

joshklop:


specs/protocol/system-config.md
```diff
@@ -358,6 +360,14 @@ Allows the owner to update the operator fee parameters.
 - MUST update the operatorFeeScalar and operatorFeeConstant values.
 - MUST emit a ConfigUpdate event with the UpdateType.OPERATOR_FEE_PARAMS type.
 
+### setMinBaseFee
+
+Allows the owner to update the minimum base fee parameter.
+
+- MUST revert if called by an address other than the owner.
+- MUST update the minBaseFee value.
+- MUST emit a ConfigUpdate event with the UpdateType.MIN_BASE_FEE type.
```
geoknee, line 369: I would just link to the jovian specs here which specify the update type (which is `6`). `UpdateType.MIN_BASE_FEE` is not a part of the spec, but of the implementation. It's not defined on this page or anywhere else in the specs AFAIK.

specs/protocol/system-config.md
```diff
@@ -43,6 +43,7 @@
   - [setGasLimit](#setgaslimit)
   - [setEIP1559Params](#seteip1559params)
   - [setOperatorFeeScalars](#setoperatorfeescalars)
+  - [setMinBaseFee](#setminbasefee)
```
geoknee, line 46: Meta comment: some parts of the spec follow the pattern of "initial version, change1, change2" and others follow "latest version". You are following this mix of patterns here which is totally fine, but I do think we should sort it out at some point. One awkward thing is that this addition is really about the "next" version of the `SystemConfig`, not the current one (and that is not obvious). At least when we put stuff under "Jovian" it's a bit clearer that it is not active yet.

specs/protocol/configurability.md
```diff
@@ -300,6 +301,12 @@ contracts deployed on layer 1.<br/>
 Note that the operator fee scalar and constant are primarily used for non-standard configurations,
 like op-succinct, so their standard values are 0.
 
+### [Minimum Base Fee](jovian/exec-engine.md#minimum-base-fee)
+
+**Description:** Minimum base fee <br/>
+**Administrator:** [System Config Owner](#admin-roles)<br/>
+**Requirement:** 0 <br/>
```
geoknee: I'm not sure we _require_ this for standard chains. Do we have any paper trail of a decision on that?

specs/protocol/system-config.md
joshklop: Good point. In other parts of the specs we say things like "starting with Jovian" or "after Jovian", but this `protocol/system-config.md` doc doesn't use that practice. For example, we could say that a particular update type is only present after a particular hardfork. This way we don't need to remember to update the specs again after the hardfork is released. Wdyt?

geoknee: Sounds good to me!

specs/protocol/system-config.md
joshklop: Will create an issue consistent with our discussion on the da footprint PR.

specs/protocol/configurability.md
joshklop: Will remove this diff consistent with our decision on the da footprint PR.

geoknee (approved):

PR #767, opened 2025-09-12
jovian: add da footprint limit specs

joshklop:
Depends on #766


specs/protocol/configurability.md
```diff
@@ -307,6 +308,12 @@ like op-succinct, so their standard values are 0.
 **Administrator:** [System Config Owner](#admin-roles)<br/>
 **Requirement:** 0 <br/>
 
+### [DA Footprint Gas Scalar](exec-engine.md#da-footprint)
+
+**Description:** scales the cumulative da size of all transactions in a block. <br/>
+**Administrator:** [System Config Owner](#admin-roles)<br/>
+**Requirement:** 0 <br/>
```
geoknee: Just flagging this, I think we will soon decide on a default value which could go here. Although I think we might permit standard chains to customize here... need to work with @ben-chain to figure out what makes sense.

specs/protocol/jovian/exec-engine.md
```diff
@@ -77,3 +80,50 @@ This retains the purity of the function that computes the next block's base fee
...
+    prevRandao: DATA (32 bytes)
+    suggestedFeeRecipient: DATA (20 bytes)
+    withdrawals: array of WithdrawalV1
+    parentBeaconBlockRoot: DATA (32 bytes)
+    transactions: array of DATA
+    noTxPool: bool
+    gasLimit: QUANTITY or null
+    eip1559Params: DATA (8 bytes) or null
+    minBaseFee: QUANTITY or null
+    daFootprintGasScalar: QUANTITY (16 bits) or null
```
geoknee: I wonder why we sometimes:
* add the number of bits/bytes in parentheses
* use QUANTITY versus DATA

specs/protocol/jovian/exec-engine.md
```diff
@@ -77,3 +80,50 @@ This retains the purity of the function that computes the next block's base fee
...
+}
+```
+
+The `daFootprintGasScalar` MUST be `null` prior to the Jovian fork, and MUST be non-`null` after the Jovian fork.
+
+### Rationale
+
+While the current L1 fee mechanism charges for DA usage based on the size of a transaction, it does not influence future
+base fee calculations. As a result, excessive DA usage is not efficiently reflected in the fee market, ultimately leading to
+occasional throttling and undesirable priority fee auctions (PFAs).
```
geoknee: I don't think we have any reference to throttling in the specs so far. Perhaps just refer to "out of protocol block building policies" ?

specs/protocol/jovian/exec-engine.md
```diff
@@ -77,3 +80,50 @@ This retains the purity of the function that computes the next block's base fee
 header, while still allowing them to be dynamically configured. Dynamic configuration is handled
 similarly to `gasLimit`, with the derivation pipeline providing the appropriate `SystemConfig`
 contract values to the block builder via `PayloadAttributesV3` parameters.
+
+## DA Footprint Limit
+
+Jovian changes the meaning of the `gasUsed` block header field when a block's DA footprint exceeds its gas
+used. The total DA footprint of a block is calculated by scaling the cumulative DA footprint of its transactions
```
geoknee:
```suggestion
This feature allows for the estimated data-availability (DA) footprint of a block to influence the block's `gasUsed` property (which is no longer always equal to the sum of the gas used in each transaction included in the block). In this way, the fee market is coupled to the estimated data-availability footprint via the EIP-1559 mechanism, allowing for an excess in data-availability requirements to cause a rise in the base fee. 

The total DA footprint of a block is calculated by scaling the cumulative DA footprint of its transactions
```

specs/protocol/system-config.md
```diff
@@ -368,6 +370,14 @@ Allows the owner to update the minimum base fee parameter.
 - MUST update the minBaseFee value.
 - MUST emit a ConfigUpdate event with the UpdateType.MIN_BASE_FEE type.
 
+### setDAFootprintGasScalar
+
+Allows the owner to update the da footprint gas gas scalar.
```
geoknee:
```suggestion
Allows the owner to update the da footprint gas scalar.
```

specs/protocol/system-config.md
```diff
@@ -368,6 +370,14 @@ Allows the owner to update the minimum base fee parameter.
...
+
+- MUST revert if called by an address other than the owner.
+- MUST update the daFootprintGasScalar value.
+- MUST emit a ConfigUpdate event with the UpdateType.DA_FOOTPRINT_GAS_SCALAR type.
```
geoknee, line 379: Same comment as on your other PR, I think it is better to refer to a numerical value for the update type, which we can do by linking to the jovian part of the spec.

geoknee (commented):
Looking great so far!

Don't we need material on how we set a) upgrade the L1Block contract during activation and b) update the scalar value stored in it?

specs/protocol/jovian/exec-engine.md
```diff
@@ -77,3 +80,50 @@ This retains the purity of the function that computes the next block's base fee
...
+
+The `daFootprintGasScalar` is loaded in a similar way to the `operatorFeeScalar` and `operatorFeeConstant`
+[included](../isthmus/exec-engine#operator-fee) in the Isthmus fork. It can be read in two interchangable ways:
+
+- read from the deposited L1 attributes (`daFootprintGasScalar`) of the current L2 block
+- read from the L1 Block Info contract (`0x4200000000000000000000000000000000000015`)
+  - using the solidity getter function `daFootprintGasScalar`
+  - using a direct storage-read: big-endian `uint16` in slot `9` at offset `0`.
+
+### DA Footprint Gas Scalar in `PayloadAttributesV3`
```
geoknee: I don't think we need this. Min base fee goes here because we need to tell the engine to insert it into the block header. For quantities read from L2 _state_ (like operator fee scalar/constant and the da footprint scalar), this does not need to go across the engine API.

geoknee (changes requested):

geoknee:
When the time comes, be nice to acknowledge @teddyknox and @axelKingsley who worked on earlier versions of this spec. If you're happy @joshklop, we can add them as co-authors.

joshklop:
> When the time comes, be nice to acknowledge @teddyknox and @axelKingsley who worked on earlier versions of this spec. If you're happy @joshklop, we can add them as co-authors.

I attempted to add Axel as a co-author on the commit, though I might need to use his email instead of his username. I didn't realize that Teddy also worked on this, so I'll add him as well.

specs/protocol/jovian/exec-engine.md
joshklop: I think the `DATA` vs. `QUANTITY` distinction comes from the execution API [specs](https://github.com/ethereum/execution-apis/blob/f91fd7b105b062fd49a3cc23063f5c7becff3afe/src/engine/common.md?plain=1#L133-L135)

specs/protocol/system-config.md
joshklop: I agree linking may very well be better. However, the rest of the document uses this format. Would you like to break convention here?

specs/protocol/jovian/l1-attributes.md
```diff
@@ -0,0 +1,40 @@
...
+| sequenceNumber    | uint64  | 12-19          |         |
+| l1BlockTimestamp  | uint64  | 20-27          |         |
+| l1BlockNumber     | uint64  | 28-35          |         |
+| basefee           | uint256 | 36-67          | 2       |
+| blobBaseFee       | uint256 | 68-99          | 3       |
+| l1BlockHash       | bytes32 | 100-131        | 4       |
+| batcherHash       | bytes32 | 132-163        | 5       |
+| operatorFeeScalar   | uint32  | 164-167      | 6       |
+| operatorFeeConstant | uint64  | 168-175      |         |
+| daFootprintGasScalar | uint16  | 176-178     | 7       |
```
joshklop: What does `Segment` refer to here? I guessed `7`.

specs/protocol/jovian/l1-attributes.md
```diff
@@ -0,0 +1,40 @@
...
+Note that the first input argument, in the same pattern as previous versions of the L1 attributes transaction,
+is the function selector: the first four bytes of `keccak256("setL1BlockValuesJovian()")`.
+
+In the first L2 block after the Jovian activation block, the Jovian L1 attributes are used.
+
+The pre-Jovian values are migrated over 1:1.
+Blocks after the Isthmus activation block contain all pre-Isthmus values 1:1,
+and also set the following new attributes:
+
+- The `daFootprintGasScalar` is set to `<todo>`.
```
joshklop: Need to update this default when we have one.

specs/protocol/jovian/derivation.md
joshklop, line 1: I used Teddy's neat tool to generate this file. We will probably have to update it again depending on when we merge this specs PR.

It's possible I got something wrong (especially the nonce). If someone could give this a second look that would be great.

specs/protocol/system-config.md
geoknee: How about we follow the convention here, and file an issue to circle back and clean up the whole doc?

specs/protocol/jovian/derivation.md
```diff
@@ -0,0 +1,102 @@
...
+
+The `L1Block` contract is deployed.
+
+A deposit transaction is derived with the following attributes:
+
+- `from`: `0x4210000000000000000000000000000000000004`
+- `to`: `null`
+- `mint`: `0`
+- `value`: `0`
+- `nonce`: `0`
```
geoknee, line 35: I note that we don't actually specify the [nonce in the Isthmus specs (](https://specs.optimism.io/protocol/isthmus/derivation.html#l1block-deployment)perhaps we should).

Seems like the pattern is to use a completely fresh `from` account, such that we can always use a `0` nonce.

specs/protocol/jovian/derivation.md
geoknee: Should be fine to merge this PR and return later with any tweaks once the implementation work lands.

specs/protocol/jovian/exec-engine.md
```diff
@@ -77,3 +79,31 @@ This retains the purity of the function that computes the next block's base fee
...
+(as calculated by the [Fjord LZ Estimation](../fjord/exec-engine.md#fjord-l1-cost-fee-changes-fastlz-estimator) by
+a configurable scalar value, the `daFootprintGasScalar`.
+
+The `daFootprintGasScalar` is loaded in a similar way to the `operatorFeeScalar` and `operatorFeeConstant`
+[included](../isthmus/exec-engine.md#operator-fee) in the Isthmus fork. It can be read in two interchangable ways:
+
+- read from the deposited L1 attributes (`daFootprintGasScalar`) of the current L2 block
+- read from the L1 Block Info contract (`0x4200000000000000000000000000000000000015`)
+  - using the solidity getter function `daFootprintGasScalar`
+  - using a direct storage-read: big-endian `uint16` in slot `9` at offset `0`.
```
blockchaindevsh, line 125: Where is `daFootprintGasScalar` used? It seems like it hasn’t been specified.

specs/protocol/configurability.md
joshklop: from call: delete this diff for now. File an issue to decide what `daFootprintGasScalar` looks like in the standard config.

specs/protocol/jovian/l1-attributes.md
joshklop: from call: EVM words. This should be `6`.

specs/protocol/jovian/l1-attributes.md
joshklop: from call: let's use `400` for now. Create an issue to update.

specs/protocol/jovian/derivation.md
joshklop: from call: ask team why a fresh account is used by convention... is there a deeper reason?

specs/protocol/jovian/exec-engine.md
joshklop: Thanks for this feedback. I'll update the wording to clarify.

specs/protocol/jovian/exec-engine.md
```diff
@@ -77,3 +79,31 @@ This retains the purity of the function that computes the next block's base fee
 header, while still allowing them to be dynamically configured. Dynamic configuration is handled
 similarly to `gasLimit`, with the derivation pipeline providing the appropriate `SystemConfig`
 contract values to the block builder via `PayloadAttributesV3` parameters.
+
+## DA Footprint Limit
+
+Jovian changes the interpretation of an L2 block's `gasUsed` field. Before, the `gasUsed` was equal to the total gas used
+by all transactions in a block. After Jovian, `gasUsed` is equal to a block's "DA footprint" iff the footprint exceeds
+the total gas used by transactions. As a result, blocks with high DA usage will cause the base fee to increase in
+subsequent blocks.
```
geoknee:
```suggestion
From Jovian, `gasUsed` is equal to a block's "DA footprint" if the footprint exceeds
the total gas used by transaction, and equal to the sum of the gas used by each transaction otherwise. As a result, blocks with high DA usage may cause the base fee to increase in
subsequent blocks.
```

specs/protocol/jovian/exec-engine.md
```diff
@@ -77,3 +79,31 @@ This retains the purity of the function that computes the next block's base fee
...
+
+A block's DA footprint is calculated by scaling the cumulative DA footprint of its transactions
+(as calculated by the [Fjord LZ Estimation](../fjord/exec-engine.md#fjord-l1-cost-fee-changes-fastlz-estimator) by
+a configurable scalar value, the `daFootprintGasScalar`. The larger the scalar, the higher the price of DA.
```
geoknee:
```suggestion
a configurable scalar value, the `daFootprintGasScalar`.
```
I would just drop this, since it feels a bit imprecise to me.

specs/protocol/jovian/exec-engine.md
```diff
@@ -77,3 +79,31 @@ This retains the purity of the function that computes the next block's base fee
...
+[included](../isthmus/exec-engine.md#operator-fee) in the Isthmus fork. It can be read in two interchangable ways:
+
+- read from the deposited L1 attributes (`daFootprintGasScalar`) of the current L2 block
+- read from the L1 Block Info contract (`0x4200000000000000000000000000000000000015`)
+  - using the solidity getter function `daFootprintGasScalar`
+  - using a direct storage-read: big-endian `uint16` in slot `9` at offset `0`.
+
+### Rationale
+
+While the current L1 fee mechanism charges for DA usage based on the size of a transaction, it does not influence future
```
geoknee:
```suggestion
While the current L1 fee mechanism charges for DA usage based on an estimate of the DA footprint of a transaction, it does not influence future
```

specs/protocol/jovian/exec-engine.md
```diff
@@ -77,3 +79,31 @@ This retains the purity of the function that computes the next block's base fee
...
+base fee calculations. As a result, excessive DA usage is not efficiently reflected in the fee market, leading to suboptimal
+resource prices.
+
+By changing the meaning of the `gasUsed` field in times of high DA demand, the fee market can properly adjust without reverting
+to an inferior PFA experience.
```
geoknee:
```suggestion
to priority fee auctions (an inferior experience for users).
```
Should expand the acronym here.

specs/protocol/jovian/l1-attributes.md
```diff
@@ -0,0 +1,40 @@
...
+Note that the first input argument, in the same pattern as previous versions of the L1 attributes transaction,
+is the function selector: the first four bytes of `keccak256("setL1BlockValuesJovian()")`.
+
+In the first L2 block after the Jovian activation block, the Jovian L1 attributes are used.
+
+The pre-Jovian values are migrated over 1:1.
+Blocks after the Isthmus activation block contain all pre-Isthmus values 1:1,
+and also set the following new attributes:
+
+- The `daFootprintGasScalar` is set to `400`.
```
geoknee: Possibly a follow-on error from the Isthmus specs, but this doesn't seem quite right to me. Don't we want to set the scalar to `400` only by default, and otherwise use a value (if set) from the SystemConfig?

Currently it sounds like we set it to `400` in every block... which would make it non-configurable.

joshklop: Follow-on error indeed. Will update.

specs/protocol/jovian/l1-attributes.md
```diff
@@ -0,0 +1,38 @@
...
+| daFootprintGasScalar | uint16  | 176-177     |         |
+
+Note that the first input argument, in the same pattern as previous versions of the L1 attributes transaction,
+is the function selector: the first four bytes of `keccak256("setL1BlockValuesJovian()")`.
+
+In the first L2 block after the Jovian activation block, the Jovian L1 attributes are used.
+
+The pre-Jovian values are migrated over 1:1.
+Blocks after the Jovian activation block contain all pre-Jovian values 1:1, and also set `daFootprintGasScalar`.
+The default value for the `daFootprintGasScalar` is `400`.
```
geoknee:
```suggestion
Blocks after the Jovian activation block contain all pre-Jovian values 1:1, and also set `daFootprintGasScalar` to the value from the SystemConfig. If that value is `0`, then a default of `600` should be used. 
```

specs/protocol/exec-engine.md
sebastianst, line 1: We don't add the gas scalar to the payload attributes, only to the L1InfoDeposit tx.

specs/protocol/system-config.md
sebastianst: Above, the doc introduces the numerical values (around line 190). It's probably best to only define the numerical value once above and then reuse the identifies here and everywhere else. So I think the only change we should do is to actually name the type in the list above the same as they are used below. Currently, the list names the types using camel-case, but then later the upper-case type names are used. Let's fix this in a follow-up.

specs/protocol/jovian/system-config.md
```diff
@@ -72,3 +78,45 @@ This function returns the currently configured minimum base fee in wei.
 ```solidity
 function minBaseFee() external view returns (uint64);
 ```
+
+## DA Footprint Configuration
+
+Jovian adds a configuration value to `SystemConfig` to control the `daFootprintGasScalar`.
+The value is a multiplier on the cumulative da footprint of a block's transactions, scaling it so
+it can be compared to the gas limit.
```
sebastianst:
```suggestion
Jovian adds a configuration value to the `SystemConfig` to control the `daFootprintGasScalar`.
The value is a multiplier on the cumulative DA footprint of a block's transactions, scaling it up to the
gas dimension so it is comparable to the gas limit.
```

specs/protocol/jovian/system-config.md
```diff
@@ -72,3 +78,45 @@ This function returns the currently configured minimum base fee in wei.
...
+Upon update, the contract emits the `ConfigUpdate` event above, enabling nodes
+to derive the configuration from L1 logs.
+
+### Interface
+
+#### DA Footprint Gas Scalar Parameters
+
+##### `daFootprintGasScalar`
+
+This function returns the currently configured DA footprint gas scalar.
```
sebastianst, line 112: We should add below something like
> If unset, it returns 0. Derivation of the L1 info deposit transaction translates unset zero values to the default value (`600`).

specs/protocol/jovian/overview.md
```diff
@@ -15,9 +15,14 @@ This document is not finalized and should be considered experimental.
 ## Execution Layer
 
 - [Minimum Base Fee](./exec-engine.md#minimum-base-fee)
+- [DA Footprint Limit](./exec-engine.md#da-footprint-limit)
 
 ## Consensus Layer
 
+- [Network upgrade transactions](./derivation.md#network-upgrade-transactions) applied during derivation
+- Auto-upgrading and extension of the [L1 Attributes Predeployed Contract][l1-attributes.md]
```
sebastianst:
```suggestion
- Auto-upgrading and extension of the [L1 Attributes Predeployed Contract][./l1-attributes.md]
```
seems broken otherwise.

specs/protocol/jovian/exec-engine.md
```diff
@@ -77,3 +79,30 @@ This retains the purity of the function that computes the next block's base fee
 header, while still allowing them to be dynamically configured. Dynamic configuration is handled
 similarly to `gasLimit`, with the derivation pipeline providing the appropriate `SystemConfig`
 contract values to the block builder via `PayloadAttributesV3` parameters.
+
+## DA Footprint Limit
```
sebastianst:
```suggestion
## DA Footprint Block Limit
```

specs/SUMMARY.md
```diff
@@ -68,6 +68,7 @@
       - [Predeploys](./protocol/isthmus/predeploys.md)
     - [Jovian](./protocol/jovian/overview.md)
       - [Execution Engine](./protocol/jovian/exec-engine.md)
+      - [Derivation](./protocol/jovian/derivation.md)
       - [System Config](./protocol/jovian/system-config.md)
```
sebastianst: Missing `l1-attributes.md`.

specs/protocol/jovian/exec-engine.md
```diff
@@ -8,6 +8,8 @@
   - [Minimum Base Fee in Block Header](#minimum-base-fee-in-block-header)
   - [Minimum Base Fee in `PayloadAttributesV3`](#minimum-base-fee-in-payloadattributesv3)
   - [Rationale](#rationale)
+- [DA Footprint Limit](#da-footprint-limit)
```
sebastianst:
```suggestion
- [DA Footprint Block Limit](#da-footprint-block-limit)
```

specs/protocol/jovian/exec-engine.md
```diff
@@ -77,3 +79,30 @@ This retains the purity of the function that computes the next block's base fee
...
+
+## DA Footprint Limit
+
+From Jovian, `gasUsed` is equal to a block's "DA footprint" if the footprint exceeds
+the total gas used by transaction, and equal to the sum of the gas used by each transaction otherwise. As a result,
+blocks with high DA usage may cause the base fee to increase in subsequent blocks.
+
+A block's DA footprint is calculated by scaling the cumulative DA footprint of its transactions
+(as calculated by the [Fjord LZ Estimation](../fjord/exec-engine.md#fjord-l1-cost-fee-changes-fastlz-estimator) by
+a configurable scalar value, the `daFootprintGasScalar`.
```
sebastianst: I think this section should be expanded and should read more like the [Proposed Solution](https://github.com/ethereum-optimism/design-docs/blob/seb/calldata-block-limit/protocol/da-footprint-block-limit.md#proposed-solution) section of the design doc. In particular, it should be more specific about how exactly the `da_footprint` of a transaction is calculated and that a block's `total_da_footprint` is the sum over all transactions _without deposit transactions_, which is an important detail.

The details about how the scalar is loaded during execution are good (and missing from the design doc).

specs/protocol/jovian/l1-attributes.md
sebastianst:
```suggestion
Blocks after the Jovian activation block contain all pre-Jovian values 1:1, and also set `daFootprintGasScalar` to the value from the `SystemConfig`. If that value is `0`, then a default of `600` is set. 
```
Avoiding "should" as this is a rule, not a _should_.

specs/protocol/jovian/exec-engine.md
```diff
@@ -77,3 +79,30 @@ This retains the purity of the function that computes the next block's base fee
...
+
+The `daFootprintGasScalar` is loaded in a similar way to the `operatorFeeScalar` and `operatorFeeConstant`
+[included](../isthmus/exec-engine.md#operator-fee) in the Isthmus fork. It can be read in two interchangable ways:
+
+- read from the deposited L1 attributes (`daFootprintGasScalar`) of the current L2 block
+- read from the L1 Block Info contract (`0x4200000000000000000000000000000000000015`)
+  - using the solidity getter function `daFootprintGasScalar`
+  - using a direct storage-read: big-endian `uint16` in slot `9` at offset `0`.
+
+### Rationale
```
sebastianst, line 129: This section is generally good. Maybe we can expand a bit on it by some text from the design doc.

sebastianst (commented):
Generally good!
- The exec-engine section should be expanded to be more detailed.
- The payload attribs change needs to be removed.

specs/protocol/jovian/system-config.md
geoknee: Agree, but a link to the other part of the spec might be nicer and have less duplication (we expect to change the default and could forget to update it in multiple places)?

specs/protocol/system-config.md
joshklop: Sounds like a good suggestion, I'll add it to the issue.

specs/protocol/jovian/system-config.md
sebastianst: Even better, then we don't need to mention the value here again.

joshklop: Good points. I'll add a link for now.

> Derivation of the L1 info deposit transaction translates unset zero values to the default value

Does derivation translate the value, or should we use a default value in the system config? 

```solidity
uint16 daFootprintGasScalar = 600;
// ...
function setDAFootprintGasScalar(uint16 _daFootprintGasScalar) external onlyOwner { 
    // ...
}
```

specs/protocol/jovian/exec-engine.md
joshklop: I received some feedback from @geoknee about this section being a bit too wordy in a previous version. Is there anything specific that you think is missing?

specs/protocol/jovian/system-config.md
joshklop: From discussion with @geoknee out-of-band: 

We'll interpret `0` as `600` at all times. It is thus not possible to entirely disable the feature (since the lowest value is `1`), but that doesn't matter since only triple-digit values will have a noticeable impact.

specs/protocol/jovian/exec-engine.md
```diff
@@ -83,12 +83,19 @@ contract values to the block builder via `PayloadAttributesV3` parameters.
...
 (as calculated by the [Fjord LZ Estimation](../fjord/exec-engine.md#fjord-l1-cost-fee-changes-fastlz-estimator) by
-a configurable scalar value, the `daFootprintGasScalar`.
+a configurable scalar value, the `daFootprintGasScalar`:
+
+```python
+da_usage_estimate = max(minTransactionSize, intercept + fastlzCoef*fastlzSize / 1e6)
+da_footprint = da_usage_estimate * da_footprint_gas_scalar
+```
+
+Here, `fastlzSize` is the length of the FastLZ-compressed RLP-encoding of a transaction.
```
geoknee: I reworked this section here https://github.com/ethereum-optimism/specs/pull/776

geoknee (approved):

PR #778, opened 2025-09-18
fix jovian specs

geoknee:
We had generated the upgrade transactions with an account which had already been used

```
cast nonce 0x4210000000000000000000000000000000000004 --rpc-url=https://mainnet.optimism.io
1
```
This means the upgrade transactions would revert. There is already a contract at the address we were trying to deploy to.

**Description**

**Tests**

**Additional context**

**Metadata**


scripts/upgrades/gen_jovian_upgrade_tx_specs.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash -e
```
geoknee: This file is consumed by the script in development here #673. We don't have to commit this file now, but in general once that script is merged we will want to commit files like this along with the autogenerated markdown.

scripts/upgrades/gen_jovian_upgrade_tx_specs.sh
```diff
@@ -0,0 +1,37 @@
...
+        --contract-name "$contract_name" \
+        --from-address $FROM_ADDRESS \
+        --from-address-nonce $FROM_ADDRESS_NONCE \
+        --git-commit-hash $GIT_COMMIT_HASH \
+        --eth-rpc-url https://optimism.rpc.subquery.network/public \
+        --proxy-address "$proxy_address" \
+        --copy-contract-bytecode true
+    # increment the from address itself by 1 in hex here
+    FROM_ADDRESS=$(inc_hex "$FROM_ADDRESS")
+done
```
sebastianst: these look like generic functionality, could this be put into a reusable script so that only the env vars and the contracts array are in a forks specific generator script?

specs/protocol/jovian/derivation.md
```diff
@@ -22,28 +22,31 @@ derivation pipeline in the following order:
...
-- `from`: `0x4210000000000000000000000000000000000004`
+- `from`: `0x4210000000000000000000000000000000000006`
 - `to`: `null`
 - `mint`: `0`
 - `value`: `0`
 - `nonce`: `0`
 - `gasLimit`: `444775`
 - `data`: `0x0x608060405234801561001057600080...` ([full bytecode](../../../specs/static/bytecode/jovian-l1-block-deployment.txt))
-- `sourceHash`: `0x98faf23b9795967bc0b1c543144739d50dba3ea40420e77ad6ca9848dbfb62e8`,
+- `sourceHash`: `0x98faf23b9795967bc0b1c543144739d50dba3ea40420e77ad6ca9848dbfb62e8`,
```
sebastianst: space at the end

specs/protocol/jovian/derivation.md
```diff
@@ -61,7 +64,7 @@ make build-contracts
 jq -r ".bytecode.object" packages/contracts-bedrock/forge-artifacts/L1Block.sol/L1Block.json
 ```
 
-This transaction MUST deploy a contract with the following code hash
+This transaction MUST deploy a contract with the following code hash
```
sebastianst: space

scripts/upgrades/gen_jovian_upgrade_tx_specs.sh
geoknee: Yes probably, probably a tip for @teddyknox.

specs/protocol/jovian/derivation.md
geoknee: We should configure the linter to catch this, this is autogenerated md so we ideally don't want to require manual tidy up of the output.

scripts/upgrades/gen_jovian_upgrade_tx_specs.sh
sebastianst: OK tracking here https://github.com/ethereum-optimism/specs/issues/779

sebastianst (approved):

PR #782, opened 2025-09-19
feat: add revenue sharing specs

0xDiscotech:
**Description**

Introduce the `FeeSplitter` predeploy as part of a new opt-in revenue-sharing feature. The goal is to make revenue sharing simpler and more modular, with configurable values and automated handling where possible.

Includes supporting contracts like `SuperchainRevSharesCalculator`, `L1Withdrawer`, and `FeesDepositor`, which act as periphery components of the system and cover the first expected use case.

`FeeVaults` were updated to be more flexible with setters, while keeping backwards compatibility.

**Additional context**

[design-docs](https://github.com/ethereum-optimism/design-docs/pull/328/files)


tynes:
Do you mind moving these files out of the `jovian` directory since we will no longer be coupling this to a hardfork, we could put it in `protocol/revshare`. If you could also add entries [here](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/predeploys.md) that would be amazing

0xDiscotech:
> Do you mind moving these files out of the jovian directory since we will no longer be coupling this to a hardfork, we could put it in protocol/revshare. If you could also add entries [here](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/predeploys.md) that would be amazing

Ofc, pushed here [da9e535](https://github.com/ethereum-optimism/specs/pull/782/commits/da9e535e7ca8cac89210e7cac2cccea77ff4ea20)

tynes (approved):

PR #788, opened 2025-09-22
jovian: Fix DA footprint specs, change default scalar to 400

sebastianst:
**Description**

- Fixes the DA footprint formula
- Improves Rationale section
- Lower default value to 400

### DIFF
--- a/specs/protocol/jovian/exec-engine.md
+++ b/specs/protocol/jovian/exec-engine.md
@@ -83,26 +83,34 @@ contract values to the block builder via `PayloadAttributesV3` parameters.
 
 ## DA Footprint Block Limit
 
-A DA footprint block limit is introduced to limit the total amount of estimated compressed
+A _DA footprint block limit_ is introduced to limit the total amount of estimated compressed
 transaction data that can fit into a block.
- For each transaction, a new resource called DA footprint is tracked, next to its gas usage.
- It is scaled to the gas dimension so that its block total can also be limited by
- the block gas limit, like a block's total gas usage.
+For each transaction, a new resource called DA footprint is tracked, next to its gas usage.
+It is scaled to the gas dimension so that its block total can also be limited by
+the block gas limit, like a block's total gas usage.
 
 Let a block's `daFootprint` be defined as follows:
 
 ```python
-def daFootprint(block)
+def daFootprint(block: Block) -> int:
   daFootprint = 0
-  for tx in block.txs:
-      if !tx.IsDepositTx
-        daUsageEstimate = max(minTransactionSize, intercept + fastlzCoef * tx.fastlzSize / 1e6)
-        daFootprint += daUsageEstimate * daFootprintGasScalar
+
+  for tx in block.transactions:
+      if tx.type == DEPOSIT_TX_TYPE:
+          continue
+
+      daUsageEstimate = max(
+          minTransactionSize,
+          (intercept + fastlzCoef * tx.fastlzSize) // 1e6
+      )
+      daFootprint += daUsageEstimate * daFootprintGasScalar
+
   return daFootprint 
 ```
 
-where `intercept`, `minTransactionSize`, `fastLzCoef` and `fastlzSize`
-are defined in the [Fjord specs](../fjord/exec-engine.md) and `/` represents integer division.
+where `intercept`, `minTransactionSize`, `fastlzCoef` and `fastlzSize`
+are defined in the [Fjord specs](../fjord/exec-engine.md), `DEPOSIT_TX_TYPE` is `0x7E`,
+and `//` represents integer floor division.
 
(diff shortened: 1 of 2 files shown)


specs/protocol/jovian/exec-engine.md
```diff
@@ -128,9 +139,12 @@ It takes on a default value as described in the section on [L1 Attributes](./l1-
...
-While the current L1 fee mechanism charges for DA usage based on an estimate of the DA footprint of a transaction, it
-does not influence future base fee calculations. As a result, excessive DA usage is not efficiently reflected in the fee
-market, leading to suboptimal resource prices.
-
-By changing the meaning of the `gasUsed` field in times of high DA demand, the fee market can properly adjust without reverting
-to priority fee auctions (an inferior experience for users).
+While the current L1 fee mechanism charges for DA usage based on an estimate of the DA footprint of a transaction, no
+protocol mechanism currently reflects the limited _DA throughput_. E.g. on Ethereum with Pectra enabled, the available
+blob throughput is `~96 kB/s` (with a target of `~64 kB/s`), but the calldata floor gas price of `40` for calldata-heavy
+transactions allows for more incompressible transaction data throughput than even that of most single OP Stack chains.
```
geoknee: This isn't quite scanning for me. Is the point that the calldata pricing on L2 is such that, with incompressible data, a single OP Stack chain can produce more DA pressure per unit time than Ethereum mainnet can accomodate?

specs/protocol/jovian/exec-engine.md
```diff
@@ -128,9 +139,12 @@ It takes on a default value as described in the section on [L1 Attributes](./l1-
...
+This is currently mitigated at the policy level by batcher-sequencer throttling. However, since batcher throttling
+artificially constricts block building, this can lead to actually falling base fees, unnecessary losses for chain
```
geoknee:
```suggestion
This is currently mitigated at the policy level by batcher-sequencer throttling: a mechanism which
artificially constricts block building. This can cause base fees to fall, which implies unnecessary losses for chain
```

I prefer something like this since the specs don't otherwise mention throttling, so it needs a little introduction.

specs/protocol/jovian/exec-engine.md
```diff
@@ -113,6 +121,9 @@ The `gasUsed` must continue to be less than or equal to the block gas limit, mea
 (since the `daFootprint` must also be less than or equal to the block gas limit),
 blocks may have no more than `gasLimit/daFootprintGasScalar` total estimated DA usage.
 
+Note that this has the effect that the base fee is now directly influenced by DA usage because it is updated according
+to a block's `gasUsed` field.
```
geoknee:
```suggestion
Note that, since the base fee continues to be updated according to a block's `gasUsed` field, the base fee may be directly influenced by DA usage.
```

Nit: slight ambiguity around "it is updated", and I also prefer to emphasize that the update mechanism remains unchanged / already exists.

specs/protocol/jovian/exec-engine.md
```diff
@@ -83,26 +83,34 @@ contract values to the block builder via `PayloadAttributesV3` parameters.
...
- the block gas limit, like a block's total gas usage.
+For each transaction, a new resource called DA footprint is tracked, next to its gas usage.
+It is scaled to the gas dimension so that its block total can also be limited by
+the block gas limit, like a block's total gas usage.
 
 Let a block's `daFootprint` be defined as follows:
 
 ```python
-def daFootprint(block)
+def daFootprint(block: Block) -> int:
```
geoknee, line 95: Nit: does the `Block` type add much here? Do we need to define it?

geoknee (approved):
LGTM, few optional suggestions

specs/protocol/jovian/exec-engine.md
sebastianst: Yes.

specs/protocol/jovian/exec-engine.md
sebastianst: I was just following the example of EIPs (see e.g. [4844](https://eips.ethereum.org/EIPS/eip-4844#execution-layer-validation)), which specify input and output types. I wanted to add clarity by specifying the output type (`int`), so I found it consistent to also specify the input type.

specs/protocol/jovian/exec-engine.md
geoknee: OK. I think what is throwing me off is that we are talking about a throughput being _higher_ (more) than an OP Stack chain. But I was expecting you to compare to an L1 chain. Perhaps you can try to rephrase it?

sebastianst: Now I see that the sentence didn't make sense at all 😅 rewritten this section, PTAL

geoknee: Nice, works for me now!

PR #791, opened 2025-10-02
jovian: Clarify that upgrade block must not have non-deposits

sebastianst:
**Description**

Clarify the upgrade block specs. Upgrade block must not contain non-deposit user transactions.


specs/protocol/jovian/derivation.md
```diff
@@ -13,14 +13,23 @@
...
+The upgrade transaction details below are based on the monorepo at commit hash
+`7daaabbf2c2cce892aa171bc7e1331ad31bcc8ca`, and will be updated once a contracts release is made.
 
 ## Network Upgrade Transactions
 
+On the Jovian activation block, a set of deposit transaction-based upgrade transactions are deterministically generated
+by the derivation pipeline in the following order:
+
+- L1 Attributes Transaction (still calling the old `L1Block.setL1BlockValuesIsthmus()`)
+- User deposits from L1
```
geoknee: Isn't this really "potentially User deposits from L1", depending on if this is the first block in the epoch?

specs/protocol/jovian/derivation.md
```diff
@@ -13,14 +13,23 @@
...
+- Network Upgrade Transactions
+  - L1Block deployment
+  - Update L1Block Proxy ERC-1967 Implementation
+
+To not modify or interrupt the system behavior around gas computation, this block must not include any non-deposit user
+transactions by setting `noTxPool: true`. This rule must be checked during derivation.
```
geoknee: I would rather put this material above the `## Network Upgrade Transactions` section, otherwise it is unclear whether the L1 Attributes Transaction counts as a network upgrade transaction or not (I guess it does not).

specs/protocol/jovian/derivation.md
sebastianst: Yes, only potentially. I just copied this wording from previous forks, and it's not wrong.

specs/protocol/jovian/derivation.md
sebastianst: makes sense

specs/protocol/jovian/derivation.md
sebastianst: rewritten, PTAL

specs/protocol/jovian/derivation.md
```diff
@@ -10,17 +10,33 @@
...
-will be updated once a contracts release is made.
+The first block with a timestamp at or after the Jovian activation time is considered the _Jovian activation block_.
 
-On the activation block, a set of deposit transaction based upgrade transactions are deterministically generated by the
-derivation pipeline in the following order:
+To not modify or interrupt the system behavior regarding gas computations, the activation block must not include any
+non-deposit transactions. Sequencer must enforce this by setting `noTxPool` to `true` in the payload attributes. This
+rule must be checked during derivation at the batch verification stage, and if the batch for the activation block
+contains any transactions, it must be `DROP`ped following the batch-dropping rules introduced in the
+[Holocene upgrade span-batch rules](../holocene/derivation.md#span-batches).
```
geoknee: Did you link to the right section here? Span batches are optional, what would the rule be for singular batches?

geoknee (approved):

specs/protocol/jovian/derivation.md
sebastianst: Yes, singluar batches are also just dropped. TBH I copied this sentence from the Interop specs with the same rule.

sebastianst: Decided to just remove this reference. It's implicitly clear that Holocene rules just apply post-Holocene, no need to mention it again.

therens42-collab (approved):

PR #793, opened 2025-10-06
jovian: Update L1Block upgrade transactions

sebastianst:
**Description**

Updates L1Block upgrade transactions based on monorepo commit 41614ab527df1cf74f9820f60d8f588ade769679

Note that the GPO upgrade is still missing from the script env file and will be added later when it's monorepo PR is merged.


scripts/upgrades/jovian.sh
```diff
@@ -2,7 +2,7 @@
 # Configuration file for interop upgrade transaction specifications
```
geoknee, line 2: out of scope but this still says "interop"

sebastianst:
superseded by https://github.com/ethereum-optimism/specs/pull/799

Issue #797, opened 2025-10-11
Jovian: Same deployer address


varun-doshi:
**Bug Description**

Seems like the deployer address for `L1BLock Deployment` and `GasPriceOracle Deployment` are the same in Jovian

https://github.com/ethereum-optimism/specs/blob/7c4cb3a2092d6618d4c6fd4c9c8909ea6288b596/specs/protocol/jovian/derivation.md?plain=1#L58

https://github.com/ethereum-optimism/specs/blob/7c4cb3a2092d6618d4c6fd4c9c8909ea6288b596/specs/protocol/jovian/derivation.md?plain=1#L143

Which means for `nonce=0` both would output the same address: `0x3Ba4007f5C922FBb33C454B41ea7a1f11E83df2C`

as per:
https://github.com/ethereum-optimism/specs/blob/7c4cb3a2092d6618d4c6fd4c9c8909ea6288b596/specs/protocol/jovian/derivation.md?plain=1#L72

https://github.com/ethereum-optimism/specs/blob/7c4cb3a2092d6618d4c6fd4c9c8909ea6288b596/specs/protocol/jovian/derivation.md?plain=1#L157

From looking at the Isthmus specs, I assume the last digit of a deployer address increases monotonically i.e. the correct deployer for `GasPriceOracle Deployment` in Jovian should be `0x4210000000000000000000000000000000000007`

Please correct me if I'm mistaken

PR #798, opened 2025-10-11
fix: possible deployer mismatch Jovian

varun-doshi:
**Description**

Possible fix for #797 

Feel free to close this if this was not a valid issue


sebastianst:
thanks, this is being addressed in https://github.com/ethereum-optimism/specs/pull/799 where we update the generator script and use a monorepo commit that contains both contracts.

PR #861, opened 2025-11-14
feat: L2CM specs

0xiamflux:
Add specifications for the `L2ContractsManager` project. These enable deterministic hard fork upgrades of L2 predeploys.

- **L2ProxyAdmin**: L2-specific implementation of universal `ProxyAdmin` with `upgradePredeploys()` and ERC1967-only overrides
- **L2ContractsManager**: Orchestrates predeploy upgrades, deployed per hard fork as part of NUT bundles
- **ConditionalDeployer**: A wrapper around the `Create2Deployer` predeploy, allowing for non-reverting deterministic deployments in case of address collision.
- **TransactionGenerationScript**: A periphery contract that generates Network Upgrade Bundles for its execution during hard forks.

Includes invariants, assumptions, function specifications, and security considerations.

[Design document](https://github.com/ethereum-optimism/design-docs/pull/351)

### DIFF
--- a/specs/experimental/l2-contracts-manager.md
+++ b/specs/experimental/l2-contracts-manager.md
@@ -24,13 +24,24 @@
 
 ## Summary
 
-A contract responsible for orchestrating the upgrade of all the supported predeploys for a given hard fork. Contains the logic necessary to perform the upgrades and any additional work they might need for the correct setup. The base upgrade logic lives in an abstract contract that MUST be inherited. Given that each hard fork may require different upgrade logic or different configuration handling, a new `L2ContractsManager` is deployed for each hard fork. The manager is deployed as part of the NUT bundle for the target hard fork, ensuring that the upgrade logic is versioned and tied to what the specific hard fork requires.
+A contract responsible for orchestrating the upgrade of all the supported predeploys for a given hard fork.
+Contains the logic necessary to perform the upgrades and any additional work they might need for the correct setup.
+The base upgrade logic lives in an abstract contract that MUST be inherited. Given that each hard fork may require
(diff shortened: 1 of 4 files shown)


specs/experimental/predeploys.md
```diff
@@ -0,0 +1,156 @@
...
+  - [`changeProxyAdmin`](#changeproxyadmin)
+  - [`upgrade`](#upgrade)
+  - [`upgradeAndCall`](#upgradeandcall)
+- [Security Considerations](#security-considerations)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+## L2ProxyAdmin
```
0xOneTony, line 29: @maurelian seems like if we want to exclude the different proxy types we would need to create an L2ProxyAdmin from scratch because we cant override the functions if we inherit from ProxyAdmin.
If we dont care about removing the different proxy types then we can inherit from ProxyAdmin and just add the `upgradePredeploys` spec

specs/experimental/l2-contracts-manager.md
```diff
@@ -0,0 +1,183 @@
...
+## Definitions
+
+### NUT Bundle
+
+A collection of Network Upgrade Transactions stored in JSON format and executed in a specific order.
+The bundle includes transactions for deploying implementation contracts via `ConditionalDeployer`,
+deploying the `L2ContractsManager` for the target hard fork, and calling `L2ProxyAdmin.upgradePredeploys()`.
+The bundle is client-agnostic and can be consumed by any implementation:
+
+```json
```
0xOneTony, line 77: We will probably need a few more field on the JSON like a commit hash from where this bundle got created, what else?

PR #863, opened 2025-11-20
Clarify span batch outdated L1 origin rules

sebastianst:
**Description**

This clarifies the specs regarding outdated L1 origin rules for span batches before and after Holocene.

This is technically not a spec change because the batch queue rules already had this rule which applies to the singular batches coming out of span batches.

A note is added regarding a footgun for implementations, which got fixed here:
- https://github.com/ethereum-optimism/optimism/pull/18283
- https://github.com/op-rs/kona/pull/3059

### DIFF
--- a/specs/protocol/delta/span-batches.md
+++ b/specs/protocol/delta/span-batches.md
@@ -332,7 +332,7 @@ Span-batch rules, in validation order:
         if and only if the L1 epoch hash is correct.
     - `start_epoch_num < prev_l2_block.l1_origin.number` -> `drop`:
       epoch number cannot be older than the origin of parent block
-- Max Sequencer time-drift checks:
+- Max Sequencer time-drift & other L1 origin checks:
   - Note: The max time-drift is enforced for the _batch as a whole_, to keep the possible output variants small.
   - Variables:
     - `block_input`: an L2 block from the span-batch,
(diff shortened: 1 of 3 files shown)


specs/protocol/holocene/derivation.md
```diff
@@ -147,6 +147,21 @@ which the span batch was derived, is also immediately dropped (see also [Fast Ch
 Invalidation](#fast-channel-invalidation)). However, a `past` span batch is only dropped, without
 dropping the remaining channel.
 
+> [!Note]
+> A word regarding overlapping span batches: the existing batch queue rules already contain the rule
+> to drop batches whose L1 origin is older than that of the L2 safe head. The Delta span batch
+> checks also have an equivalent rule that applies to all singular batcher past the safe head.
+> Now full span batch checks aren't done any more in Holocene, but the batch queue rules are still
+> applied to singular batcher that are streamed out of span batches, so in particular this rule also
```
geoknee:
```suggestion
> applied to singular batches that are streamed out of span batches, so in particular this rule also
```

specs/protocol/delta/span-batches.md
```diff
@@ -341,6 +341,8 @@ Span-batch rules, in validation order:
       It may reach to the next origin outside the L1 origins of the span.
   - Rules:
     - For each `block_input` whose timestamp is greater than `safe_head.timestamp`:
+      - `block_input.l1_origin.number < safe_head.l1_origin.number` -> `drop`: enforce increasing L1
+        origins (also for non-overlapping section of overlapping span batches).
```
geoknee: The part in parentheses I think is going to be hard to understand when returning to this without context. Can you say what the span batch is overlapping with? That might help.

geoknee (approved):

specs/protocol/delta/span-batches.md
sebastianst: The term "overlapping span batch" is always used to denote overlap with the existing safe chain, also throughout this page, so I don't think there's a risk of misunderstanding. But now I think I'm just gonna remove those parentheses entirely. It's more confusing than helpful.

PR #875, opened 2026-01-21
feat: OPCMv2 spec

smartcontracts:
**Description**
Written to some extent alongside claude, although it's accurate.

### DIFF
--- a/specs/experimental/contracts/L1/opcm.md
+++ b/specs/experimental/contracts/L1/opcm.md
@@ -41,6 +41,8 @@
     - [Impact](#impact-7)
   - [i01-009: Time Independence](#i01-009-time-independence)
     - [Impact](#impact-8)
+  - [i01-010: PermissionedDisputeGame Fallback Preservation](#i01-010-permissioneddisputegame-fallback-preservation)
+    - [Impact](#impact-9)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
--- a/specs/experimental/contracts/L1/opcm.md
+++ b/specs/experimental/contracts/L1/opcm.md
@@ -87,8 +87,10 @@ interface of the upgrader rather than the contracts it deploys.
 
 A migration mechanism used during upgrades to set state values. Extra Instructions can be used to
 bootstrap new state that has no predecessor value (e.g., when an upgrade introduces a new variable)
-or to change state that already has a predecessor value. Extra Instructions are specific to the
-OPCM that defines them and must not apply to OPCMs of higher major versions.
+or to change state that already has a predecessor value. Each OPCM must specify which Extra
+Instructions are permitted, and those Extra Instructions must not be permitted in OPCMs of higher
+major versions. Permanent inputs should be added to the interface rather than implemented via Extra
+Instructions.
 
 ### Superchain ProxyAdmin Owner
 
--- a/specs/experimental/contracts/L1/opcm.md
+++ b/specs/experimental/contracts/L1/opcm.md
@@ -52,11 +52,11 @@ references to the implementation contracts for that release.
 
 OPCM provides three core capabilities:
 
-- **System Deployment**: Deploying all contracts for a new OP Stack [System](#system) at the
+- **System Deployment**: Deploying all L1 contracts for a new OP Stack [System](#system) at the
   version the OPCM represents. This capability is permissionless.
-- **System Upgrade**: Upgrading an existing [System](#system) to the version the OPCM represents.
-  This capability is designed to be invoked via delegatecall by the ProxyAdmin owner, as OPCM
-  itself holds no special authority.
+- **System Upgrade**: Upgrading the L1 contracts of an existing [System](#system) to the version
+  the OPCM represents. This capability is designed to be invoked via delegatecall by the ProxyAdmin
+  owner, as OPCM itself holds no special authority.
 - **SuperchainConfig Upgrade**: Upgrading the SuperchainConfig contract. This capability is
   designed to be invoked via delegatecall by the
   [Superchain ProxyAdmin Owner](#superchain-proxyadmin-owner).
@@ -85,10 +85,10 @@ interface of the upgrader rather than the contracts it deploys.
 
 ### Extra Instructions
 
-A migration mechanism used during upgrades to bootstrap new state that has no predecessor value.
-When an upgrade introduces a new variable that did not exist in the previous version, Extra
-Instructions allow users to provide initial values for these variables. Extra Instructions are
-specific to the OPCM that defines them and must not apply to OPCMs of higher major versions.
+A migration mechanism used during upgrades to set state values. Extra Instructions can be used to
+bootstrap new state that has no predecessor value (e.g., when an upgrade introduces a new variable)
+or to change state that already has a predecessor value. Extra Instructions are specific to the
+OPCM that defines them and must not apply to OPCMs of higher major versions.
 
 ### Superchain ProxyAdmin Owner
 
@@ -100,12 +100,13 @@ ProxyAdmin owner separate from individual chain ProxyAdmin owners.
 
 ### a01-001: Ethereum Environment
 
-OPCM is designed to run on Ethereum networks. Behavior on non-Ethereum EVM chains is not guaranteed.
+OPCM is designed to run on Ethereum L1 mainnet and Sepolia. Behavior on other networks is not
+guaranteed.
 
 #### Mitigations
 
-- Documentation clearly states Ethereum as the target environment
-- Testing is performed against Ethereum mainnet and official testnets
+- Documentation clearly states Ethereum L1 mainnet and Sepolia as the target environments
+- Testing is performed against Ethereum L1 mainnet and Sepolia
 
 ### a01-002: Implementation Contracts Trusted
 
@@ -114,8 +115,8 @@ represent.
 
 #### Mitigations
 
-- Implementation contracts are deployed through audited deployment scripts
-- Verification scripts confirm deployed bytecode matches the release commit
+- Audited verification scripts confirm deployed bytecode matches the release commit. Verification
+  scripts must be re-audited when changed.
 - Implementation addresses are immutable once OPCM is deployed
 
 ### a01-003: Code at Release Commit is Safe
@@ -138,7 +139,6 @@ constraints.
 
 - System upgrade capability only functions when delegatecalled, as OPCM holds no authority itself
 - ProxyAdmin ownership is typically controlled by governance multisigs
-- Upgrade transactions are subject to timelock delays where applicable
 
 ### a01-005: Superchain ProxyAdmin Owner Trusted
 
@@ -202,10 +202,12 @@ transferring control to unauthorized parties or breaking the security model of t
 
 Chains must follow the correct upgrade flow. Valid upgrade paths are:
 
-- Same major version to a higher minor version (e.g., v1.2.0 to v1.3.0)
+- Same major version to a higher minor version (e.g., v1.2.0 to v1.3.0). Skipping minor versions is
+  permitted, as a new increased minor version represents a replacement of the previous minor
+  version.
 - Current major version to the next major version (e.g., v1.x.y to v2.0.0)
 
-Downgrades and skipping required OPCMs are not permitted.
+Downgrades and skipping major versions are not permitted.
 
 #### Impact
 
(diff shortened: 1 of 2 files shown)


specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,280 @@
...
+Each OPCM instance represents a specific [Contract Release](#contract-release) and holds immutable
+references to the implementation contracts for that release.
+
+OPCM provides three core capabilities:
+
+- **System Deployment**: Deploying all contracts for a new OP Stack [System](#system) at the
+  version the OPCM represents. This capability is permissionless.
+- **System Upgrade**: Upgrading an existing [System](#system) to the version the OPCM represents.
+  This capability is designed to be invoked via delegatecall by the ProxyAdmin owner, as OPCM
+  itself holds no special authority.
```
mds1: We should clarify that these are just for L1 contracts specifically

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,280 @@
...
+The version of the OPContractsManager contract itself, which may differ from the
+[Contract Release](#contract-release) version. The OPCM Version reflects the capabilities and
+interface of the upgrader rather than the contracts it deploys.
+
+### Extra Instructions
+
+A migration mechanism used during upgrades to bootstrap new state that has no predecessor value.
+When an upgrade introduces a new variable that did not exist in the previous version, Extra
+Instructions allow users to provide initial values for these variables. Extra Instructions are
+specific to the OPCM that defines them and must not apply to OPCMs of higher major versions.
```
mds1: Can extra instructions also be used to change state that already has some predecessor value? Worth answering in this paragraph to remove any ambiguity about the spec

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,280 @@
...
+
+The ProxyAdmin owner for the SuperchainConfig contract. Since SuperchainConfig is shared across
+multiple chains on a network (many-to-one relationship), it has its own governance-controlled
+ProxyAdmin owner separate from individual chain ProxyAdmin owners.
+
+## Assumptions
+
+### a01-001: Ethereum Environment
+
+OPCM is designed to run on Ethereum networks. Behavior on non-Ethereum EVM chains is not guaranteed.
```
mds1: Does this imply OPCM should support running on L2s like OPMainnet because those are EVM chains? Or should we narrow this and say it's only designed to run on Ethereum mainnet and Sepolia?

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,280 @@
...
+- Testing is performed against Ethereum mainnet and official testnets
+
+### a01-002: Implementation Contracts Trusted
+
+The implementation contracts referenced by OPCM are trusted and correct for the release they
+represent.
+
+#### Mitigations
+
+- Implementation contracts are deployed through audited deployment scripts
```
mds1: Is this true? Have we ever actually audited our deployment scripts?

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,280 @@
...
+- Verification scripts confirm deployed bytecode matches the release commit
```
mds1: The verification script is audited, and should be re-audited when changed, which may be worth mentioning here

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,280 @@
...
+### a01-004: ProxyAdmin Owner Trusted for Upgrades
+
+The ProxyAdmin owner invoking system upgrades is trusted and operating within governance
+constraints.
+
+#### Mitigations
+
+- System upgrade capability only functions when delegatecalled, as OPCM holds no authority itself
+- ProxyAdmin ownership is typically controlled by governance multisigs
+- Upgrade transactions are subject to timelock delays where applicable
```
mds1: Where is this applicable?

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,280 @@
...
+transferring control to unauthorized parties or breaking the security model of the system.
+
+### i01-004: Upgrade Ordering
+
+Chains must follow the correct upgrade flow. Valid upgrade paths are:
+
+- Same major version to a higher minor version (e.g., v1.2.0 to v1.3.0)
+- Current major version to the next major version (e.g., v1.x.y to v2.0.0)
+
+Downgrades and skipping required OPCMs are not permitted.
```
mds1: One thing missing from here is whether you are allowed to skip minor versions or not. My understanding is that currently you are allowed to do that, eg U15 -> U16a

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,280 @@
...
+
+**Severity: High/Critical**
+
+If violated, chains could skip critical migration steps or security patches, potentially leaving
+them in an unsupported or vulnerable state that diverges from the expected upgrade path.
+
+### i01-005: Initialization Safety
+
+Contracts deployed or upgraded by OPCM must never be left uninitialized or open to initialization
+by unauthorized parties.
```
mds1: Just tightening this up. No one should be able to re-initialize, even some privileged role 
```suggestion
Contracts deployed or upgraded by OPCM must never be left uninitialized or open to re-initialization.
```

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,280 @@
...
+#### Impact
+
+**Severity: High/Critical**
+
+If violated, attackers could initialize contracts with malicious parameters, potentially taking
+ownership of critical system components or corrupting system state.
+
+### i01-006: SuperchainConfig Backwards Compatibility
+
+SuperchainConfig must maintain backwards compatibility with all past versions of the OP Stack.
```
mds1: We should define the oldest op-contracts release here, there was a time where versions of the OP stack did not have releases and I don't think we need to support those. My gut is find the MCP L1 release and call that the oldest one that we need to support, but maybe we should be a little more rigorous here

specs/experimental/contracts/L1/opcm.md
smartcontracts: Yes, they can be

specs/experimental/contracts/L1/opcm.md
smartcontracts: Right, I was trying to capture that with Ethereum networks, but we probably want to be more specific

specs/experimental/contracts/L1/opcm.md
smartcontracts: Nope, not true, need to fix

mds1 (approved):

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,283 @@
...
+The version of the OPContractsManager contract itself, which may differ from the
+[Contract Release](#contract-release) version. The OPCM Version reflects the capabilities and
+interface of the upgrader rather than the contracts it deploys.
+
+### Extra Instructions
+
+A migration mechanism used during upgrades to set state values. Extra Instructions can be used to
+bootstrap new state that has no predecessor value (e.g., when an upgrade introduces a new variable)
+or to change state that already has a predecessor value. Extra Instructions are specific to the
+OPCM that defines them and must not apply to OPCMs of higher major versions.
```
maurelian: This last sentence is pretty vague in the absence of the actual implementation. I think it's difficult to fully address but perhaps:

```suggestion
or to change state that already has a predecessor value. Each OPCM must specify which Extra Instructions are permitted, and those Extra Instruction
must not be permitted in OPCMs of higher major versions (rather they should be added to the interface as first class citizens). 
```

That might be too verbose or redundant with later content.

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,283 @@
...
+The implementation contracts referenced by OPCM are trusted and correct for the release they
+represent.
+
+#### Mitigations
+
+- Audited verification scripts confirm deployed bytecode matches the release commit. Verification
+  scripts must be re-audited when changed.
+- Implementation addresses are immutable once OPCM is deployed
+
+### a01-003: Code at Release Commit is Safe
```
maurelian, line 126: This point feels a bit extraneous tbh, like it doesn't impact on the implementation of the OPCM, it's just part of the threat model/supply chain safety.

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,283 @@
...
+
+#### Mitigations
+
+- SuperchainConfig upgrade capability only functions when delegatecalled
+- Superchain governance controls the Superchain ProxyAdmin Owner
+- SuperchainConfig changes affect multiple chains, requiring heightened governance scrutiny
+
+## Invariants
+
+### i01-001: Code Integrity
```
maurelian, line 160: somewhat similar to the above, this is about the supply chain. Perhaps we need specs for  `DeployImplementations` and `DeployOpchain`.

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,283 @@
...
+and preserved without modification.
+
+#### Impact
+
+**Severity: High/Critical**
+
+If violated, upgrades could silently change critical system properties like ownership, potentially
+transferring control to unauthorized parties or breaking the security model of the system.
+
+### i01-004: Upgrade Ordering
```
maurelian, line 205: I think this is better included as an Assumption.

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,283 @@
...
+re-initialization.
+
+#### Impact
+
+**Severity: High/Critical**
+
+If violated, attackers could initialize contracts with malicious parameters, potentially taking
+ownership of critical system components or corrupting system state.
+
+### i01-006: SuperchainConfig Backwards Compatibility
```
maurelian, line 235: this is a good callout I would not have thought to include here.

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,283 @@
...
+
+Given valid parameters, a single system deployment or upgrade operation must always be executable
+within the gas limits of an Ethereum block.
+
+#### Impact
+
+**Severity: High**
+
+If violated, it would be impossible to deploy new chains or upgrade existing chains, effectively
+breaking the core functionality of OPCM and blocking all future upgrades.
```
maurelian, line 257: Do we have testing for this?

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,283 @@
...
+
+### i01-008: Batch Operation Gas Efficiency
```
maurelian, line 259: same as previous

specs/experimental/contracts/L1/opcm.md
```diff
@@ -0,0 +1,283 @@
...
+Multiple upgrade operations (approximately 5) should be executable within a single transaction.
+
+#### Impact
+
+**Severity: Low**
+
+If violated, upgrading multiple chains would require separate transactions, increasing operational
+complexity and cost but not blocking functionality.
+
+### i01-009: Time Independence
```
maurelian, line 270: another good callout

specs/experimental/contracts/L1/opcm.md
smartcontracts: I think this is actually an invariant we need to explicitly maintain. If it's an assumption then we wouldn't have any controls on upgrade ordering.

specs/experimental/contracts/L1/opcm.md
smartcontracts: Somewhat, yes. We test that deploy/upgrade fits below the gas limit. We don't have more complex invariant/fuzz testing that it works with all possible valid parameters.

specs/experimental/contracts/L1/opcm.md
smartcontracts: Hmmm we should probably have testing for this

specs/experimental/contracts/L1/opcm.md
smartcontracts: This one I want to keep in here. It is about DeployImplementations and DeployOpchain. The idea of this invariant is that it is explicitly meant to capture this type of information about those two scripts and therefore implies something about the behavior of the scripts. IMO this invariant is accurate and is a core quality of OPCM, but we use many different tools to maintain that property.

specs/experimental/contracts/L1/opcm.md
smartcontracts: Going to keep as-is because I do want to keep track of this higher-level quality about OPCM. It's not really the most useful assumption in some sense but I'd rather be overly verbose than not verbose enough.

specs/experimental/contracts/L1/opcm.md
maurelian: Actually, yes, you are correct. I had forgotten that we had added that feature.

maurelian (approved):

PR #876, opened 2026-01-27
specs: add L2 contract upgrades overview and ConditionalDeployer

maurelian:
Adds foundational specification for L2 predeploy contract upgrades:

- Overview: Describes deterministic hard fork-driven upgrade mechanism
- ConditionalDeployer: Idempotent CREATE2 deployment with collision handling

Includes 4 invariants, 2 assumptions, and interface definition.
Part 1 of 6-part specification.

---
**Stack:**
- **→ #876 (Layer 1: Overview + ConditionalDeployer)** ← You are here
- #877 (Layer 2: L2ProxyAdmin)
- #878 (Layer 3: L2ContractsManager)
- #879 (Layer 4: Network Upgrade Transaction Bundle)
- #880 (Layer 5: Custom Upgrade Block Gas Limit)
- #881 (Layer 6: Upgrade Process)


specs/protocol/l2-contract-upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+**Severity: Critical**
+
+If deployments are not idempotent, upgrade transactions that attempt to deploy unchanged implementations would revert
+or deploy the implementation to an unexpected address. In the latter case, the proxy would then be upgrade incorrectly,
+as we must predict the implementation address in advance in order to correctly generate the NUT bundle.
+
+#### i03-003: Non-Reverting Collision Handling
+
+When a [CREATE2 Collision](#create2-collision) is detected (contract already deployed at the target address), the
+ConditionalDeployer MUST return successfully without reverting and without modifying blockchain state.
```
0xOneTony, line 133: maybe worth mentioning what address it will return in case of collision

specs/protocol/l2-contract-upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+This specification defines the mechanism for upgrading L2 predeploy contracts through deterministic, hard fork-driven
+Network Upgrade Transactions (NUTs). The system enables safe, well-tested upgrades of L2 contracts with both
+implementation and upgrade paths written in Solidity, ensuring determinism, verifiability, and testability across all
+client implementations.
+
+The upgrade system maintains the existing pattern of injecting Network Upgrade Transactions at specific fork block
+heights while improving the development and testing process. Upgrade transactions are defined in JSON bundles that are
+tracked in git, generated from Solidity scripts, and executed deterministically at fork activation.
+
+## ConditionalDeployer
```
0xOneTony, line 41: `ConditionDeployer` is gonna be a periphery contract rather than a Predeploy right?

maurelian: What do you mean by periphery contract, like a pre-install?

My thoughts:
1. We should also add it to the L2Genesis, as we want it available in future hardforks.
2. It doesn't need to be upgradable. If we need to replace it we can just deploy a different copy of it using the same method.

specs/protocol/l2-contract-upgrades.md
maurelian: Do you think it should return the deployment address? In some ways it doesn't matter since we can't actually use the return value, as it needs to be pre-computed off-chain (all we can do is inject transactions, we can't read their results).

0xiamflux: I think the cleaner approach would be to just not return anything, as you mention for this use case we cannot use the address anyway and returning the address in cases where no contract is deployed (because it was deployed before) feels off.

PR #877, opened 2026-01-27
specs: add L2ProxyAdmin to upgrade spec

maurelian:
Adds L2ProxyAdmin component for batch upgrade orchestration:

- Defines upgradePredeploys() function for DELEGATECALL-based batch upgrades
- Exclusive Depositor Account authorization
- Backwards compatibility with existing ProxyAdmin interface

Includes 4 invariants, 3 assumptions, and interface definition.
Part 2 of 6-part specification.

---
**Stack:**
- #876 (Layer 1: Overview + ConditionalDeployer)
- **→ #877 (Layer 2: L2ProxyAdmin)** ← You are here

### DIFF
--- a/specs/protocol/l2-contract-upgrades.md
+++ b/specs/protocol/l2-contract-upgrades.md
@@ -24,6 +24,27 @@
       - [Impact](#impact-2)
     - [i04-004: Collision Detection Accuracy](#i04-004-collision-detection-accuracy)
       - [Impact](#impact-3)
+- [L2ProxyAdmin](#l2proxyadmin)
+  - [Overview](#overview-2)
+  - [Definitions](#definitions-1)
+    - [Depositor Account](#depositor-account)
+    - [Predeploy](#predeploy)
+  - [Assumptions](#assumptions-1)
+    - [a01-001: Depositor Account is Controlled by Protocol](#a01-001-depositor-account-is-controlled-by-protocol)
+      - [Mitigations](#mitigations-2)
+    - [a02-002: L2ContractsManager is Trusted](#a02-002-l2contractsmanager-is-trusted)
+      - [Mitigations](#mitigations-3)
+    - [a03-003: Predeploy Proxies Follow Expected Patterns](#a03-003-predeploy-proxies-follow-expected-patterns)
+      - [Mitigations](#mitigations-4)
+  - [Invariants](#invariants-1)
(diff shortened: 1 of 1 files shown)


specs/protocol/l2-contract-upgrades.md
```diff
@@ -164,3 +186,130 @@ interface IConditionalDeployer {
...
+
+##### Impact
+
+**Severity: Critical**
+
+If the DELEGATECALL is not properly executed, upgrades could fail silently leaving proxies in inconsistent states, or
+worse, the ProxyAdmin's own storage could be corrupted. This could result in loss of admin control over predeploys or
+enable exploitation of corrupted state.
+
+#### i03-003: Backwards Compatibility Maintained
```
0xOneTony, line 270: Does this also mean we will keep support of the `ResolvedDelegate` and `Chugsplash` proxies?

specs/protocol/l2-contract-upgrades.md
```diff
@@ -164,3 +186,130 @@ interface IConditionalDeployer {
...
+[Predeploys](#predeploy) being upgraded follow the expected proxy patterns (ERC-1967 or similar) and correctly handle
+`upgradeTo()` and `upgradeToAndCall()` operations when called by the ProxyAdmin.
+
+##### Mitigations
+
+- All L2 predeploys use standardized proxy implementations from the contracts-bedrock package
+- Proxy implementations are thoroughly tested and audited
+- Fork-based testing validates upgrade operations against actual deployed proxies
+
+### Invariants
```
0xOneTony, line 240: I think an other invariant could be that `upgradePredeploys` must be atomic and we cant have partial upgrades if for whatever reason this fails mid upgrade

specs/protocol/l2-contract-upgrades.md
maurelian: we need to keep the full interface available, but we don't necessary need to keep the logic that enables upgrading ResolvedDelegate and Chugsplash proxies.

specs/protocol/l2-contract-upgrades.md
maurelian: We have that for the [L2ContractsManager](https://github.com/ethereum-optimism/specs/blob/3b108b11bf0fc3a6aed378e4aba9d367d054ade4/specs/protocol/l2-contract-upgrades.md#L437), which I think is sufficient

Issue #781, opened 2025-09-19
Enshrine new convention for Network Upgrade Transactions which deploy new contracts


geoknee:
Currently, for transactions such as these (example https://specs.optimism.io/protocol/isthmus/derivation.html#l1block-deployment) we choose the `from` address to be a unique account which has `nonce=0`. The choice of `from` affects the deployment address of the contract. Using a fresh account means we can safely set `nonce:0` in the upgrade transaction. 

That unique account tends to be chosen in the `0x421...` namespace. Choosing it correctly involves checking the entire spec history to ensure we aren't reusing an address. This is error prone, particularly when forks get reordered. 

A better pattern is 

> take the first 20 bytes of the source hash as sender address. That way you have a guaranteed unique deployer address, and thus nonce 0, and thus a unique and deterministic deployed contract address. Without making it coupled to bytecode, so the proxy upgrade tx input stays stable on dev changes

h/t to @protolambda for this idea. 

The source hash is defined here https://specs.optimism.io/protocol/deposits.html#source-hash-computation

Note: the idea of using a CREATE2 type deployment was considered, but discarded in favour of the above approach (in particular it would couple the deployment address and tx input to the contract bytecode, which is undesirable).

protolambda:
Note on address collision resistance of the suggestion:

- `CREATE` addresses are computed as: `keccak256(rlp([sender, nonce]))` (the RLP never starts with many zeroes, and certain ranges of bytes, alike to legacy-tx-typing)
-  `CREATE2` addresses are computed as: `keccak256( 0xff ++ address ++ salt ++ keccak256(init_code))[12:]` (the 0xff separates it from the RLP)
- upgrade source-hashes are computed as: `keccak256(bytes32(uint256(2)), keccak256(intent))` (and I assume the `sourceHash[12:]` as address here)

So the start of the keccak preimage always differs (non-zero RLP byte in legacy create case, `0xff` in eth create2 case, `0x000...002` in this update case) and so there are no possible address collisions.


If we want to be extra-safe, we could use the `CREATE2` computation, set the deployer address to a system-address, and the source-hash as salt. Then it fits in the existing address computation framework, and won't collide in the future in unexpected ways, and users are stopped from colliding with it by using an inaccessible system address as create2-deployer.

PR #884, opened 2026-01-28
specs: add L2 contract upgrades

maurelian:


specs/protocol/l2-contract-upgrades.md
```diff
@@ -0,0 +1,1171 @@
...
+{
+  "metadata": {
+    "version": "1.0.0",
+    "fork": "ForkName",
+    "generatedAt": "2024-01-15T10:30:00Z",
+    "sourceCommit": "abc123def456...",
+    "compiler": {
+      "version": "0.8.25",
+      "settings": {...}
+    }
```
maurelian: Don't think we need this

specs/protocol/l2-contract-upgrades.md
```diff
@@ -0,0 +1,1171 @@
...
+
+**Severity: Critical**
+
+If transactions are misordered, executions will fail when attempting to call non-existent contracts or reference
+incorrect addresses. This would cause the entire upgrade to fail at fork activation, halting the chain.
+
+#### i04-004: Correct Nonce Sequencing
+
+Transaction nonces for the [Depositor Account](#depositor-account) MUST form a contiguous sequence with no gaps or
+duplicates. Each transaction must use the next available nonce.
```
maurelian: I need to seek more clarity about how the sender accounts and nonces should be determined for each upgrade transaction.

specs/protocol/l2-contract-upgrades.md
```diff
@@ -0,0 +1,1171 @@
...
+If the L2ContractsManager address could be manipulated, an attacker could redirect the DELEGATECALL to a malicious
+contract, achieving the same impact as i01-001 (complete compromise of all predeploys).
+
+## L2ContractsManager
+
+### Overview
+
+The L2ContractsManager is a contract deployed fresh for each upgrade that contains the upgrade logic and coordination
+for a specific set of predeploy upgrades. When invoked via DELEGATECALL from the [L2ProxyAdmin](#l2proxyadmin), it
+gathers network-specific configuration from existing predeploys, computes new implementation addresses, and executes
```
0xiamflux: I think L2CM would not be the one computing the addresses but the offchain script that generates the NUTs.

maurelian: You are correct. It's been fixed.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,696 @@
...
+  "metadata": {
+    "version": "1.0.0",
+    "fork": "ForkName",
+    "generatedAt": "2024-01-15T10:30:00Z",
+    "sourceCommit": "abc123def456...",
+    "compiler": {
+      "version": "0.8.25",
+      "settings": {...}
+    }
+  },
```
maurelian, line 418: This metadata was suggested by claude, but I don't actually see much value in it. I'll comment on each in the field reqs section below.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,696 @@
...
+      "data": "0xabcd...",
+      "gasLimit": "1000000"
+    }
+  ]
+}
+```
+
+**Field Requirements:**
+
+- `metadata.version`: Bundle format version for compatibility
```
maurelian: Describing the bundle format version might actually be useful relative to the other metadata fields.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,696 @@
...
+- `metadata.fork`: Name of the fork this bundle implements
```
maurelian: This won't be known when the bundle is generated in PRs to develop. When the bundle is canonicalized we will probably know the fork name, but setting this value at that time will add manual toil that I'm not sure is valuable.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,696 @@
...
+- `metadata.generatedAt`: ISO 8601 timestamp of generation (informational only, not used in determinism)
```
maurelian: Just adds noise. We def don't want this in develop PRs.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,696 @@
...
+- `metadata.sourceCommit`: Git commit hash of the source code that generated this bundle
```
maurelian: This would be useful to have for provenance, but it would mean that you always need to commit this bundle in a separate commit on top of contract changes, and then the CI verification would need to checkout the source commit.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,696 @@
...
+- `metadata.compiler`: Compiler version and settings used to build contracts
```
maurelian: The compiler settings are pretty complex and are defined in foundry.toml, not reasonable to copy them all here.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,696 @@
...
+- `transactions`: Array of transaction objects in execution order
+- `transactions[].nonce`: Nonce for the [Depositor Account](./l2-upgrades-2-contracts.md#depositor-account), starting
```
maurelian: This should be removed, the nonce should just be the next one.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,696 @@
...
+### Transaction Execution Sequence
+
+Within the fork activation block, transactions execute in this order:
+
+1. **ConditionalDeployer Deployment** (if needed): Deploy the ConditionalDeployer contract
+2. **Implementation Deployments**: For each predeploy being upgraded, deploy new implementation via
+   ConditionalDeployer
+3. **L2ContractsManager Deployment**: Deploy the L2ContractsManager for this upgrade
+4. **ProxyAdmin Upgrade** (if needed): Upgrade the L2ProxyAdmin implementation
+5. **Batch Upgrade Execution**: Call `L2ProxyAdmin.upgradePredeploys(l2ContractsManagerAddress)` which:
```
0xiamflux: Would the availability of this function be an assumption worth to mention?

maurelian: There are already a bunch of things that are assumptions in one section and invariants in others. It's kind of gotten out of control and I've been working on cutting it down, so I don't want to lose too much ground by adding more assumptions unless they really add something novel.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,696 @@
...
+
+**Severity: High**
+
+A flawed upgrade that reaches fork activation cannot be undone through the normal upgrade process and would require a
+new fork with corrective upgrades.
+
+#### iUP-005: Verifiable Upgrade Execution
+
+After fork activation, it MUST be possible to verify that the executed upgrade transactions match the committed bundle
+and that the resulting contract state matches expectations.
```
0xiamflux, line 203: Feels like it would be worth to mention if we take an action in the case the contracts **do not** match the expectations, but then that means there's something fundamentally wrong in the process.

maurelian: Yeah, that would violate other invariants, but it probably makes sense here just to indicate that another upgrade would probably be required to address any issues.

specs/protocol/l2-upgrades-2-contracts.md
```diff
@@ -0,0 +1,492 @@
...
+If the L2ContractsManager address could be manipulated, an attacker could redirect the DELEGATECALL to a malicious
+contract, achieving complete compromise of all predeploys.
+
+## L2ContractsManager
+
+### Overview
+
+The L2ContractsManager is a contract deployed fresh for each upgrade that contains the upgrade logic and coordination
+for a specific set of predeploy upgrades. When invoked via DELEGATECALL from the [L2ProxyAdmin](#l2proxyadmin), it
+gathers network-specific configuration from existing predeploys, computes new implementation addresses, and executes
```
0xiamflux: I think the L2CM must somehow receive the pre-computed addresses not compute them.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,576 @@
...
+
+#### aUP-001: Fork Activation Time is Coordinated
+
+All OP Stack chains coordinate fork activation times to enable consistent upgrade rollout. The fork activation
+timestamp is communicated well in advance of activation to allow node operators to prepare.
+
+##### Mitigations
+
+- Fork activation timestamps are published in protocol documentation and configuration files
+- Activation times are set and communicated far enough in advance to allow preparation
```
sebastianst: note that the canonical place for fork activation times to be set is the superchain-registry

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,576 @@
...
+
+1. **Implementation & Bundle Generation**: Contract changes are implemented and the
+   [bundle generation script](#bundle-generation-script) produces a deterministic JSON bundle containing all upgrade
+   transactions. This bundle will be checked into the monorepo at
+   `packages/contracts-bedrock/snapshots/current-l2-upgrade-bundle.json` (or similar). CI will enforce that this
+   bundle always corresponds with bundle generated from the source code in the current commit.
+
+2. **Bundle Canonicalization**: Prior to cutting a contracts release, the `current-l2-upgrade-bundle.json` should
+   be copied to a new file named for the corresponding upgrade (ie. `holocene-upgrade-bundle.json`). The contracts
+   release should then be cut.
```
sebastianst: I'm wondering how we'll manage this in practice while iterating on a new fork. E.g. will we need to manually copy the bundle to `<fork>-upgrade-bundle.json` every time? How is it checked that this copied file is the same? And what happens when the `current` deviates from the latest fork bundle?

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,576 @@
...
+   upgrade L1 contracts. Perhaps we move to a double semver release like `op-contracts/v6.0.0-v1.0.0`
+   (`op-contracts/{{L1-Semver}}-{{L2-Semver}}`)?
+   Perhaps `op-deployer` should just be able to manage separate L1 and L2 contract releases
+   (op-contracts-l1/v6.0.0 and op-contracts-l2/v1.0.0)?
+   It might seem like it makes sense to have the the L2 contracts just get tagged by the op-node release, but that
+   breaks the current op-deployer release flow which releases on top of a contracts release.
+
+3. **Client Integration**: The canonical bundle is integrated into L2 client implementations and released with fork
+   activation logic. TBD how the client should actually consume the canonical bundle. Does the contracts package
+   become a dep, or does the build just grab the json and embed it in the binary?
```
sebastianst: Build should embed the JSON, Go supports this easily, we can lazy-parse when we hit a fork block during block execution. TBD if Rust can embed it just as easily, I guess yes. Or it's a step in the build pipeline.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,576 @@
...
+### Transaction Execution Sequence
+
+Within the fork activation block, transactions will execute in this order:
+
+1. **ConditionalDeployer Deployment** (if needed): Deploy the ConditionalDeployer contract
+2. **Implementation Deployments**: For each predeploy being upgraded, deploy new implementation via
+   ConditionalDeployer
+3. **L2ContractsManager Deployment**: Deploy the L2ContractsManager for this upgrade
+4. **ProxyAdmin Upgrade** (if needed): Upgrade the L2ProxyAdmin implementation
+5. **Batch Upgrade Execution**: Call `L2ProxyAdmin.upgradePredeploys(l2ContractsManagerAddress)` which:
```
sebastianst: Should contain the **L1 Info Deposit** as a 0. transaction to highlight that it is executed before all others.

sebastianst: what exactly is the effect of "if needed"? will there still be a tx and it will check the existence and then just not do anything if the target contract exists?

specs/protocol/l2-upgrades-1-execution.md
geoknee, line 126: You could mention the superchain registry here?

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,576 @@
...
+
+### Bundle Format
+
+The bundle is a JSON file with the following structure:
+
+```json
+{
+  "metadata": {
+    "version": "1.0.0",
+    "fork": "ForkName",
```
sebastianst: how is this fork selected?

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,576 @@
...
+      "settings": {...}
+    }
+  },
+  "transactions": [
+    {
+      "nonce": 0,
+      "to": "0x1234...",
+      "value": "0",
+      "data": "0xabcd...",
+      "gasLimit": "1000000"
```
sebastianst: don't we also need `from` for deposit txs?

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,576 @@
...
+
+#### aUP-002: Testing Environments Match Production
+
+Fork-based testing environments accurately represent production chain state, allowing upgrade testing to catch issues
+that would occur in production.
+
+##### Mitigations
+
+- Fork tests use actual mainnet chain state as starting point
+- Testing validates against multiple chains with different configurations
```
geoknee: I don't think our current testing is this deep or this broad. What are the different configurations we are interested in? Which mainnet chain state do we start with? OPM? Why is that important?

sebastianst (commented):
Part I

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,576 @@
...
+### Overview
+
+The Upgrade Process defines the complete lifecycle of an L2 predeploy upgrade, from initial development through fork
+activation and execution. The process ensures that upgrades are developed safely, tested thoroughly, and executed
+deterministically across all OP Stack chains.
+
+This end-to-end process integrates all components of the upgrade system: contract development, bundle generation,
+testing, verification, and execution at fork activation.
+
+### Definitions
```
maurelian, line 108: Uh let me just assumption and or invariant to add is that the transaction payload is identical for all chains. There may be ephemeral differences in execution such as data red, but the payloads are identical.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,576 @@
...
+    "generatedAt": "2024-01-15T10:30:00Z",
+    "sourceCommit": "abc123def456...",
+    "compiler": {
+      "version": "0.8.25",
+      "settings": {...}
+    }
+  },
+  "transactions": [
+    {
+      "nonce": 0,
```
maurelian: Don't think we need the nonce here anymore.

specs/protocol/l2-upgrades-1-execution.md
maurelian: No, I think we can default to the standard depositor account, and use CREATE2 to ensure that the nonce is not an input to the implementation address.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,576 @@
...
+
+### Transaction Execution Sequence
+
+Within the fork activation block, transactions will execute in this order:
+
+1. **ConditionalDeployer Deployment** (if needed): Deploy the ConditionalDeployer contract
+2. **Implementation Deployments**: For each predeploy being upgraded, deploy new implementation via
+   ConditionalDeployer
+3. **L2ContractsManager Deployment**: Deploy the L2ContractsManager for this upgrade
+4. **ProxyAdmin Upgrade** (if needed): Upgrade the L2ProxyAdmin implementation
```
maurelian: This will need to come from the zero address, and that in turn necessitates the ability to specify the `from` address, though we may choose to default to the depositor in the go/rust code.

maurelian: Reasoning: we want to default to the depositor, and using the CREATE2 Deterministic Deployment proxy allows us to do that without worrying about the nonce.

However for this one contract upgrade we need to make use of the `address(0)` loophole in the Proxy.sol implementation in order to upgrade the L2ProxyAdmin.

maurelian: Relevant: https://github.com/ethereum-optimism/specs/issues/781#issuecomment-3312216017

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,576 @@
...
+- `metadata.version`: Bundle format version for compatibility
+- `metadata.fork`: Name of the fork this bundle implements
+- `metadata.generatedAt`: ISO 8601 timestamp of generation (informational only, not used in determinism)
+- `metadata.sourceCommit`: Git commit hash of the source code that generated this bundle.
+- `metadata.compiler`: Compiler version and settings used to build contracts
+- `transactions`: Array of transaction objects in execution order
+- `transactions[].nonce`: Nonce for the [Depositor Account](./l2-upgrades-2-contracts.md#depositor-account), starting
+  from current nonce
+- `transactions[].to`: Target address (contract being called or deployed)
+- `transactions[].value`: ETH value to send (typically "0")
```
sebastianst: we probably can leave this out? I cannot think of an upgrade tx that needs to send ether.

specs/protocol/l2-upgrades-1-execution.md
maurelian: >Should contain the L1 Info Deposit as a 0. transaction to highlight that it is executed before all others.

Good callout, I've added that L1 Info deposit to the beginning of the block.

> what exactly is the effect of "if needed"? 

That ('if needed') is poor phrasing, I'm updating it to 'one-time only' to indicate that these are necessary to enable infrastructure which will support this mechanism of upgrades in general. Once these steps are complete they do not need to be included in future upgrades.

specs/protocol/l2-upgrades-1-execution.md
maurelian: With a bit more thought and conversation during the discussion with George, I am now leaning towards simply always using the current bundle and not having any bundle to copy over to. The thinking is that:
1. all contracts should _always be release ready_.
2. new functionality, can be set behind a "feature flag" which is only activated by a special boolean after a fork (this would be similar to the [isJovian](https://github.com/ethereum-optimism/optimism/blob/8096e17e5df78c5e23365baa2503959510ce0de0/packages/contracts-bedrock/src/L2/GasPriceOracle.sol#L226) boolean value and [setJovian](https://github.com/ethereum-optimism/optimism/blob/laughing-wing/op-node/rollup/derive/jovian_upgrade_transactions.go) upgrade txn work).

This would mean that there is no additional toil required in order to cut an node release.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 72699d7 - Kept the bundle format version field as you suggested. It is useful for compatibility tracking.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 72699d7 - Removed the `fork` field from metadata.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 72699d7 - Removed the `generatedAt` field from metadata.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 72699d7 - Removed the `sourceCommit` field from metadata to avoid the workflow complexity you mentioned.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 72699d7 - Removed compiler settings from metadata. The settings are defined in foundry.toml.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 72699d7 - Removed the `nonce` field from transaction objects.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 72699d7 - Marked `value` field as optional (defaults to "0" if omitted).

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 72699d7 - Added clarification that if contracts don't match expectations, it would violate other invariants and likely require another upgrade to address the issues.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 72699d7 - Added aUP-004 assumption about transaction payloads being identical across all chains.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 72699d7 - Added optional `from` field to bundle format. It defaults to the Depositor Account but can be set to address(0) for the L2ProxyAdmin upgrade to use the zero-address upgrade path in Proxy.sol.

specs/protocol/l2-upgrades-2-contracts.md
maurelian: Fixed in 72699d7 - Clarified in aL2CM-002 that the L2ContractsManager receives pre-computed addresses via its constructor (from the off-chain bundle generation script) rather than computing them itself.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,581 @@
...
+### Overview
+
+The Custom Upgrade Block Gas Limit mechanism provides guaranteed gas availability for executing upgrade transactions at
+fork activation, independent of the regular block gas limit and system transaction gas constraints. This ensures that
+complex multi-contract upgrades can execute completely within the [fork activation block](#fork-activation-block)
+without running out of gas.
+
+Standard L2 blocks are constrained by `systemTxMaxGas` (typically 1,000,000 gas), which is insufficient for executing
+the deployment and upgrade transactions in a typical predeploy upgrade. The custom gas limit bypasses this constraint
+for upgrade blocks specifically.
```
sebastianst, line 473: Note that in practice, also all remaining gas in the block after the user deposit gas of 20M (?) is available, which is why past upgrades that consumed more than 1M were possible. But it relied on the implicit assumption that chains had enough gas available via their block gas limit.

specs/protocol/l2-upgrades-1-execution.md
maurelian: This field is not helpful, removed it.

specs/protocol/l2-upgrades-1-execution.md
maurelian: We are honestly being very conservative here, but given the broad range of contract versions that are running on the various chains and the severity of a failure in the upgrade path we feel that it is necessary to verify the upgrade path on each chain, at least those in the superchain registry. This will need to be at least semi-automated of course, but won't be necessary on future upgrades since the chains will finally be running the same versions of contracts.

specs/protocol/l2-upgrades-1-execution.md
maurelian: > simply always using the current bundle and not having any bundle to copy over to.

Okay, after further thought there is an issue with this idea. Basically this should work for the first L2CM upgrade, but when the next upgrade comes around, and we cut a release, we will be overwriting the previous JSON bundle with the newer one unless we are explicitly defining a bundle to track for each fork.

maurelian: Fixed in 5dc4702: Updated the spec to reflect the continuous release readiness approach. Removed the bundle canonicalization/copying step in favor of always using `current-l2-upgrade-bundle.json` as the authoritative source. New functionality should be behind feature flags that are activated after a fork.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 5dc4702: Clarified that the bundle JSON should be embedded into L2 client binaries at build time using Go's `//go:embed` or Rust's `include_str!`. The bundle is lazy-parsed when the derivation pipeline encounters a fork activation block.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 5dc4702: Added a note in the Custom Upgrade Block Gas Limit section explaining that past upgrades relied on remaining block gas availability, and this spec makes the gas allocation explicit to avoid such implicit dependencies.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 5dc4702: Softened the testing claims in aUP-002 mitigations to reflect current state. Changed to "should" language and specified that fork tests should use actual mainnet chain state (e.g., OP Mainnet) as starting point.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,579 @@
...
+4. **L2ContractsManager Deployment**: Deploy the L2ContractsManager for this upgrade
+5. **Batch Upgrade Execution**: Call `L2ProxyAdmin.upgradePredeploys(l2ContractsManagerAddress)` which:
+   - Executes DELEGATECALL to L2ContractsManager.upgrade()
+   - L2ContractsManager gathers configuration from existing predeploys
+   - For each predeploy, calls `proxy.upgradeTo()` or `proxy.upgradeToAndCall()`
+   - Verifies all upgrades completed successfully
+
+All of these transactions execute before any user-submitted transactions in the block.
+
+**Note:** Transactions which are labeled as "one-time only" are necessary to enable infrastructure which will support this mechanism of upgrades in general. Once these steps are complete they do not need to be included in future upgrades.
```
maurelian: Since these transactions are one time only, we should look for a way to keep them out of the bundle itself and instead hardcode them in the client implementation. 

Otherwise, we will need to update the bundle generation after the first upgrade.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,579 @@
...
+### Assumptions
+
+#### aUP-001: Fork Activation Time is Coordinated
+
+All OP Stack chains coordinate fork activation times to enable consistent upgrade rollout. The fork activation
+timestamp is communicated well in advance of activation to allow node operators to prepare.
+
+##### Mitigations
+
+- Fork activation timestamps are define in the superchain-registry
```
0xiamflux: Small typo, it should be 'defined'.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,579 @@
...
+
+#### Fork Activation Block
+
+The L2 block at which a protocol upgrade becomes active, identified by a specific L2 block timestamp. This block
+contains the Network Upgrade Transactions that implement the protocol changes.
+
+#### Bundle Generation Script
+
+A Solidity script (typically using Forge scripting) that deterministically computes all transaction data for an
+upgrade. The script deploys contracts, computes addresses, and assembles transaction calldata.
```
0xiamflux: The generation script shouldn't deploy any contract.

specs/protocol/l2-upgrades-2-contracts.md
```diff
@@ -0,0 +1,479 @@
...
+addresses where a contract already exists (collision detected). False negatives (failing to detect existing contracts)
+and false positives (detecting non-existent contracts) are both prohibited.
+
+##### Impact
+
+**Severity: High**
+
+False negatives would cause failed deployments while false positives would prevent legitimate deployments, both
+breaking the upgrade process.
+
```
0xiamflux, line 208: Is it worth to add a 'contract at address availability' invariant? i.e. after the execution of the Conditional Deployer, the address returned by it contains the runtime code derived from the init code passed to it.

specs/protocol/l2-upgrades-2-contracts.md
```diff
@@ -0,0 +1,479 @@
...
+- All L2 predeploys use standardized proxy implementations with well-tested upgrade functions
+- Fork-based testing exercises upgrade operations against actual deployed proxies
+- Non-upgradeable predeploys (if they exist) will be excluded from the upgrade process
+
+#### aL2CM-004: Feature Flags Are Correctly Configured
+
+When [feature flags](#feature-flag) are used to customize upgrade behavior, the FeatureFlags contract is properly
+configured in the environment and returns consistent values throughout the upgrade execution.
+Production features which are enable must be exposed by the L1Block contract's interface.
```
0xiamflux: typo, should be 'enabled'

specs/protocol/l2-upgrades-1-execution.md
0xiamflux: Agreed

specs/protocol/l2-upgrades-2-contracts.md
```diff
@@ -0,0 +1,479 @@
...
+#### iCD-003: Non-Reverting Collision Handling
+
+When a [CREATE2 Collision](#create2-collision) is detected (contract already deployed at the target address), the
+ConditionalDeployer MUST return successfully without reverting and without modifying blockchain state.
+
+##### Impact
+
+**Severity: Medium**
+
+If collisions cause reverts, the presence of reverting transactions in an upgrade block would cause confusion.
```
0xOneTony, line 194: Is that the only impact? even though this seems unlikely to happen, i guess worst case is that the contract is already deployed with the same bytecode and nothing changes, which is not really an issue i guess, but curious if there should be some follow up action or how we should handle this case.

specs/protocol/l2-upgrades-2-contracts.md
```diff
@@ -0,0 +1,479 @@
...
+corrupted, resulting in loss of admin control over predeploys.
+
+#### iL2PA-003: Backwards Compatibility Maintained
+
+The upgraded L2ProxyAdmin implementation MUST maintain the existing interface for standard proxy administration
+functions. Existing functionality for upgrading individual proxies, changing proxy admins, and querying proxy state
+MUST continue to work as before.
+
+Note: Backwards compatibility requires maintaining the full ProxyAdmin interface, but does not require supporting
+upgrades of legacy proxy types (ResolvedDelegate and Chugsplash proxies).
```
0xOneTony: 🔥 maybe also worth noting that currently no predeploy is using any of the 2 legacy proxy types

specs/protocol/l2-upgrades-2-contracts.md
```diff
@@ -0,0 +1,479 @@
...
+
+## Overview
+
+This specification defines the mechanism for upgrading L2 predeploy contracts through deterministic, hard fork-driven
+Network Upgrade Transactions (NUTs). The system enables safe, well-tested upgrades of L2 contracts with both
+implementation and upgrade paths written in Solidity, ensuring determinism, verifiability, and testability across all
+client implementations.
+
+The upgrade system maintains the existing pattern of injecting Network Upgrade Transactions at specific fork block
+heights while improving the development and testing process. Upgrade transactions are defined in JSON bundles that are
```
0xOneTony: maybe worth linking the bundle format from the execution spec or in the future the path of the bundles in git

specs/protocol/l2-upgrades-2-contracts.md
```diff
@@ -0,0 +1,479 @@
...
+upgrades of legacy proxy types (ResolvedDelegate and Chugsplash proxies).
+
+##### Impact
+
+**Severity: High**
+
+If backwards compatibility is broken, existing tooling and scripts that interact with the ProxyAdmin could fail,
+preventing emergency responses and breaking operational procedures.
+
+## L2ContractsManager
```
0xOneTony, line 331: even though i think this isnt an issue now, do you think we should have an assumption about the order of the upgrade calls execution?

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 64ee642: Corrected typo to "defined".

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 64ee642: Updated the definition to clarify that the script "computes CREATE2 addresses, generates deployment initcode, and assembles transaction calldata" rather than deploying contracts.

specs/protocol/l2-upgrades-2-contracts.md
maurelian: Fixed in 64ee642: Corrected typo to "enabled".

specs/protocol/l2-upgrades-2-contracts.md
maurelian: Fixed in 64ee642: Added a note that currently no predeploy uses these legacy proxy types.

specs/protocol/l2-upgrades-2-contracts.md
maurelian: Fixed in 64ee642: Added a cross-reference link to the Bundle Format section in the execution spec.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,580 @@
...
+##### Impact
+
+**Severity: Critical**
+
+A failed upgrade can lead to a chain halt, for example if a new function is not added to the L1Block
+contract, causing the L1 Attributes deposit transactions to fail.
+
+#### iUP-002: Deterministic Execution
+
+The upgrade must be deterministic, regardless of chain-specific configuration, or chain state at the time of
```
smartcontracts: Are there any potential cases where contracts want to store `block.number` or something similar at upgrade time? This would violate the invariant. If not, that's fine, we probably just need a semgrep rule or something

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,580 @@
...
+### Transaction Execution Sequence
+
+Within the fork activation block, transactions will execute in this order:
+
+0. **L1 Info Deposit transaction**: This is the transaction which occurs at the start of each block, it is unchanged by
+  this work.
+1. **ConditionalDeployer Deployment** (one-time only): Deploy the ConditionalDeployer contract
+2. **Implementation Deployments**: For each predeploy being upgraded, deploy new implementation via
+   ConditionalDeployer
+3. **ProxyAdmin Upgrade** (one-time only): Upgrade the L2ProxyAdmin implementation
```
smartcontracts, line 241: Maybe I'll find out later but leaving a note here, what is the purpose of this step?

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,580 @@
...
+4. **L2ContractsManager Deployment**: Deploy the L2ContractsManager for this upgrade
+5. **Batch Upgrade Execution**: Call `L2ProxyAdmin.upgradePredeploys(l2ContractsManagerAddress)` which:
```
smartcontracts: This is a single entrypoint, correct?

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,580 @@
...
+
+All of these transactions execute before any user-submitted transactions in the block.
+
+**Note:** Transactions which are labeled as "one-time only" are necessary to enable infrastructure which will support this mechanism of upgrades in general. Once these steps are complete they do not need to be included in future upgrades.
+
+## Network Upgrade Transaction Bundle
+
+### Overview
+
+The Network Upgrade Transaction (NUT) Bundle is a JSON-formatted data structure containing the complete set of
```
smartcontracts, line 260: lol

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,580 @@
...
+#### aNUTB-004: JSON Format is Correctly Parsed
+
+All L2 client implementations (Go, Rust, etc.) correctly parse the JSON bundle format and extract transaction fields
+identically. Parsing inconsistencies would cause consensus failures.
+
+##### Mitigations
+
+- JSON schema is simple and uses standard field types
+- Test vectors validate parsing across client implementations
+- Acceptance tests should validate bundle execution across implementations
```
smartcontracts, line 345: And I guess, forge is not compromised

specs/protocol/l2-upgrades-1-execution.md
maurelian: Actually, I think that storing `block.number` would be fine, and this invariant needs to be clarified. The key is that the upgrade transaction payloads be deterministic, but it's OK if there are differences in memory or storage state that occur after execution. 

For example the L2StandardBridge will have a different L1StandardBridge address written to storage on each chain, but that is OK because it is read from the env, not inserted into the upgrade calldata.

specs/protocol/l2-upgrades-2-contracts.md
maurelian: Do you mean saying that the L2CM is not called until various other contracts have been deployed etc so that they are available in the state? If so I think that is worth including yes.

specs/protocol/l2-upgrades-2-contracts.md
maurelian: IIRC, in a previous discussion with @sebastianst and @geoknee they stated that if a particular upgrade transaction, it would not have any impact on other upgrade transactions. This makes sense as each upgrade tx is just inserted into the upgrade block. In the same way that different deposit transactions, and different sequencer  transactions don't affect each other when one fails, nor do these upgrade transactions.

So, as long as the expected code is at the expected address (per `iCD-002: Idempotent Deployment Operations` above), then the impact here is no more than Medium, or even arguably Low.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,580 @@
...
+
+#### Derivation Pipeline
+
+The component of L2 client implementations responsible for constructing L2 blocks from L1 data and protocol rules. The
+derivation pipeline determines block attributes including gas limits and inserts upgrade transactions at fork
+activations.
+
+### Assumptions
+
+#### aUBGL-001: Upgrade Gas Requirements Are Bounded
```
smartcontracts, line 502: Note that it will be challenging if not impossible to make the gas usage deterministic because other users can manipulate the state a bit

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,580 @@
...
+##### Impact
+
+**Severity: Critical**
+
+If insufficient gas is allocated, upgrade transactions will fail mid-execution, leaving predeploys in partially
+upgraded states and halting the chain at fork activation.
+
+#### iUBGL-002: Deterministic Gas Allocation
+
+The gas allocation for upgrade blocks MUST be deterministic and identical across all L2 nodes executing the fork
```
smartcontracts, line 543: The gas allocation can be deterministic but will again note that the gas USAGE may not be, which I believe is fine but I want to point it out

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -0,0 +1,580 @@
...
+If custom gas allocation applies to non-upgrade blocks, it could enable DOS attacks by allowing transactions to consume
+excessive gas or bypass fee markets.
+
+### Gas Allocation Specification
+
+The custom upgrade block gas allocation is implemented in the derivation pipeline with the following behavior:
+
+**Gas Allocation Value:**
+
+- The upgrade block gas allocation SHOULD be set to 50,000,000 gas (50M gas)
```
smartcontracts: Any reason why we set this to 50m rather than having a dynamic value that can be configured in the JSON?

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 0dc6bdc: Clarified that one-time only transactions should be kept out of the bundle and hardcoded in the client implementation.

specs/protocol/l2-upgrades-2-contracts.md
```diff
@@ -0,0 +1,481 @@
...
+or deploy the implementation to an unexpected address, breaking the upgrade.
+
+#### iCD-003: Non-Reverting Collision Handling
+
+When a [CREATE2 Collision](#create2-collision) is detected (contract already deployed at the target address), the
+ConditionalDeployer MUST return successfully without reverting and without modifying blockchain state.
+
+##### Impact
+
+**Severity: Medium**
```
smartcontracts, line 192: Seems higher than this, no? If it reverts, the upgrade breaks.

specs/protocol/l2-upgrades-2-contracts.md
```diff
@@ -0,0 +1,481 @@
...
+The [Depositor Account](#depositor-account) is exclusively controlled by the L2 protocol's derivation and execution
+pipeline. No external parties can submit transactions from this address.
+
+##### Mitigations
+
+- The [Depositor Account](#depositor-account) address has no known private key
+- Transactions from this address can only originate from the protocol's derivation pipeline processing L1 deposit events
+- The address is hardcoded in the protocol specification and client implementations
+
+#### aL2PA-002: L2ContractsManager does not Modify Storage
```
smartcontracts: Seems like this is more about the code not being malicious rather than specifically about modifying storage

specs/protocol/l2-upgrades-2-contracts.md
```diff
@@ -0,0 +1,481 @@
...
+operation fails, the entire DELEGATECALL MUST revert, leaving all predeploys in their pre-upgrade state.
+
+##### Impact
+
+**Severity: Critical**
+
+If upgrades are not atomic, a partial failure could leave some predeploys upgraded and others not, creating an
+inconsistent system state that breaks inter-contract dependencies.
+
+#### iL2CM-004: Correct Upgrade Method Selection
```
smartcontracts: Why are we not following the same clear-and-reinitialize pattern that we have in OPCM? Having a specially constructed upgrade transaction per upgrade was a large source of our problems with the OPCMv1

specs/protocol/l2-upgrades-1-execution.md
maurelian: If I understand the question, all upgrades happen atomically downstream of the call to `L2ProxyAdmin.upgradePredeploys()`. 

The term "Batch" might not be helpful here. 

```suggestion
5. **Upgrade Execution**: Call `L2ProxyAdmin.upgradePredeploys(l2ContractsManagerAddress)` which will atomically:
```

specs/protocol/l2-upgrades-2-contracts.md
maurelian: We are going to do that, so you are right that this is incorrect.

specs/protocol/l2-upgrades-1-execution.md
maurelian: This is outdated, that is the plan now.

specs/protocol/l2-upgrades-2-contracts.md
maurelian: no, if it reverts the other transactions are unaffected.
https://github.com/ethereum-optimism/specs/pull/884#discussion_r2760731995

specs/protocol/l2-upgrades-2-contracts.md
maurelian: Good point, that is a more generic assumption that we need to make.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Hmmm, for a given chain, the gas usage should also be deterministic. For different chains, that's true.

specs/protocol/l2-upgrades-1-execution.md
maurelian: I think that's fine, we can easily bake a good buffer in for each tx.

specs/protocol/l2-upgrades-1-execution.md
maurelian: It makes step 5 possible.
Recall that the difference between the L1 OPCM flow is that on L2 we are unable to do a delegate call from within the L2 PAO, so instead we need to do the delegate call from within the L2ProxyAdmin, and we need new logic to make that possible.

specs/protocol/l2-upgrades-2-contracts.md
maurelian: Fixed in 50d9382: Added iCD-005 invariant to ensure contract availability after ConditionalDeployer deployment.

specs/protocol/l2-upgrades-2-contracts.md
maurelian: Fixed in 50d9382: Added a note in the L2ContractsManager overview clarifying that all prerequisite contracts must be deployed before the L2ContractsManager is called, and that the transaction execution sequence ensures this ordering.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 50d9382: Clarified iUP-002 to specify that transaction payloads must be deterministically generated and identical across all chains, while execution may result in different storage/memory state (e.g., storing block.number or reading chain-specific config).

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 50d9382: Added aNUTB-001b assumption that the build toolchain (Solidity compiler, Forge, and other build tools) is not compromised.

specs/protocol/l2-upgrades-2-contracts.md
maurelian: Fixed in 50d9382: Updated aL2PA-002 to be more broadly about the L2ContractsManager code not being malicious (covering storage corruption, unauthorized operations, and vulnerabilities), rather than just storage modification.

specs/protocol/l2-upgrades-2-contracts.md
maurelian: Fixed in 50d9382: Updated iL2CM-004 to reflect the clear-and-reinitialize pattern: upgradeTo(StorageSetter), reset initialized value, then upgradeToAndCall(initialize()).

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in 50d9382: Updated the gas allocation specification to reflect that the value is specified in the bundle JSON (rather than hardcoded at 50M), allowing it to be dynamically configured per upgrade.

specs/protocol/l2-upgrades-1-execution.md
maurelian: Fixed in https://github.com/ethereum-optimism/specs/commit/50d93821f60ed0afc06f95182bd381b07a5e755b: Updated the gas allocation specification to reflect that the value is specified in the bundle JSON (rather than hardcoded at 50M), allowing it to be dynamically configured per upgrade.

smartcontracts (approved):

PR #888, opened 2026-02-04
Update preimage hints

pauldowman:
**Description**

- Adds documentation for the `l1-blob` hint.
- Updates the format of the `l1-precompile` hint, and also renames it to `l1-precompile-v2` which is used in op-program, but not in Kona (which has the same format but named `l1-precompile`). 

Question for reviewers: should the spec keep the name as `l1-precompile` instead of `l1-precompile-v2`, since that's what Kona uses, which will become the default fault proof program?

**Tests**

**Additional context**

**Metadata**

### DIFF
--- a/specs/fault-proof/index.md
+++ b/specs/fault-proof/index.md
@@ -28,7 +28,8 @@
     - [`l1-block-header <blockhash>`](#l1-block-header-blockhash)
     - [`l1-transactions <blockhash>`](#l1-transactions-blockhash)
     - [`l1-receipts <blockhash>`](#l1-receipts-blockhash)
-    - [`l1-precompile <precompile ++ inputbytes>`](#l1-precompile-precompile--inputbytes)
+    - [`l1-blob <blobhash ++ metadata>`](#l1-blob-blobhash--metadata)
+    - [`l1-precompile-v2 <address ++ requiredGas ++ inputbytes>`](#l1-precompile-v2-address--requiredgas--inputbytes)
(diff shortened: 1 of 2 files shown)


geoknee:
Made an update here https://github.com/ethereum-optimism/specs/pull/890

geoknee:
> Question for reviewers: should the spec keep the name as `l1-precompile` instead of `l1-precompile-v2`, since that's what Kona uses, which will become the default fault proof program?

We could keep the specs in sync with op-program until we officially switch to kona as the default?

geoknee (approved):

ajsutton:
I'm not really sure why these are in the specs at all. I don't think they're consistent between op-program and kona at all (you can't use op-program host with kona prestates or visa versa). At one point I thought it would be useful to have things be host agnostic and standard but it hasn't been a problem to keep them matched up in practice and we're moving to only having kona in the future anyway.

pauldowman:
> I'm not really sure why these are in the specs at all. I don't think they're consistent between op-program and kona at all (you can't use op-program host with kona prestates or visa versa). At one point I thought it would be useful to have things be host agnostic and standard but it hasn't been a problem to keep them matched up in practice and we're moving to only having kona in the future anyway.

Agreed, since host and client are coupled, but I updated it since it was already in here. The other option is remove the whole section.

ajsutton:
Let's stick with what op-program is using for now but I do think we should just remove this entirely in the future instead of updating to align with what kona uses.

PR #896, opened 2026-03-11
feat: zk dispute game 

ashitakah:
The spec for the smart contract side of the ZK dispute game integration into the OP Stack.                                                                                                                                                  
                                                                                                                                                                                          
Four spec files are added covering the `ZKDisputeGame` contract: the overall design and OPCM integration, the full game lifecycle and bond accounting, the `IZKVerifier` interface, and the ZK program definition (inputs, outputs, `absolutePrestate`).

The game resolves disputes in a single round - a proposer posts an output root, anyone can challenge it, and a prover submits a ZK proof to defend the claim or lets the window expire. It follows the same patterns as `FaultDisputeGame` (MCP clones, `DelayedWETH`, permissionless access, same safety checks) while adding verifier agnosticism through `IZKVerifier`, parent chaining, and SP1 PLONK as the initial proving backend.


specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,175 @@
...
+| -------------------------------------------- | ------ | --------------------------- | --------------------------- | ---------------- |
+| Unchallenged, deadline expires               | NORMAL | `initBond`                  | —                           | —                |
+| Unchallenged, proof provided                 | NORMAL | `initBond`                  | —                           | nothing          |
+| Challenged, no proof                         | NORMAL | nothing                     | `initBond + challengerBond` | —                |
+| Challenged, proof, prover == proposer        | NORMAL | `initBond + challengerBond` | nothing                     | _(same)_         |
+| Challenged, proof, prover != proposer        | NORMAL | `initBond`                  | nothing                     | `challengerBond` |
+| Parent `CHALLENGER_WINS`, child challenged   | NORMAL | nothing                     | `initBond + challengerBond` | —                |
+| Parent `CHALLENGER_WINS`, child unchallenged | NORMAL | nothing (burned)            | —                           | —                |
```
ajsutton: What do we actually do with burned funds - do we send them to address(0) or are they just left in DelayedWETH and not paid out?

Either is fine functionally, but if the funds are left in DelayedWETH that will affect monitoring since we normally check the DelayedWETH contract has a balance equal to the sum of outstanding bonds. Accounting for old games that left funds behind would be expensive. We could just monitor for DelayedWETH being under-collateralised though.

The amount of funds being burnt would be very low in practice.

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,175 @@
...
+required `initBond`. The factory deploys an MCP clone, which then validates the proposal.
+
+A game may start from the anchor state by setting `parentIndex = type(uint32).max`, or it may
+reference a parent game.
+
+The `_extraData` passed to `DisputeGameFactory.create()` has this layout:
+
+| Field               | Type      | Description                                        |
+| ------------------- | --------- | -------------------------------------------------- |
+| `l2SequenceNumber`  | `uint256` | L2 block number asserted by this game's root claim |
```
ajsutton: Can we make this a uint64 please? Block numbers will never exceed uint64 and being able to specify a uint256 makes it much harder to deal with in off chain components.  The `l2SequenceNumber()` method will still return a uint256 but we will have a natural guarantee that it always fits in a uint64.

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,175 @@
...
+### Parent Validation
+
+When a parent is referenced (`parentIndex != type(uint32).max`), `initialize()` MUST revert if
+any of the following checks fail:
+
+- Parent MUST NOT be blacklisted.
+- Parent MUST NOT be retired (i.e., `createdAt > retirementTimestamp`).
+- Parent MUST be the same game type (`ZK_GAME_TYPE`).
+- Parent MUST NOT have resolved as `CHALLENGER_WINS`.
+- Parent's `l2SequenceNumber` MUST be at or above the anchor state's `l2SequenceNumber`.
```
ajsutton: `l2SequenceNumber` must be _above_ the anchor state's `l2SequenceNumber`. If it is equal it should just use the anchor state instead. I suspect we had this detail wrong in the design doc actually (sorry).

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,175 @@
...
+proposal chains from being completed after a game type transition.
+
+## Challenge
+
+Challenging is fully permissionless. Anyone may call `challenge()` before the challenge deadline.
+The call MUST include `challengerBond` ETH, which `challenge()` deposits into `DelayedWETH` on
+the caller's behalf.
+
+- `challenge()` MUST revert if `gameOver()` returns `true`.
+- Only one challenge is allowed per game.
```
ajsutton, line 98: nit: Should we phrase this as MUST revert if the game is not in the `Unchallenged` state?

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,175 @@
...
+current deadline, regardless of whether the game has been challenged.
+
+- `prove()` MUST revert if `gameOver()` returns `true` (covers both an already-submitted proof
+  and an expired deadline).
+- The verifier call MUST revert for invalid proofs.
+- On success, `status` transitions to `UnchallengedAndValidProofProvided` or
+  `ChallengedAndValidProofProvided` (depending on whether the game was challenged), and
+  `gameOver()` returns `true` immediately.
+- If the game was challenged and the prover is different from the proposer, the prover earns
+  `challengerBond` upon resolution.
```
ajsutton: Wouldn't the prover always get `challengerBond`? If they are also the proposer they'd also get the proposer bond refunded but seems cleaner to just spec them as two separate payments.

If we do spec it this way we need to state what happens if they are the proposer.

Likely we should just omit this and cover payment of bonds in the resolution section.

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Resolution
+
+Resolution is permissionless. Anyone may call `resolve()` once `gameOver()` returns `true` and
+the parent game is resolved.
+
+- `resolve()` MUST revert if `resolvedAt != 0` (i.e., game is already resolved).
+- The parent game MUST be resolved before the child.
```
ajsutton: nit: Maybe phrase this as MUST revert if the parent is not resolved.

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,175 @@
...
+| Challenged, no proof                         | NORMAL | nothing                     | `initBond + challengerBond` | —                |
+| Challenged, proof, prover == proposer        | NORMAL | `initBond + challengerBond` | nothing                     | _(same)_         |
+| Challenged, proof, prover != proposer        | NORMAL | `initBond`                  | nothing                     | `challengerBond` |
+| Parent `CHALLENGER_WINS`, child challenged   | NORMAL | nothing                     | `initBond + challengerBond` | —                |
+| Parent `CHALLENGER_WINS`, child unchallenged | NORMAL | nothing (burned)            | —                           | —                |
+| Game blacklisted                             | REFUND | `initBond`                  | `challengerBond`            | —                |
+| Game retired                                 | REFUND | `initBond`                  | `challengerBond`            | —                |
+| Game type changed mid-play                   | REFUND | `initBond`                  | `challengerBond`            | —                |
```
ajsutton: What does it mean for the game type to change mid-play? A game's type can't change once it's created. Maybe you mean the respected game type changed, but that shouldn't affect resolution and bond payments (we'd retire all games at the same time as changing the respected game type if we wanted to refund bonds).

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,175 @@
...
+## Closing
+
+After resolution, bonds are distributed through a two-phase process identical to the
+`FaultDisputeGame`:
+
+**`closeGame()`** (permissionless, also called internally by `claimCredit`):
+
+- MUST revert if `AnchorStateRegistry` is paused.
+- MUST revert with `GameNotResolved` if `resolvedAt == 0`.
+- MUST wait for the finality delay: `block.timestamp - resolvedAt > DISPUTE_GAME_FINALITY_DELAY_SECONDS`.
```
ajsutton: Should we spec this as must wait for the AnchorStateRegistry.isFinalized to return true for this game? It's the AnchorStateRegistry that tracks the air gap and we want to delegate everything to that.

Net result is the same, just a conceptual abstraction.

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,175 @@
...
+- If the game has a Valid Claim, registers the game as the new anchor state via `AnchorStateRegistry`.
+- Determines NORMAL or REFUND mode and unlocks bonds in `DelayedWETH` accordingly.
+
+**`claimCredit(recipient)`**:
+
+1. Triggers `closeGame()` if not yet closed.
+2. Calls `DelayedWETH.unlock(recipient)` to queue the withdrawal. If `recipient` has no credit
+   allocated by this game, the unlock has no effect.
+3. After the `DelayedWETH` delay, `withdraw()` transfers ETH to the recipient.
```
ajsutton: Probably worth noting that it:
*  _does not_ revert if  the recipient has not yet unlocked their credit (even if they have no credit). This lets op-challenger use claimCredit to close the game.
* _does_ revert if the recipient has already unlocked their credit and has none to claim
* _does_ revert if the recipient's credit is not yet available to withdraw (DelayedWETH.withdraw reverts)
* _does_ revert if the bond transfer fails

That should match what FaultDisputeGame does. It's a bit fiddly but is what makes op-challenger automatically handle closing games and unlocking credit as part of its bond claimer without actually sending a heap of transactions that do nothing while the bonds are still in the air gap.

specs/fault-proof/stage-one/zk/zk-dispute-game.md
```diff
@@ -0,0 +1,387 @@
...
+Under normal operation, the anchor state advances on a timescale (12+ hours minimum) that is much
+larger than typical proposal frequency (e.g., 1 hour), making orphan risk from parent validation
+negligible.
+
+#### Mitigations
+
+- A rational proposer would never use a parent whose `l2SequenceNumber` is below the anchor, as
+  it unnecessarily increases the proving range.
+- [Parent validation](game-mechanics.md#parent-validation) requires the parent's
+  `l2SequenceNumber` to be at or above the anchor state, preventing chains from building on stale
```
ajsutton: We should update this to be strictly above the anchor state as well.

specs/fault-proof/stage-one/zk/zk-dispute-game.md
```diff
@@ -0,0 +1,387 @@
...
+## Invariants
+
+### iZKG-001: A Valid Proof Always Wins
+
+If a valid ZK proof is submitted before the current deadline, the game MUST resolve as
+`DEFENDER_WINS` (assuming a valid parent chain).
+
+#### Impact
+
+**Severity: Critical**
```
ajsutton: We would not consider this a critical issue in immunefi terms because it is still protected by the air gap. I don't think this actually maps to immunefi, but I suspect it shouldn't be critical here either because of that same fallback. probably just high.
Same for all the other cases where the air gap isn't violated.

specs/fault-proof/stage-one/zk/zk-dispute-game.md
```diff
@@ -0,0 +1,387 @@
...
+the bridge.
+
+### iZKG-003: Bond Safety via DelayedWETH
+
+All bonds MUST be deposited into and withdrawn from `DelayedWETH`. The game contract MUST NOT
+hold raw ETH bonds.
+
+#### Impact
+
+**Severity: High**
```
ajsutton: This probably is critical as there is no way to recover bonds kept on the game contract itself and there's no way to upgrade the contract. The bonds would just be lost forever.

ajsutton (commented):
Looking good generally.

ajsutton:
Oh a few extra invariants we could document:

Resolution requires resolved parent (ordering invariant). The spec states this in prose ("The parent game MUST be resolved before the child") but doesn't list it as an invariant. It's load-bearing
   for iZKG-005 — parent invalidity propagation only works if resolution order is enforced. If a child could resolve before its parent, it could finalize as DEFENDER_WINS before discovering the
  parent is invalid.

```
  iZKG-009: Child Resolution Requires Resolved Parent
  resolve() MUST revert if the parent game has not yet resolved. The resolution dependency chain MUST be honored in topological order.
  Impact — Critical: Without this, iZKG-005 cannot hold. A child could finalize a withdrawal before its parent is invalidated.
```

  Single challenge per game. Already noted but worth repeating as the most obviously missing invariant. A second challenge could reset the prove deadline, disrupt bond accounting, or overwrite the
  challenger address.

```
  iZKG-010: At Most One Challenge Per Game
  challenge() MUST revert if the game has already been challenged.
  Impact — High: A second challenge could reset the prove deadline (giving challengers unbounded time), overwrite the original challenger's address (stealing their bond credit), or double the bond
  liability.
```

Bond conservation / no fund leakage. Not sure of the right wording for this one given that some funds do get burnt. The spec describes who gets what in each scenario but never states the meta-property: total payouts equal total deposits. No bonds can be permanently locked
  (except intentional burning) or created from nothing.

```
  iZKG-011: Bond Conservation
  For any resolved game, the sum of all distributed bonds MUST equal initBond + challengerBond (or initBond if unchallenged). No value may be created or permanently locked beyond the defined "burn"
  path.
  Impact — High: Violation means either fund loss for participants or an exploitable source of unbacked withdrawals from DelayedWETH.
```

  Monotonic state progression. The game status can only advance forward through the state machine, never regress. This is implied by the transition table but is a safety property in its own right —
  if state could regress (e.g., Challenged → Unchallenged), bond accounting and deadline logic would break.


Plus the assumption:
  Proof generation liveness within the prove window. The spec assumes bonds are economically rational (aZKG-004) and the verifier is sound (aZKG-001), but never assumes that a valid proof can be
  generated within maxProveDuration. If proving infrastructure is unavailable — network outage, prover service down, proving takes longer than the window — a valid proposer loses their bond to a
  challenger despite having a correct claim. This is a distinct concern from bond sizing. Something like:

```
  aZKG-007: Proof Generation Is Feasible Within the Prove Window
  A prover with access to the required L1/L2 data can generate a valid proof within maxProveDuration under normal conditions.
  Mitigations: maxProveDuration must be set with headroom above worst-case proving times. Multiple independent proving services reduce single-point-of-failure risk.
```

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: [a62cd1b](https://github.com/ethereum-optimism/specs/pull/896/commits/a62cd1b7587a873d8932552889a733d7e40734de)

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: [a62cd1b](https://github.com/ethereum-optimism/specs/pull/896/commits/a62cd1b7587a873d8932552889a733d7e40734de)

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: [a62cd1b](https://github.com/ethereum-optimism/specs/pull/896/commits/a62cd1b7587a873d8932552889a733d7e40734de)

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: [a62cd1b](https://github.com/ethereum-optimism/specs/pull/896/commits/a62cd1b7587a873d8932552889a733d7e40734de)

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: [a62cd1b](https://github.com/ethereum-optimism/specs/pull/896/commits/a62cd1b7587a873d8932552889a733d7e40734de)

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: [a62cd1b](https://github.com/ethereum-optimism/specs/pull/896/commits/a62cd1b7587a873d8932552889a733d7e40734de)

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: [a62cd1b](https://github.com/ethereum-optimism/specs/pull/896/commits/a62cd1b7587a873d8932552889a733d7e40734de)

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: [a62cd1b](https://github.com/ethereum-optimism/specs/pull/896/commits/a62cd1b7587a873d8932552889a733d7e40734de)

specs/fault-proof/stage-one/zk/zk-dispute-game.md
ashitakah: [a62cd1b](https://github.com/ethereum-optimism/specs/pull/896/commits/a62cd1b7587a873d8932552889a733d7e40734de)

specs/fault-proof/stage-one/zk/zk-dispute-game.md
ashitakah: [a62cd1b](https://github.com/ethereum-optimism/specs/pull/896/commits/a62cd1b7587a873d8932552889a733d7e40734de)

specs/fault-proof/stage-one/zk/zk-dispute-game.md
ashitakah: [a62cd1b](https://github.com/ethereum-optimism/specs/pull/896/commits/a62cd1b7587a873d8932552889a733d7e40734de)

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: [fix: revert unchallenged](https://github.com/ethereum-optimism/specs/pull/896/commits/e1bee78678bede0ea91cc8eb7f77741506eb0051)

ajsutton (approved):
LGTM.

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,176 @@
...
+which games can finalize withdrawals (via `isGameClaimValid`), but MUST NOT prevent in-progress
+proposal chains from being completed after a game type transition.
+
+## Challenge
+
+Challenging is fully permissionless. Anyone may call `challenge()` before the challenge deadline.
+The call MUST include `challengerBond` ETH, which `challenge()` deposits into `DelayedWETH` on
+the caller's behalf.
+
+- `challenge()` MUST revert if the game is not in the `Unchallenged` state.
```
Joxess, line 97: We need to add that it should revert also if `gameOver()` returns `true` as well

specs/fault-proof/stage-one/zk/zk-interface.md
```diff
@@ -0,0 +1,94 @@
...
+        _proofBytes
+    );
+
+    // record proof submission and set game over
+}
+```
+
+All public values come from immutable on-chain game state, so no caller-supplied data affects what
+the verifier checks. `l1Head` is set to `blockhash(block.number - 1)` in `initialize()`, which
+runs in the same transaction as the clone deployment in `DisputeGameFactory.create()`.
```
Joxess: We could adjust this explanation to say the `l1Head` is set by `DisputeGameFactory.create()` directly

Joxess:
```suggestion
All public values come from immutable on-chain game state, so no caller-supplied data affects what
the verifier checks. `l1Head` is captured as `blockhash(block.number - 1)` by `DisputeGameFactory.create()` and packed into the clone's CWIA data.
```

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,176 @@
...
+A game may start from the anchor state by setting `parentIndex = type(uint32).max`, or it may
+reference a parent game.
+
+The `_extraData` passed to `DisputeGameFactory.create()` has this layout:
+
+| Field              | Type     | Description                                                                    |
+| ------------------ | -------- | ------------------------------------------------------------------------------ |
+| `l2SequenceNumber` | `uint64` | L2 block number asserted by this game's root claim                             |
+| `parentIndex`      | `uint32` | Index of the parent game; `type(uint32).max` if starting from the anchor state |
+
```
Joxess, line 67:
```suggestion
During `initialize()`, the game snapshots whether its game type is the currently respected game type (`wasRespectedGameTypeWhenCreated`). `AnchorStateRegistry.isGameClaimValid()` reads this flag when determining if a game's root claim can advance the anchor state or finalize withdrawals via the Portal. The snapshot ensures that a game type transition does not retroactively invalidate games that were created under the previous respected type.
```

Joxess: I think this is worth adding @ashitakah @ajsutton

specs/fault-proof/stage-one/zk/zk-dispute-game.md
```diff
@@ -0,0 +1,463 @@
...
+    - [Mitigations](#mitigations-1)
+  - [aZKG-003: Parent Chaining Preserves Correctness](#azkg-003-parent-chaining-preserves-correctness)
+    - [Mitigations](#mitigations-2)
+  - [aZKG-004: Bonds Are Economically Rational](#azkg-004-bonds-are-economically-rational)
+    - [Mitigations](#mitigations-3)
+  - [aZKG-005: Guardian Acts Honestly and Timely](#azkg-005-guardian-acts-honestly-and-timely)
+    - [Mitigations](#mitigations-4)
+  - [aZKG-006: Anchor State Advances Slowly Relative to Proposal Frequency](#azkg-006-anchor-state-advances-slowly-relative-to-proposal-frequency)
+    - [Mitigations](#mitigations-5)
+  - [aZKG-007: Proof Generation Is Feasible Within the Prove Window](#azkg-007-proof-generation-is-feasible-within-the-prove-window)
```
Joxess, line 36: Should we add another assumption that mentions that proposer claims are based on finalized L1 data? right now the `l1Head` commited during game creation could change if L1 is reorged at some point (the contract state is consistent, but the `rootClaim` may be based on L1 data that is no longer canonical). But this is mostly operational though 

@ashitakah @ajsutton

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: I like it! For me it is a yes

specs/fault-proof/stage-one/zk/zk-dispute-game.md
ashitakah: If L1 reorgs after game creation, the committed l1Head may no longer be canonical, making the `rootClaim` unprovable. This is an operational risk for the proposer though.

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: f1669fc

specs/fault-proof/stage-one/zk/zk-interface.md
ashitakah: f1669fc

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: f1669fc

specs/fault-proof/stage-one/zk/zk-dispute-game.md
ajsutton: In terms of game creation, If L1 reorgs then either:
* The game will no longer be created because it was re-orged out
* The game will be create with a L1 head from the newly canonical chain

It's not possible to have a game that has a non-canonical L1 head. But that's separate to whether a L1 reorg could cause the game creation tx to now be included on L1 prior to the batch data required for the proposal. That would make it an invalid proposal and the proposer could lose their bond.

So yes, a cautious proposer would wait for the batch data to be finalized on L1 and by default op-proposer does that, but it's not a strict requirement and I don't think we need an assumption for it. The proposer might accept a risk of some proposals being invalidated by a reorg in order to propose sooner and speed up withdrawals (we do this on testnets but its mostly just a legacy of pre-fault proofs times when testnet proposals instantly finalized). The system still works fine, just the proposer may lose bonds or wind up challenging its own proposal. op-node has a follow distance for L1 anyway (5 blocks by default) so the chances of seeing a block as `safe` on L2 and having it re-orged out is incredibly small.

Joxess (approved):

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,185 @@
...
+During `initialize()`, the game snapshots whether its game type is the currently respected game type
+(`wasRespectedGameTypeWhenCreated`). `AnchorStateRegistry.isGameClaimValid()` reads this flag when
+determining if a game's root claim can advance the anchor state or finalize withdrawals via the
+Portal. The snapshot ensures that a game type transition does not retroactively invalidate games
+that were created under the previous respected type.
+
+### Parent Validation
+
+When a parent is referenced (`parentIndex != type(uint32).max`), `initialize()` MUST revert if
```
wwared, line 76: To double-check my understanding: if parentIndex == type(uint32).max, we don't perform any of the checks below because the anchor state is always considered a valid starting point for a new game, correct? (No need to change anything, just asking)

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,185 @@
...
+- `challenge()` MUST revert if the game is not in the `Unchallenged` state.
+- Only one challenge is allowed per game.
+- Calling `challenge()` resets the deadline to `block.timestamp + maxProveDuration`.
+
+## Proving
+
+Proving is fully permissionless. Anyone may call `prove(proofBytes)` at any point before the
+current deadline, regardless of whether the game has been challenged.
+
+- `prove()` MUST revert if the game is already resolved (i.e., `status != IN_PROGRESS`).
```
wwared: nit: Where is `IN_PROGRESS` as a status defined? `enum GameStatus` above doesn't have it. Should this be `status != Resolved`? Or should it be `status != Resolved && status != UnchallengedAndValidProofProvided && status != ChallengedAndValidProofProvided`? Or should this be `gameOver() == false`?

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,185 @@
...
+  its own proof status.
+- Otherwise (parent `DEFENDER_WINS`), the outcome is determined as follows:
+
+| Game state at resolution                                 | Outcome           | Bond distribution                                              |
+| -------------------------------------------------------- | ----------------- | -------------------------------------------------------------- |
+| Unchallenged, deadline expired                           | `DEFENDER_WINS`   | Proposer recovers `initBond`                                   |
+| Unchallenged, valid proof provided                       | `DEFENDER_WINS`   | Proposer recovers `initBond`; prover receives nothing          |
+| Challenged, no proof by deadline                         | `CHALLENGER_WINS` | Challenger receives `initBond + challengerBond`                |
+| Challenged, valid proof, prover == proposer              | `DEFENDER_WINS`   | Proposer recovers `initBond + challengerBond`                  |
+| Challenged, valid proof, prover != proposer              | `DEFENDER_WINS`   | Proposer recovers `initBond`; prover receives `challengerBond` |
```
wwared, line 131: nit: Does it matter to make these cases explicitly separate? In the code would it be equivalent to always send `initBond` to the proposer and `challengerBond` to the prover?

I am perfectly fine keeping it like this if the reason is to make it more explicit and easier to understand, just wondering if there is some other underlying reason to keep the cases separate besides clarity

specs/fault-proof/stage-one/zk/game-mechanics.md
```diff
@@ -0,0 +1,185 @@
...
+
+1. Triggers `closeGame()` if not yet closed.
+2. Calls `DelayedWETH.unlock(recipient)` to queue the withdrawal. If `recipient` has no credit
+   allocated by this game, the unlock call is a no-op and the function does NOT revert. This
+   allows op-challenger to call `claimCredit` to close the game without knowing in advance whether
+   the recipient has outstanding credit.
+3. Calls `DelayedWETH.withdraw(recipient)` to transfer ETH to the recipient. This call MUST revert
+   if the recipient has no credit to withdraw, if the credit is not yet available (i.e., the
+   `DelayedWETH` delay has not elapsed), or if the ETH transfer fails.
+
```
wwared: I'm confused by the combination of:

> the unlock call is a no-op and the function does NOT revert. This allows op-challenger to call `claimCredit` to close the game without knowing in advance whether the recipient has outstanding credit

and

> This call MUST revert if the recipient has no credit to withdraw, if the credit is not yet available (i.e., the `DelayedWETH` delay has not elapsed), or if the ETH transfer fails

If the `DelayedWETH.withdraw(recipient)` call happens unconditionally and reverts under those conditions, then wouldn't it be impossible for op-challenger to call `claimCredit` to close the game without knowing whether the recipient has outstanding credit?

Am I missing something? I feel like either op-challenger needs to know if there is outstanding credit, it should call `closeGame` separately, or we need to conditionally call `DelayedWETH.withdraw(recipient)` in `claimCredit`

specs/fault-proof/stage-one/zk/zk-dispute-game.md
```diff
@@ -0,0 +1,463 @@
...
+| **DelayedWETH**                    | Bond custody with a deposit → unlock → withdraw lifecycle. Provides a time window for the Guardian to freeze funds post-resolution. |
+| **IZKVerifier**                    | Generic verifier interface. The concrete deployment for the initial release uses Succinct's PLONK verifier.                         |
+
+## Actors
+
+| Actor                       | Role                                                                                                                        |
+| --------------------------- | --------------------------------------------------------------------------------------------------------------------------- |
+| **Proposer**                | Fully permissionless. Creates games via `DisputeGameFactory.create()` with the required `initBond`.                         |
+| **Challenger**              | Fully permissionless. Disputes a proposal by calling `challenge()` and depositing `challengerBond`.                         |
+| **Prover**                  | Fully permissionless. Submits a valid ZK proof via `prove(proofBytes)`. May be the same address as the proposer.            |
```
wwared: nit:

```suggestion
| **Prover**                  | Fully permissionless. Submits a valid ZK proof via `prove(proofBytes)`. May be the same address as the proposer or the challenger. |
```

specs/fault-proof/stage-one/zk/zk-dispute-game.md
```diff
@@ -0,0 +1,463 @@
...
+
+The following fields are packed into `gameArgs` in order:
+
+| Field                  | Type       | Description                                        |
+| ---------------------- | ---------- | -------------------------------------------------- |
+| `absolutePrestate`     | `bytes32`  | ZK program identity (e.g., SP1 verification key)   |
+| `verifier`             | `address`  | Address of the `IZKVerifier` contract              |
+| `maxChallengeDuration` | `Duration` | Time window for challenges after game creation     |
+| `maxProveDuration`     | `Duration` | Time window for proof submission after a challenge |
+| `challengerBond`       | `uint256`  | Bond required to challenge a proposal              |
```
wwared, line 172: nit/question: Should `initBond` also be part of the game args or `ZKDisputeGameConfig`? Or do we want to use the same value for all chains? cc @ajsutton

specs/fault-proof/stage-one/zk/zk-dispute-game.md
```diff
@@ -0,0 +1,463 @@
...
+- The `DISPUTE_GAME_FINALITY_DELAY_SECONDS` airgap between resolution and `closeGame` provides
+  the Guardian a window to act.
+- `DelayedWETH` provides an additional window after `closeGame` to freeze funds.
+
+### aZKG-006: Anchor State Advances Slowly Relative to Proposal Frequency
+
+Under normal operation, the anchor state advances on a timescale (12+ hours minimum) that is much
+larger than typical proposal frequency (e.g., 1 hour), making orphan risk from parent validation
+negligible.
+
```
wwared: I agree and think this is correctly stated, but a question I have: is there any technical means that enforces that the anchor state advances slowly compared to proposals? As I understand it, the anchor state advances every time closeGame() is called on a valid resolved game, and in theory that could happen with the same frequency as proposals are submitted. Am I missing something?

specs/fault-proof/zk-fault-proof-vm.md
```diff
@@ -0,0 +1,105 @@
...
+Unlike interactive fault proofs, the ZK VM requires no multi-round on-chain interaction. A single
+`prove()` call to the dispute game is sufficient to cryptographically establish correctness,
+provided the verifier is sound.
+
+## ZK Program
+
+The ZK program is the circuit executed off-chain to generate a proof. It takes a set of
+[public values](#inputs) as inputs and verifies that applying L2 derivation and executing the
+resulting transactions against the starting state yields the claimed output root at the specified
+L2 block number, given the observed L1 data up to `l1Head`.
```
wwared, line 44: I would add the following clarification here, or something along the lines of:


```suggestion
L2 block number, given the observed L1 data up to `l1Head`.

The ZK program also receives a program-specific set of private input values known only to the prover and not the verifier. This represents things like necessary block and transaction data for fully replaying the transaction between the starting root to the claimed output root. This extra data is necessary for the correct execution of the ZK program, but the verifier does not need to check it and it is not posted on-chain alongside the proof. To ensure a malicious prover cannot use malicious values that lead to an invalid or unsound execution proof, the ZK program MUST verify that the private values are somehow directly or cryptographically linked to some of the public values. The ZK program should NOT trust any private prover-only values without explicitly checking it against some public value that the verifier has access to. This includes checks like hashing block data and comparing against the verifier hashes and ensuring all the blocks form a chain that starts and/or ends at one of the hashes in the public values.

Updates to the ZK program MUST ensure that the private input values can never be maliciously manipulated to lead to invalid proofs. Updates to ZK programs can modify what private input it receives and as long as the public input remains the same, the on-chain ZKDisputeGame does not need to be updated.
```

This is not really important for the correctness of the on-chain verifier (since the verifier by definition is never exposed to any of the prover-private values), but it is very important for the correctness of the ZK program. Changes to the ZK program must always keep this invariant or this can lead to unsound valid proofs being generated. If you think it makes sense, this might be a good assumption to add too maybe?

The private prover inputs to the ZK program are equivalent/analogous to the data availability requirements of the existing fault proof program.

specs/fault-proof/zk-fault-proof-vm.md
```diff
@@ -0,0 +1,105 @@
...
+
+The program is the ZK equivalent of the [Fault Proof Program](index.md#fault-proof-program): it
+runs the same L2 derivation logic (`op-node` + `op-geth`) but inside a zkVM, producing a proof
```
wwared: nit: The derivation logic being ran is actually `kona-node` or similar, since the Rust kona code will be what is actually used inside the zkVM (most zkVMs can more easily run Rust code than Go code currently). To simplify this, I would change it to:

```suggestion
runs the same L2 derivation logic as the fault proof program but inside a zkVM, producing a proof
```

specs/fault-proof/zk-fault-proof-vm.md
```diff
@@ -0,0 +1,105 @@
...
+- The ZK program is compiled to run inside the SP1 zkVM.
+- `absolutePrestate` = the SP1 program verification key (`programVKey`).
+- The on-chain verifier is Succinct's PLONK verifier, wrapped behind the `IZKVerifier` interface.
+
+## Proof Generation
+
+Proofs are generated off-chain by a prover that:
+
+1. Fetches the required L1 and L2 data up to `l1Head`.
+2. Executes the ZK program inside the zkVM with the public values as inputs.
```
wwared: nit:
```suggestion
2. Executes the ZK program inside the zkVM with the public values as inputs and provides the required L1 and L2 data as private values to the ZK program.
```

specs/fault-proof/zk-fault-proof-vm.md
wwared: Something like this maybe?

  iZKG-013: Private values are derived and/or linked to public values
  The ZK program MUST always either directly derive or cryptographically link any prover-only private input values to the public values. The public values are the only ones available to the on-chain verifier, and thus a malicious prover MUST NOT be able to use private values to create an invalid proof. This requires anchoring private values to the public values somehow.
  Impact - Critical: Violation means that a malicious prover can create a proof that verifies on-chain successfully, but represents possibly invalid state roots.

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: CopyShareYes, that's correct. The parent validation checks only apply when parentIndex != type(uint32).max.

When parentIndex == type(uint32).max, the game roots directly from the anchor state, which is treated as an unconditionally valid starting point, no blacklist, retirement, game type, or sequence number checks are performed.

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: Yep you absolutely right. I will change it. Nice catch!

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: Split into two rows to make the reward split explicit without requiring the reader to infer the prover == proposer edge case. I like to maintain separate, however im open to change it if you think that is not necessary.

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: As we change a bit the implementation we need to change this part of the spec.

specs/fault-proof/stage-one/zk/zk-dispute-game.md
ashitakah: initBond is set on the DisputeGameFactory via setInitBond(gameType, amount), which is a per-chain contract, so it's already configurable per chain, just not part of ZKDisputeGameConfig.

ashitakah: If we want to set the same value for all chains we could do it in the deploy I think, but correct me if im wrong

specs/fault-proof/stage-one/zk/zk-dispute-game.md
ashitakah: You're right, there's no technical mechanism enforcing this. the minimum time for the anchor to advance is maxChallengeDuration + DISPUTE_GAME_FINALITY_DELAY_SECONDS, but multiple resolved games could close in rapid succession. Updated the wording to make this an operational expectation rather than a guarantee.

specs/fault-proof/zk-fault-proof-vm.md
ashitakah: IMO This invariant belongs in zk-fault-proof-vm.md since it's a constraint on the ZK program itself, not on the on-chain dispute game. Added it there as iZKVM-001.

specs/fault-proof/zk-fault-proof-vm.md
ashitakah: [35e3cdb](https://github.com/ethereum-optimism/specs/pull/896/commits/35e3cdbb193e9bc2711acf45978c6faab1238eb1)

specs/fault-proof/zk-fault-proof-vm.md
ashitakah: [35e3cdb](https://github.com/ethereum-optimism/specs/pull/896/commits/35e3cdbb193e9bc2711acf45978c6faab1238eb1)

specs/fault-proof/zk-fault-proof-vm.md
ashitakah: [35e3cdb](https://github.com/ethereum-optimism/specs/pull/896/commits/35e3cdbb193e9bc2711acf45978c6faab1238eb1)

specs/fault-proof/stage-one/zk/zk-dispute-game.md
ashitakah: [35e3cdb](https://github.com/ethereum-optimism/specs/pull/896/commits/35e3cdbb193e9bc2711acf45978c6faab1238eb1)

specs/fault-proof/stage-one/zk/zk-dispute-game.md
ashitakah: [35e3cdb](https://github.com/ethereum-optimism/specs/pull/896/commits/35e3cdbb193e9bc2711acf45978c6faab1238eb1)

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: [35e3cdb](https://github.com/ethereum-optimism/specs/pull/896/commits/35e3cdbb193e9bc2711acf45978c6faab1238eb1)

specs/fault-proof/stage-one/zk/game-mechanics.md
ashitakah: [35e3cdb](https://github.com/ethereum-optimism/specs/pull/896/commits/35e3cdbb193e9bc2711acf45978c6faab1238eb1)

specs/fault-proof/stage-one/zk/game-mechanics.md
wwared: It's fine to keep it separate for clarity, just wanted to double check there wasn't another reason for it

specs/fault-proof/stage-one/zk/game-mechanics.md
wwared: The two-phase mechanism is more clear now, thanks

specs/fault-proof/stage-one/zk/zk-dispute-game.md
wwared: Ah I see that DGF includes options for setting it, that answers my question!

I don't think we'll ever want to set the same value for all chains, was just confused where it was coming from

wwared (approved):
LGTM

PR #897, opened 2026-03-26
feat: l2cm update specs

0xOneTony:
- Updated bundle format to follow what is being generated by the solidity script
- Updated how gas allocation value for the block is being handled

### DIFF
--- a/specs/protocol/l2-upgrades-1-execution.md
+++ b/specs/protocol/l2-upgrades-1-execution.md
@@ -429,9 +429,6 @@ The bundle is a JSON file with the following structure:
 }
 ```
 
-Fields within each transaction object MUST be serialized in alphabetical order (`data`, `from`, `gasLimit`,
-`intent`, `to`) to ensure deterministic JSON output.
-
 **Field Requirements:**
 
 - `metadata.version`: Bundle format version for compatibility tracking
--- a/specs/protocol/l2-upgrades-1-execution.md
+++ b/specs/protocol/l2-upgrades-1-execution.md
@@ -437,10 +437,10 @@ Fields within each transaction object MUST be serialized in alphabetical order (
 - `metadata.version`: Bundle format version for compatibility tracking
 - `transactions`: Array of transaction objects in execution order
 - `transactions[].data`: Transaction calldata as hex string
-- `transactions[].from`: Sender address. Set to the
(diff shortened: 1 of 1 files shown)


specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -419,25 +419,32 @@ The bundle is a JSON file with the following structure:
...
+`intent`, `to`) to ensure deterministic JSON output.
+
 **Field Requirements:**
 
 - `metadata.version`: Bundle format version for compatibility tracking
 - `transactions`: Array of transaction objects in execution order
-- `transactions[].to`: Target address (contract being called)
 - `transactions[].data`: Transaction calldata as hex string
+- `transactions[].from`: Sender address. Set to the
+  [Depositor Account](./l2-upgrades-2-contracts.md#depositor-account) for most transactions. Must be set to
```
0xniha: I'd remove the "Set to the Depositor Account for most transactions":
```suggestion
- `transactions[].from`: Sender address. Must be set to
```

0xniha: @maurelian  do you agree that `from` shouldn’t be optional?
We removed the optionality since Solidity doesn’t support optional args, and `address(0)` is used for upgrade txs, so it can’t serve as a fallback.
Are there any cases where we would use this argument as optional?

0xOneTony: updated https://github.com/ethereum-optimism/specs/pull/897/commits/11d9ee1131a36babac1c016ce4671b67a720a2dd

maurelian: It is a pretty important value, so I think it is OK to not make it optional.

specs/protocol/l2-upgrades-1-execution.md
```diff
@@ -419,25 +419,32 @@ The bundle is a JSON file with the following structure:
...
+      "gasLimit": 1000000,
+      "intent": "Deploy Example Implementation",
+      "to": "0x1234..."
     }
   ]
 }
 ```
 
+Fields within each transaction object MUST be serialized in alphabetical order (`data`, `from`, `gasLimit`,
+`intent`, `to`) to ensure deterministic JSON output.
```
maurelian: I think this is more of a solidity implementation issue than anything and we don't need to include it here.

```suggestion
```

maurelian (approved):
One suggestions, approving now to unblock.

PR #900, opened 2026-04-03
Interop: Remove "Supervisor / Managed Node" ; "Simplify Replacement Logic" ; "Specify Superroot API Behavior"

axelKingsley:
This is a second attempt at an earlier spec change. This time it's just the most obviously correct updates and none of the editorial or controversial updates.


karlfloersch (approved):
🚢

PR #908, opened 2026-05-04
Post-Execution transaction and SDM specs

### LINKED ISSUE #20182: SDM: Specs
Add SDM specs at https://specs.optimism.io/

nonsense:
This changeset defines **post-execution transaction** specs and the version-1 **Sequencer-Defined Metering (SDM)** specs.

   ## Changes

   - Add Post-Execution Transaction envelope spec (`0x7D`)
   - Add SDM version-1 payload schema
   - Define refund application, canonical gas accounting, and fee settlement
   - Add glossary and summary entries
   - Clarify that verifiers trust refunds included by the sequencer rather than recomputing them

Fixes: https://github.com/ethereum-optimism/optimism/issues/20182


specs/protocol/lagoon/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+
+```text
+[version, blockNumber, gasRefundEntries]
+```
+
+| Field              | Type                | Description                                                                                                   |
+| ------------------ | ------------------- | ------------------------------------------------------------------------------------------------------------- |
+| `version`          | `uint8`             | MUST be `1`.                                                                                                  |
+| `blockNumber`      | `uint64`            | The L2 block number containing this payload (see [post-exec.md § Block Number](./post-exec.md#block-number)). |
+| `gasRefundEntries` | `list<SDMGasEntry>` | Non-zero per-transaction gas refunds.                                                                         |
```
einar-oplabs, line 67:
```suggestion
| `version`          | `u8`             | MUST be `1`.                                                                                                  |
| `blockNumber`      | `u64`            | The L2 block number containing this payload (see [post-exec.md § Block Number](./post-exec.md#block-number)). |
| `gasRefundEntries` | `Vec<SDMGasEntry>` | Gas refunds. Exactly one entry per refund. A transaction has either zero or one refund.                                                                        |
```

specs/protocol/lagoon/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+
+### `SDMGasEntry`
+
+Each entry in `gasRefundEntries` is an RLP list of two fields:
+
+| Field       | Type     | Description                                      |
+| ----------- | -------- | ------------------------------------------------ |
+| `index`     | `uint64` | Transaction index within the block (zero-based). |
+| `gasRefund` | `uint64` | Gas refund for the transaction at `index`.       |
```
einar-oplabs, line 76:
```suggestion
| `index`     | `u64` | Transaction index within the block (zero-based). |
| `gasRefund` | `u64` | Gas refund for the transaction at `index`.       |
```

specs/protocol/lagoon/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Validity Rules
+
+A version-1 payload is invalid if any of the following hold:
+
+1. `gasRefundEntries` is empty.
```
einar-oplabs, line 82: What happens if there are no normal transactions in the block?

specs/protocol/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+| `index`     | `uint64` | Transaction index within the block (zero-based). |
+| `gasRefund` | `uint64` | Gas refund for the transaction at `index`.       |
+
+### Validity Rules
+
+A version-1 payload is invalid if any of the following hold:
+
+1. `gasRefundEntries` is empty.
+2. Any `SDMGasEntry` has `gasRefund == 0`.
+3. Entries are not strictly ordered by increasing `index`.
```
einar-oplabs:
```suggestion
3. Entries are not ordered by strictly increasing `index`.
```

specs/protocol/lagoon/sdm.md
einar-oplabs: Or if there is one normal transaction, but it does not qualify for a refund?

specs/protocol/lagoon/sdm.md
einar-oplabs: nit: I don't think it makes sense to use golang types in the spec going forward.

specs/protocol/lagoon/sdm.md
einar-oplabs: to answer my own questions:

1. there will be no PostExec txn.
2. there will be no entry and possibly not PostExec txn.

specs/protocol/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+Every transaction in the block is classified as exactly one of:
+
+| Kind       | Source                                                | May have refund? |
+| ---------- | ----------------------------------------------------- | ---------------- |
+| `Normal`   | A regular user transaction.                           | Yes              |
+| `Deposit`  | A [deposited transaction][g-deposited] (type `0x7E`). | No               |
+| `PostExec` | The post-exec transaction (type `0x7D`).              | No               |
+
+Deposits buy gas on L1 and have no L2-side gas-price to refund. The post-exec transaction carries SDM data only;
+it charges no fees and consumes no gas.
```
einar-oplabs:
```suggestion
it charges no fees, consumes no gas and is not executed as code.
```

specs/protocol/lagoon/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+
+For each `Normal` transaction at index `i`, define `refund(i)` as:
+
+- the `gasRefund` value in the payload entry whose `index == i`, if one exists; otherwise
+- `0`.
+
+The refund value is sequencer-defined block data. Clients use the included value directly when executing and
+validating the block.
+
+## Canonical Gas
```
einar-oplabs, line 114: I still feel we are re-defining what canonical means, which is potentially confusing for non-SDM aware users. I would suggest either 

    Under SDM we redefine canonical gas to mean ..

or consistently call it `sdm_canonical_gas`.

specs/protocol/lagoon/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Security Considerations
+
+**Sequencer-defined amounts.** Refund amounts are part of the sequencer's block data. Consensus only constrains their
+encoding, target transaction, and application bounds.
+
+**Cross-block replay.** The post-exec transaction's `blockNumber` field anchors each payload to its containing
+block. A payload from one block re-injected into another fails the envelope `blockNumber` check.
+
+**Settlement underflow.** Any settlement debit underflow invalidates the block.
```
einar-oplabs, line 216: Can this actually happen with the defined settlement rules, e.g.

    r * max(p - b, 0)

or is it just for adversarial-sequencer scenario?

specs/protocol/lagoon/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+### Signer and Signature
+
+A post-exec transaction has no signer. It is authorized by the protocol — by virtue of being synthesized by the
+sequencer at a position that satisfies the [block-level structural rules](#block-level-structural-rules) — rather
+than by a signature.
+
+For compatibility with tooling that expects an ECDSA signature on every typed transaction:
+
+- The recovered sender address is the zero address `0x0000000000000000000000000000000000000000`.
+- The signature triple `(v, r, s)` is reported as `(0, 0, 0)` in API responses.
```
einar-oplabs: Do we have a test for this?

specs/protocol/lagoon/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+keccak256(0x7D || rlp_encoded_payload)
+```
+
+The hash is computed over the full EIP-2718 encoding, so the type byte is included in the hash preimage. The
+payload's [block number](#block-number) is what distinguishes otherwise identical payloads in different blocks.
+
+### Transaction Field Projections
+
+Many of the fields a transaction is normally expected to expose are inapplicable to a post-exec transaction. When a
+post-exec transaction is observed through a generic transaction interface, its fields project as follows:
```
einar-oplabs: This does not seem to match the encoded data, and so is it really a projection?  It seems to me like we are deserializing into the wrong structure.  Is this observed view ever useful and/or meaningful? Do we have a test?

specs/protocol/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+
+A post-exec transaction emits a receipt with type byte `0x7D`. The RLP-encoded consensus fields of the receipt are
+identical to those of an EIP-1559 receipt:
+
+- `postStateOrStatus` ([EIP-658])
+- `cumulativeGasUsed`
+- `logsBloom`
+- `logs`
+
+A post-exec transaction is constructed by the protocol, executes synthetically, emits no logs, and consumes no gas
```
einar-oplabs:
```suggestion
A post-exec transaction is constructed by the protocol, does not execute, emits no logs, and consumes no gas.  It is metadata.
```

einar-oplabs (approved):

specs/protocol/lagoon/post-exec.md
nonsense: Yes, we do.

specs/protocol/lagoon/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+
+[g-transaction-type]: ../glossary.md#transaction-type
+[g-sequencer]: ../glossary.md#sequencer
+[g-post-exec-payload]: ../glossary.md#post-exec-payload
+[g-post-exec-schema-version]: ../glossary.md#post-exec-payload-schema-version
+[g-sdm]: ../glossary.md#sequencer-defined-metering
+
+## Overview
+
+Post-execution transactions are an [EIP-2718] [transaction type][g-transaction-type] that allows a block to carry
```
geoknee:
```suggestion
Post-execution transactions are an [EIP-2718] [transaction type][g-transaction-type] that allow a block to carry
```

specs/protocol/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+[g-transaction-type]: ../glossary.md#transaction-type
+[g-sequencer]: ../glossary.md#sequencer
+[g-post-exec-payload]: ../glossary.md#post-exec-payload
+[g-post-exec-schema-version]: ../glossary.md#post-exec-payload-schema-version
+[g-sdm]: ../glossary.md#sequencer-defined-metering
+
+## Overview
+
+Post-execution transactions are an [EIP-2718] [transaction type][g-transaction-type] that allows a block to carry
+sequencer-provided consensus data. Unlike user-submitted transactions, a post-exec transaction is synthesized by the
```
geoknee:
```suggestion
sequencer-provided consensus data. Unlike user-submitted transactions or deposited transactions, a post-exec transaction is synthesized by the
```
you could link to the relevant part of the glossary.

specs/protocol/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+| `blockNumber` | `uint64` | The L2 block number this payload is anchored to.                               |
+| _trailing_    | _varies_ | Defined by the active [schema version](#defined-schema-versions).              |
+
+The leading two fields define the envelope; all remaining fields belong to the schema selected by `version`.
+
+### Schema Version
+
+`version` is the [post-exec payload schema version][g-post-exec-schema-version]. When the payload schema needs to
+gain or change fields, a new version number is assigned and the new layout is documented as a
+[defined schema version](#defined-schema-versions).
```
geoknee:
```suggestion
`version` is the [post-exec payload schema version][g-post-exec-schema-version]. When the new payload schema calls for additional or different fields, a new version number is assigned and the new layout is documented as a
[defined schema version](#defined-schema-versions).
```

Nit: slightly clearer for me.

specs/protocol/lagoon/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+`version` is the [post-exec payload schema version][g-post-exec-schema-version]. When the payload schema needs to
+gain or change fields, a new version number is assigned and the new layout is documented as a
+[defined schema version](#defined-schema-versions).
+
+### Block Number
+
+`blockNumber` anchors the payload to the L2 block number of the containing block. The anchoring serves two
+purposes:
+
+1. It guarantees that two otherwise identical payloads produced for different blocks have distinct
```
geoknee: Two "different" blocks with identical block number, identical transactions (including identical post exec payloads) would have the **same** transaction hash even if the overall block hash is different. For example, if the parent block hash is different. Is that a problem? 

Even if it is not a problem, might be worth tightening up the language because "different blocks" is a bit ambiguous here and the sentence could be construed as incorrect.

specs/protocol/lagoon/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+### Encoding
+
+The EIP-2718 encoding of a post-exec transaction is:
+
+```text
+0x7D || rlp_encoded_payload
+```
+
+where `rlp_encoded_payload` is the RLP encoding of the [post-exec payload][post-exec-payload-envelope] as a list,
+and `||` denotes byte concatenation. The transaction encoding contains no outer list around the payload.
```
geoknee: I'm unsure what this final sentence means. Can you elaborate (or remove if it is not necessary)?

specs/protocol/lagoon/post-exec.md
geoknee: I agree, I think we could probably just cut this section entirely or state that the payload does not _need_ to include fields that transactions might be expected to have (e.g. the `to` field).

specs/protocol/lagoon/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+
+A post-exec transaction never charges fees, never debits or credits an account by virtue of being a transaction,
+and consumes no gas from the block gas pool. Any side effects on account balances are defined by the active schema,
+not by this envelope.
+
+### Signer and Signature
+
+A post-exec transaction has no signer. It is authorized by the protocol — by virtue of being synthesized by the
+sequencer at a position that satisfies the [block-level structural rules](#block-level-structural-rules) — rather
+than by a signature.
```
geoknee: Is it worth tying this to a) receiving unsafe payloads signed by the sequencer and b) deriving batches from transactions submitted by (signed by) the batch submitter? I think this is how the authorization works in practice. The transaction itself is not signed, but it is transmitted on a transport layer (either p2p gossip or L1) where there is some signing happening (albeit on an outer "wrapper" object).

specs/protocol/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+2. It detects misordered or replayed payloads at decode time, before any schema-specific validation runs.
+
+The normative rule that `blockNumber` equals the containing block's number is stated in
+[Block-Level Structural Rules](#block-level-structural-rules).
+
+### Defined Schema Versions
+
+| `version` | Schema                                 | Status       |
+| --------- | -------------------------------------- | ------------ |
+| `1`       | [Sequencer-Defined Metering](./sdm.md) | Experimental |
```
geoknee: I would personally vote to not have this "status" field. Schemas are either in or out, and SDM will be the initial schema.

specs/protocol/lagoon/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+The following rules hold for every block, regardless of which schema version is active. Any violation invalidates
+the block.
+
+1. **At most one.** A block contains at most one transaction with type byte `0x7D`.
+2. **Last in block.** When a `0x7D` transaction is present, it MUST be the final transaction of the block.
+3. **Anchored to block.** The `blockNumber` field of the embedded payload MUST equal the L2 block number of the
+   containing block.
+4. **Recognized schema.** The payload's `version` MUST be a [defined schema version](#defined-schema-versions).
+5. **Schema must be active.** When no schema version is active for the block's timestamp, the block MUST NOT
+   contain a `0x7D` transaction.
```
geoknee, line 189: Perhaps link this to the idea of schemas activating in hard forks? In the implementation, do we already have a two-stage check on SDM? i.e. we first check that the hardfork activating post-exec transactions happened, and then we check if the hardfork activating SDM happened? Initially this will be the same hardfork, but for the next schema it will be a different one.

specs/protocol/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+
+A post-exec transaction emits a receipt with type byte `0x7D`. The RLP-encoded consensus fields of the receipt are
+identical to those of an EIP-1559 receipt:
+
+- `postStateOrStatus` ([EIP-658])
+- `cumulativeGasUsed`
+- `logsBloom`
+- `logs`
+
+A post-exec transaction is constructed by the protocol, does not execute, emits no logs, and consumes no gas.  It is metadata.
```
geoknee:
```suggestion
A post-exec transaction is constructed by the protocol, does not modify the state, emits no logs, and consumes no gas.  It is metadata.
```

nit: slightly better IMO since it is not always clear what we mean by "execute"?

specs/protocol/lagoon/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Receipt
+
+A post-exec transaction emits a receipt with type byte `0x7D`. The RLP-encoded consensus fields of the receipt are
+identical to those of an EIP-1559 receipt:
+
+- `postStateOrStatus` ([EIP-658])
+- `cumulativeGasUsed`
+- `logsBloom`
+- `logs`
```
geoknee, line 202: Does this comply with https://eips.ethereum.org/EIPS/eip-2718?

geoknee: I guess it does. Why do we include these fields? Most of them are constant, and the remainder seem redundant. Is it to retain compatibility with 3rd party tooling?

specs/protocol/lagoon/post-exec.md
```diff
@@ -0,0 +1,233 @@
...
+
+The post-exec receipt participates in the block's receipts trie like any other receipt. Schema-specific data is
+not surfaced through the post-exec receipt itself; it is surfaced through extensions to the receipts of the user
+transactions whose execution the schema describes, as defined by the active schema.
+
+[EIP-658]: https://eips.ethereum.org/EIPS/eip-658
+
+## Derivation
+
+Post-exec transactions are L2-only and are not derived from L1. They are constructed by the sequencer during block
```
geoknee: Nit: "not derived from L1" is throwing me off here. Can we find a better wording for this section to remove ambiguity?

geoknee (commented):
Part 1 on post-exec. Looking great so far!

Meta comment: I think we should consider moving the content under the "lagoon" hardfork section. Even big changes like this don't usually get a spot outside of a hardfork section.

specs/protocol/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+[EIP-1559]: https://eips.ethereum.org/EIPS/eip-1559
+
+## Activation
+
+SDM is gated by a per-chain timestamp activation rule. When SDM is active for a block, the block MAY contain a
+post-exec transaction; when SDM is inactive, the block MUST NOT contain a post-exec transaction (see
+[post-exec.md § Block-Level Structural Rules](./post-exec.md#block-level-structural-rules)).
+
+SDM has no scheduled hardfork activation. It MUST NOT activate as part of any hardfork up to and including Karst.
+Until SDM is scheduled into a hardfork, every production block MUST be produced and validated with SDM inactive.
```
geoknee: This can probably be updated with the "lagoon" activation. This would be implicit if we put the section underneath a "lagoon" hardfork section.

specs/protocol/lagoon/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+| `blockNumber`      | `uint64`            | The L2 block number containing this payload (see [post-exec.md § Block Number](./post-exec.md#block-number)). |
+| `gasRefundEntries` | `list<SDMGasEntry>` | Non-zero per-transaction gas refunds.                                                                         |
+
+### `SDMGasEntry`
+
+Each entry in `gasRefundEntries` is an RLP list of two fields:
+
+| Field       | Type     | Description                                      |
+| ----------- | -------- | ------------------------------------------------ |
+| `index`     | `uint64` | Transaction index within the block (zero-based). |
```
geoknee, line 75: Do the indexes need to be ordered? Are duplicates allowed?

specs/protocol/lagoon/sdm.md
geoknee, line 76: Is there a requirement for this field to be nonzero?

geoknee: I see below that it is.

specs/protocol/lagoon/sdm.md
geoknee: I see below that ordering is required, and strictly increasing rules out duplicates.

specs/protocol/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+A version-1 payload is invalid if any of the following hold:
+
+1. `gasRefundEntries` is empty.
+2. Any `SDMGasEntry` has `gasRefund == 0`.
+3. Entries are not ordered by strictly increasing `index`.
+4. Any entry's `index` does not refer to a [`Normal`](#transaction-classification) transaction in the block.
+
+The envelope-level rules in [post-exec.md § Block-Level Structural Rules](./post-exec.md#block-level-structural-rules)
+apply in addition to the rules above. If the sequencer assigns no gas refunds, the block has no post-exec
+transaction.
```
geoknee:
```suggestion
apply in addition to the rules above.
```
I think future uses of post-exec might invalidate this sentence so suggest removing it.

specs/protocol/lagoon/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+| `gasRefund` | `uint64` | Gas refund for the transaction at `index`.       |
+
+### Validity Rules
+
+A version-1 payload is invalid if any of the following hold:
+
+1. `gasRefundEntries` is empty.
+2. Any `SDMGasEntry` has `gasRefund == 0`.
+3. Entries are not ordered by strictly increasing `index`.
+4. Any entry's `index` does not refer to a [`Normal`](#transaction-classification) transaction in the block.
```
geoknee: I wonder if we can avoid using "normal", which is effectively new nomenclature here. What if we just state the byte range that the `type` of the referred to transaction is allowed to fall into?

geoknee: Or if we still find it useful to use "normal", we can at least use the byte range to define it more concretely.

geoknee (commented):
Part 2: still need to review `sdm.md` from "Gas Refund Semantics" onwards.

specs/protocol/sdm.md
nonsense: Are you sure? I think the specs should reflect the `now` not the `future`, or? I think we should remove it only when we have PostExec carry other data, and indeed is not solely used for SDM?

geoknee: How about we say "As long as SDM is the only active payload type: If the sequencer assigns no gas refunds, the block has no post-exec
transaction."?

geoknee: Agreed about describing the "now", I can just see us not ever returning to tweak this part in future. So if we make it "forwards compatible" that is nice.

specs/protocol/lagoon/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+(post-Isthmus).
+
+| Recipient                         | Adjustment | Amount                                                              |
+| --------------------------------- | ---------- | ------------------------------------------------------------------- |
+| Sender (`tx.from`)                | credit     | `r * p + (operatorFee(evmGasUsed) - operatorFee(canonicalGasUsed))` |
+| Block beneficiary                 | debit      | `r * max(p - b, 0)`                                                 |
+| Base fee vault                    | debit      | `r * b`                                                             |
+| Operator fee vault (post-Isthmus) | debit      | `operatorFee(evmGasUsed) - operatorFee(canonicalGasUsed)`           |
+
+The sender credit equals the sum of the recipient debits. The L1 fee vault is not adjusted: L1 cost is independent
```
geoknee: I could not verify this invariant. Specifically when `p <= b`, the equation does not seem to balance?

specs/protocol/lagoon/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+of L2 gas usage.
+
+### Application Rules
+
+Settlement is applied after the transaction's EVM frame finishes and before the transaction state delta is
+committed. It is atomic with the transaction and does not produce a separate receipt.
+
+If any debit would underflow, the block is invalid. In normal operation this cannot occur, because each recipient
+was just paid the corresponding amount by the EVM in the same transaction.
```
geoknee: What do you mean by "normal operation"? Is there any mode where this can occur, other than in the case of a bug?

specs/protocol/lagoon/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+## Receipt Extension
+
+The post-exec transaction's own receipt carries no SDM-specific fields (see
+[post-exec.md § Receipt](./post-exec.md#receipt)).
+
+The JSON-RPC receipts returned for `Normal` transactions are extended with a single additional field:
+
+| Field         | Type                            | Description                                                                                                                     |
+| ------------- | ------------------------------- | ------------------------------------------------------------------------------------------------------------------------------- |
+| `opGasRefund` | `Quantity` (`uint64`) or `null` | The SDM gas refund credited to the transaction, or `null` when SDM was inactive for the block or the transaction had no refund. |
```
geoknee, line 184: Is this a breaking change for ecosystem tooling? For example, `cast receipt`?

specs/protocol/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+| ------------- | ------------------------------- | ------------------------------------------------------------------------------------------------------------------------------- |
+| `opGasRefund` | `Quantity` (`uint64`) or `null` | The SDM gas refund credited to the transaction, or `null` when SDM was inactive for the block or the transaction had no refund. |
+
+The value is sourced from the embedded post-exec payload's `gasRefundEntries` for the transaction's `index`.
+`opGasRefund` is not part of the receipt's RLP encoding and is not committed to the receipts trie.
+
+## Backwards Compatibility
+
+Until SDM is scheduled into a hardfork, blocks are produced and validated identically to a chain that has never
+specified SDM: no post-exec transactions appear, no canonical-gas adjustment is performed, and no settlement runs.
```
geoknee:
```suggestion
Prior to the hardfork timestamp where SDM activates, blocks are produced and validated identically to a chain that has never
specified SDM: no post-exec transactions appear, no canonical-gas adjustment is performed, and no settlement runs.
```

specs/protocol/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+
+When SDM is scheduled, two changes become observable:
```
geoknee:
```suggestion
When SDM is active, two changes become observable:
```

specs/protocol/sdm.md
```diff
@@ -0,0 +1,199 @@
...
+- The receipts of `Normal` transactions in such blocks gain the `opGasRefund` field; clients that ignore unknown
+  fields are unaffected.
+
+Mempool and transaction-pool interfaces are unchanged: post-exec transactions are not user-submittable and do not
+propagate over the public transaction-gossip protocol.
+
+## Security Considerations
+
+**Sequencer-defined amounts.** Refund amounts are part of the sequencer's block data. Consensus only constrains their
+encoding, target transaction, and application bounds.
```
geoknee:
```suggestion
**Sequencer-defined amounts.** Refund amounts are part of the sequencer's block data. The only consensus-defined constraints are on their
encoding, target transaction, and application bounds; otherwise the sequencer has complete freedom to allocate refunds according to arbitrary policies. 
```

geoknee: Not sure if this captures your intended meaning. What are "application bounds" in this context?

specs/glossary.md
```diff
@@ -834,6 +839,60 @@ In these specifications, "execution engine" always refer to the L2 execution eng
...
+
+# Post-Execution Transactions
+
+[post-exec]: glossary.md#post-execution-transactions
+
+A class of [EIP-2718] [transaction types][transaction-type] that carry sequencer-provided consensus data.
+Post-execution transactions are emitted by the [sequencer], are appended to a block after the last user transaction,
+and are not propagated through the public mempool.
+
+## Post-Execution Transaction
```
geoknee: Do we need both of these? IMO better to just keep the `post-exec-tx` one and drop `post-exec`.

specs/protocol/lagoon/sdm.md
geoknee: I agree "canonical gas" is not really sufficiently self-describing and could be taken to mean different things. So best to either come up with a more descriptive name, or explicitly define it (or both).

geoknee (commented):
👏 Great stuff @nonsense !

Left mostly minor suggestions. Main things I would block on are:
* nesting these specs under "lagoon"
* equation balancing query below

specs/protocol/lagoon/sdm.md
nonsense: we could use "User" instead of "Normal", what do you think? I think `Transaction Classification` clearly explains what's meant here, no?

specs/protocol/lagoon/post-exec.md
nonsense: I am not sure I follow -- we have a check for single hard fork right now, not for 2 hardforks -- we check activation of this feature based on Interop/Lagoon activation.

specs/protocol/post-exec.md
nonsense: I updated it to `A post-exec transaction is constructed by the protocol; it is not executed as EVM code, emits no logs, and consumes no gas from the block gas pool`. It does modify some state, even if it is not executed as EVM code.

specs/protocol/lagoon/post-exec.md
nonsense: I reviewed the actual RPCs, and updated the spec -- the RPC object is now minimal.

specs/protocol/lagoon/sdm.md
nonsense: I explicitly defined it. I am hesitant to rename it now, as this is what the implementation is also using.

specs/glossary.md
```diff
@@ -834,6 +839,65 @@ In these specifications, "execution engine" always refer to the L2 execution eng
...
+
+## Canonical Gas
+
+[canonical-gas]: glossary.md#canonical-gas
+
+Under [Sequencer-Defined Metering](#sequencer-defined-metering), the gas a transaction is accounted for after its
+gas refund is applied: `canonicalGasUsed = evmGasUsed - refund`, where `evmGasUsed` is the raw gas reported by the
+EVM. Canonical gas is the value recorded in transaction receipts and summed into the block's `cumulativeGasUsed`
+and `gasUsed`. It is distinct from the raw EVM gas, and unrelated to the "canonical chain" sense of _canonical_
+used elsewhere in this glossary.
```
geoknee, line 897:
```suggestion
Under [Sequencer-Defined Metering](#sequencer-defined-metering), canonical gas is the net gas consumed by a transaction for accounting purposes (i.e. after its
gas refund is applied). Specifically: `canonicalGasUsed = evmGasUsed - refund`, where `evmGasUsed` is the raw gas reported by the
EVM. Canonical gas is the value recorded in transaction receipts and summed into the block's `cumulativeGasUsed`
and `gasUsed`fields. It is distinct from the raw EVM gas, and unrelated to the "canonical chain" sense of _canonical_
used elsewhere in this glossary.
```

geoknee: (optional, take it if you like it!)

geoknee (approved):

geoknee (approved):

PR #918, opened 2026-06-11
Clean up interop specs and fix typos

geoknee:
- Rename "interop network upgrade" and related timestamps to "Lagoon activation timestamp" and standardize "activation block" references
- Remove several "Future Considerations" and empty Security TODO sections
- Enforce chain id representation as uint256 in Chain ID section (but more succinctly)
- Fix keccak spelling/usage (keccak256) and minor wording improvements
- Recommend L2toL2CrossDomainMessenger as preferred entrypoint


specs/interop/dependency-set.md
```diff
@@ -31,29 +29,13 @@ to consume logs that it has produced much more cheaply than providing a block ha
...
-replay attacks between chains. This EIP does not specify the max size of a chain id, although
-[EIP-2294](https://eips.ethereum.org/EIPS/eip-2294) attempts to add a maximum size. Since this EIP is
-stagnant, all representations of chain ids MUST be the `uint256` type.
-
-In the future, OP Stack chains reserve the right to use up to 32 bytes to represent a chain id. The
-configuration of the chain should deterministically map to a chain id and with careful architecture
-changes, all possible OP Stack chains in the superchain will be able to exist counterfactually.
-
-It is a known issue that not all software in the Ethereum ecosystem can handle 32 byte chain ids.
+All representations of chain ids MUST be the `uint256` type.
```
geoknee: This felt quite wordy so I just chopped it down to what is actually being specified.

specs/interop/dependency-set.md
```diff
@@ -31,29 +29,13 @@ to consume logs that it has produced much more cheaply than providing a block ha
...
 The dependency set is managed in the client software. Adding a chain to the dependency set is
 considered an upgrade to the network. It is not possible to remove chains from the dependency set.
 
-## Future Considerations
-
-### Layer 1 as Part of the Dependency Set
-
-The layer one MAY be part of the dependency set in the future. This means that any event
-created on layer one is consumable on layer two.
-
```
geoknee, line 56: I'm using the principle here that "we might do something in future" does not belong in the specs.

specs/interop/token-bridging.md
```diff
@@ -185,90 +182,3 @@ function relayERC20(SuperchainERC20 _token, address _from, address _to, uint256
   emit RelayedERC20(address(_token), _from, _to, _amount, _source);
 }
 ```
-
-## Future Considerations
```
geoknee, line 189: Another instance of "we might do something in future". Probably belongs more in a design doc.

specs/interop/dependency-set.md
ajsutton: A bunch of our own software doesn't actually meet this requirement. I suspect we want to say something like:

All chain IDs used in interop dependency sets must fit within a uint64. Software should be designed to support up to uint256 chain IDs.

ajsutton (approved):
LGTM. Just one nit on chain ID so we don't specify something that we don't support and drown in AI bug reports.

specs/interop/dependency-set.md
geoknee:
```suggestion
All chain IDs used in interop dependency sets must fit within a `uint64`. Software should be designed to support up to uint256 chain IDs.
```

PR #922, opened 2026-07-08
system-config: remove batchInbox getter and setGasConfig

mds1:
Spec side of ethereum-optimism/optimism#21641, which removes `batchInbox()` and the deprecated `setGasConfig` from the SystemConfig at op-contracts v8.0.0.

Changes:

1. The `batchInbox` and `setGasConfig` sections stay as historical record rather than being deleted, each marked `(-op-contracts/v8.0.0)` and noting that it was removed in `SystemConfig` v4.0.0. Their TOC entries stay too.
2. `initialize`'s batch inbox MUST is split into a versioned pair: `(-op-contracts/v8.0.0)` sets the Batch Inbox address, `(+op-contracts/v8.0.0)` clears the legacy slot. Nothing in the OP Stack read that copy — derivation takes the batch inbox from the rollup config — and keeping it meant OPCMv2's reinitialization could recompute it with the new-chain addressing scheme, which is what rotated it for every pre-OPCM chain during U19.
3. Version markers now spell out which scheme they mean. `(+op-contracts/vX.Y.Z)` and `(-op-contracts/vX.Y.Z)` are op-contracts releases; a version written as `SystemConfig` vX.Y.Z is the contract's own semver. There is a one-line note to that effect at the top of the Function Specification section, and the existing markers in `system-config.md` and `stage-1.md` are normalized to the full identifier.

   The existing `(+v4.1.0)` markers were genuinely ambiguous, so to be explicit about why they are releases: at `op-contracts/v4.1.0` the feature-flag and optional-`ETHLockbox` pause behavior those markers describe is present, while `SystemConfig` is at 3.7.0 and `SuperchainConfig` at 2.3.0. Neither contract has ever been at 4.1.0 — `SuperchainConfig` is still at 2.4.3 today — and `setFeatureEnabled` is absent at `op-contracts/v4.0.0`.
4. `ConfigUpdate` type 1 records that as of `SystemConfig` v4.0.0 the `overhead` entry is always zero, and that the payload stays two words so its length is unchanged. `_setGasConfigEcotone` hardcodes the zero because op-node ignores the overhead after Ecotone, but the event data has to stay 64 bytes.
5. The post-Ecotone **Overhead** note says it is no longer writable, since `setGasConfig` was its only writer, and that `setGasConfigEcotone` emits zero in its place. The variable and getter are kept, so `overhead()` still works.
6. The Batch Inbox address heading in `configurability.md` loses its stale `SystemConfig.sol#L176` permalink and says the value is defined in the rollup configuration and, as of `SystemConfig` v4.0.0, is no longer stored onchain.

Not doing here, per the review discussion: a separate per-contract _Changelog_ section. The `(+v)`/`(-v)` markers already carry this change now that they say which versioning scheme they use. Happy to open a follow-up issue for the Changelog idea if we want it.

### DIFF
--- a/specs/protocol/stage-1.md
+++ b/specs/protocol/stage-1.md
@@ -117,15 +117,15 @@ identifier determines which systems or chains are affected by the pause:
 - When the identifier is a non-zero address, the pause applies specifically to the chain or set of
   chains associated with that identifier.
 
-(-v4.1.0) The identifier must be an `ETHLockbox` address or `address(0)`. This allows for targeted
+(-op-contracts/v4.1.0) The identifier must be an `ETHLockbox` address or `address(0)`. This allows for targeted
 pausing of either specific chains, the interop set (which shares an `ETHLockbox` contract), or all
 chains that share the same `SuperchainConfig` when `address(0)` is used as the identifier.
 
-(-v4.1.0) OP Chains are expected to integrate with the `SuperchainConfig` via their `SystemConfig`
+(-op-contracts/v4.1.0) OP Chains are expected to integrate with the `SuperchainConfig` via their `SystemConfig`
 contract, which will check for the status of the pause by passing along the address of the
 `ETHLockbox` being used within that system as the Pause Identifier.
 
-(+v4.1.0) The identifier must be an `ETHLockbox` address, an `OptimismPortal` address, or
+(+op-contracts/v4.1.0) The identifier must be an `ETHLockbox` address, an `OptimismPortal` address, or
 `address(0)`. This allows for targeted pausing of specific chains, the interop set (which shares an
 `ETHLockbox` contract), or all chains that share the same `SuperchainConfig` when `address(0)` is
 used as the identifier. When the `ETHLockbox`
--- a/specs/protocol/system-config.md
+++ b/specs/protocol/system-config.md
@@ -32,11 +32,13 @@
   - [optimismPortal](#optimismportal)
   - [optimismMintableERC20Factory](#optimismmintableerc20factory)
   - [getAddresses](#getaddresses)
+  - [batchInbox](#batchinbox)
   - [startBlock](#startblock)
   - [paused](#paused)
   - [superchainConfig](#superchainconfig)
   - [setUnsafeBlockSigner](#setunsafeblocksigner)
   - [setBatcherHash](#setbatcherhash)
+  - [setGasConfig](#setgasconfig)
   - [setGasConfigEcotone](#setgasconfigecotone)
   - [setGasLimit](#setgaslimit)
   - [setEIP1559Params](#seteip1559params)
@@ -104,8 +106,8 @@ After the Ecotone upgrade:
 
 - The **Scalar** attribute encodes additional scalar information in a versioned encoding scheme
 - The **Overhead** value is ignored and does not affect the L2 state-transition output. As of
-  SystemConfig 4.0.0 it is also immutable, since `setGasConfig` was the only function that
-  wrote it and has been removed
+  `SystemConfig` v4.0.0 it is also no longer writable, since `setGasConfig` was its only writer,
+  and `setGasConfigEcotone` emits zero in its place
 
 #### Post-Ecotone Scalar Encoding
 
@@ -183,7 +185,8 @@ In version `0`, the following update types are supported:
 - Type `0`: `batcherHash` overwrite, as `bytes32` payload
 - Type `1`: Pre-Ecotone, `overhead` and `scalar` overwrite, as two packed `uint256` entries. After
   Ecotone upgrade, `overhead` is ignored and `scalar` is interpreted as a versioned encoding that
-  updates `baseFeeScalar` and `blobBaseFeeScalar`
+  updates `baseFeeScalar` and `blobBaseFeeScalar`. As of `SystemConfig` v4.0.0 the `overhead` entry
+  is always zero, and the payload stays two words so its length is unchanged
 - Type `2`: `gasLimit` overwrite, as `uint64` payload
 - Type `3`: `unsafeBlockSigner` overwrite, as `address` payload
 - Type `4`: `eip1559Params` overwrite, as `uint256` payload encoding denomination and elasticity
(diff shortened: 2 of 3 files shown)


specs/protocol/system-config.md
```diff
@@ -318,15 +316,6 @@ Allows the owner to update the [Batcher Hash](#batcher-hash).
...
 
-### setGasConfig
-
-Allows the owner to update the gas configuration parameters (pre-Ecotone).
-
-- MUST revert if called by an address other than the owner.
-- MUST revert if the scalar exceeds the maximum allowed value (no upper 8 bits should be set).
-- MUST update the overhead and scalar values.
-- MUST emit a ConfigUpdate event with the UpdateType.FEE_SCALARS type.
-
```
sebastianst, line 329: We usually leave historical information in the specs, which is also why this section was still there with the "(pre-Ecotone)" caveat. Let's better leave it in here and instead add another sentence that this function has been deleted in the SystemConfig 4.0.0, which gets deployed on standard chains with Upgrade 20.

specs/protocol/system-config.md
```diff
@@ -270,10 +272,6 @@ Returns the address of the OptimismMintableERC20Factory contract.
 
 Returns a consolidated struct containing all the L1 contract addresses.
 
-### batchInbox
-
-Returns the address of the [Batch Inbox](#batch-inbox).
```
sebastianst, line 275: Same comment, let's leave it in but say, deleted as of SystemConfig 4.0.0

specs/protocol/system-config.md
sebastianst: what version does this refer to?

sebastianst: what version does this refer to?

specs/protocol/system-config.md
```diff
@@ -201,9 +201,11 @@ If a System Config Update cannot be parsed for any reason, it is not applied and
...
 - MUST set the owner of the contract to the provided `_owner` address.
 - MUST set the SuperchainConfig contract address.
-- MUST set the batcher hash, gas config, gas limit, unsafe block signer, resource config, batch
-  inbox, L1 contract addresses, and L2 chain ID.
+- MUST set the batcher hash, gas config, gas limit, unsafe block signer, resource config,
+  L1 contract addresses, and L2 chain ID.
 - MUST set the start block to the current block number if it hasn't been set already.
+- MUST clear the legacy batch inbox storage slot. The OP Stack reads the batch inbox address from
+  the rollup configuration, so the value stored here was redundant and could drift from it.
 - MUST validate the resource configuration parameters against system constraints.
```
sebastianst, line 218: Not sure how we'll want to keep historical context here...

sebastianst (commented):
We usually keep historical information in the protocol specs, but I can see how this gets increasingly awkward, see comments. Wondering if we want instead switch to a description of the present state only and keep a separate _Changelog_ section where we describe the diffs of each contract change.

specs/protocol/system-config.md
mds1: Restored — the `setGasConfig` section is back with its TOC entry, marked `(-op-contracts/v8.0.0)` and saying it was removed in `SystemConfig` v4.0.0, pointing at `setGasConfigEcotone` instead. The MUSTs are unchanged.

specs/protocol/system-config.md
mds1: Same — `batchInbox` is back with its TOC entry, marked `(-op-contracts/v8.0.0)`, noting it was removed in `SystemConfig` v4.0.0 because the address lives in the rollup config, which is what the OP Stack reads.

specs/protocol/system-config.md
mds1: op-contracts releases, not a contract semver, clarified this in the text with a new convention: `(+op-contracts/vX.Y.Z)` / `(-op-contracts/vX.Y.Z)` for releases and a spelled-out `SystemConfig` vX.Y.Z for a contract version

mds1: Same as above — op-contracts release. Normalized to `(+op-contracts/v4.1.0)`.

specs/protocol/system-config.md
mds1: Kept it, with the batch inbox split out of the combined MUST into a versioned pair so the old behavior sits next to the new one:

```
- (-op-contracts/v8.0.0) MUST set the [Batch Inbox](#batch-inbox) address.
- (+op-contracts/v8.0.0) MUST clear the legacy batch inbox storage slot. The OP Stack reads the
  batch inbox address from the rollup configuration, so the onchain copy was redundant and could
  drift from it.
```

That seemed better than a prose aside, since the surrounding bullets are all MUSTs.

specs/protocol/stage-1.md
```diff
@@ -117,15 +117,15 @@ identifier determines which systems or chains are affected by the pause:
...
 pausing of either specific chains, the interop set (which shares an `ETHLockbox` contract), or all
 chains that share the same `SuperchainConfig` when `address(0)` is used as the identifier.
 
-(-v4.1.0) OP Chains are expected to integrate with the `SuperchainConfig` via their `SystemConfig`
+(-op-contracts/v4.1.0) OP Chains are expected to integrate with the `SuperchainConfig` via their `SystemConfig`
 contract, which will check for the status of the pause by passing along the address of the
 `ETHLockbox` being used within that system as the Pause Identifier.
 
-(+v4.1.0) The identifier must be an `ETHLockbox` address, an `OptimismPortal` address, or
+(+op-contracts/v4.1.0) The identifier must be an `ETHLockbox` address, an `OptimismPortal` address, or
```
maurelian, line 128: I have never noticed this `+/- semver` convention before. I would assume that it signifies when a particular clause was added or removed, but the `-` above. don't seem like things which were removed.

specs/protocol/stage-1.md
```diff
@@ -117,15 +117,15 @@ identifier determines which systems or chains are affected by the pause:
 - When the identifier is a non-zero address, the pause applies specifically to the chain or set of
   chains associated with that identifier.
 
-(-v4.1.0) The identifier must be an `ETHLockbox` address or `address(0)`. This allows for targeted
+(-op-contracts/v4.1.0) The identifier must be an `ETHLockbox` address or `address(0)`. This allows for targeted
 pausing of either specific chains, the interop set (which shares an `ETHLockbox` contract), or all
 chains that share the same `SuperchainConfig` when `address(0)` is used as the identifier.
 
-(-v4.1.0) OP Chains are expected to integrate with the `SuperchainConfig` via their `SystemConfig`
+(-op-contracts/v4.1.0) OP Chains are expected to integrate with the `SuperchainConfig` via their `SystemConfig`
```
maurelian, line 124: Note that https://github.com/ethereum-optimism/optimism/pull/22095 would change this.

specs/protocol/stage-1.md
maurelian: OK, I found the addition that documents the convention where +/- means before/after.

maurelian (approved):

sebastianst (approved):

PR #930, opened 2026-08-18
pcd: align specs with the shipped deployment pipeline

0xiamflux:
## Context
The PCD spec pages landed before super root support and the `prepare` / `prestate` /
`continue` hardening merged. This updates all three against the code that shipped. Spec-only.

- Both game families are deployable at genesis. New definitions for Super Root, Starting Anchor
  Root, Selected Prestate and Fallback Prestate replace `Prestate / Prestate Hash`, and
  `iOPD-002` pins the root and its `l2SequenceNumber` per family.
- OPCM's game-type check is scoped to the `SUPER_ROOT_GAMES_MIGRATION` mode, plus the three
  value checks PCD added: non-zero anchor root, no `0xdead` placeholder, non-zero prestate.
- `DeployOPChainInput` gains `startingAnchorRoot` and `cannonAbsolutePrestate`. A permissionless
  deploy enables the respected game and its guardian fallback.
- Command behavior now matches the pipeline: anchor pinning, prestate invalidation on every
  `prepare`, and `continue`'s digest, fork-simulation, anchor and post-deploy checks.
- The `BLOCKHASH` and second-RPC guards never shipped and are replaced by what the code does.

Claims verified against `615b1c75dd` and `9714d4dd2c` on `develop` and op-deployer's three
commands.

### DIFF
--- a/specs/experimental/predictive-chain-deployments/contracts.md
+++ b/specs/experimental/predictive-chain-deployments/contracts.md
@@ -285,19 +285,7 @@ Before PCD, enabling any permissionless game type during an initial deployment r
 `OPContractsManagerV2_InvalidGameConfigs`, on the premise that no prestate exists yet. PCD invalidates that premise.
 A real prestate now exists at deployment time.
 
-The game-type check is replaced by one scoped to the OPCM's mode:
-
-```solidity
-// Initial deployments must select game types compatible with the active mode.
-bool validForInitialDeploy = superRootGamesMigrationEnabled
-    ? (isSuperPermissionedGame || isSuperCannonKonaGame)
-    : (isPermissionedCannonGame || isCannonKonaGame);
-if (_isInitialDeployment && _cfg.disputeGameConfigs[i].enabled && !validForInitialDeploy) {
-    revert OPContractsManagerV2_InvalidGameConfigs();
-}
-```
-
-PCD then:
+The game-type check is replaced by one scoped to the OPCM's mode. PCD then:
 
 - MUST allow the permissionless game type of the OPCM's mode to be `enabled` during an initial deployment.
 - MUST reject a game type from the other mode, and MUST keep `CANNON` and `ZK_DISPUTE_GAME` rejected at initial
--- a/specs/experimental/predictive-chain-deployments/contracts.md
+++ b/specs/experimental/predictive-chain-deployments/contracts.md
@@ -27,7 +27,7 @@
   - [DeployOPChain script](#deployopchain-script)
     - [Address prediction](#address-prediction)
     - [Broadcasting a permissionless deployment](#broadcasting-a-permissionless-deployment)
-    - [Script input struct](#script-input-struct)
+    - [Script inputs](#script-inputs)
   - [OPCMv2 config validation](#opcmv2-config-validation)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
@@ -238,12 +238,11 @@ substitutions are safe (see [aPCD-001](#apcd-001-opcm-address-determinism)).
 
 #### Broadcasting a permissionless deployment
 
-Today the script hard-codes a permissioned-only initial deployment:
+Before PCD the script hard-coded a permissioned-only initial deployment:
 
-- It `require`s the input game type to be `PERMISSIONED_CANNON`, reverting with *"only PERMISSIONED_CANNON game type
-  is supported for initial deployment"*.
-- It builds the six `disputeGameConfigs` with the permissionless types disabled.
-- It sets `startingRespectedGameType` to the permissioned type and `startingAnchorRoot` to a placeholder.
+- It rejected any input game type other than `PERMISSIONED_CANNON`.
+- It built the six `disputeGameConfigs` with the permissionless types disabled.
+- It set `startingRespectedGameType` to the permissioned type and `startingAnchorRoot` to a placeholder.
 
 In order to support permissionless deployments the script is updated to:
 
@@ -263,24 +262,17 @@ In order to support permissionless deployments the script is updated to:
   addresses match the prediction (see
   [iOPD-001](./op-deployer.md#iopd-001-the-deployed-l1-system-matches-the-committed-artifacts)).
 
-#### Script input struct
+#### Script inputs
 
-Carrying those values into the script requires extending its Solidity input struct. `Types.DeployOPChainInput`
-(`scripts/libraries/Types.sol`) today exposes a single `disputeGameType` and `disputeAbsolutePrestate`. It gains two
-fields, and one existing field changes meaning:
+A permissionless deployment supplies three values the script did not previously take:
 
-| Field | Exists | Meaning |
-| --- | --- | --- |
-| `startingAnchorRoot` | new | The `Proposal` seeded into the `AnchorStateRegistry` |
-| `cannonAbsolutePrestate` | new | The [fallback prestate](./overview.md#fallback-prestate) for the guardian fallback game |
-| `disputeAbsolutePrestate` | repurposed | The [selected prestate](./overview.md#selected-prestate) of the respected game type |
+- the [starting anchor root](./overview.md#starting-anchor-root) seeded into the `AnchorStateRegistry`,
+- the [selected prestate](./overview.md#selected-prestate) of the respected game type, and
+- for the output root family only, the [fallback prestate](./overview.md#fallback-prestate) of the guardian fallback
+  game.
 
-`cannonAbsolutePrestate` is the **fallback**, not the permissionless prestate.
-The permissionless prestate is carried in `disputeAbsolutePrestate`, which previously carried the permissioned game's
-prestate.
-
-The script MUST reject a `CANNON_KONA` input whose two prestate values are equal. They commit to different fault-proof
-programs, **so equal values always indicate a misconfigured producer.**
+The script MUST reject an output root family input whose selected prestate equals its fallback prestate. They commit
+to different fault-proof programs, **so equal values always indicate a misconfigured producer.**
 
 ### OPCMv2 config validation
 
@@ -310,18 +302,19 @@ PCD then:
 - MUST allow the permissionless game type of the OPCM's mode to be `enabled` during an initial deployment.
 - MUST reject a game type from the other mode, and MUST keep `CANNON` and `ZK_DISPUTE_GAME` rejected at initial
   deployment (see [Supported initial game types](#supported-initial-game-types)).
-- MUST validate the values a permissionless initial deployment now carries. These checks are new, before PCD both
-  values were always placeholders:
+- MUST apply the checks PCD adds:
   - `startingAnchorRoot.root` MUST be non-zero on any initial deployment, and its `l2SequenceNumber` MUST leave room
     for a `uint64` successor.
   - The `0xdead` placeholder root MUST be rejected when an enabled config is one of `CANNON`, `CANNON_KONA` or
     `SUPER_CANNON_KONA`. Only a permissioned-only deployment may still carry the placeholder.
   - An enabled config among those same three game types MUST carry a non-zero `absolutePrestate`, decoded from its
     `gameArgs`.
-- MUST keep every other pre-existing validation intact:
+  - `SUPER_PERMISSIONED` MUST carry a zero init bond, whether or not it is enabled.
+- MUST keep every other validation OPCM already performs intact:
   - exactly six configs in the fixed game-type order,
-  - the init-bond rules, including `SUPER_PERMISSIONED` carrying a zero init bond,
-  - the `ZK_DISPUTE_GAME` dev-flag gate, and
+  - the init-bond rules for every other game type,
+  - the `ZK_DISPUTE_GAME` dev-flag gate plus its non-zero prestate requirement, both reachable only through the
+    upgrade path, and
   - `startingRespectedGameType` must correspond to an enabled config.
 
 The value checks reject an unset or placeholder input. They cannot reject a wrong one, since any non-zero prestate and
--- a/specs/experimental/predictive-chain-deployments/contracts.md
+++ b/specs/experimental/predictive-chain-deployments/contracts.md
@@ -8,6 +8,7 @@
 - [Definitions](#definitions)
   - [ChainContracts struct](#chaincontracts-struct)
   - [FullConfig](#fullconfig)
+  - [Supported initial game types](#supported-initial-game-types)
 - [Assumptions](#assumptions)
   - [aPCD-001: OPCM address determinism](#apcd-001-opcm-address-determinism)
     - [Mitigations](#mitigations)
@@ -26,6 +27,7 @@
   - [DeployOPChain script](#deployopchain-script)
     - [Address prediction](#address-prediction)
     - [Broadcasting a permissionless deployment](#broadcasting-a-permissionless-deployment)
+    - [Script input struct](#script-input-struct)
   - [OPCMv2 config validation](#opcmv2-config-validation)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
@@ -33,8 +35,8 @@
 ## Overview
 
 This document describes the contracts changes that are part of the implementation of the
-[Predictive Chain Deployments (PCD) design](https://github.com/ethereum-optimism/design-docs/pull/385). The bulk of
-the work is on the `op-deployer` side and the Solidity surface is deliberately minimal: one on-chain contract change
+[Predictive Chain Deployments (PCD) design](https://github.com/ethereum-optimism/design-docs/blob/main/ecosystem/op-deployer/predictive-chain-deployments/pcd-design.md).
+The bulk of the work is on the `op-deployer` side. The Solidity surface is limited to one on-chain contract
 (`OPContractsManagerV2`) and one Forge script (`DeployOPChain.s.sol`).
 
 1. **[`DeployOPChain` script](#deployopchain-script)**: broadcasts a permissionless deployment with a real
@@ -46,7 +48,8 @@ the work is on the `op-deployer` side and the Solidity surface is deliberately m
 
 ## Definitions
 
-See [Salt Mixer](./overview.md#salt-mixer) and [Anchor Block](./overview.md#anchor-block).
+See [Salt Mixer](./overview.md#salt-mixer), [Anchor Block](./overview.md#anchor-block),
+[Selected Prestate](./overview.md#selected-prestate) and [Fallback Prestate](./overview.md#fallback-prestate).
 
 ### ChainContracts struct
 
@@ -71,6 +74,7 @@ struct ChainContracts {
 ```
 
 The `op-deployer` consumes a subset of these addresses when building the genesis and rollup artifacts:
+
 - `l1CrossDomainMessenger`, `l1StandardBridge`, `l1ERC721Bridge` are inputs to `L2Genesis.s.sol` (the
   [L2 allocs](./overview.md#l2-allocs)).
 - `optimismPortal` and `systemConfig` are written into `rollup.json`.
@@ -100,10 +104,25 @@ struct FullConfig {
 }
 ```
 
-Once the PCD pipeline has built the `startingAnchorRoot` and the `absolutePrestate`, it fills these fields with real
-values instead of placeholders. `startingAnchorRoot` is the genesis [output root](./overview.md#output-root) with
-`l2SequenceNumber = 0`. The `absolutePrestate` is carried inside the permissionless `disputeGameConfigs`
-entry we are currently enabling.
+Once the PCD pipeline has built the [starting anchor root](./overview.md#starting-anchor-root) and the
+[selected prestate](./overview.md#selected-prestate), it fills these fields with real values instead of placeholders.
+The prestate is carried inside the permissionless `disputeGameConfigs` entry being enabled.
+
+### Supported initial game types
+
+`disputeGameConfigs` always holds six entries in a fixed order. Which of them may be `enabled` at initial deployment
+depends on whether the OPCM has the `SUPER_ROOT_GAMES_MIGRATION` dev feature on:
+
+| Index | Game type | Enabled at initial deployment |
+| --- | --- | --- |
+| 0 | `CANNON` | Never. Legacy slot |
+| 1 | `PERMISSIONED_CANNON` | Output root mode only |
+| 2 | `CANNON_KONA` | Output root mode only |
+| 3 | `SUPER_PERMISSIONED` | Super root mode only |
+| 4 | `SUPER_CANNON_KONA` | Super root mode only |
+| 5 | `ZK_DISPUTE_GAME` | Never at initial deployment. Reachable through the upgrade path |
+
+`startingRespectedGameType` must be one of the four selectable types and **must belong to the OPCM's mode**.
 
 ## Assumptions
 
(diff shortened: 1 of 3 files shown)


specs/experimental/predictive-chain-deployments/contracts.md
```diff
@@ -100,33 +104,53 @@ struct FullConfig {
...
-`l2SequenceNumber = 0`. The `absolutePrestate` is carried inside the permissionless `disputeGameConfigs`
-entry we are currently enabling.
+Once the PCD pipeline has built the [starting anchor root](./overview.md#starting-anchor-root) and the
+[selected prestate](./overview.md#selected-prestate), it fills these fields with real values instead of placeholders.
+The prestate is carried inside the permissionless `disputeGameConfigs` entry being enabled.
+
+### Supported initial game types
+
+`disputeGameConfigs` always holds six entries in a fixed order. Which of them may be `enabled` at initial deployment
+depends on whether the OPCM has the `SUPER_ROOT_GAMES_MIGRATION` dev feature on:
```
0xarktos, line 114: Since this is expected to be removed we could add a note to state that only super root games will eventually be supported and link the relevant monorepo issue

0xarktos (approved):
LGTM, one small comment

specs/experimental/predictive-chain-deployments/contracts.md
0xiamflux: We can track the update of the specs in the same issue just so the specs always reflect the current state of things, wdyt?

0xarktos: Sounds good, let's do that!

specs/experimental/predictive-chain-deployments/overview.md
```diff
@@ -47,32 +50,70 @@ Deploying a new chain today has a cyclic dependency:
...
+
+Every chain keeps its own registry in all three cases. In a super root deployment the registries hold the same value.
+
+### Selected Prestate
+
+The absolute prestate of the game type a chain deploys with. It is the Cannon hash of the initial MIPS machine state,
+produced by compiling the Kona fault-proof program with the embedded chain config. The hash is deterministic for a
+given chain config.
+
+A permissioned-only deployment carries no selected prestate.
```
OptimismBot: This is not what ships, and the same claim recurs at `op-deployer.md:195`, `:253`, `:266` and `:306`.

`DeployOPChain.s.sol` `checkInput` requires the prestate unconditionally, with no permissioned-only exemption:

```solidity
require(_i.disputeAbsolutePrestate.raw() != bytes32(0), "DeployOPChainInput: disputeAbsolutePrestate not set");
```

and `pipeline/opchain.go` `BuildDeployOPChainInput` fills it for a permissioned-only chain from the resolved proof params — `requirements.RequiresPrestate` is false, so `chainState.Prestate` never overwrites it:

```go
case embedded.GameTypePermissionedCannon:
    cannonAbsolutePrestate = proofParams.DisputeAbsolutePrestate
```

So a permissioned-only `CANNON_KONA`-family deployment does carry a real, non-zero selected prestate — the permissioned game's op-program prestate from the intent / standard defaults. What PCD leaves empty is only `chainState.Prestate`, the *PCD-committed* prestate. As written, an implementer reads this as "zero is acceptable here" and trips the script's `require`.

Suggest scoping the claim, e.g. "A permissioned-only deployment carries no PCD-committed prestate; its `disputeAbsolutePrestate` continues to come from the intent or the standard defaults, and MUST still be non-zero."

Worth one adjacent sentence for the super-root case too: for `SUPER_PERMISSIONED`, `cannonAbsolutePrestate` is zero (matching "Super root deployments have no fallback prestate") and `SuperPermissionedDisputeGameConfig` has no prestate field at all — yet `disputeAbsolutePrestate` must still be non-zero and is then entirely unused.

specs/experimental/predictive-chain-deployments/contracts.md
```diff
@@ -230,22 +289,41 @@ permissionless config can be included.
...
+  - `startingAnchorRoot.root` MUST be non-zero on any initial deployment, and its `l2SequenceNumber` MUST leave room
+    for a `uint64` successor.
+  - The `0xdead` placeholder root MUST be rejected when an enabled config is one of `CANNON`, `CANNON_KONA` or
+    `SUPER_CANNON_KONA`. Only a permissioned-only deployment may still carry the placeholder.
+  - An enabled config among those same three game types MUST carry a non-zero `absolutePrestate`, decoded from its
+    `gameArgs`.
+- MUST keep every other pre-existing validation intact:
   - exactly six configs in the fixed game-type order,
-  - the init-bond rules,
+  - the init-bond rules, including `SUPER_PERMISSIONED` carrying a zero init bond,
```
OptimismBot: This rule is new in PCD, not pre-existing, so it belongs in the new-checks list above rather than under "MUST keep every other pre-existing validation intact".

[`615b1c75dd`](https://github.com/ethereum-optimism/optimism/commit/615b1c75ddc4e9a7a73a8a9ea45dabf3e7b05201) — one of the two commits this PR verifies against — introduced it, splitting the previously unconditional "enabled ⇒ non-zero bond" rule in two:

```diff
-                    && _cfg.disputeGameConfigs[i].enabled && _cfg.disputeGameConfigs[i].initBond == 0
+            if (isSuperPermissionedGame && _cfg.disputeGameConfigs[i].initBond != 0) {
+                revert OPContractsManagerV2_InvalidGameConfigs();
+            }
...
+                !isSuperPermissionedGame && _cfg.disputeGameConfigs[i].enabled
+                    && _cfg.disputeGameConfigs[i].initBond == 0
```

The pre-existing rule was "any enabled config needs a non-zero bond"; the `SUPER_PERMISSIONED` carve-out (bond MUST be zero, enabled or not) arrived with PCD.

specs/experimental/predictive-chain-deployments/contracts.md
```diff
@@ -230,22 +289,41 @@ permissionless config can be included.
...
+- MUST allow the permissionless game type of the OPCM's mode to be `enabled` during an initial deployment.
+- MUST reject a game type from the other mode, and MUST keep `CANNON` and `ZK_DISPUTE_GAME` rejected at initial
+  deployment (see [Supported initial game types](#supported-initial-game-types)).
+- MUST validate the values a permissionless initial deployment now carries. These checks are new, before PCD both
+  values were always placeholders:
+  - `startingAnchorRoot.root` MUST be non-zero on any initial deployment, and its `l2SequenceNumber` MUST leave room
+    for a `uint64` successor.
+  - The `0xdead` placeholder root MUST be rejected when an enabled config is one of `CANNON`, `CANNON_KONA` or
+    `SUPER_CANNON_KONA`. Only a permissioned-only deployment may still carry the placeholder.
+  - An enabled config among those same three game types MUST carry a non-zero `absolutePrestate`, decoded from its
```
OptimismBot, line 298: The new-value-checks list is missing a fourth check that shipped alongside these three: an enabled `ZK_DISPUTE_GAME` must also carry a non-zero prestate, decoded as a `ZKDisputeGameConfig` rather than a `FaultDisputeGameConfig`.

```solidity
if (_cfg.disputeGameConfigs[i].enabled && isZkDisputeGame) {
    IOPContractsManagerUtils.ZKDisputeGameConfig memory zkGameConfig =
        abi.decode(_cfg.disputeGameConfigs[i].gameArgs, (IOPContractsManagerUtils.ZKDisputeGameConfig));
    if (zkGameConfig.absolutePrestate.raw() == bytes32(0)) {
        revert OPContractsManagerV2_InvalidGameConfigs();
    }
}
```

It is unreachable at initial deployment (`validForInitialDeploy` never admits ZK) but does apply on the upgrade path, where the "Reachable through the upgrade path" row of the game-type table sends readers. Right now the only ZK item in the list below is the dev-flag gate, so this check is unspecified. Note also that the ZK prestate is a verification key, not a Cannon MIPS hash — worth not folding it into the same bullet as the three Cannon-family types.

specs/experimental/predictive-chain-deployments/op-deployer.md
```diff
@@ -186,61 +208,116 @@ command can run.
...
+**Before broadcasting any transaction:**
+
+- MUST verify the current intent still agrees with the frozen snapshot, and that the dependency set still matches.
+- MUST re-download both artifact bundles by locator and verify their content digests against the frozen values.
+- MUST fail when the deployer or OPCM differs from the one recorded by `prepare`.
+- MUST halt a permissionless deployment when the prestate is unset or holds the reserved fallback value. The prestate
+  gate is conditional on a permissionless game type: a **permissioned-only** deployment carries no selected prestate
+  and proceeds without one.
+- MUST halt a permissionless deployment when the starting anchor root is unset or still holds the `0xdead` placeholder.
+- MUST re-check the predicted addresses by simulating the deployment against a fresh fork of L1 (pre-broadcast
```
OptimismBot, line 327: This pre-broadcast list reads as an exhaustive set of hard gates, which makes the omission of the deployment-window re-check misleading. `continue` does re-check it, but only warns and proceeds (`continue.go:255`):

```go
if uint64(*expected.GenesisTime) <= latest.Time {
    r.cfg.Logger.Warn(
        "committed genesis time has elapsed",
        "chainID", chainID.Hex(),
        "genesisTime", uint64(*expected.GenesisTime),
        "l1HeadTime", latest.Time,
    )
}
```

`prepare` is the only hard gate (`prepare.go:414`, "the deployment window has elapsed", comparing against the safe head rather than latest). That matters because [Deployment window](#deployment-window) states correctness "requires `genesis_time` to still be in the future when `OPCM.deploy()` is mined", and FM5's mitigation credits only `prepare` — so a `continue` run long enough after `prepare` broadcasts past an elapsed window with nothing but a log line. FM5 is Low/Low and the chain fills the gap with empty blocks, so warn-and-proceed looks deliberate; it just needs to be stated. Suggest a non-MUST bullet, e.g. "SHOULD warn, without halting, when the committed genesis time has already elapsed against the L1 head."

specs/experimental/predictive-chain-deployments/contracts.md
```diff
@@ -195,33 +221,66 @@ Using the same script for both ensures the predicted addresses are equal to the
...
+Only `msg.sender`, `saltMixer`, `l2ChainId` and the OPCM address determine the deployed addresses, so these
+substitutions are safe (see [aPCD-001](#apcd-001-opcm-address-determinism)).
+
 #### Broadcasting a permissionless deployment
 
 Today the script hard-codes a permissioned-only initial deployment:
 
-- It `require`s the initial game type to be `PERMISSIONED_CANNON` (or `SUPER_PERMISSIONED_CANNON` in super-root mode),
-  reverting with *"only PERMISSIONED_CANNON game type is supported for initial deployment"*.
+- It `require`s the input game type to be `PERMISSIONED_CANNON`, reverting with *"only PERMISSIONED_CANNON game type
```
OptimismBot: Two things are now stale here. Minor, but this section is the before/after narrative, so it is the most likely place for a reader to be misled about what is current.

The quoted revert string no longer exists anywhere in the tree. `_initialDeployGameSelection` now reverts with:

```solidity
require(permissionless_ || permissioned, "DeployOPChain: unsupported dispute game type");
require(
    GameTypes.isSuperGame(_disputeGameType) == _isSuperRoot,
    "DeployOPChain: dispute game type does not match OPCM super root mode"
);
```

And "Today the script hard-codes a permissioned-only initial deployment" (line 240) now describes history rather than the current tree. The OPCMv2 section below already handles this correctly with "Before PCD, enabling any permissionless game type ..." — matching that past-tense framing here would keep the two sections consistent.

OptimismBot (commented):
**COMMENT** — I checked the substantive claims on this page against `develop` (the shipped code, at `fc4ddf6`, which includes both [`615b1c75dd`](https://github.com/ethereum-optimism/optimism/commit/615b1c75ddc4e9a7a73a8a9ea45dabf3e7b05201) and [`9714d4dd2c`](https://github.com/ethereum-optimism/optimism/commit/9714d4dd2c77052fad62505598db3568b1458f7d)). The alignment is genuinely good — the `_assertValidFullConfig` snippet is verbatim-accurate, and I independently confirmed essentially every `MUST` on the op-deployer page against `prepare.go` / `prestate.go` / `continue.go` / `continue_verify.go` / `pipeline/opchain.go` / `pipeline/output_root.go`, including the removals: neither the `BLOCKHASH` guard nor the second-RPC cross-check exists anywhere in the tree, so dropping them is correct.

Five inaccuracies remain, none of them merge-blocking. The first is the only one I'd want fixed before merge, since it is the exact defect class this PR exists to remove and it appears in four places.

1. **"A permissioned-only deployment carries no selected prestate" contradicts the shipped code.** `DeployOPChain.checkInput` requires `disputeAbsolutePrestate != bytes32(0)` *unconditionally*, and `BuildDeployOPChainInput` passes `proofParams.DisputeAbsolutePrestate` for `PERMISSIONED_CANNON`. What is left empty is only `chainState.Prestate`, the PCD-committed prestate. Details inline; the claim appears at `overview.md:108`, `op-deployer.md:195`, `op-deployer.md:253`, `op-deployer.md:266` and `op-deployer.md:306`.

2. **`SUPER_PERMISSIONED`'s zero init bond is listed as pre-existing, but it shipped with PCD** (`contracts.md:323`) — see inline.

3. **The new ZK prestate check is missing from the OPCMv2 list** (`contracts.md:319`) — see inline.

4. **`continue` warns rather than fails on an elapsed deployment window** (`op-deployer.md:269`) — see inline.

5. **Stale "Today the script ..." framing and a revert string that no longer exists** (`contracts.md:243`) — see inline.

Everything else I spot-checked held: the `0xdead` placeholders line up across all three layers (`Constants.PLACEHOLDER_STARTING_ANCHOR_ROOT`, `ScriptConstants.DEFAULT_OUTPUT_ROOT().root`, `opcm.DefaultStartingAnchorRoot.Root`, and `opcm.PermissionedCannonFallbackPrestatePlaceholder` are all `0xdead00…00`); the dry-run placeholders are `standard.PlaceholderAddress = {0x01}` for the six roles and `{0x01}` as the permissionless sentinel root; the salt really is `keccak256(abi.encode(l2ChainId, bytes20(msg.sender) ++ saltMixer, contractName))`, so `aPCD-001`'s new mitigation is sound; `output_root.go` errors explicitly (rather than silently using an empty root) when the genesis header has no withdrawals root, so the Isthmus requirement is accurate; and super-root anchors do use the timestamp as `l2SequenceNumber` with every member seeded identically. All eight post-deploy validation items in the `continue` list are actually checked — the fallback game's *prestate* only via `StandardValidator` on the permissionless path, which is the only path where the spec claims it.

`markdownlint`, `doctoc` (no TOC drift), `cspell`, and both new design-doc links are clean.

Requested by @mds1 via Slack

specs/experimental/predictive-chain-deployments/op-deployer.md
```diff
@@ -186,61 +208,116 @@ command can run.
...
+  preflight) and abort on mismatch.
+- MUST verify the simulated broadcast is exactly one call, to the pinned OPCM, from the pinned deployer.
```
OptimismBot: This line is the clearest example of the pattern: it is a real security property — it is what stops a compromised or buggy script from smuggling extra broadcasts into the deployment — and it is stated only as a bullet in a command's procedure list, with no ID, no severity, and no impact.

The skill lists "Captures invariants - States properties that must always hold" and "Includes severity and impact" among its NON-NEGOTIABLE standards, and reserves the Invariants section for exactly that. Across the diff there are 48 added `MUST` statements and no new invariants; `iPCD-001`–`003` and `iOPD-001`–`004` are the same set as before the PR. An auditor who reads the Invariants sections to learn what to verify will not see any of the guarantees PCD actually shipped.

The three I would promote, because they are safety properties rather than procedure:

- **This one.** The simulated broadcast is exactly one call, to the pinned OPCM, from the pinned deployer.
- **`continue` MUST re-download both artifact bundles by locator and verify their content digests against the frozen values** (line 268). This is an artifact-integrity property, and it is not covered by `aOPD-003`, which assumes op-deployer *state* is trusted — the bundles are a separate trust surface, and the fact that a digest check exists at all says the spec does not extend that assumption to them.
- **`prestate` MUST reject the reserved fallback prestate value as a selected prestate** (line 251), together with `continue`'s matching halt on line 265. Committing the fallback placeholder as the real prestate makes the respected game unplayable from block 0, which is the same Critical impact `iOPD-003` already describes.

Each needs the template's `#### Impact` plus a severity. The remaining `MUST`s read as procedure and are fine where they are; the command sections can then reference the invariant IDs the way `continue` already references `aOPD-002`.

specs/experimental/predictive-chain-deployments/contracts.md
```diff
@@ -195,33 +221,66 @@ Using the same script for both ensures the predicted addresses are equal to the
...
+  enables **two** configs. `CANNON_KONA` enables `PERMISSIONED_CANNON`, `SUPER_CANNON_KONA` enables
+  `SUPER_PERMISSIONED`. The fallback carries the [fallback prestate](./overview.md#fallback-prestate).
+- MUST pass the real [starting anchor root](./overview.md#starting-anchor-root) instead of the placeholder, and MUST
+  reject the placeholder for a permissionless deployment.
 - MUST continue to support permissioned-only deployments.
 - MUST use a `from` address consistent with the [address-prediction](#address-prediction) dry-run, so the deployed
   addresses match the prediction (see
   [iOPD-001](./op-deployer.md#iopd-001-the-deployed-l1-system-matches-the-committed-artifacts)).
 
+#### Script input struct
```
OptimismBot: This section is the one place in the PR that goes furthest from the skill, which states in its second paragraph that a contract spec "is **not** a function-by-function API reference" and, under Writing Principles, to avoid "specific parameter names or types".

The section names the Solidity file (`scripts/libraries/Types.sol`), the struct, and three field names, then tabulates them by `Field | Exists | Meaning`. The `Exists` column is the part that will not age well: `new` and `repurposed` are only meaningful relative to the struct's pre-PCD shape, so the table is a changelog entry embedded in a spec. Once someone renames `cannonAbsolutePrestate` or adds a fourth field, the table is wrong and nothing in CI catches it.

The properties underneath are worth keeping, and they can be stated without the struct:

- A permissionless deployment supplies a starting anchor proposal, the selected prestate of its respected game type, and — for the output-root family — a distinct fallback prestate for the guardian fallback game.
- The selected prestate and the fallback prestate MUST be distinct, because they commit to different fault-proof programs.

The second bullet is currently the closing sentence of this section and is the only genuinely normative content in it. Note the same field-level framing appears in `### FullConfig` and `### ChainContracts struct` above, but those predate this PR — this section is new, so it is the one worth reconsidering now.

specs/experimental/predictive-chain-deployments/op-deployer.md
```diff
@@ -186,61 +208,116 @@ command can run.
...
-  - an **intent override** declared in the intent, which the command resolves.
-- MUST fail if neither source is provided.
-- If both are provided, it MUST fail when the two hashes disagree.
-- Re-running `prepare` with a different anchor MUST invalidate a previously-committed prestate and force a rebuild
-  before `continue` (see [FM6](#failure-modes)).
+`rollup.json`, and `depsets.json` rendered from the prepared state. The `prestate` command does not compile or hash
+anything itself. It only commits the built hash into op-deployer state, which `continue` reads.
+
+- MUST write the prestate hash to op-deployer state from any of three sources:
+  - the `--dispute-absolute-prestate` **command flag**, or the matching environment variable,
```
OptimismBot: This is a move away from the skill's standard rather than toward it, and the diff makes it explicit. Before:

> - a **command flag** that passes the hash directly, or
> - an **intent override** declared in the intent, which the command resolves.

After, the same two bullets became three carrying `--dispute-absolute-prestate`, `globalDeployOverrides.faultGameAbsolutePrestate` and `deployOverrides.faultGameAbsolutePrestate`. The skill's translation table asks for the opposite direction, and the old wording was already capability-focused.

The property that matters here is the resolution rule, not the spelling of the keys: the prestate is committed from exactly one of three sources — a command-line override, a global intent override, or a per-chain intent override — and the resolved value MUST be unambiguous, so two or more sources that disagree fail. That survives a flag rename; the current text does not.

The concrete names are genuinely useful, but they already live in [Developer experience](#developer-experience), which names `--dispute-absolute-prestate` and `faultGameAbsolutePrestate` — so this is duplication as well as coupling. Consolidating there keeps both audiences served.

Same applies more mildly to the state and intent keys the PR introduces elsewhere on this page: `interopDepSet` (line 344), `preparedDeployment` (line 256), `l1StartBlockHash` (lines 234, 349), and `opcmAddress` / `superchainConfigProxy` (line 214). `op-deployer.md` specifies an orchestrator rather than a contract, so some of this is fair — I would not pursue it beyond the `prestate` list.

specs/experimental/predictive-chain-deployments/contracts.md
```diff
@@ -230,22 +289,41 @@ permissionless config can be included.
...
+A real prestate now exists at deployment time.
+
+The game-type check is replaced by one scoped to the OPCM's mode:
 
 ```solidity
-// During initial deployment, only permissioned types can be enabled, because no
-// prestate exists for permissionless games.
-if (_isInitialDeployment && !isPermissioned && _cfg.disputeGameConfigs[i].enabled) {
+// Initial deployments must select game types compatible with the active mode.
+bool validForInitialDeploy = superRootGamesMigrationEnabled
```
OptimismBot: The skill's Writing Principles say to avoid "implementation patterns" and specific names, and this snippet now carries three local booleans (`validForInitialDeploy`, `superRootGamesMigrationEnabled`, and the `isSuperPermissionedGame` family it references but does not define) that exist only inside `_assertValidFullConfig`'s loop body.

It is also redundant. The five `MUST` bullets immediately below already say everything the snippet says, in the register the skill asks for, and they say it more completely — the snippet shows the mode gate but not the anchor-root or prestate value checks that the bullets cover. So the snippet adds coupling without adding information, and it is the part most likely to drift: renaming a local variable silently falsifies the spec.

Dropping the code block and keeping the prose would read cleanly, since the paragraph above it ("The game-type check is replaced by one scoped to the OPCM's mode") already carries the framing. If the concrete form is worth keeping for reviewers, a permalink to the function at a pinned commit does that without the staleness.

Acknowledged that the pre-PCD version of this section also had a snippet, so this is continuing an existing pattern rather than introducing one — but the replacement is twice as long, and the surrounding prose got good enough in this PR that the snippet is no longer carrying weight.

specs/experimental/predictive-chain-deployments/op-deployer.md
```diff
@@ -155,22 +173,26 @@ genesis cannot be proven, and fault proofs are broken from block 0.
...
 #### Impact
 
 **Severity: Critical**
 
 A mismatched prestate makes the two sides of a dispute disagree on the starting state. Fault proofs are broken from
-block 0, with no on-chain guard to catch it at deploy time. Recovering from this is expensive. See
-[FM4](#failure-modes) and [FM6](#failure-modes).
+block 0. Recovering from this is expensive.
+
+OPCM rejects a **zero** prestate on an enabled Cannon-family game (see
```
OptimismBot: Good content, wrong section. The template defines `#### Impact` as what happens if the invariant is violated, and the first paragraph does exactly that ("Fault proofs are broken from block 0. Recovering from this is expensive."). This second paragraph is about *detection* — which guard catches the violation, and which cannot:

> OPCM rejects a **zero** prestate on an enabled Cannon-family game [...] It cannot reject a wrong one, since any nonzero value passes. Reproducing the prestate before `continue` is the only thing that catches a prestate that is set but incorrect.

Under the template that reasoning belongs in a Mitigations block, and there is a natural home for it: `aOPD-004: The prestate build is reproducible` already has a Mitigations list, and "reproducing the prestate before `continue` is the only thing that catches a set-but-incorrect prestate" is precisely the argument for why that assumption has to hold. [Deployment guards](#deployment-guards) is the other candidate, and it already carries the companion sentence about the on-chain checks.

Minor, and I would not hold the PR for it — but the observation that the on-chain guard is a liveness check rather than a correctness one is the most load-bearing sentence added to this page, and burying it in an Impact block is where a reader is least likely to find it.

OptimismBot (commented):
**COMMENT** — follow-up pass against the [`contract-specs` skill](https://github.com/ethereum-optimism/op-claude-marketplace/blob/main/plugins/op-eng-skills/skills/contract-specs/SKILL.md) (and the [`professional-docs`](https://github.com/ethereum-optimism/op-claude-marketplace/blob/main/plugins/op-eng-skills/skills/professional-docs/SKILL.md) standards it inherits), which I had not seen when I posted [my first review](https://github.com/ethereum-optimism/specs/pull/930#pullrequestreview-5068662996). The correctness findings there still stand; these are separate and all structural.

Two things the PR gets right against the skill, worth saying first. The four new Definitions (Super Root, Starting Anchor Root, Selected Prestate, Fallback Prestate) are exactly the "define domain-specific terms" work the skill asks for, and Selected/Fallback Prestate are deliberately *not* the code's identifiers — that is the implementation-agnostic naming the skill wants. The canonical template is also intact: every assumption carries Mitigations, every invariant carries Impact with a Severity, and the `a{PREFIX}-###` / `i{PREFIX}-###` IDs are well-formed. `professional-docs` is clean too: no emoji or decorative Unicode in the diff.

The findings below all trace to the skill's two NON-NEGOTIABLE standards — "captures invariants" and "is implementation-agnostic".

1. **The PR adds 48 normative `MUST`s and zero invariants.** The Invariants sections are unchanged (`iPCD-001`–`003`, `iOPD-001`–`004`); every new requirement landed as a prose bullet under `prepare` / `prestate` / `continue` / `Change Specification`. Several are genuine safety properties. Detail and candidates inline.

2. **`#### Script input struct` is a field-level API reference** (`contracts.md:266`) — the skill opens by saying a spec "is **not** a function-by-function API reference".

3. **The `prestate` section trades capability language for concrete CLI and config identifiers** (`op-deployer.md:247`), moving away from the skill's translation table — and this one is a visible before/after in the diff.

4. **The expanded Solidity snippet in OPCMv2 config validation** (`contracts.md:300`) now carries local variable names, and is redundant with the prose `MUST`s directly beneath it.

5. **`iOPD-003`'s Impact section now carries mitigation and detection analysis** (`op-deployer.md:187`) rather than consequences of violation.

None of these are merge blockers, and 4 and 5 are close to judgment calls. 1 is the one I would act on: it is the difference between an auditor reading the Invariants section and seeing the properties, versus having to read the whole pipeline narrative to find them.

One caveat on scope. The skill is written for a single-contract spec, and these are three pages spanning a Forge script, an on-chain contract, and a Go orchestrator. The standards land hardest on `contracts.md`; for `op-deployer.md` some implementation coupling is legitimately unavoidable, so weigh 3 accordingly.

Requested by @mds1 via Slack

mds1 (approved):

PR #933, opened 2026-08-21
specs: replace flashblocks with subblocks

### LINKED ISSUE #929: replace flashblocks spec with subblocks spec
We should only the public facing parts of subblocks. See the docs here:

https://docs.optimism.io/op-stack/features/subblocks
https://docs.optimism.io/app-developers/guides/transactions/subblocks-and-gas-usage

For example we should sate the state root doesn't need to be / isn't populated. And cover the wire format and JSON-RPC. It can be very lightweight.

bitwiseguy:
## Summary

- Replace the detailed Flashblocks specification with a minimal, consumer-facing Subblocks specification.
- Document the WebSocket payload format, sequencing guarantees, JSON-RPC behavior, and transaction inclusion implications.
- Clarify that Subblocks are a backwards-compatible successor to Flashblocks, except that `state_root`, `block_hash`, `withdrawals_root`, and `withdrawals` are zeroed.
- Update the specs book summary and spelling dictionary.

## Related Issues
* Closes https://github.com/ethereum-optimism/specs/issues/929


specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Consumer guarantees
+
+Within one in-progress block, subblocks are append-only:
+
+- A sequence starts at index `0` and subsequent indices increase by one.
+- All subblocks in the sequence have the same `payload_id` and block number.
+- Transactions from each subblock are appended in stream order. A consumer obtains the in-progress transaction list by
+  concatenating each subblock's `diff.transactions`.
+- The first subblock contains the block's deposited transactions and any other sequencer transactions that execute
```
geoknee, line 40: You should be able to find a crosslink to another part of the specs for this e.g. https://specs.optimism.io/glossary.html?highlight=depos#deposited-transaction

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+  before mempool transactions. Later subblocks add mempool transactions.
+
+If the in-progress block is abandoned, the next sequence supersedes it. Consumers MUST NOT treat a subblock as safe or
```
geoknee, line 43: What does "sequence" mean here? Do you mean the next payload in the sequence?

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+| `block_hash` | `0x0000000000000000000000000000000000000000000000000000000000000000` |
+| `withdrawals_root` | `0x0000000000000000000000000000000000000000000000000000000000000000` |
+| `withdrawals` | `[]` |
+
+These fields remain on the wire only for backwards compatibility. Consumers MUST treat them as unavailable and MUST
+NOT use them as commitments to the in-progress block or state. In particular, a zero `state_root` or `block_hash` is
+not the value of the eventual sealed block.
+
+A consumer that needs the in-progress state executes the streamed transactions against the parent state. The state
+root and block hash become authoritative only after the L2 block is sealed.
```
geoknee, line 141:
```suggestion
A consumer that needs the in-progress state may reconstruct it by executing the streamed transactions against the parent state. The state
root and block hash become authoritative only after the L2 block is sealed.
```
Can consumers tell that the block is sealed only by consuming the subblocks stream? If so, how do they do that?

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+
+## JSON-RPC
+
+Most applications consume subblocks through a subblock-aware RPC node instead of subscribing to the stream directly.
```
geoknee, line 145:
```suggestion
Applications may consume subblocks via a subblock-aware RPC node instead of subscribing to the stream directly.
```

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+- `eth_getCode`
+- `eth_getStorageAt`
+- `eth_getTransactionCount`
+
+Successive `eth_getBlockByNumber("pending", ...)` calls during the same L2 block may return an expanding transaction
+list as new subblocks arrive.
+
+Methods that identify a transaction directly, including `eth_getTransactionByHash` and
+`eth_getTransactionReceipt`, also search preconfirmed subblock transactions before returning `null`. Their responses
+use the standard Ethereum JSON-RPC shapes, but block-derived fields are provisional until the block is sealed.
```
geoknee, line 164:
```suggestion
Methods that identify a transaction directly, including `eth_getTransactionByHash` and
`eth_getTransactionReceipt`, may also return preconfirmed subblock transactions. Their responses
use the standard Ethereum JSON-RPC shapes, but block-derived fields are provisional until the block is sealed.
```

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+
+The fields have the following semantics:
+
+- `payload_id` identifies one payload build and is constant throughout the sequence.
+- `index` is a JSON number identifying the subblock's position in the sequence.
+- `base` contains immutable block properties and is present only at index `0`.
+- `diff.transactions` contains only the EIP-2718 encoded transactions added by this subblock.
+- `diff.gas_used`, `diff.receipts_root`, and `diff.logs_bloom` describe the cumulative in-progress block after applying
+  this subblock.
+- `diff.blob_gas_used`, when present, is the cumulative blob gas used by the in-progress block.
```
geoknee, line 118: Good to link to this section here https://specs.optimism.io/protocol/jovian/exec-engine.html#da-footprint-block-limit, since this field doesn't actually mean blob gas used (no blob gas on L2 anyway).

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+- `diff.post_exec_tx`, when present, is the latest cumulative post-execution transaction. It is carried separately from
```
geoknee, line 119: I would state "from the lagoon fork" https://specs.optimism.io/protocol/lagoon/post-exec.html

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+
+## WebSocket stream
+
+The sequencer publishes subblocks through a WebSocket endpoint. Each subblock is a JSON object in a WebSocket text
+frame. Field names use `snake_case`; byte strings, addresses, hashes, transactions, and quantities use their standard
+Ethereum JSON encodings unless stated otherwise below.
+
+The JSON shape deliberately retains Flashblocks-era names and fields for wire compatibility. Implementations commonly
+represent the envelope as `OpFlashblockPayload` and its `diff` as either `OpFlashblockPayloadDelta` or the legacy
+`ExecutionPayloadFlashblockDeltaV1` type.
```
geoknee, line 54: So the shape in the websocket stream is important, but I don't think the implementation bits are. IIUC we would be free to rename these types in the implementation with no user impact.
```suggestion
The JSON shape deliberately retains Flashblocks-era names and fields for wire compatibility.
```

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+    fee_recipient: Bytes20
+    prev_randao: Bytes32
+    block_number: uint64
+    gas_limit: uint64
+    timestamp: uint64
+    extra_data: Bytes
+    base_fee_per_gas: uint256
+}
+
+ExecutionPayloadFlashblockDeltaV1 {
```
geoknee, line 90:
```suggestion
Payload {
    payload_id: Bytes8
    index: uint64
    base: Optional[PayloadBase]
    diff: Delta
    metadata: Metadata
}

PayloadBase {
    parent_beacon_block_root: Bytes32
    parent_hash: Bytes32
    fee_recipient: Bytes20
    prev_randao: Bytes32
    block_number: uint64
    gas_limit: uint64
    timestamp: uint64
    extra_data: Bytes
    base_fee_per_gas: uint256
}

Delta {
```

I _think_ we can just give these whatever names we like here and don't need to use the fb terminology. They aren't visible to consumers.

Here's a frame from the ws stream:
```json
{
  "payload_id": "0x044ced4279a1aa17",
  "index": 10,
  "diff": {
    "state_root": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "receipts_root": "0xebfdb43d2893b5260feeee55047863679b07d79de806c02c29b014a2fa924170",
    "logs_bloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "gas_used": "0xb496",
    "block_hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "transactions": [],
    "withdrawals": [],
    "withdrawals_root": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "blob_gas_used": "0x0"
  },
  "metadata": {
    "block_number": 1281784,
    "new_account_balances": {
      "0x0000f90827f1c53a10cb7a02335b175320002935": "0x0",
      "0x000f3df6d732807ef1319fb7b8bb8522d0beac02": "0x0",
      "0x4200000000000000000000000000000000000015": "0x0",
      "0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001": "0x0"
    },
    "receipts": {}
  }
}
```

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+frame. Field names use `snake_case`; byte strings, addresses, hashes, transactions, and quantities use their standard
+Ethereum JSON encodings unless stated otherwise below.
+
+The JSON shape deliberately retains Flashblocks-era names and fields for wire compatibility. Implementations commonly
+represent the envelope as `OpFlashblockPayload` and its `diff` as either `OpFlashblockPayloadDelta` or the legacy
+`ExecutionPayloadFlashblockDeltaV1` type.
+
+### Sequence
+
+The first payload in a sequence has `index` equal to `0` and includes `base`. Later payloads increment `index` and omit
```
geoknee, line 58:
```suggestion
The first `Payload` in a sequence has `index` equal to `0` and includes `base`. Later payloads increment `index` and omit
```

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+  concatenating each subblock's `diff.transactions`.
+- The first subblock contains the block's deposited transactions and any other sequencer transactions that execute
+  before mempool transactions. Later subblocks add mempool transactions.
+
+If the in-progress block is abandoned, the next sequence supersedes it. Consumers MUST NOT treat a subblock as safe or
+final chain data.
+
+## WebSocket stream
+
+The sequencer publishes subblocks through a WebSocket endpoint. Each subblock is a JSON object in a WebSocket text
```
geoknee, line 48:
```suggestion
The sequencer publishes subblocks through a WebSocket endpoint. Each subblock is a JSON `Payload` object (defined below) in a WebSocket text
```

specs/protocol/subblocks.md
geoknee: We should also consider whether we want to use `Subblock` instead of `Payload`. Thoughts?

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+the block gas limit, the transaction remains pending for a later block.
+
+Applications SHOULD NOT assume that a transaction with a gas limit below the full block gas limit can be included in
+the next subblock.
+
+## Backwards compatibility
+
+Subblocks are the backwards-compatible successor to Flashblocks. The WebSocket envelope, JSON field names, and
+JSON-RPC behavior are unchanged, and legacy names such as `OpFlashblockPayload`,
+`ExecutionPayloadFlashblockDeltaV1`, and `--flashblocks-url` remain in use so existing consumers can continue to decode
```
geoknee, line 180: See above, I'm not sure these names are actually in the wire format?

geoknee: We can of course mention that the wire format remains compatible with these legacy types defined at https://github.com/flashbots/rollup-boost/blob/main/specs/flashblocks.md (and provide the link).

geoknee (commented):
Looks good but I think we can tighten up the names of the types we are using and how they relate to the legacy types.

Also, the removed specs have a couple of sections which I think may be worth retaining:

* https://specs.optimism.io/protocol/flashblocks.html#flashblock-validity-rules
* https://specs.optimism.io/protocol/flashblocks.html#flashblock-system-invariants

But I'm keen to know your thoughts on that.

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+}
+
+ExecutionPayloadFlashblockDeltaV1 {
+    state_root: Bytes32
+    receipts_root: Bytes32
+    logs_bloom: Bytes256
+    gas_used: uint64
+    block_hash: Bytes32
+    transactions: List[Bytes]
+    withdrawals: List[Withdrawal]
```
geoknee, line 97: We aren't defining the `Withdrawal` type here. Can we do so or provide a link?

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,185 @@
...
+    withdrawals_root: Bytes32
+    blob_gas_used: Optional[uint64]
+    post_exec_tx: Optional[Bytes]
+}
+
+Metadata {
+    block_number: uint64
+    new_account_balances: Map[Bytes20, uint256]
+    receipts: Map[Bytes32, Receipt]
```
geoknee, line 106: Ideally we can define `Receipt` as well, at least via a link (to the ethereum spec) or via https://specs.optimism.io/glossary.html?highlight=receipt#receipt

specs/protocol/subblocks.md
bitwiseguy: I meant the sequence of subblocks. Changed to use the term "in-progress block" since "sequence" is a loaded term

specs/protocol/subblocks.md
bitwiseguy: No you can't tell if a block is sealed solely by consuming subblocks. Updated this paragraph to make that clear

specs/protocol/subblocks.md
bitwiseguy: I went with `Subblock` instead of `Payload`

specs/protocol/subblocks.md
bitwiseguy: Used `Subblock` here instead of `Payload`

bitwiseguy:
> Also, the removed specs have a couple of sections which I think may be worth retaining:
https://specs.optimism.io/protocol/flashblocks.html#flashblock-validity-rules
https://specs.optimism.io/protocol/flashblocks.html#flashblock-system-invariants

I added the following two bullets to the "Consumer guarantees" section:
```
- The sequencer MUST NOT publish distinct subblocks with the same `payload_id` and `index`.
- The cumulative in-progress block after each subblock MUST satisfy all applicable OP Stack execution rules.
```

specs/protocol/subblocks.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Sequence
+
+The first `Subblock` for an in-progress block has `index` equal to `0` and includes `base`. Each subsequent `Subblock`
+increments `index` by one and omits `base`. The `payload_id`, `base`, and `metadata.block_number` identify the
+in-progress L2 block.
+
+A consumer joining the stream after index `0` cannot reconstruct the complete in-progress block. It SHOULD ignore
+payloads until it receives the next index `0`. A consumer SHOULD also discard its current sequence if an index is
+skipped, duplicated, or paired with a different payload ID or block number.
```
geoknee:
```suggestion
### Sequence

The ordered set of `Subblocks` for a given in-progress block is known as a _sequence_.
The first `Subblock` in the sequence has `index` equal to `0` and includes the`base` field. Each subsequent `Subblock`
increments `index` by one and omits `base`. The `payload_id`, `base`, and `metadata.block_number` identify the
in-progress L2 block.

A consumer joining the stream after index `0` cannot reconstruct the complete in-progress block. It SHOULD ignore
payloads until it receives the next index `0`. A consumer SHOULD also discard its current sequence if an index is
skipped, duplicated, or paired with a different payload ID or block number.
```

I think a more formal definition of "sequence" like this will really help readability, since we refer to it quite a bit below.

geoknee:
> > Also, the removed specs have a couple of sections which I think may be worth retaining:
> > https://specs.optimism.io/protocol/flashblocks.html#flashblock-validity-rules
> > https://specs.optimism.io/protocol/flashblocks.html#flashblock-system-invariants
> 
> I added the following two bullets to the "Consumer guarantees" section:
> 
> ```
> - The sequencer MUST NOT publish distinct subblocks with the same `payload_id` and `index`.
> - The cumulative in-progress block after each subblock MUST satisfy all applicable OP Stack execution rules.
> ```

I think this is fine for now. There really aren't many guarantees with subblocks or flashblocks so it's best to not over promise here.

geoknee (commented):
👏 Thanks @bitwiseguy !

geoknee (approved):

specs/protocol/subblocks.md
bitwiseguy: I decided to use "Ordering" instead of "Sequence" to avoid the overloaded term and confusion with "sequencer". Made that renaming change in multiple places

PR #939, opened 2026-09-07
fault-proof: Move super root dispute game out of the interop section

ajsutton:
Super root dispute games ship before interop activates. A chain proposes super roots with a dependency set of one chain, and the consolidation step then has no executing message to validate. Each chain keeps its own `DisputeGameFactory` and `AnchorStateRegistry` until it joins the op-governed dependency set.

The fault proof page moves from `specs/interop/` to `specs/fault-proof/stage-one/super-fault-dispute-game.md`, with a `book.toml` redirect so the published URL still resolves. The page drops the statements that make the game interop specific and states when consolidation has nothing to validate.

The `l2-receipts` and `l2-block-data` pre-image hints stay gated on interop. Both are sent only from consolidation, which the client skips entirely when interop is inactive. The chain ID requirement on the other L2 hints now depends on the state transition the program verifies, not on the interop hardfork.

`OptimismPortal` gains a note on `migrateToSharedDisputeGame` beside the shared `ETHLockbox` migration. This also corrects two references to `migrateToSuperRoots`, which no longer exists.

Open as a draft for terminology review.


specs/fault-proof/stage-one/super-fault-dispute-game.md
```diff
@@ -34,6 +34,15 @@ pre-interop [fault dispute game][fault-dispute-game] with some modifications:
 - The L2 block number preimage oracle local key now always provides the timestamp of the proposal.
 - The L2 chain ID preimage oracle local key is no longer used and is never populated by the dispute game.
 
+A super game type is the respected game type in place of an output root game type. A permissionless chain uses the
+`SUPER_CANNON_KONA` game type. A permissioned chain uses the `SUPER_PERMISSIONED` game type.
```
ajsutton:
```suggestion
```

specs/fault-proof/stage-one/super-fault-dispute-game.md
```diff
@@ -80,13 +92,17 @@ For each step, the valid post state `TransitionState` is calculated by the algor
...
+  not
   checked at this stage.
   - No block is derived if `Step` is greater than the number of chains in the super root.
   - If the derivation process is unable to derive the next L1 block because the L1 head is reached, the post state is
     the [invalid state](#invalid-state)
 - Increment `Step`
 
+A game for a single chain therefore derives a block at `Step = 0` and derives no block for the remaining 126 steps.
+Those steps only increment `Step`.
+
```
ajsutton:
```suggestion
```

specs/fault-proof/stage-one/super-fault-dispute-game.md
```diff
@@ -99,10 +109,14 @@ recursively replacing any blocks with executing messages that became invalid bec
 The post state is defined as a super output where `timestamp` is the `SuperOutput` timestamp + 1, and the output roots
 are set to the output roots of the validated blocks (including any required replacements).
 
+When interop is not active, the `CrossL2Inbox` predeploy has no code, so a block contains no
+[executing message](../../interop/messaging.md#executing-messages). This step then has nothing to validate. It replaces
+no block, and the post state carries the pending progress through unchanged.
```
wwared: **Claude:** **Should-fix — distinguish the predeploy proxy from its implementation.**

The statement that `CrossL2Inbox` “has no code” before interop is incorrect. The [predeploy specification](https://github.com/ethereum-optimism/specs/blob/32db693586ac51299e24d8fe2692c658036fb60a/specs/protocol/predeploys.md#L42-L44) requires proxy bytecode throughout the first 2,048 addresses of the predeploy namespace, including the inbox at `0x4200000000000000000000000000000000000022`. Genesis installs that proxy independently of interop activation; installing the inbox implementation is what is gated.

The conclusion that there are no executing messages is correct, but the distinction matters: the address has nonempty code, and a call to the uninitialized proxy reverts rather than behaving like a call to an empty account. Please describe the absent implementation instead of an empty predeploy account.

Suggested replacement:

```suggestion
When interop is not active, the `CrossL2Inbox` implementation is not installed behind its predeploy proxy, so a block
contains no [executing message](../../interop/messaging.md#executing-messages). Consolidation therefore replaces no block,
and the post state carries the pending progress through unchanged.
```

specs/interop/optimism-portal.md
```diff
@@ -58,6 +66,35 @@ Returns the `ProxyAdmin` owner that manages the `ETHLockbox`.
...
+- MUST revert if the system is paused
+- MUST revert if the new `AnchorStateRegistry` is the same as the current one
+- MUST revert if either address is zero
+- MUST revert if the new `ETHLockbox` has not authorized this portal
+- MUST emit a `PortalMigrated` event with the old and new `ETHLockbox` and `AnchorStateRegistry`
+- SHOULD be called atomically with [`ETHLockbox.migrateLiquidity`](./eth-lockbox.md#migrateliquidity) in the same
+  transaction, or the portal may not be able to unlock enough ETH to finalize withdrawals
+
+The migration is one way. It invalidates every withdrawal proof that is already submitted, so users MUST prove those
+withdrawals again.
```
wwared: **Claude:** **Should-fix — make the intended exactly-once usage explicit.**

`migrateToSharedDisputeGame` is designed to be called exactly once, and that is the expectation for chain operators. Although the implementation permits multiple calls, it is not designed for repeated migration. The wording “The migration is one way” can be read as a contract-enforced guarantee rather than an operational requirement.

Please state the intended exactly-once usage explicitly and distinguish it from what the function enforces. This should not suggest that repeat migrations are an intended use of the API. Keep the warning that joining the shared dispute-game system invalidates proofs against the previous factory.

Suggested replacement:

```suggestion
`migrateToSharedDisputeGame` is designed for a one-time migration. Chain operators are expected to call it exactly once
when joining the op-governed dependency set. Although the function permits repeated calls, it is not designed for repeated
migration. Switching to the shared `DisputeGameFactory` invalidates withdrawal proofs against games from the previous
factory, so users MUST prove those withdrawals again.
```

wwared (approved):
LGTM. Two non-blocking nits/suggestions below, AI did a better job explaining them than I could

PR #943, opened 2026-09-09
opcm: add OPContractsManagerMigrator spec

### LINKED ISSUE #937: specs: interop specify OPContractsManagerMigrator
There is no spec for `OPContractsManagerMigrator` or other spec that mentions the `migrate()`. Add one to following the standard contract spec process, covering `OPContractsManagerMigrationValidator` as well.

0xiamflux:
## Description
Specifies `OPContractsManagerMigrator` and `OPContractsManagerMigrationValidator`, the OPCM
components that merge independent OP Stack chains into a single interop set.

Closes #937

### DIFF
--- a/specs/experimental/contracts/L1/opcm-migrator.md
+++ b/specs/experimental/contracts/L1/opcm-migrator.md
@@ -485,13 +485,15 @@ rejection.
 
 ### iMIG-013: Custom gas token chains are not migrated
 
-A chain using a custom gas token must not join an [Interop Set](#interop-set), and must not have its
-portal-held balance migrated into any shared `ETHLockbox`.
+A [Member Chain](#member-chain) must not use a custom gas token. Migration must reject a chain whose
+`SystemConfig` has the custom gas token feature enabled before any state is mutated.
 
 #### Impact
 
 **Severity: Critical**
 
-A custom gas token chain's portal holds that token, not ETH. Pooling it into an `ETHLockbox` mixes
-an unrelated asset into the set's ETH liquidity, where it backs withdrawals it does not correspond
-to and cannot be recovered.
+A custom gas token chain's native asset is not ETH. Joining a set enables the interop feature on the
+member, which activates the [interop ETH predeploys](../../../interop/eth-bridging.md) on its L2. A
+cross-chain ETH transfer burns the sending chain's native asset and mints ETH on the receiving
+chain. That ETH is withdrawable from the set's single shared `ETHLockbox`, so every unit such a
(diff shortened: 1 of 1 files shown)


specs/experimental/contracts/L1/opcm-migrator.md
```diff
@@ -0,0 +1,497 @@
...
+    RA["A: old DGF / ASR / ETHLockbox"]
+    RB["B: old DGF / ASR / ETHLockbox"]
+  end
+  PA --> LB
+  PA --> ASR
+  PB --> LB
+  PB --> ASR
+  ASR --> DGF
+  ASR --> LB
+  WETH --> LB
```
Joxess: This part is not reflecting the fact that the delayedweth is reused from chain A

Joxess: It would be nice to have a short explanation for each arrow (pause, guardian, eth custody) etc

Joxess: makes sense to add the L1CDM / L1SB and SuperchainConfig? it is true that it grows the mermaid, but I think it would be useful to contain the whole pic

Joxess: Like delayedweth ruses chain A; while RB["B: old DGF / ASR / DWETH / ETHLockbox"]

specs/experimental/contracts/L1/opcm-migrator.md
```diff
@@ -0,0 +1,497 @@
...
+  RA -. pause .-> LB
+  RB -. pause .-> LB
+```
+
+### Pause Identity
```
Joxess: pause identifier?

specs/experimental/contracts/L1/opcm-migrator.md
```diff
@@ -0,0 +1,497 @@
...
+A chain using a custom gas token must not join an [Interop Set](#interop-set), and must not have its
+portal-held balance migrated into any shared `ETHLockbox`.
+
+#### Impact
+
+**Severity: Critical**
+
+A custom gas token chain's portal holds that token, not ETH. Pooling it into an `ETHLockbox` mixes
+an unrelated asset into the set's ETH liquidity, where it backs withdrawals it does not correspond
+to and cannot be recovered.
```
Joxess: Actually cgt chains don't hold any token/value directly. I believe a cgt chain can potentially be migrated, but the decision is, for now, out of scope because it needs to be addressed with other kinds of workarounds

PR #944, opened 2026-09-14
lagoon: specify the span batch encoding of post-exec transactions

geoknee:
Closes [protocol-team#318](https://github.com/ethereum-optimism/protocol-team/issues/318).

## The gap

The post-exec transaction type `0x7D` is fully specified as an EIP-2718 envelope in `specs/protocol/lagoon/post-exec.md`. Its span batch representation was specified nowhere — the word "span" did not appear anywhere under `specs/protocol/lagoon/`.

Delta lists `tx_datas` encodings for legacy, type 1 and type 2. Isthmus adds type 4 with an exact field order. Nothing added `0x7D`.

Delta's generic sentence — that `tx_datas` matches "the encoding of the fields as in the EIP-2718 format of the `TransactionType`" — does not close it. It settles only `tx_datas`, and the document does not treat it as sufficient for a new type anyway, since Isthmus still added an explicit row. It settles none of the other slots. `contract_creation_bits` is defined as marking a contract creation, which a post-exec transaction is not, yet both clients set the bit to `1` so that no `tx_tos` entry is consumed. Nothing authorised that.

This is a specification gap, not a client bug. The two clients agree byte for byte. No document stated the layout they chose.

## What this adds

A new `specs/protocol/lagoon/derivation.md`, following the Isthmus precedent for adding a transaction type to span batches:

- **`tx_datas` row.** `0x7D ++ rlp_encoded_payload` — the payload RLP list verbatim, so the element is the transaction's whole EIP-2718 encoding rather than a reduced one. Unlike every other type, no fields are omitted, because there are none to omit.
- **Every transposed slot.** `contract_creation_bits` = 1, no `tx_tos` entry, `y_parity_bits` = 0, `tx_sigs` r = s = 0, `tx_nonces` = 0, `tx_gases` = 0, no `protected_bits` entry.
- **Reconstruction**, and why the batch/safe-block overlap check is well defined for these blocks.

Also wired into `SUMMARY.md` and the Lagoon overview, with a cross-link from `post-exec.md`.

## The three questions in the issue

1. **Which `tx_datas` form?** The verbatim EIP-2718 body, as both clients use.
2. **`contract_creation_bits` for a type with no `to`?** `1`. The document explains why this is not a contradiction of Delta's wording: what the bit governs is whether the transaction consumes a `tx_tos` entry. For every pre-Lagoon type, "is a contract creation" and "has no `to`" are the same condition, so Delta states the former; post-exec is the first type where they diverge, and the operative reading is the latter.
3. **Must the unused slots be zero?** Yes, and strictly, in both directions: a batcher MUST zero them, and a decoder MUST verify they are zero and reject the batch otherwise. The same applies to a `contract_creation_bits` bit of `0` for a post-exec transaction.

That third answer changed during review. I first wrote the decoder side permissively — ignore the values, never reject — reasoning that the clients discard those fields during reconstruction, so no value can change the derived blocks, and that turning a client convention into a drop rule is usually a consensus change in disguise. @geoknee pushed for strictness and is right. The general caution does not bite here, because `0x7D` cannot legally appear in a batch before Lagoon and no chain in the superchain-registry sets `lagoon_time` — there is no posted batch for a stricter rule to reinterpret. Meanwhile permissiveness has a real cost: it makes a span batch's encoding malleable, giving one sequence of L2 blocks unboundedly many valid encodings that differ in bytes a batcher picks freely, with `tx_sigs` alone reserving 64 unused bytes per transaction to pick them in.

The issue also suggested the overlap check is non-deterministic without a zero rule. It isn't: `checkSpanBatchOverlap` compares reconstructed transaction bytes, and reconstruction discards those slots. The document states that positively rather than repeating the concern.

## Client changes are required before Lagoon activates

This is now a spec that leads the implementations rather than one that records them. Neither client rejects what this document says a decoder MUST reject:

- **op-node** — `fullTxs` (`span_batch_txs.go`) consumes a `tx_tos` entry for a `0` bit and `convertToFullTx` then discards it; no slot values are checked.
- **kona** — `full_txs` (`batch/transactions.rs`) and `to_full_tx_bytes` (`batch/tx_data/wrapper.rs`) do the same.

The encoder side already complies on both sides, and op-batcher reaches `addPostExecTx` through `AppendSingularBatch`, so no batcher change is needed.

One caveat that decides how the client work should be done: these slots are only reachable at reconstruction time, so the check lands in `fullTxs` / `full_txs` — the exact path of the open [protocol-team#222](https://github.com/ethereum-optimism/protocol-team/issues/222), where a non-critical error there makes op-node drop only the batch and continue the channel while kona drops the whole remaining channel. Adding this rule naively would add a new batcher-triggerable trigger for that divergence. Either #222 lands first, or the rejection surfaces as a batch-queue validity verdict rather than a decode failure.

## Verification

Every normative claim was checked against both clients in both directions, then re-checked by an adversarial review pass:

- op-node — `span_batch_txs.go` (`addPostExecTx`, `AddTxs`, `fullTxs`, `recoverV`), `span_batch_tx.go` (`spanBatchPostExecTxData`, `encodeTyped`, `convertToFullTx`), `op-core/types/post_exec_tx.go`, `batches.go` (overlap check), `channel_in_reader.go` (drop path).
- kona — `batch/tx_data/wrapper.rs`, `batch/transactions.rs`, `batch/raw.rs`, `rust/op-alloy/.../post_exec.rs`.

`markdownlint`, `doctoc`, `cspell` and the filename check all pass; every internal anchor resolves.

## Out of scope

The batch queue drop rule for future types above `2` still needs a Lagoon amendment — that is [protocol-team#319](https://github.com/ethereum-optimism/protocol-team/issues/319). It applies to singular batches too, where no encoding question arises, so it belongs in its own change. This document marks the boundary and notes the amendment is still owed, rather than assuming a rule that does not exist yet.

One adjacent observation, left alone deliberately: `post-exec.md` says post-exec transactions are submitted to the DA layer "alongside the user transactions and any deposited transactions". Deposits are never in batch data, so that clause is loose. It is pre-existing text and not this change's business, but the new paragraph sits just below it, so it is now more visible.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


specs/protocol/lagoon/derivation.md
```diff
@@ -0,0 +1,139 @@
...
+transaction consumes an entry from `tx_tos`: a `0` bit consumes the next address, a `1` bit consumes none. For
+every transaction type defined before Lagoon, "is a contract creation" and "has no `to` field" are the same
+condition, which is why the [Delta definition](../delta/span-batches.md#span-batch-format) states the former. A
+post-exec transaction is the first type for which they differ. The operative reading is the latter: the bit is `1`
+whenever the transaction has no recipient, so that `tx_tos` stays exactly as long as the number of transactions
+that have one.
+
+A decoder MUST NOT reject a batch whose bit is `0` for a post-exec transaction. It consumes the next `tx_tos` entry
+as it would for any `0` bit, and discards it: a post-exec transaction has no recipient field for the address to
+occupy. As with the slots below, the bit cannot change the transaction the decoder reconstructs.
```
geoknee: I'd rather we make this very strict and say the decoder MUST reject such a batch. There is no reason to allow the decoding to be more permissive than the encoding (on the contrary actually, it is safer to make it very strict).

specs/protocol/lagoon/derivation.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Unused signature and gas accounting slots
+
+`y_parity_bits`, `tx_sigs`, `tx_nonces` and `tx_gases` are positional: every transaction in the span occupies one
+slot in each, whether or not the corresponding field exists. A post-exec transaction has no signature, nonce or gas
+limit, so:
+
+- A batcher MUST write zero into each of these slots for a post-exec transaction, as given in the table above. This
+  makes the span batch encoding of a given sequence of L2 blocks unique.
+- A decoder MUST ignore the values in these slots for a post-exec transaction, and MUST NOT reject a batch because
```
geoknee: I'd rather we say the decode MUST require the values to match the spec and MUST reject a batch otherwise.

specs/protocol/lagoon/derivation.md
geoknee: Agreed, changed in 842155c — a decoder now MUST reject a batch whose bit is `0` for a post-exec transaction.

You are right that I had this calibrated wrong. My reason for the permissive wording was that turning a client convention into a drop rule is normally a consensus change in disguise. That does not apply here: no chain in the superchain-registry sets `lagoon_time`, and `0x7D` cannot legally appear in a batch before Lagoon, so there is no already-posted batch for a stricter rule to reinterpret. Tightening is free precisely because the type has not shipped.

Worth flagging for the client side: `op-node` and `kona` both accept a `0` bit today (`fullTxs` consumes a `tx_tos` entry and `convertToFullTx`/`to_full_tx_bytes` then discard it), so both need a matching change before Lagoon activates.

specs/protocol/lagoon/derivation.md
geoknee: Agreed, changed in 842155c — a decoder now MUST verify `y_parity_bits`, `tx_sigs`, `tx_nonces` and `tx_gases` are zero for a post-exec transaction and MUST reject the batch otherwise.

The argument that convinced me is malleability rather than safety-in-decoding: tolerating those values gives one sequence of L2 blocks unboundedly many valid encodings, and `tx_sigs` alone reserves 64 unused bytes per transaction for a batcher to fill freely. Strictness also costs nothing on the encoder side — op-batcher reaches `addPostExecTx` via `AppendSingularBatch` and already zero-fills all four.

One implementation caveat for whoever picks up the client work, which is not a spec question but will decide whether the change is safe to land. These slots are only reachable at reconstruction time, so the check has to live in `fullTxs` / `full_txs` — and that is the exact path in [protocol-team#222](https://github.com/ethereum-optimism/protocol-team/issues/222), where a non-critical error makes op-node drop just the batch and continue the channel (`channel_in_reader.go`, "dropping malformed span batch") while kona drops the whole remaining channel. Implementing this rule naively therefore adds a new batcher-triggerable trigger for that open divergence. Either #222 lands first, or the rejection needs to surface as a batch-queue validity verdict rather than a decode failure.

specs/protocol/lagoon/derivation.md
```diff
@@ -0,0 +1,140 @@
...
+the block body.
+
+## Transposed Envelope Fields
+
+A post-exec transaction has no envelope fields to transpose. Its slots in the span batch `txs` structure take the
+following values:
+
+| Slot                     | Value for a `0x7D` transaction                       |
+| ------------------------ | ---------------------------------------------------- |
+| `contract_creation_bits` | `1`                                                  |
```
geoknee, line 94: This is pretty... weird to read.

Could we rename `contract_creation_bits` to `nonzero_to_bits`?

Could that be made as a backwards compatible change?

In general does the addition of post exec transactions justify a new span batch format?

geoknee: Fair — most of that paragraph exists only to apologise for the name, which is a sign the name is the problem.

**Rename: yes, but that specific name inverts it.** Today the bit is `1` when there is *no* `to`. `nonzero_to_bits` reads as `1` = `to` is present, which is the opposite polarity, so a straight find-and-replace would silently invert the meaning for anyone reading it fresh. Polarity-preserving options: `zero_to_bits` (closest to your instinct), `no_to_bits`, or `to_omitted_bits`. Flipping the bit *and* renaming would read best of all, but that is a wire change, not a rename.

**Backwards compatible: completely, as a rename.** Nothing in the span batch is self-describing — `txs` is a positional concatenation (`contract_creation_bits ++ y_parity_bits ++ tx_sigs ++ ...`), no tags, no names on the wire. So this is purely editorial: no fork gate, no consensus impact, and client identifiers (`contractCreationBits`, `contract_creation_bits`) can follow at leisure. Two small costs: it edits Delta, a shipped fork document, so it is worth keeping a "formerly `contract_creation_bits`" note so searches and third-party implementations still land; and until clients rename, spec and code vocabulary differ.

Suggest doing it as its own PR against `delta/span-batches.md` rather than growing this one, after which the Lagoon prose collapses to a single sentence stating the rule. Happy to open that.

**A new span batch format: I do not think this justifies one.** I measured it rather than guessing. Per post-exec transaction the unused slots cost 66 raw bytes (64 `tx_sigs` + zero `tx_nonces`/`tx_gases` varints). After channel compression that is **2–4 bytes with brotli** and 0.3–8 with zlib, across 0–20 user txs per block, because the slots are a run of zeros that LZ eats almost entirely. At 2s blocks that is roughly 170 KB/day — on the order of one extra blob per day.

Worth noting the strictness rule we just agreed is what buys that: because the slots MUST be zero, the run is always compressible. Had we allowed batchers to write anything there, the cost would have been the full 66 bytes.

Against a few bytes per block, a new format version costs a version-negotiation surface, two decode paths maintained forever (fault proofs re-derive historical channels indefinitely), a fresh fork gate, and a new surface for exactly the cross-client divergence class we are already working through in [#222](https://github.com/ethereum-optimism/protocol-team/issues/222) and [#293](https://github.com/ethereum-optimism/protocol-team/issues/293). Span batch v2 has sat specified-but-unactivated since Delta for a much larger saving (`fee_recipients`), which suggests the bar is well above this.

If we do open a new version later, the post-exec slots are a reason to include in it, not a reason to start it.

Method caveat: synthetic data modelled on the `txs` layout (random 64-byte sigs, small nonce/gas varints, ERC20-shaped calldata), 300 blocks per configuration, not replayed from a real channel — so treat 2–4 bytes as the order of magnitude, not a precise figure.

geoknee: Opened as #945, stacked on this branch (base `gk/lagoon-span-batch-post-exec`, so its diff shows only the rename — retarget to `main` once this merges).

Went with `zero_to_bits` rather than `nonzero_to_bits` to keep the polarity: the bit is `1` when the `to` is absent. `no_to_bits` or `to_omitted_bits` would do equally well if either reads better — it is a one-line change.

The Delta definition is reworded to describe what the bit actually selects (consumes no entry from `tx_tos`), with a note recording the former name so searches and third-party implementations still land. The Lagoon paragraph that prompted your "weird to read" collapses from seven lines to two, which was the point of doing this.

Left the two substantive questions answered above rather than acted on: no new span batch format, and the client-side renames can follow whenever convenient since nothing is on the wire.

specs/protocol/lagoon/derivation.md
```diff
@@ -0,0 +1,140 @@
...
+
+Decoding is deliberately no more permissive than encoding. The values in these slots cannot reach the reconstructed
+transaction, so tolerating them would cost nothing in the short term — but it would make a span batch's encoding
+malleable: the same sequence of L2 blocks would have unboundedly many valid encodings, differing in bytes that a
+batcher chooses freely. `tx_sigs` alone reserves 64 bytes per transaction that no post-exec transaction uses. A
+strict decoder keeps the encoding canonical, denies a batcher that space as a channel for arbitrary data, and
+leaves nothing that a later upgrade would have to tighten retroactively.
+
+This strictness is available precisely because `0x7D` is new. A post-exec transaction cannot legally appear in a
+batch before Lagoon, so no rule stated here reinterprets any batch that has already been posted.
```
sebastianst: I'd remove this paragraph. It's interesting context, but not necessary to be part of the spec.

einar-oplabs (approved):
Just for my own understanding: we are specifying exactly what the positional values must be for a `PostExecTx` in the span batch data structure below.  This seems like the way to go.

```
kona — rust/kona/crates/protocol/protocol/src/batch/transactions.rs:18

pub struct SpanBatchTransactions {
    pub total_block_tx_count: u64,
    pub contract_creation_bits: SpanBatchBits,
    pub tx_sigs: Vec<Signature>,
    pub tx_nonces: Vec<u64>,
    pub tx_gases: Vec<u64>,
    pub tx_tos: Vec<Address>,
    pub tx_data: Vec<Vec<u8>>,
    pub protected_bits: SpanBatchBits,
    pub tx_types: Vec<OpTxType>,
    pub legacy_tx_count: u64,
}
```

specs/protocol/lagoon/derivation.md
```diff
@@ -0,0 +1,146 @@
...
+amendment: its `transaction_list` holds each transaction's EIP-2718 encoding verbatim, so a post-exec transaction
+appears there as `0x7D ++ rlp_encoded_payload` like any other typed transaction, with nothing transposed and no
+slots to fill. Only the span batch format, which splits each transaction across per-field slots, needed the rules
+above.
+
+Whether a batch is permitted to contain a post-exec transaction at all is a separate, batch-level question, and one
+that applies to both batch formats. It is governed by the `batch.transactions` drop rules in
+[Batch Queue](../derivation.md#batch-queue). Those rules drop any transaction of a future type greater than `2`,
+with the type `4` exception added by [Isthmus](../isthmus/derivation.md#activation); they require a corresponding
+Lagoon amendment for `0x7D`.
```
einar-oplabs: **Opus 5 (high)**: This reproduces only half of the Isthmus precedent. Isthmus's span batch update has an `## Activation` section stating that type-4 transactions are accepted only if Isthmus is active *at the timestamp of the batch*, and — critically — that the check "must happen at the level of individual batches that are derived from span batches, not to span batches as a whole", so a span batch may legally straddle activation. There is no equivalent here.

Two consequences:

1. The strictness argument below ("a post-exec transaction cannot legally appear in a batch before Lagoon") is asserted rather than specified. No document states it today — protocol-team#319 is what would.
2. As the spec reads with this PR merged, `batch.transactions` still drops any transaction of type greater than `2` except `4`. So this document specifies a precise span batch encoding for a transaction that the batch queue rule forbids outright. That is not only out of scope; the two documents contradict each other until #319 lands.

Even if the drop rule itself belongs in #319, the *activation granularity* is span batch encoding business and would sit naturally here.

specs/protocol/lagoon/derivation.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Unused signature and gas accounting slots
+
+`y_parity_bits`, `tx_sigs`, `tx_nonces` and `tx_gases` are positional: every transaction in the span occupies one
+slot in each, whether or not the corresponding field exists. A post-exec transaction has no signature, nonce or gas
+limit, so:
+
+- A batcher MUST write zero into each of these slots for a post-exec transaction, as given in the table above.
+- A decoder MUST verify that each of them is zero, and MUST reject a batch in which any of them is not. Such a
+  batch is invalid and is dropped.
```
einar-oplabs: **Opus 5 (high)**: "a batch" is unqualified here and in the `contract_creation_bits` rule above. Isthmus flags this exact distinction as load-bearing: rejection at span batch granularity discards every block in the span, rejection at singular batch granularity discards one.

That matters more than usual here, because these slots are only reachable in `fullTxs` / `full_txs`, which operate span-wide — so the obvious implementation is precisely the granularity Isthmus warns against. Worth pinning explicitly which one a decoder MUST apply.

specs/protocol/lagoon/derivation.md
```diff
@@ -0,0 +1,146 @@
...
+A post-exec transaction is not a contract creation, so this deserves a word. What the bit governs is whether the
+transaction consumes an entry from `tx_tos`: a `0` bit consumes the next address, a `1` bit consumes none. For
+every transaction type defined before Lagoon, "is a contract creation" and "has no `to` field" are the same
+condition, which is why the [Delta definition](../delta/span-batches.md#span-batch-format) states the former. A
+post-exec transaction is the first type for which they differ. The operative reading is the latter: the bit is `1`
+whenever the transaction has no recipient, so that `tx_tos` stays exactly as long as the number of transactions
+that have one.
+
+A decoder MUST reject a batch in which the bit is `0` for a post-exec transaction. Such a batch is invalid and is
+dropped, exactly as one whose `tx_datas` element carries an unusable transaction type is.
```
einar-oplabs, line 115: **Opus 5 (high)**: "is dropped" is the sentence op-node and kona already read differently: op-node drops the batch and continues the channel, kona drops the remaining channel (the protocol-team#222 divergence the PR description notes).

Isthmus's Activation section addresses propagation directly — "if Holocene is active, this will also lead to the remaining span batch, and channel that contained it, to get dropped". Since this document adds new triggers for that path, saying how far the drop propagates seems worth the sentence.

specs/protocol/lagoon/derivation.md
```diff
@@ -0,0 +1,146 @@
...
+
+A post-exec transaction is unlike every previously batched type: it carries only an opaque
+[post-exec payload][g-post-exec-payload], with no nonce, gas limit, recipient or signature. Most of the slots the
+span batch format reserves per transaction have no natural value for it. This section specifies all of them.
+
+Nothing outside the `txs` structure changes. In particular, a post-exec transaction is an ordinary member of its
+block's transaction list, so it is counted in that block's `block_tx_counts` entry and in the
+`MAX_SPAN_BATCH_ELEMENT_COUNT` total, exactly like a user transaction. Because a post-exec transaction is
+[the final transaction of its block](./post-exec.md#block-level-structural-rules), it occupies the last index of
+that block's slice of the span.
```
einar-oplabs, line 43: **Opus 5 (high)**: This is descriptive — no MUST, and no decoder obligation attached. So what happens to a span batch carrying a `0x7D` at a non-final index within a block, or two of them in one block? Rejected at decode, or a block that fails later?

The asymmetry is what stands out. The document mandates rejecting a nonzero `tx_nonces` byte that provably cannot affect the derived block, but leaves a structural violation that *can* affect it unaddressed. Either make it a MUST here, or state that structural validation belongs to `post-exec.md` and is not a batch-decode concern.

specs/protocol/lagoon/derivation.md
```diff
@@ -0,0 +1,146 @@
...
+## Transaction Data
+
+This corresponds with a new encoding of the `tx_datas` list as specified in
+[the Delta span batch spec](../delta/span-batches.md#span-batch-format), adding a new transaction type:
+
+Transaction type `0x7D` ([post-exec](./post-exec.md)): `0x7D ++ rlp_encoded_payload`
+
+where `rlp_encoded_payload` is the RLP encoding of the [post-exec payload][g-post-exec-payload] as a list, exactly
+as defined by the transaction's [EIP-2718 encoding](./post-exec.md#encoding). As for every other `tx_datas`
+element, the bytes following the type byte MUST be a single RLP list.
```
einar-oplabs, line 72: **Opus 5 (high)**: This rule is correct and both clients already enforce it at element framing — op-node rejects a non-list kind in `ReadTxData` (`span_batch.go:687`), kona checks `!rlp_header.list` in `read_tx_data` (`batch/transactions.rs:143`).

What is unstated is where validation stops. Below that framing check both clients treat the payload as fully opaque: op-node captures it with `s.Raw()` (`span_batch_tx.go:77`), kona copies `payload_length + header.length()` bytes (`batch/tx_data/wrapper.rs:29`). Neither inspects the version byte or the payload schema at decode time.

Stating one validation rule and stopping reads like the beginning of a contract. One sentence saying payload validity is `post-exec.md`'s business and is not checked at batch decode would close it.

specs/protocol/lagoon/derivation.md
```diff
@@ -0,0 +1,146 @@
...
+
+The bit for a post-exec transaction MUST be `1`.
+
+A post-exec transaction is not a contract creation, so this deserves a word. What the bit governs is whether the
+transaction consumes an entry from `tx_tos`: a `0` bit consumes the next address, a `1` bit consumes none. For
+every transaction type defined before Lagoon, "is a contract creation" and "has no `to` field" are the same
+condition, which is why the [Delta definition](../delta/span-batches.md#span-batch-format) states the former. A
+post-exec transaction is the first type for which they differ. The operative reading is the latter: the bit is `1`
+whenever the transaction has no recipient, so that `tx_tos` stays exactly as long as the number of transactions
+that have one.
```
einar-oplabs, line 112: **Opus 5 (high)**: The argument is convincing, but the correction lands only in a hardfork-specific document. After this PR, Delta still defines the bit as "is a contract creation" while Lagoon says the operative meaning is "has no recipient" — and the next person adding a transaction type will read Delta.

Consider a one-line amendment to the Delta definition itself (a `1` bit means the transaction consumes no `tx_tos` entry) so the fix is permanent rather than something each future type has to rediscover.

einar-oplabs (approved):
I asked Claude to identify things that were still ambiguously specified.

Interestingly, it seems to me that e.g. the encoder is assuming certain invariants in the spec, e.g. that the `PostExecTx` comes last and is unique.  But it will likely still encode malformed input, e.g. two `PostExecTx` in a block.  I am not sure if we should reuse our verification logic, accept it as is (out of scope for encoder) or at least state clearly whatever assumptions we make in the spec.

geoknee:
Thanks @einar-oplabs — answering both reviews.

**On the first:** yes, exactly that. The document pins the positional values every slot of `SpanBatchTransactions` takes for a `PostExecTx`, because nothing else did. Heads up that `contract_creation_bits` is renamed to `zero_to_bits` in #945, stacked on this branch — the bit never tested contract creation, it selects whether a `tx_tos` entry is consumed, and post-exec is the first type where the two come apart. Spec-side only; the kona and op-node identifiers can follow whenever convenient.

**On the second:** good catch, and it resolves in a way worth writing down, so I have — `ba6bc21`.

You are right that the encoder assumes uniqueness and last-position without checking. But the invariants are enforced, just not in the codec. All four block-level rules live in `parse_post_exec_payload_from_transactions` (`op-alloy/crates/consensus/src/post_exec.rs:138-190`):

| Rule in `post-exec.md` | Error |
| --- | --- |
| At most one | `MultiplePostExecTxs` |
| Last in block | `PostExecTxNotLast` |
| Anchored to block | `BlockNumberMismatch` |
| Not before Lagoon | `UnexpectedPostExecTx` |

op-reth calls it from `post_exec_mode_from_transactions` (`crates/evm/src/lib.rs:156`) while executing the block, which is exactly where `post-exec.md` says these rules bite ("any violation invalidates the block").

So the end-to-end behaviour for a hand-crafted malformed batch is already defined: the span batch decodes (the codec is format-only), produces payload attributes, and the engine returns `INVALID`. Holocene's [Steady Block Derivation](https://specs.optimism.io/protocol/holocene/derivation.html#engine-queue) then replaces the payload with a deposit-only one and drops the remaining span batch and its channel. Non-halting, and the same on both sides.

Of your three options I went with the third and explicitly against the first. Reusing the verification logic in the batch layer would give one rule two enforcement points with *different* outcomes — batch drop at decode time, versus invalid block plus deposit-only replacement plus channel drop. Those derive different chains, which is the divergence class already open as [protocol-team#222](https://github.com/ethereum-optimism/protocol-team/issues/222). One rule, one enforcement point, is worth more here than defence in depth.

Encoder-side I think "out of scope" is right rather than merely acceptable: a batcher only ever encodes blocks the engine already accepted, so a check there guards against a local bug, not a protocol condition. A debug assertion, if anyone wants one, not a spec rule.

The new paragraph says the position is an assumption the encoding *inherits* rather than enforces, names what happens to a batch that breaks it, and points at both enforcement points.

One thing that fell out of checking this and is worth flagging separately: **op-geth has no post-exec implementation at all** — `grep -r PostExecTx` across the repo returns nothing. So this enforcement currently exists only on the op-reth path. That is an implementation-status gap rather than a spec one, but it does mean the rules in `post-exec.md` are unenforced for an op-geth-based chain today.

specs/protocol/lagoon/derivation.md
geoknee: Addressed in `ba6bc21`, which landed just after your review — I took the second option.

The document now says the last-index position is an assumption the encoding inherits rather than enforces, that the span batch format can faithfully encode a block with two `0x7D` transactions or one that is not last, and that such a batch is well formed *as a batch*. The structural rules are caught where `post-exec.md` states them, on the block.

On the asymmetry you flag: I think it is the right way round, though the document did not previously earn that. A nonzero `tx_nonces` byte is a property of the *batch encoding* and nothing else validates it, so if the codec does not reject it nothing ever will. A misplaced `0x7D` is a property of the *block*, and `op-alloy`'s `parse_post_exec_payload_from_transactions` already rejects it (`MultiplePostExecTxs`, `PostExecTxNotLast`) during execution. Duplicating that into the codec would give one rule two enforcement points with different outcomes — batch drop at decode, versus invalid block plus deposit-only replacement plus channel drop — which derive different chains. That is the [#222](https://github.com/ethereum-optimism/protocol-team/issues/222) class, so I would rather not add to it.

specs/protocol/lagoon/derivation.md
geoknee: Agreed, and fixed in `538ff5e`. The document now says explicitly that the framing check is where batch decoding's interest in the payload ends, and that a decoder MUST NOT inspect the `version` byte or validate the payload against a schema while decoding a batch — below the outer RLP list the element is opaque and is reproduced verbatim.

Your reading of both clients matches mine: op-node captures the payload with `s.Raw()`, kona copies `payload_length + header.length()` bytes, and neither looks inside at this layer.

Worth noting where that stops being true, since it is adjacent and not obvious. Once the reconstructed transaction is decoded into a typed envelope, the two diverge: kona's `TxPostExec::rlp_decode` parses and validates the payload, while op-node's `PostExecTx` keeps it opaque and its own source documents the split as a "Go-lenient/Rust-strict wire-acceptance divergence". That is [protocol-team#247](https://github.com/ethereum-optimism/protocol-team/issues/247), and it sits one layer above this document — which is a further reason for the batch codec to state plainly that it does not participate.

specs/protocol/lagoon/derivation.md
geoknee: Already done, in #945 — stacked on this branch, opened before your review so it is easy to miss.

It renames the field to `zero_to_bits` in the Delta definition itself and rewords that definition to describe what the bit actually selects: the transaction has no `to` field and therefore consumes no entry from `tx_tos`. A note records the former name so searches and third-party implementations still land on it.

You are right that the Lagoon-only correction was the weaker fix, and for exactly the reason you give — the next person adding a transaction type reads Delta. That PR also deletes the paragraph of apology this one needed, which was the tell that the name rather than the prose was the problem.

Purely editorial: nothing in a span batch is self-describing, so no encoding, polarity or fork gate is touched. The client identifiers can follow whenever convenient.

specs/protocol/lagoon/derivation.md
geoknee: Half taken, half deliberately declined — `538ff5e`.

Taken: the rejection granularity is now pinned. These slots are positional across the whole span, so a violation cannot be attributed to the one block whose transaction carries it; the span batch is what becomes invalid, and the document says so rather than saying "a batch".

Declined: how far that invalidity propagates. Three reasons.

It is not specific to post-exec. Every decode-stage failure in a span batch raises the same question — an unusable transaction type, a short `tx_tos`, a malformed bitlist. Settling it inside a Lagoon document would fix for `0x7D` what stays unstated for all of them, and would put the general rule in a place nobody looking for it would read.

The two clients disagree, and that disagreement is the open question in [#222](https://github.com/ethereum-optimism/protocol-team/issues/222) rather than an oversight: op-node's `ChannelInReader` drops the batch and keeps reading the channel, kona's `ChannelReader` discards the rest of it. Writing either behaviour here would adjudicate that issue by side effect, in a document neither client team would think to re-read when #222 is decided.

And the Isthmus sentence you quote is scoped to a *validity* verdict from the batch queue, where Holocene's fast channel invalidation genuinely applies. A decode-time rejection never reaches the batch queue, which is precisely the gap #222 documents. Reusing that sentence's phrasing here would assert the propagation it describes for a path it does not cover.

So the document now states that the propagation is a property of malformed span batches in general and is not settled here. When #222 lands, the answer should go in the general rule and this document should need no edit.

specs/protocol/lagoon/derivation.md
geoknee: Agreed and pinned in `538ff5e`, in both places — the slot rules here, and the `zero_to_bits` rule above in #945 so the two are not phrased differently once that merges.

The answer is span batch granularity, and your reasoning is why: these slots are positional across the whole span, so a violation genuinely cannot be attributed to the single block whose transaction carries it without extra bookkeeping neither client does. That makes span-wide rejection the honest rule rather than merely the convenient one — the distinction Isthmus flags does not have two available answers here the way it does for a per-block acceptance check.

Which is also why I stopped there and did not specify how far the rejection propagates beyond the span batch: same sentence, different question, and that one is [#222](https://github.com/ethereum-optimism/protocol-team/issues/222). Reasoning in the reply on the neighbouring thread.

specs/protocol/lagoon/derivation.md
geoknee: Both points are right, and the second is sharper than the PR admitted. Partly addressed in `538ff5e`; partly a deliberate split I want to justify rather than quietly keep.

**The contradiction is real and is now stated outright.** The document previously implied it by saying the Batch Queue rules "require a corresponding Lagoon amendment"; it now says plainly that until that amendment lands the two documents disagree, because the Batch Queue rules forbid the transaction this document gives an encoding for. A reader should not have to infer that from a polite clause.

**On activation granularity — you are right that it is unspecified, and I checked that both clients already implement it.** op-node computes `isSDM := cfg.IsSDM(blockTimestamp)` *inside* the per-block loop of the span batch check (`batches.go:411`), and kona gates on `cfg.is_sdm_active(self.timestamp)` per `SingleBatch` (`single.rs:183`). So a span batch may already legally straddle the Lagoon activation and carry post-exec transactions only in the blocks at or after it — exactly the Isthmus rule, implemented, undocumented.

I recorded that fact in the Batch Acceptance section, with the note that it belongs with the acceptance rule rather than here, and did not write it as a normative rule. The reason is that it does not stand up alone: "`0x7D` is accepted only if Lagoon is active at the timestamp of the individual batch" presupposes that `0x7D` is ever accepted, which is the very thing the Batch Queue rules currently deny. Specifying the qualifier while the thing it qualifies is still forbidden would make the contradiction worse rather than better. Isthmus could bundle them because it wrote both at once; [#319](https://github.com/ethereum-optimism/protocol-team/issues/319) should do the same.

**On the strictness argument being asserted rather than specified** — half right, and fixed in `932cdcc`. The support exists, it was just not the rule I appeared to be leaning on. `post-exec.md` already states normatively that a block before the Lagoon activation timestamp MUST NOT contain a `0x7D` transaction, so no posted batch carries one. The claim now cites that instead of reading as an appeal to the batch queue drop rule, which is indeed the one rule that does not yet say it.

If you would rather see the whole activation story land here instead, I am happy to write it — it is a short section and the client behaviour is verified. It would mean this PR also amending the Batch Queue rule, which is #319's stated scope, so I would want that to be a deliberate call rather than something I widen into on my own.

PR #946, opened 2026-09-16
Refresh the smart contract architecture overview

mds1:
The overview (https://specs.optimism.io/protocol/overview.html#architecture-overview) still used the older dispute-game contracts and omitted the L2 contract upgrade path. Update the L1 diagrams for U20 with ETHLockbox enabled, and document the Karst L2 upgrade flow.

- Show super-root games, withdrawal validation, ETH custody, and the correct pause and owner permissions.
- Add the L2ContractsManager upgrade flow and supporting predeploys, OperatorFeeVault, and beacon-root and block-history notes.
- Separate dense diagrams by purpose, use consistent colors, and move pause and permission details into nearby notes. Correct the ERC721 bridge labels and distinguish fixed implementations from upgradeable proxies.


specs/protocol/overview.md
```diff
@@ -156,92 +136,92 @@ graph LR
...
+- Green contracts have fixed implementations. The factory creates dispute games as clones with immutable arguments;
+  `MIPS64` and `PreimageOracle` are deployed directly.
+- `SuperFaultDisputeGame` uses game type `SUPER_CANNON_KONA` (`9`). `SuperPermissionedDisputeGame` uses
+  `SUPER_PERMISSIONED` (`5`) and accepts proposals only from its configured proposer. It resolves immediately in favor of
+  the proposal, without challenges or bonds. Withdrawal finality and Guardian checks still apply.
+- Users deposit or withdraw ETH and tokens through the bridges. They can also deposit directly through `OptimismPortal`,
+  where they prove and execute withdrawals.
+- The bridges, messenger, portal, `ETHLockbox`, `AnchorStateRegistry`, and `DelayedWETH` read pause state through
+  [SystemConfig](./system-config.md). It combines the global pause state from [SuperchainConfig](./superchain-config.md)
+  with the chain-specific pause state, identified by `ETHLockbox`. The portal also reads its configuration from `SystemConfig`.
```
wwared: **Claude:** **Should-fix — pause routing.** Against deployed U20 code this holds only for the bridges and messenger (`L1StandardBridge.sol:123`, `L1ERC721Bridge.sol:66`, `L1CrossDomainMessenger.sol:64` all call `systemConfig.paused()`). The portal, `AnchorStateRegistry` and `DelayedWETH` read `ethLockbox.paused()` directly (`OptimismPortal2.sol:302`, `AnchorStateRegistry.sol:136`, `DelayedWETH.sol:102`), and `ETHLockbox` is the hub — `superchainConfig.isLocalOrGlobalPaused(address(this))` (`ETHLockbox.sol:112`). `SystemConfig.paused()` is *downstream* of the lockbox (`SystemConfig.sol:568`).

In fairness, the sentence is faithful to this repo's own text — `optimism-portal.md:360`, `anchor-state-registry.md:417` and `eth-lockbox.md:156` all still say `SystemConfig`, and `system-config.md:301-311` does normatively make `SystemConfig.paused()` the global∨local combiner keyed on `ETHLockbox`. The spec book, not this PR, is stale, which is why I didn't grade this blocking. Two fixes worth making regardless:

1. `DelayedWETH` is unsupported by *either* source — `bond-incentives.md:205` says `SuperchainConfig.paused()`, the code says `ETHLockbox`.
2. `ETHLockbox` appearing in one sentence as both a *reader of* `SystemConfig` and the *identifier used by* `SystemConfig` reads as circular. Naming `ETHLockbox` as the pause hub is correct under both the spec and the implementation.

specs/protocol/overview.md
```diff
@@ -156,92 +136,92 @@ graph LR
...
+- The Guardian pauses or unpauses `SuperchainConfig`. It can blacklist or retire games and set the respected game type
```
wwared: Do we also want to mention the Pause Deputy role here? They can also pause the chain, and it is a different role than the Guardian right?

specs/protocol/overview.md
```diff
@@ -156,92 +136,92 @@ graph LR
...
+  in `AnchorStateRegistry`. The factory owner configures game types and bonds; the `ProxyAdmin` owner can hold or recover
+  bonds in `DelayedWETH`.
```
wwared: I think we only support configurations where the "factory owner" is the same entity as the "`ProxyAdmin` owner". Here's Claude's explanation of this: 
 the Proxy Admin Owner is *the only* address authorized for `DisputeGameFactory.setImplementation`, `setInitBond`, `DelayedWETH.hold` and `recover`; OPCM transfers factory ownership to the PAO (`OPContractsManagerV2.sol:1006`) and the standard validator enforces it (`StandardValidatorUtils.sol:271`, `DF-30`).

Suggestion:

```suggestion
  in `AnchorStateRegistry`. The `ProxyAdmin` owner also owns `DisputeGameFactory` in the standard configuration: it registers game-type implementations and init bonds, and can hold or recover bonds in `DelayedWETH`.
```

specs/protocol/overview.md
```diff
@@ -48,101 +49,80 @@ This document assumes you've read the [background](../background.md).
 
 ## Architecture Overview
 
+Blue nodes are upgradeable contracts, green nodes have fixed implementations, and orange nodes are actors or protocol
+transactions. Grey nodes show other contracts, addresses, or configuration. Dotted arrows indicate reads.
```
wwared: **Claude:** **Should-fix — the legend isn't satisfiable in the Core L1 section.** It promises "orange nodes are actors or protocol transactions" and "dotted arrows indicate reads", but both new L1 diagrams have zero actor nodes and zero read edges — `classDef actor` isn't even declared there. `Guardian`, `SuperchainConfig` and `SystemConfig` now appear as diagram nodes *nowhere on the page*.

The prose bullets do cover every privileged capability, so nothing is lost outright, but the control plane is no longer visible at a glance in the section whose job is architecture. A small third "control plane" diagram would close both gaps: Guardian/Pause Deputy → `SuperchainConfig`; Guardian → `AnchorStateRegistry` (blacklist / retire / respected game type); PAO → `ProxyAdmin`, `DisputeGameFactory`, `DelayedWETH`, `ETHLockbox`, `OptimismPortal`.

specs/protocol/overview.md
```diff
@@ -48,101 +49,80 @@ This document assumes you've read the [background](../background.md).
...
-        L2Nodes(Layer 2 Nodes)
-    end
-
-    L2Nodes -.->|fetch transaction batches| BatchDataEOA
-    L2Nodes -.->|fetch deposit events| OptimismPortal
+These diagrams assume a single ETH-gas chain with `ETHLockbox` enabled.
 
-    Batcher -->|publish transaction batches| BatchDataEOA
+```mermaid
+%%{init: {"theme": "base", "themeVariables": {"lineColor": "#64748b", "edgeLabelBackground": "#f1f5f9"}, "themeCSS": ".nodeLabel a { color: inherit !important; }"}}%%
```
wwared: **Claude:** **Should-fix — `themeCSS` removes the only signal that diagram nodes are links.** The live site sets `.content a { text-decoration: none; }` (`specs.optimism.io/css/general.css:109`), so colour is the sole affordance. With `color: inherit !important`, `OptimismPortal` (linked) and `Implementation` / `L2 Node` / `Predeploy Implementations` (not linked) render identically.

I understand the motive — link blue on the new coloured fills — but keep some affordance:

```
"themeCSS": ".nodeLabel a { color: inherit !important; text-decoration: underline; }"
```

FWIW I verified the directive itself works: parsed all 7 diagrams against the version this book actually serves (`specs/static/mermaid.min.js`, mermaid 10.9.1) — all OK, and `getConfig()` after `parse()` confirms `themeCSS` survives the directive sanitizer rather than being silently dropped.

Separately (nit): this PR is the first use of `%%{init}%%` anywhere in `specs/**`; 15+ other diagrams still use the old stroke-only classDefs. Worth a line in `CONTRIBUTING.md` if this is the new house style, otherwise this page diverges visually from the rest of the book indefinitely.

specs/protocol/overview.md
```diff
@@ -48,101 +49,80 @@ This document assumes you've read the [background](../background.md).
...
-        Batcher(<a href="./batcher.html">Batcher</a>)
-    end
-
-    subgraph "Layer 2 Interactions"
-        L2Nodes(Layer 2 Nodes)
-    end
-
-    L2Nodes -.->|fetch transaction batches| BatchDataEOA
-    L2Nodes -.->|fetch deposit events| OptimismPortal
+These diagrams assume a single ETH-gas chain with `ETHLockbox` enabled.
```
wwared, line 60: Not sure if we want to, but we could add a small note here (or somewhere in this section) mentioning that the `ETHLockbox` can also be shared between chains. I'm leaning towards not adding it since that's interop-specific.

specs/protocol/overview.md
```diff
@@ -48,101 +49,80 @@ This document assumes you've read the [background](../background.md).
...
+    OptimismPortal <-->|lock/unlock ETH| ETHLockbox
+
+    classDef proxy fill:#e8f1ff,stroke:#3971b8,color:#17375e,stroke-width:1.5px;
+    classDef other fill:#f1f3f5,stroke:#98a2b3,color:#344054,stroke-width:1.5px;
+    class L1StandardBridge,L1ERC721Bridge,L1CrossDomainMessenger,OptimismPortal,ETHLockbox proxy;
+    class ExternalERC20,ExternalERC721 other;
+```
+
+The portal uses dispute games to verify withdrawals. Permissionless games dispute super roots at L2 timestamps;
+the portal reads the output root for its chain from the selected game.
```
wwared: **Claude:** **Nit — this reads as if the portal picks a trusted game.** The prover supplies `_disputeGameIndex` (`OptimismPortal2.sol:386`); safety comes from the `AnchorStateRegistry` predicate set, none of which is mentioned: factory-registered, respected game type at creation, not blacklisted, not retired, not paused, resolved `DEFENDER_WINS`, past the dispute-game finality delay (`AnchorStateRegistry.sol:307-355`), plus the portal's own proof-maturity delay.

The interesting property is the inverse of what's written — anyone may point at any game and the registry decides. Worth saying, because it's what makes line 143's "Withdrawal finality and Guardian checks still apply" load-bearing for the permissioned path.

The super-root/timestamp half is correct (`super-fault-dispute-game.md:25`).

specs/protocol/overview.md
```diff
@@ -156,92 +136,92 @@ graph LR
   - The `L1StandardBridge` contract sits behind the [`L1ChugSplashProxy`](https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts-bedrock/src/legacy/L1ChugSplashProxy.sol)
     contract, a legacy proxy contract type used within older versions of the OP Stack. This proxy type is used exclusively
     for the `L1StandardBridge` contract to maintain backwards compatibility.
+- Green contracts have fixed implementations. The factory creates dispute games as clones with immutable arguments;
+  `MIPS64` and `PreimageOracle` are deployed directly.
+- `SuperFaultDisputeGame` uses game type `SUPER_CANNON_KONA` (`9`). `SuperPermissionedDisputeGame` uses
+  `SUPER_PERMISSIONED` (`5`) and accepts proposals only from its configured proposer. It resolves immediately in favor of
+  the proposal, without challenges or bonds. Withdrawal finality and Guardian checks still apply.
```
wwared: **Claude:** Verified this against source and it's **exactly right**, which is the highest-risk claim in the PR — flagging that explicitly so it doesn't get second-guessed later:

- `SUPER_CANNON_KONA = 9` and `SUPER_PERMISSIONED = 5` match `dispute/lib/Types.sol:70,82`, and `SUPER_PERMISSIONED` correctly has no `_CANNON` suffix.
- `initialize()` sets `status = DEFENDER_WINS; resolvedAt = createdAt` and emits `Resolved` at creation; `resolve()` is a `pure` no-op; `msg.value != 0` reverts; there is no challenger role or `weth()` anywhere in the contract (`SuperPermissionedDisputeGame.sol:48-74`). Proposer-only via `tx.origin != proposer()`.

One editorial call for you: `SuperFaultDisputeGame` is generic over game type — `SUPER_CANNON` (4) and `SUPER_ASTERISC_KONA` (7) use the same implementation family. Pinning `9` is right for the U20 standard config (validators require exactly 9 and 5 registered, `SUPER_CANNON` unregistered) but is narrower than the contract.

specs/protocol/overview.md
```diff
@@ -156,92 +136,92 @@ graph LR
...
+- Users deposit or withdraw ETH and tokens through the bridges. They can also deposit directly through `OptimismPortal`,
+  where they prove and execute withdrawals.
```
wwared: nit: Maybe "finalize withdrawals" is slightly more pedantically correct.

```suggestion
- Users deposit or withdraw ETH and tokens through the bridges. They can also deposit directly through `OptimismPortal`,
  where they prove and finalize withdrawals.
```

specs/protocol/overview.md
```diff
@@ -48,101 +49,80 @@ This document assumes you've read the [background](../background.md).
...
+    classDef fixed fill:#e5f3ec,stroke:#3b8363,color:#173f2e,stroke-width:1.5px;
+    class OptimismPortal,DisputeGameFactory,AnchorStateRegistry,DelayedWETH proxy;
+    class SuperFaultDisputeGame,SuperPermissionedDisputeGame,MIPS64,PreimageOracle fixed;
+    style Games fill:#f8fafc,stroke:#cbd5e1,color:#334155;
 ```
 
 #### Notes for Core L1 Smart Contracts
 
-- The `Batch Inbox Address` described above (**highlighted in GREY**) is _not_ a smart contract and is instead an arbitrarily
+- The `Batch Inbox Address` shown below (**highlighted in GREY**) is _not_ a smart contract and is instead an arbitrarily
```
wwared: nit: The `Batch Inbox Address` node only shows up in the L2 Node Components section, not in this section anymore. We could move it down below to minimize confusion

specs/protocol/overview.md
```diff
@@ -156,92 +136,92 @@ graph LR
...
-- Contracts highlighted as "L2 System Contracts" are updated or mutated automatically as part of the chain derivation
-  process. Users typically do not mutate these contracts directly, except in the case of the `FeeVault` contracts where
+- L1 attributes transactions update `L1Block`, and the execution engine credits transaction fees to the fee vaults.
+  `GasPriceOracle` reads fee parameters from `L1Block`.
+  Users typically do not mutate these contracts directly, except in the case of the `FeeVault` contracts where
   any user may trigger a withdrawal of collected fees to the pre-determined withdrawal address.
+- Fee vault withdrawals can target L1 through `L2ToL1MessagePasser`, or L2 directly, depending on the vault's configuration.
+- The execution engine also updates the [beacon roots contract](./exec-engine.md#ecotone-beacon-block-root) with the
+  L1 origin's parent beacon block root and the
+  [history storage contract](./isthmus/derivation.md#eip-2935-contract-deployment) with L2 block hashes.
```
wwared: **Claude:** **Nit — anchor precision.** `#eip-2935-contract-deployment` documents only the one-time deployment transaction, not the per-block write that actually populates the history contract with block hashes. The claim is true by EIP-2935 semantics but isn't stated at the cited anchor. (The beacon-roots half checks out against `exec-engine.md:492-504`.)

specs/protocol/overview.md
```diff
@@ -156,92 +136,92 @@ graph LR
...
 - Smart contracts that sit behind `Proxy` contracts are **highlighted in BLUE**. Refer to the
   [Smart Contract Proxies](#smart-contract-proxies) section below to understand how these proxies are designed.
```
wwared: nit: I think this is a duplicated comment (about the `Proxy` contracts being in blue)? I'd remove it and keep it only at the legend description at the top of the document, but not too important either way.

specs/protocol/overview.md
```diff
@@ -263,60 +244,128 @@ graph LR
...
+    classDef fixed fill:#e5f3ec,stroke:#3b8363,color:#173f2e,stroke-width:1.5px;
+    classDef actor fill:#fff1df,stroke:#bf7b2a,color:#664515,stroke-width:1.5px;
+    class Proxy proxy;
+    class ProxyAdmin,Implementation fixed;
+    style LogicalContract fill:#f8fafc,stroke:#cbd5e1,color:#334155;
+    class ProxyAdminOwner actor;
+```
+
+On L1, the `ProxyAdmin` owner uses [OP Contracts Manager](../experimental/op-contracts-manager.md) to coordinate upgrades.
+The owner's Safe delegatecalls the manager, so its calls to `ProxyAdmin` execute with the owner's authority.
```
wwared: **Claude:** **Nit — internally inconsistent with line 229.** That sentence (unchanged) says the owner "can be any EOA or smart contract"; an EOA cannot issue a `DELEGATECALL`. OPCM's own spec is normative that the flow requires a Safe (`op-contracts-manager.md:218,234`: "intended to be called via DELEGATECALL from the Proxy Admin Owner Safe"), consistent with `stage-1.md:161-176`.

The mechanics as described are correct — just reconcile one of the two sentences, e.g. "the Proxy Admin Owner, which in practice is a Safe, delegatecalls the manager".

specs/protocol/overview.md
```diff
@@ -263,60 +244,128 @@ graph LR
...
+
+#### L2 contract upgrades
+
+[Network upgrade transactions](./l2-upgrades-1-execution.md#bundle-format) run at fork activation. They deploy the new
+implementations and a version of `L2ContractsManager`, then call `L2ProxyAdmin.upgradePredeploys` from `DEPOSITOR_ACCOUNT`.
+`L2ProxyAdmin` delegatecalls the manager, which upgrades the predeploy proxies atomically and preserves their existing
+chain configuration. Only `DEPOSITOR_ACCOUNT` can call `upgradePredeploys`; the `L2ProxyAdmin` owner can still use its
+ordinary administrative upgrade methods.
```
wwared: **Claude:** Verified the whole Karst paragraph against source — all of it checks out, noting it so it doesn't get re-litigated: `upgradePredeploys` is `DEPOSITOR_ACCOUNT`-only and `delegatecall`s the manager (`L2ProxyAdmin.sol:46-58`), the owner retains `ProxyAdmin`'s ordinary methods, `L2ContractsManager.upgrade()` re-reads config from the live predeploys (`L2ContractsManager.sol:151-244`), and `ConditionalDeployer.deploy` skips when the CREATE2 target already has code (`ConditionalDeployer.sol:41-52`). Every edge in the diagram below matches `l2-upgrades-1-execution.md:232-250`.

specs/protocol/overview.md
```diff
@@ -263,60 +244,128 @@ graph LR
...
+    NetworkUpgrade(Network Upgrade<br/>Transactions)
+    ConditionalDeployer(<a href="https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L2/ConditionalDeployer.sol">ConditionalDeployer</a>)
+    DeterministicDeploymentProxy(Deterministic<br/>DeploymentProxy)
+    Implementations(Predeploy Implementations)
+    L2ContractsManager(<a href="https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L2/L2ContractsManager.sol">L2ContractsManager</a>)
+    L2ProxyAdmin(<a href="https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L2/L2ProxyAdmin.sol">L2ProxyAdmin</a>)
+    ProxyAdminOwner(ProxyAdmin Owner)
+    PredeployProxies(Predeploy Proxies)
+    Configuration(Existing Predeploy<br/>Configuration)
+    L1Block(L1Block)
```
wwared: **Claude:** **Nit —** bare node here, while the same contract is linked as `./predeploys.html#l1block` at line 164. (`L2ContractsManager` reading feature flags from `L1Block` is correct — `L1Block.isFeatureEnabled`, `L2ContractsManager.sol:186`.)

specs/protocol/overview.md
```diff
@@ -263,60 +244,128 @@ graph LR
...
+    classDef actor fill:#fff1df,stroke:#bf7b2a,color:#664515,stroke-width:1.5px;
+    classDef other fill:#f1f3f5,stroke:#98a2b3,color:#344054,stroke-width:1.5px;
+    class ConditionalDeployer,L2ProxyAdmin,PredeployProxies,L1Block,L2DevFeatureFlags proxy;
+    class DeterministicDeploymentProxy,Implementations,L2ContractsManager fixed;
+    class NetworkUpgrade,ProxyAdminOwner actor;
+    class Configuration other;
 ```
 
+`ConditionalDeployer`, `L2ProxyAdmin`, and `L2DevFeatureFlags` are proxied predeploys. Green contracts have fixed
+implementations; `L2ContractsManager` is deployed separately for each upgrade. Its upgrade calls execute in
```
wwared: nit: This is another duplicated legend comment. I'm ok with either keeping or removing it.

specs/protocol/overview.md
```diff
@@ -263,60 +244,128 @@ graph LR
...
-    Challenger -->|verify/challenge/defend games| FaultDisputeGame
-
-    classDef l2Components stroke:#333,stroke-width:2px;
-    classDef systemUser stroke:#f9a,stroke-width:2px;
-    classDef l1Contracts stroke:#bbf,stroke-width:2px;
-
-    class RollupNode,ExecutionEngine l2Components;
-    class BatchSubmitter,OutputSubmitter,Challenger systemUser;
-    class BatchDataEOA,OptimismPortal,DisputeGameFactory,FaultDisputeGame l1Contracts;
+    Challenger -.->|fetch super roots| RollupNode
```
wwared: **Claude:** **Nit — super-root sourcing is single-chain-only.** `op-node`'s `superroot_atTimestamp` is scoped to `cfg.L2ChainID` and can only report its own chain's output; for a multi-chain interop dependency set the Supervisor/Supernode is the authoritative source (`op-service/apis/supernode.go`, and `interop/superroot.md:23-30` frames the super root as a cross-chain artifact).

Correct as drawn for a single chain — but unlike the Core L1 diagrams (line 60) this one carries no scoping disclaimer. One sentence would do it.

Unrelated nit: the node id is still `OutputSubmitter` while the label is now `Proposer`.

specs/protocol/overview.md
wwared: nit: Probably should remove Geth mention while we're here
```suggestion
- **Maximal compatibility with ETH1 nodes:** The implementation should minimize any differences with a vanilla execution-layer
```

wwared: nit: Probably should remove Geth mention while we're here
```suggestion
  - The execution engine leverages the execution-layer's existing mempool and sync implementations, including snap sync.
```

wwared: nit: Probably should remove Geth mention while we're here
```suggestion
  - Running a replica is as simple as running an execution-layer node.
```

wwared: nit: I think the `ProxyAdmin` has to be a smart contract now, because otherwise I don't think the OPCM flow works since EOAs can't `delegatecall`. See line 257 too, which mentions the `delegatecall` mechanism

Suggestion, something along the lines of:

```suggestion
The `ProxyAdmin` contract is controlled by some `owner` address that is a Safe multisig in the standard configuration.
```

wwared: nit: Should remove the Geth mention while we're here:

```suggestion
Since the execution engine implements the standard Ethereum execution-layer protocol, the OP Stack reuses the existing devp2p network and transaction pool to propagate transactions. Execution clients interoperate on this network. The same network can also be used to propagate submitted blocks and support snap-sync.
```

wwared (approved):
Comments I posted above are all non-blocking, these changes match the U20 and super-roots semantics and everything looks good

specs/protocol/overview.md
wwared: Added: the Pause Deputy can trigger but not lift a pause.

specs/protocol/overview.md
wwared: Applied, rewrapped to fit the 120-char line limit.

specs/protocol/overview.md
wwared: Agreed, leaving it out as interop-specific.

specs/protocol/overview.md
wwared: Applied.

specs/protocol/overview.md
wwared: Moved to L2 Node Components; "below" -> "above".

specs/protocol/overview.md
wwared: Removed here, kept the Core L1 one that carries the legacy-proxy sub-bullets.

specs/protocol/overview.md
wwared: Removed, along with the third copy in the Core L1 notes.

specs/protocol/overview.md
wwared: Applied, wrap point moved one word earlier to stay under 120 chars.

wwared: Applied as "the execution layer's" (dropped the hyphen) to fit the line limit.

wwared: Applied.

wwared: Applied.

wwared: Applied with "peer-to-peer" instead of "devp2p", which fails cspell.

